PR #4, opened 2023-09-21
Fix miscounting the number of parameters when an issue is recorded.

grynspan:
We're accidentally reporting the number of characters in the string representation of each argument as the number of arguments.

For example, with a 1-arg function, we're currently emitting something like:

```
✘ Test f(i:) recorded an issue with 6 arguments i → 10 at fooTests.swift:13:3: Expectation failed: (i → 10) < 10
```

When we should be emitting:

```
✘ Test f(i:) recorded an issue with 1 argument i → 10 at fooTests.swift:13:3: Expectation failed: (i → 10) < 10
```


grynspan:
@swift-ci please test

PR #6, opened 2023-09-21
Easier docker commands in readme

weissi:
### Motivation:

`docker run` allows to set the working directory with `-w` which seems easier than passing `--package-path`

### Modifications:

- Suggest to use `docker run [...] -w /swift-testing`
- Remove the shell wrapping

### Result:

Easier docker commands


stmontgomery (approved):
This looks good to me, thanks!

PR #12, opened 2023-09-23
Update .spi.yml

finestructure:
This opts doc generation in to running with 5.9. We're building with the latest release (at the moment still 5.8 - until next week), which fails for this package since it's 5.9+.

The `swift_version` can be removed after we update to 5.9 as the latest release.

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -4,3 +4,4 @@ metadata:
 builder:
   configs:
     - documentation_targets: [Testing]
+      swift_version: 5.9


stmontgomery (approved):
Thanks! I had noticed the DocC wasn't posted yet on SPI and planned to investigate or reach out to you folks, so appreciate you proactively fixing this for us.

grynspan:
Thanks for the assist!

finestructure:
Happy to help :)

PR #13, opened 2023-09-23
Fix typo in Vision.md

ElonPark:
The original example in the vision document passes because `x` is greater than 1.

### Motivation:

Failure case, but condition is a success

### Modifications:

Change comparison operator to less than

### Result:

Examples that match the description


grynspan (approved):

stmontgomery (approved):
Thanks 😊

PR #20, opened 2023-09-25
Update MigratingFromXCTest.md

yermukhanbet:
Columns with Rows seemed not to be working. Could not find the proper way of columns and rows, so changed to simple before and after vertical code for this time, so it could be seen properly.


grynspan:
Thanks for your interest in this repo! This document is part of our DocC bundle and is designed to be rendered using DocC rather than (for example) GitHub. `@Column` and `@Row` are features of DocC.

PR #21, opened 2023-09-25
#expect(throws:) expression should accept non-Void return type and describe returned values

stmontgomery:
This changes `#expect(throws:)` and similar macros so that their closure parameters accept any type instead of `Void` specifically.

### Motivation:

This makes testing trivial throwing expressions of non-`Void` types easier, by not requiring syntax to discard the return value. For example:

```swift
func mightThrowXOrReturnY() throws -> Y { ... }
#expect(throws: X.self) {
  try mightThrowXOrReturnY() // ⚠️ Result of call to 'mightThrowXOrReturnY()' is unused
}
```

### Modifications:

- Change closure parameters of relevant macro-related declarations to use generic `R` type instead of `Void`.
- Change implementation of relevant `__check...` functions to include a description of return value.
- Add new unit tests.
- Update documentation to mention that return values are always discarded.

### Result:

The changes in this PR mean the warning above will no longer be emitted by the compiler. These changes also allow printing a description of the return value if the expectation fails and the expression does not throw an error, providing a more detailed and actionable failure message.

### DIFF
--- a/Sources/Testing/Expectations/Expectation+Macro.swift
+++ b/Sources/Testing/Expectations/Expectation+Macro.swift
@@ -83,7 +83,8 @@
 ///
 /// If `expression` does not throw an error, or if it throws an error that is
 /// not an instance of `errorType`, an ``Issue`` is recorded for the test that
-/// is running in the current task.
+/// is running in the current task. Any value returned by `expression` is
+/// discarded.
 ///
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test Linux

PR #23, opened 2023-09-25
Rename `memberFunctionCall` argument since it's not always a member function.

grynspan:
Rename `memberFunctionCall` argument since it's not always a member function.


grynspan:
@swift-ci please test

PR #27, opened 2023-09-25
Allow specifying the source location of `#expect()` and `#require()`.

### LINKED ISSUE #15: Provide an API to supply explicit file/line locations
### Description

When writing helper functions that call XCTest assertion methods, it's common to have those helpers take defaulted `file`/`line` arguments and pass those through to the XCT* calls; this makes the failure message appear at the call site of the helper method instead of all being grouped inside helper. A contrived example:

```swift
func assertIsNotEmpty(_ collection: some Collection, file: StaticString = #file, line: UInt = #line) {
  XCTAssertFalse(collection.isEmpty, file: file, line: line)
}

func testSomething() {
  assertIsNotEmpty([])  // failure appears here
}
```

I didn't see this explicitly supported by the APIs currently available; I looked in the vision doc for future plans but just found [this reference](https://github.com/apple/swift-testing/blob/main/Documentation/Vision.md#expectations):

> The source code location of the expectation, typically using the format #file:#line:#column. This helps test authors jump quickly to the line of code to view context, and lets IDEs present the failure in their UI at that location.

Maybe adding defaulted `file` and `line` arguments to the `#expect`/`#require` macros would be sufficient here?

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

0.0.0-initial

_No response_

grynspan:
This PR adds an additional argument to the `#expect()` and `#require()` macros that, if specified, replaces the source location derived at macro expansion time and passed to `__check()`. This change is ABI-safe for anyone who has already adopted swift-testing because it does not affect the emitted code or symbols. It is source-safe.

In the typical use case, the existing/old code path is taken at compile time and this PR has no effect. If a developer wants to specify a precise source location, they can do so.

This PR also changes the default behaviour of `#expect()` to generate an instance of `SourceLocation` using default arguments rather than by asking the macro context to provide all info. This is a little bit faster and is more accurate when a freestanding macro expression is embedded in another macro expression, such as:

```swift
#expect(throws: FooError.self) {
  #expect(try foo())
}
```

> [!NOTE]
> The arguments to `SourceLocation.init()` all currently have default values. We might want to change that in the future so that a developer cannot accidentally underspecify an instance. Where a default value is truly required, we can offer a `.here()` factory function that explicitly indicates intent.

Resolves rdar://112974442.
Resolves #15.

### DIFF
--- a/Sources/TestingMacros/ConditionMacro.swift
+++ b/Sources/TestingMacros/ConditionMacro.swift
@@ -40,6 +40,10 @@ private protocol _ConditionMacro: ExpressionMacro, Sendable {
 
 // MARK: -
 
+/// The token used as the label of the source location argument passed to
+/// `#expect()` and `#require()`.
+private let _sourceLocationLabel = TokenSyntax.identifier("sourceLocation")
+
 /// The token used as a mandatory label on any (first) trailing closure used
 /// with `#expect()` or `#require()`.
 private let _trailingClosureLabel = TokenSyntax.identifier("performing")
@@ -65,6 +69,9 @@ extension _ConditionMacro {
       // never the first argument.)
       commentIndex = macroArguments.dropFirst().lastIndex { $0.label == nil }
     }
+    let sourceLocationArgumentIndex = macroArguments.lazy
+      .compactMap(\.label)
+      .firstIndex { $0.tokenKind == _sourceLocationLabel.tokenKind }
 
     // Construct the argument list to __check().
     let expandedFunctionName: TokenSyntax
@@ -75,6 +82,7 @@ extension _ConditionMacro {
         // Include all arguments other than the "comment" argument here.
         checkArguments += macroArguments.indices.lazy
           .filter { $0 != commentIndex }
+          .filter { $0 != sourceLocationArgumentIndex }
           .map { macroArguments[$0] }
 
         // The trailing closure should be the focus of the source code capture.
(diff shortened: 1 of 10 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/TestingMacros/ConditionMacro.swift
Matejkob, line 1: I believe that the [documentation for `_ConditionMacro`](https://github.com/apple/swift-testing/blob/e64068037b8a5260e2c0105a20376527382933be/Sources/TestingMacros/ConditionMacro.swift#L22-L25) also needs to be updated because it contains information:

> /// Overloads of these macros that evaluate an expression take exactly two
> /// arguments, a "condition" argument and a `Comment?`-typed argument.

about two arguments, but now we have three, right?

grynspan: Good eye. :) I'll tweak that.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

stmontgomery (approved):
One small thing, but LGTM!

Sources/TestingMacros/ConditionMacro.swift
```diff
@@ -75,6 +83,7 @@ extension _ConditionMacro {
         // Include all arguments other than the "comment" argument here.
```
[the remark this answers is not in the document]
grynspan: Will revise. Thanks!

PR #29, opened 2023-09-26
Add support for the `SWT_SF_SYMBOLS_ENABLED` envvar.

grynspan:
On macOS, if we detect the presence of the SF Pro font, we enable SF Symbols output. This is the wrong behaviour if using a third-party terminal app that doesn't support it, or if writing to a file or pipe that we've misdetected as being a TTY.

This environment variable provides an opt-out (or explicit opt-in) for SF Symbols support on macOS. It is already documented and was meant to be supported.


grynspan:
@swift-ci please test

PR #30, opened 2023-09-26
Make `Test.all` visible as SPI.

grynspan:
`Test.all` produces a list of all tests visible at runtime via metadata inspection. By making it SPI, it can be filtered or otherwise modified, then passed to `Runner` or `Runner.Plan` to produce a partial list of tests for running.

Resolves rdar://116069136.


grynspan:
@swift-ci please test

grynspan:
I realize as I post this that it comes in at the wrong stage, so I'm going to close and open a new one with a different approach.

suzannaratcliff (approved):

PR #31, opened 2023-09-27
Do not re-add leading/trailing trivia when expanding condition macros.

grynspan:
swift-syntax 5.9.0 is able to adjust leading and trailing trivia after macro expansions, so it is no longer necessary for swift-testing to do so.

Resolves rdar://113853493.


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #32, opened 2023-09-27
Allow arguments to tests to be gathered asynchronously.

grynspan:
This change allows a developer to specify an argument to a test that is the result of an asynchronous function/operation. For example:

```swift
@Test(arguments: await downloadFilesFromInternet())
func readFile(file: File) {
  ...
}
```

These arguments are resolved at runtime during the test planning stage.

> [!NOTE]
> This change does not enable _throwing_ arguments. That's more complicated (because we need to potentially handle the errors on an individual basis rather than stopping the entire operation.) I'm leaving it for a separate PR.

### DIFF
--- a/Sources/Testing/Test+Macro.swift
+++ b/Sources/Testing/Test+Macro.swift
@@ -542,7 +542,7 @@ public func __invokeXCTestCaseMethod<T>(
 @_alwaysEmitConformanceMetadata
 public protocol __TestContainer {
   /// The set of tests contained by this type.
-  static var __tests: [Test] { get }
+  static var __tests: [Test] { get async }
 }
 
 extension Test {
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci please test

Sources/Testing/Test+Macro.swift
```diff
@@ -554,38 +554,42 @@ extension Test {
...
+    get async {
+      await withTaskGroup(of: [Self].self) { taskGroup in
+        swt_enumerateTypes({ typeName, _ in
+          // strstr() lets us avoid copying either string before comparing.
+          Self._testContainerTypeNameMagic.withCString { testContainerTypeNameMagic in
+            nil != strstr(typeName, testContainerTypeNameMagic)
+          }
+        }, /*typeEnumerator:*/ { type, context in
+          if let type = unsafeBitCast(type, to: Any.Type.self) as? any __TestContainer.Type {
+            let taskGroup = context!.assumingMemoryBound(to: ThrowingTaskGroup<[Self], any Error>.self)
```
stmontgomery, line 583: Since we're not using with**Throwing**TaskGroup, did you mean `TaskGroup` here?
```suggestion
            let taskGroup = context!.assumingMemoryBound(to: TaskGroup<[Self], any Error>.self)
```

grynspan: _whoopsie_

grynspan:
@swift-ci please test

PR #34, opened 2023-09-28
Fix a new build error as of today's toolchain.

grynspan:
`await` is no longer allowed on an `if` expression, so tweak the code that does that.


grynspan:
@swift-ci please test

stmontgomery (approved):
Interesting… Was it always intentional to disallow that usage and it was just a bug?

grynspan:
I guess it was just a bug!

PR #39, opened 2023-09-30
Change Event to reference a test by ID instead of including a full Test

stmontgomery:
Change the `Event` type to only include a `Test.ID` value instead of a full `Test` instance, and add a "context" to the `Event.Handler` function signature so that event handlers receive a context object they can use to lookup the `Test` an `Event` references if needed.

### Motivation:

This is a step towards a larger goal of allowing `Event` and some of its related types to conform to `Codable`. It also makes each individual `Event` instance lighter-weight in terms of its stored properties.

### Modifications:

- Change `Event` to have a `Test.ID` instead of a `Test` (still optional).
- Add a new `Event.Context` type which gets passed to a new second parameter of `Event.Handler`.
- Add a new (private) "runner context" type which centralizes the several pieces of task-local state managed by a `Runner`.
- Replace several places which used `Configuration.current` to be `Runner.current`, including the mechanism for falling back to all-known runners in detached tasks.
- Adjust unit tests.


stmontgomery:
@swift-ci please test

dennisweissmann (approved):

Sources/Testing/Running/Configuration+EventHandling.swift
```diff
@@ -13,12 +13,13 @@ extension Configuration {
...
   ///   - event: The event to handle.
+  ///   - context: The context associated with the event.
   ///
   /// This method is used to handle events which are generated by the testing
   /// library. Typically this only involves passing the event to this instance's
   /// `eventHandler` but this method may also be used as a customization point
   /// to change how the event is passed to the event handler.
-  func handleEvent(_ event: Event) {
-    eventHandler(event)
+  func handleEvent(_ event: consuming Event, in context: consuming Event.Context) {
```
grynspan, line 22: We might end up calling multiple event handlers (e.g. if an event occurs on a detached task) so these should be borrowing, not consuming, if anything.

Sources/Testing/Running/Runner.Context.swift
```diff
@@ -0,0 +1,204 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+extension Runner {
+  /// A type which collects the task-scoped context for a running ``Runner``
+  /// instance, the tests it runs, and other objects it interacts with.
+  ///
+  /// This type is intended for use via the task-local
+  /// ``Runner/_Context/current`` property.
+  fileprivate struct _Context: Sendable {
```
grynspan, line 17: The style we've adopted is to _not_ put an underscore before `fileprivate` symbols, just `private` ones.

Sources/Testing/Running/Runner.Context.swift
```diff
@@ -0,0 +1,204 @@
...
+    /// The test case that is running on the current task, if any.
+    var testCase: Test.Case?
+
+    /// The context related to the runner running on the current task.
+    @TaskLocal
+    fileprivate static var current: Self = .init()
+  }
+}
+
+extension Runner {
```
grynspan, line 33: By making `Runner` the task-local value, we are hoisting much more data into task-local storage (namely the test graph.) Are we okay with that?

Sources/Testing/Events/Event.Recorder.swift
```diff
@@ -411,19 +411,21 @@ extension Event.Recorder {
   ///
   /// - Parameters:
   ///   - event: The event to record.
+  ///   - eventContext: The context associated with the event.
   ///
   /// - Returns: A string description of the event, or `nil` if there is nothing
   ///   useful to output for this event.
-  func _record(_ event: Event) -> String? {
+  func _record(_ event: Event, in eventContext: Event.Context) -> String? {
```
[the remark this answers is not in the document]
grynspan: Any chance the two context types could be merged? 🤔

Sources/Testing/Events/Event.swift
```diff
@@ -143,37 +143,75 @@ extension Event {
...
-  func post(configuration: Configuration? = nil) {
-    if let configuration = configuration ?? Configuration.current {
+  /// `runner` is not `nil`, `self` is passed to the value of the
+  /// ``Configuration/eventHandler`` property of its ``Runner/configuration``.
+  /// If `runner` is `nil`, and ``Runner/current`` is _not_ `nil`, its
+  /// configuration's event handler is used instead. If there is no current
+  /// runner, the event is posted to the event handlers of all runners'
+  /// configurations which have been set as current across all tasks in the
+  /// process.
+  func post(runner: Runner? = nil) {
```
grynspan, line 197: Need a better parameter label for this arg. `to` is probably fine.

Sources/Testing/Running/Runner.Context.swift
```diff
@@ -0,0 +1,204 @@
...
+
+    /// The test case that is running on the current task, if any.
+    var testCase: Test.Case?
+
+    /// The context related to the runner running on the current task.
+    @TaskLocal
+    fileprivate static var current: Self = .init()
+  }
+}
+
```
grynspan, line 32: Pragma mark here?

Sources/Testing/Running/Runner.Context.swift
```diff
@@ -0,0 +1,204 @@
...
+  /// instance, the tests it runs, and other objects it interacts with.
+  ///
+  /// This type is intended for use via the task-local
+  /// ``Runner/_Context/current`` property.
+  fileprivate struct _Context: Sendable {
+    /// The runner that is running on the current task, if any.
+    var runner: Runner?
+
+    /// The test that is running on the current task, if any.
+    var test: Test?
```
grynspan, line 22: Sharing all these values in a single task-local is going to change the behaviour of tests/runners nested in other runners (i.e. our own tests.) It means that when we temporarily swap out the configuration to run our own unit tests, we'll be swapping out more state than just the configuration, and we need to make sure that a test failure in our own test target can handle that correctly like before.

Sources/Testing/Events/Event.swift
```diff
@@ -143,37 +143,75 @@ extension Event {
...
+    /// - Parameters
+    ///   - event: The event whose associated test should be retrieved.
+    ///
+    /// - Returns: The ``Test`` which `event` was associated with, if any. If
+    ///   `event` was not associated with a particular test, `nil` is returned.
+    ///
+    /// This uses the event's ``Event/testID`` property to look up the
+    /// corresponding ``Test`` instance.
+    public func test(for event: Event) -> Test? {
+      guard let testID = event.testID else {
```
grynspan, line 175: This interface is awkward. Since the context for an event is specific to a particular test, we should just have that test as a stored property of the context rather than providing a lookup function.

We should also migrate the current test case in here for consistency.

Sources/Testing/Events/Event.swift
```diff
@@ -143,37 +143,75 @@ extension Event {
...
+  /// passed to an ``Event/Handler``.
+  public struct Context: Sendable {
+    /// The runner which ran the test which caused this instance's associated
+    /// event to be posted.
+    ///
+    /// This runner may be used within a ``Event/Handler`` to retrieve the
+    /// ``Test`` (if any) that a particular ``Event`` references via its
+    /// ``Event/testID`` property.
+    @_spi(ExperimentalTestRunning)
+    public var runner: Runner
```
grynspan, line 161: Is this property's value correct when we temporarily pop the current context for self-hosted tests? (We need a unit test for that.)

grynspan (changes requested):

Sources/Testing/Running/Configuration+EventHandling.swift
stmontgomery: Oddly, the compiler isn't catching this (shouldn't it?)

grynspan: Since the type is copyable, I am pretty sure it just ends up copying it. i.e. `consuming` in parameter position doesn't really have any effect on copyable types. I think.

Sources/Testing/Running/Runner.Context.swift
stmontgomery: I've amended my approach in the most recent commits to no longer capture the entire `Runner` in task local storage

Sources/Testing/Events/Event.swift
stmontgomery: Yeah, I've taken this suggestion and now capture just the current `Test` and `Test.Case` values in this `Context` instance.

Sources/Testing/Running/Runner.Context.swift
stmontgomery: I confirmed (via a new unit test I added) that the handling of this is correct, and in fact I think the new approach is _more_ correct than before: Previously we only swapped out the current `Configuration`, but now, because this group of values is stored together as one `Runner.Context` instance under a single `@TaskLocal`, I can easily swap the entire set of them to their previous value. 

To illustrate: Before this PR, the following test would've failed:

```swift
  @Test func contextScopedEventHandler() async {
    configuration.eventHandler = { _, _ in
      #expect(Test.current?.name == "contextScopedEventHandler()")  // ❌ got "foo" instead
    }

    await Test(name: "foo") {}.run(configuration: configuration)
  }

```

Sources/Testing/Events/Event.swift
stmontgomery: Addressed that [here](https://github.com/apple/swift-testing/pull/39#discussion_r1342940290)!

Sources/Testing/Running/Runner.Context.swift
grynspan: Okay, I'm satisfied with that explanation!

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Runner.Context.swift
```diff
@@ -0,0 +1,226 @@
...
+  /// Call a function while the value of ``Test/Case/current`` is set.
+  ///
+  /// - Parameters:
+  ///   - testCase: The new value to set for ``Test/Case/current``.
+  ///   - body: A function to call.
+  ///
+  /// - Returns: Whatever is returned by `body`.
+  ///
+  /// - Throws: Whatever is thrown by `body`.
+  static func withCurrent<R>(_ testCase: Self, perform body: () throws -> R) rethrows -> R {
```
grynspan, line 206: Do we use the `sync` variants? If they're just dead code, let's drop 'em.

Sources/Testing/Running/Runner.Context.swift
```diff
@@ -0,0 +1,226 @@
...
+  /// Call a function while the value of ``Test/Case/current`` is set.
+  ///
+  /// - Parameters:
+  ///   - testCase: The new value to set for ``Test/Case/current``.
+  ///   - body: A function to call.
+  ///
+  /// - Returns: Whatever is returned by `body`.
+  ///
+  /// - Throws: Whatever is thrown by `body`.
+  static func withCurrent<R>(_ testCase: Self, perform body: () async throws -> R) async rethrows -> R {
```
grynspan, line 221: It occurs to me we could probably merge these two implementations into one that takes a `WriteableKeyPath<Runner.Context, T>`. I don't know if that's worth exploring (if it is, I'll look at it for a future PR, no need to do it today.)

Sources/Testing/Running/Runner.Context.swift
stmontgomery: Ah yes, we do not need them, and removing them improves code coverage.

Sources/Testing/Events/Event.swift
```diff
@@ -123,16 +123,46 @@ public struct Event: Sendable {
...
+    self.testID = testID
     self.testCase = testCase
     self.instant = instant
   }
+
+  /// Post an ``Event`` with the specified values.
+  ///
+  /// - Parameters:
+  ///   - kind: The kind of event that occurred.
+  ///   - testID: The test for which the event occurred, if any.
```
grynspan, line 146: Incorrect parameter label.

Sources/Testing/Events/Event.swift
```diff
@@ -123,16 +123,46 @@ public struct Event: Sendable {
...
-  init(_ kind: Kind, for test: Test? = .current, testCase: Test.Case? = .current, instant: Test.Clock.Instant = .now) {
+  ///
+  /// When creating an event to be posted, use
+  /// ``post(_:for:testCase:instant:configuration)`` instead since that ensures
+  /// any task local-derived values in the associated ``Event/Context`` match
+  /// the event.
+  init(_ kind: Kind, testID: Test.ID?, testCase: Test.Case?, instant: Test.Clock.Instant = .now) {
     self.kind = kind
-    self.test = test
+    self.testID = testID
```
grynspan, line 137: If the `Test` has moved to `Context`, we should consider moving the test case as well since it can carry an arbitrarily heavy payload in the form of arguments.

Sources/Testing/Running/Runner.Context.swift
stmontgomery: Sounds good, I'll leave that to you to explore

Sources/Testing/Events/Event.swift
```diff
@@ -159,21 +216,21 @@ extension Event {
   /// instead. If there is no current configuration, the event is posted to
   /// the event handlers of all configurations set as current across all tasks
   /// in the process.
-  func post(configuration: Configuration? = nil) {
+  private func post(context: Context, configuration: Configuration? = nil) {
```
grynspan, line 219: Label for `context` should probably be `in` for consistency?

Sources/Testing/Events/Event.swift
stmontgomery: I have now removed `Event.testCase` and left `Event.Context.testCase`. I plan to continue working in this area and find a way to replace the removed property with a lighter-weight value that indicates _which_ test case the event was for, without needing to include the full `Test.Case` value.

stmontgomery:
@swift-ci please test

Sources/Testing/Events/Event.swift
```diff
@@ -159,21 +214,21 @@ extension Event {
   /// instead. If there is no current configuration, the event is posted to
   /// the event handlers of all configurations set as current across all tasks
   /// in the process.
-  func post(configuration: Configuration? = nil) {
+  private func post(in context: Context, configuration: Configuration? = nil) {
```
grynspan, line 217: `private` -> `_`

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

grynspan (approved):

PR #40, opened 2023-10-01
Refactor platforms from Package.swift

uhooi:
### Motivation:

Basically, enums and static variables should be preferred over strings because they are type-safe.
Since the apple org repository serves as an example for everyone, there is a risk of being imitated because people think it is correct to use string-based.

### Modifications:

Specify version type-safe.

### Result:

Refactoring

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -120,7 +120,7 @@ extension Array where Element == PackageDescription.SwiftSetting {
         "-Xfrontend", "-define-availability", "-Xfrontend", "_swiftVersionAPI:macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0",
       ]),
       .enableUpcomingFeature("ExistentialAny"),
-      .define("SWT_TARGET_OS_APPLE", .when(platforms: [.macOS, .iOS, .macCatalyst, .watchOS, .tvOS, .custom("visionos")])),
+      .define("SWT_TARGET_OS_APPLE", .when(platforms: [.macOS, .iOS, .macCatalyst, .watchOS, .tvOS, .visionOS])),
     ]
   }
 }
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci please test

grynspan:
Thanks for opening this PR! Rather than writing "SSIA", it would be really helpful if you could give us a quick explanation of the change. I know it may feel a bit redundant, but the PR description field is an opportunity to explain e.g. the benefits of a PR, any side effects that might occur, etc.

Package.swift
```diff
@@ -17,11 +17,11 @@ let package = Package(
...
-    .macOS("13.0"),
-    .iOS("16.0"),
-    .watchOS("9.0"),
-    .tvOS("16.0"),
-    .custom("visionos", versionString: "1.0"),
+    .macOS(.v13),
+    .iOS(.v16),
+    .watchOS(.v9),
+    .tvOS(.v16),
+    .visionOS(.v1),
```
stmontgomery, line 24: Thank you for contributing!

Note that this last one, for visionOS, required using `.custom` originally because the version of Xcode that our macOS CI system was using at that time did not yet have that API. It appears it is new enough now, but I recall there was a second usage of `.custom` later in this same file. Are you interested in updating that one from `.custom("visionos")` to `.visionOS` as well?

uhooi: I had missed it. Corrected. https://github.com/apple/swift-testing/pull/40/commits/ab019df73cd6199b18ecca16e7c8ddedaa37fc72

uhooi:
@grynspan Sorry.
I wrote a little more detail.
https://github.com/apple/swift-testing/pull/40#issuecomment-1742072084

grynspan:
Much appreciated!

grynspan:
@swift-ci please test

stmontgomery (approved):

PR #44, opened 2023-10-02
Make sure the nesting level of sections in "Migrating from XCTest" is consistent.

grynspan:
This PR adjusts the header level of some of the sections of MigratingFromXCTest.md so that they are consistent. Namely, those sections that deal with converting specific "gadgets" within a test method/function are meant to be nested under the section about converting test methods/functions.


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #45, opened 2023-10-02
Refactor `SourceCode` factory functions a bit.

grynspan:
Right now, we have two overloads of `__fromComponents()` that really should be separate functions entirely. Refactor/rename to eliminate ambiguity during macro expansion.


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #49, opened 2023-10-03
Use `@autoclosure` in `__checkBinaryOperation()`.

grynspan:
This change simplifies the macro expansions of `#expect()` and `#require()` when they are passed binary operation expressions. In prerelease versions of the Swift 5.9 toolchain, we ran into issues with autoclosures and worked around them by capturing right-hand operands in full closures. Those issues have been resolved. By switching to autoclosures we can solve some grammatical issues that crop up with explicit closures, such as:

```swift
@Test func foo() {
  let x = [2, 4, 6]
  x.forEach {
    #expect(0 == $0 % 2) // 🛑 Contextual closure type '() -> Int' expects 0
                         // arguments, but 1 was used in closure body
  }
}
```

Partially resolves rdar://114275884.

### DIFF
--- a/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
+++ b/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
@@ -109,7 +109,7 @@ public func __checkValue(
 private func _callBinaryOperator<T, U, R>(
   _ lhs: T,
   _ op: (T, () -> U) -> R,
-  _ rhs: @autoclosure () -> U
+  _ rhs: () -> U
 ) -> (result: R, rhs: U?) {
   // The compiler normally doesn't allow a nonescaping closure to call another
   // nonescaping closure, but our use cases are safe (e.g. `true && false`) and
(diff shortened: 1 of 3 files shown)


Sources/Testing/Expectations/ExpectationChecking+Macro.swift
```diff
@@ -108,17 +108,17 @@ public func __checkValue(
 /// - Throws: Whatever is thrown by `op`.
 private func _callBinaryOperator<T, U, R>(
   _ lhs: T,
-  _ op: (T, () -> U) throws -> R,
```
grynspan, line 111: We don't expand expressions containing `try` (currently), so this can't ever be throwing in practice.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

stmontgomery (approved):

stmontgomery:
Could we add a new test with the snippet you showed in the PR description?

grynspan:
> Could we add a new test with the snippet you showed in the PR description?

Only yesterday's toolchain has a fix necessary for that code to compile, so I'm leaving it out for now. Will revisit when some more dust has settled on the fix.

PR #51, opened 2023-10-03
Fix a warning when building the GitStatus target on Windows.

grynspan:
On Windows, we get a warning about `var arguments` in this target never being mutated. This PR resolves that warning.


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #52, opened 2023-10-04
Add a protocol defining custom descriptions for types during expectation failure.

grynspan:
This PR introduces a new protocol, `CustomTestStringConvertible`, to which types can conform in order to control their representations during an expectation failure. The protocol contains a single property of type `String`.

A default implementation is provided for `Optional` so that `Optional(x)` is not logged in favour of simply `x`. Developers can provide additional implementations for their own types.

I could use feedback on the name of the protocol and its property, in particular!

Example test:

```swift
@Test func f() {
  let x: Int? = 7
  #expect(x == 8)
}
```

Failure message (without this change):

> Expectation failed: (x → Optional(7)) == (8 → Optional(8))

It should just be:

> Expectation failed: (x → 7) == 8


grynspan:
@swift-ci please test

Sources/Testing/SourceAttribution/CustomExpressionExpandable.swift
```diff
@@ -0,0 +1,60 @@
...
+/// ```swift
+/// #expect(food.addSeasoning())
+/// ```
+///
+/// The expanded representation of the condition expression will be derived from
+/// `String(describing: food)`. If that string is unsuitable for display in a
+/// test's output, then the type can be made to conform to
+/// ``CustomExpressionExpandable`` and the value of the instance's
+/// ``descriptionInExpectationFailure`` property will be used instead.
+public protocol CustomExpectationFailureRepresentable {
```
stmontgomery, line 35: Not sure yet, but I could imagine generalizing this just a little further and naming the protocol `CustomFailureRepresentable` so that it encompasses representations of values in _any_ failure context, not just _expectation_ failures specifically.

Sources/Testing/SourceAttribution/CustomExpressionExpandable.swift
```diff
@@ -0,0 +1,60 @@
...
+  /// A description of this instance to use when describing it in an expectation
+  /// failure.
+  var descriptionInExpectationFailure: String { get }
```
stmontgomery, line 38: Based on above idea, this could become `descriptionInFailureRepresentation`?

Sources/Testing/SourceAttribution/CustomExpressionExpandable.swift
grynspan: I can buy this. Let's check around and see what other folks think?

Sources/Testing/SourceAttribution/CustomExpressionExpandable.swift
grynspan: Perhaps just `failureRepresentation`.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/SourceAttribution/CustomTestStringConvertible.swift
```diff
@@ -0,0 +1,154 @@
...
+extension String {
+  /// Initialize this instance so that it can be presented in a test's output.
+  ///
+  /// - Parameters:
+  ///   - value: The value to describe.
+  ///
+  /// ## See Also
+  ///
+  /// - ``CustomTestStringConvertible``
+  public init<T>(describingForTest value: T) {
```
stmontgomery, line 101: Not `some Any`?

stmontgomery: I wonder if we _need_ this to be `public`? The testing library certainly needs to call it, but do users? I can imagine scenarios where it might become useful but it might be best for this to be internal until we find a concrete need.

Sources/Testing/SourceAttribution/CustomTestStringConvertible.swift
```diff
@@ -0,0 +1,154 @@
...
+    // The mangled type name SPI doesn't handle generic types very well, so we
+    // ask for the dynamic type of `value` (type(of:)) instead of just T.self.
+    lazy var valueType = type(of: value as Any)
+    if let value = value as? any CustomTestStringConvertible {
+      self = value.testDescription
+    } else if let value = value as? any CustomStringConvertible {
+      self.init(describing: value)
+    } else if let value = value as? any CustomDebugStringConvertible {
+      self.init(reflecting: value)
```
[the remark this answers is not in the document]
grynspan: I did think of that, but I left it out mostly because I don't think it's a very common protocol to conform to… but it's trivial to add it. Strong feelings here?

Sources/Testing/SourceAttribution/CustomTestStringConvertible.swift
grynspan: It's parallel to `String(describing:)` and `CustomStringConvertible`. I _think_ it makes sense to expose both at the same visibility (whether that's `public`, `@_spi`, or something else.)

grynspan:
@swift-ci please test

Sources/Testing/SourceAttribution/CustomTestStringConvertible.swift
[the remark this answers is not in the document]
stmontgomery: Actually, could these 3 known stdlib protocol cases be collapsed into a single `else if` clause, since `String(describing:)` is known to handle them?
```swift
// If it's one of the known types that `String(describing:)` prioritizes, defer to that
} else if value is TextOutputStreamable
       || value is CustomStringConvertible
       || value is CustomDebugStringConvertible {
  self.init(describing: value)
} else if /* C enum */ {
 ...
} else if /* Swift enum */ {
 ...
} else {
  // Use the generic description of the value.
  self.init(describing: value)
}
```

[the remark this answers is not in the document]
grynspan: Yeah, we can do that.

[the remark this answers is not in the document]
grynspan: Actually, no, let's not, because we lose access to the specializations on `String.init(describing:)` and that means the runtime will end up needing to do twice as many conformance checks, which isn't free. See: https://github.com/apple/swift/blob/06984b430775bd3ae76d705a741d2427e6166fe1/stdlib/public/core/Mirror.swift#L625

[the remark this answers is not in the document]
stmontgomery: Ah right, makes sense. Well, I think I'd lean towards adding the `TextOutputStreamable` specialization at least

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

PR #53, opened 2023-10-07
Convert `StaticString`s that refer to a #filePath to `String`s so that they are `Codable`

dennisweissmann:
Preparation for a follow-up PR that makes `Event` conform to `Codable`.

### Motivation:

Some data structures need to conform to `Codable` so that they can be transmitted between the test process and connected processes.

### Modifications:

For consistency I converted all `StaticString`s that refer to a `#filePath` to `String`s, not just the ones I expect need to become `Codable`.

### Result:

This change enables us in a future PR to make certain types conform to `Codable`.

Resolves rdar://116629725


dennisweissmann:
@swift-ci please test

dennisweissmann:
@swift-ci please test

grynspan:
Would it make sense to wait for my `XCTIssue` stuff to land, then just delete the `#if` in XCTS?

Sources/Testing/Running/XCTestScaffold.swift
```diff
@@ -163,7 +163,8 @@ public enum XCTestScaffold: Sendable {
       // if the issue represents an assertion failure or a thrown error.
       if !issue.isKnown {
         let sourceLocation = issue.sourceLocation ?? .init()
-        XCTFail(String(describing: issue), file: sourceLocation._filePath, line: UInt(sourceLocation.line))
+
+        recordFailure(inTest: testCase, withDescription: String(describing: issue), inFile: sourceLocation._filePath, atLine: sourceLocation.line, expected: false)
```
stmontgomery, line 167: I think you'd want to pass `true` for `expected:`.

And this predates this PR, but if `issue.sourceLocation` is `nil`, I think it should not default to `?? .init()` because that would capture the source location inside `XCTestScaffold.swift`. Instead, I think it should use `inFile: "<unknown>, atLine: 0`. Those are the traditional values representing "no source location known".

dennisweissmann:
@swift-ci please test

dennisweissmann:
@swift-ci please test

stmontgomery (approved):

grynspan:
@swift-ci please test

Sources/Testing/Running/XCTestScaffold.swift
```diff
@@ -158,12 +158,11 @@ public enum XCTestScaffold: Sendable {
         testCase.rawValue.record(XCTIssue(issue, processLaunchedByXcode: isProcessLaunchedByXcode))
       }
 #else
-      // NOTE: XCTestCase.recordFailure(withDescription:inFile:atLine:expected:)
```
grynspan, line 161: Please keep this comment here so that we don't accidentally misuse the expected parameter later.

grynspan (changes requested):
One change requested. Otherwise, looks good to me.

dennisweissmann:
@swift-ci please test

grynspan (approved):

grynspan:
Don't forget to change this in XCTestScaffold.swift:

```swift
      XCTSourceCodeLocation(
        filePath: String(describing: sourceLocation._filePath),
```

grynspan (changes requested):
Actually, you should search the package for anything using `_filePath` and get rid of any redundant `String(describing:)` calls you find. :)

dennisweissmann:
> Actually, you should search the package for anything using `_filePath` and get rid of any redundant `String(describing:)` calls you find. :)

Thanks, done!

dennisweissmann:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan (approved):

stmontgomery (approved):

dennisweissmann:
Since CI is broken right now due to https://github.com/apple/swift/issues/69096, I double checked that this is building and testing successfully on macOS, Jonathan is looking into Linux.

dennisweissmann:
Linux building and testing was successful as well.

PR #56, opened 2023-10-10
Preserve the old signature of `__TestContainer.__tests` temporarily.

grynspan:
We have some precompiled test targets that used the previous definition of `__tests` which we can't recompile just yet. Preserve the previous protocol requirement temporarily until we have a chance to update everything.

We'll revert (most of) this change in a couple of weeks.

### DIFF
--- a/Sources/Testing/Test+Macro.swift
+++ b/Sources/Testing/Test+Macro.swift
@@ -542,7 +542,29 @@ public func __invokeXCTestCaseMethod<T>(
 @_alwaysEmitConformanceMetadata
 public protocol __TestContainer {
   /// The set of tests contained by this type.
-  static var __tests: [Test] { get async }
+  ///
+  /// - Note: This property will be removed in a future update to the testing
+  ///   library. It remains here for binary compatibility with recently-compiled
+  ///   test targets.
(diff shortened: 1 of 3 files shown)


suzannaratcliff (approved):
Looks good!

grynspan:
@swift-ci please test

Sources/Testing/Test+Macro.swift
```diff
@@ -542,7 +542,29 @@ public func __invokeXCTestCaseMethod<T>(
...
+  ///   it directly.
+  static func __tests() async -> [Test]
+}
+
+extension __TestContainer {
+  public static var __tests: [Test] {
+    []
+  }
+
+  public static func __tests() async -> [Test] {
```
stmontgomery, line 565: I understand the need for the default implementation of `var __tests` above, but do we need this one?

grynspan: Yes, if we have an old compiled test running on a newer build of swift-testing, we'll hit this implementation.

stmontgomery (approved):

grynspan:
@swift-ci please test

PR #58, opened 2023-10-12
Link to Swift issues about using `_mangledTypeName()`.

grynspan:
Adds comments linking to issues filed against the Swift repo that if resolved would help us avoid using this unsupported interface.


stmontgomery (approved):

grynspan:
As this change affects comments only, skipping CI runs that we know are failing right now.

PR #60, opened 2023-10-12
Fix the 256-color terminal check so it doesn't require `_StringProcessing`.

grynspan:
Use `strstr()` so that we can detect 256-color terminal support on older macOS versions or with older toolchains that don't have the `_StringProcessing` module.

I did _not_ change another use of `contains()` in ExpectationChecking+Macro.swift because it may be sensitive to Unicode nuances whereas a search for the ASCII string ``"256"` isn't.

I also took the opportunity to replace one other usage of `#available(explicit_versions_here)` with an availability define. I sorted the availability defines in the package manifest too.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -114,11 +114,13 @@ extension Array where Element == PackageDescription.SwiftSetting {
         "-strict-concurrency=complete",
         "-require-explicit-sendable",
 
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

briancroom (approved):

stmontgomery (approved):

grynspan:
@swift-ci please test

grynspan:
Failures are due to a known compiler issue. Tested at desk.

PR #63, opened 2023-10-14
Support setting a debugger breakpoint to pause when a test fails

stmontgomery:
Add a function with a well-defined symbol that gets invoked whenever a failure occurs. This allows an attached debugger to place a breakpoint on that symbol and pause execution whenever a test failure is recorded.

### Motivation:

It's often helpful when investigating a test failure to attach using a debugger and place a breakpoint to step through the code. And sometimes, you want to pause exactly at the spot where a test failure occurs, instead of at the beginning of a particular test. This makes that workflow possible by establishing a function for which you can create a symbolic breakpoint, so that execution will pause whenever a test failure occurs.

### Modifications:

- Introduce a new function which is called whenever a test failure is recorded. Open to naming improvement suggestions!
- Allow configuring this "failure breakpoint" in `Configuration` to facilitate unit testing.
- Add new unit tests.

### Result:

It is possible to set a symbolic breakpoint on this new function and pause upon failure.

Resolves rdar://116948062

### DIFF
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -141,14 +141,14 @@ extension Issue {
 /// the testing library calls this function synchronously. This facilitates
 /// interactive debugging of test failures by allowing a symbolic breakpoint to
 /// be added specifying the symbol name of this function
-/// (`SWTFailureBreakpoint`), so that the debugger may pause execution and allow
-/// a user to inspect the process state. This function performs no action of its
-/// own.
+/// (`swt_failureBreakpoint`), so that the debugger may pause execution and
+/// allow a user to inspect the process state. This function performs no action
+/// of its own.
 ///
 /// This function is not part of the public interface of the testing library,
 /// but it is exported and its symbol name must remain stable.
-@_cdecl("SWTFailureBreakpoint")
-@inline(never)
+@_cdecl("swt_failureBreakpoint")
+@inline(never) @_optimize(none)
 @usableFromInline
 func failureBreakpoint() {
   // Empty.
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -71,6 +71,15 @@ extension Issue {
     }
 
     Event.post(.issueRecorded(self), configuration: configuration)
+
+    if !isKnown {
+      // Since this is not a known issue, invoke the failure breakpoint.
+      //
+      // Do this after posting the event above, to allow the issue to be printed
+      // to the console first (assuming the event handler does this), since that
+      // can help explain the failure.
+      failureBreakpoint()
+    }
   }
 
   /// Record an issue when a running test fails unexpectedly.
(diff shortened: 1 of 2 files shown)


stmontgomery:
@swift-ci please test

grynspan:
Is the intent here just to provide a named function that can be used as a breakpoint target? If so, we can probably call it directly from post(). We will also want to annotate the C++ function with a few attributes to make sure clang doesn't optimize it away.

stmontgomery:
@swift-ci please test

stmontgomery:
> Is the intent here just to provide a named function that can be used as a breakpoint target? If so, we can probably call it directly from post().

I wanted to make it a `Configuration` property for a couple reasons:

- It facilitates unit testing, which the new tests in this PR demonstrate.
- When integrating with tools, this `Configuration` property provides the _option_ of calling a pre-existing function/symbol, which may be useful in case the tool already has some debugging support it wants to integrate with.

But customizing this new `Configuration` property is not _necessary_ to take advantage of this new feature. This PR intentionally defaults to a new, documented symbol which this library provides directly, and this makes the feature usable even when directly attached via a debugger, without other intermediary tools involved.

> We will also want to annotate the C++ function with a few attributes to make sure clang doesn't optimize it away.

Definitely! I [included](https://github.com/apple/swift-testing/pull/63/files#diff-e4c99e48fcfc7209c4f7c2d0b960874acb7e2cb318687994ff32d039da9c53a4R26) `__attribute__((noinline)) __attribute__((disable_tail_calls)) SWT_SENDABLE` — are there other/different attributes you think are needed?

grynspan:
> When integrating with tools, this Configuration property provides the option of calling a pre-existing function/symbol

Seems like this is more easily solved by having the tool, in its implemented event handler, call that function directly. I am not sure we actually need any code changes at all to support such a workflow.

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -125,3 +134,22 @@ extension Issue {
...
+/// the testing library calls this function synchronously. This facilitates
+/// interactive debugging of test failures by allowing a symbolic breakpoint to
+/// be added specifying the symbol name of this function
+/// (`SWTFailureBreakpoint`), so that the debugger may pause execution and allow
+/// a user to inspect the process state. This function performs no action of its
+/// own.
+///
+/// This function is not part of the public interface of the testing library,
+/// but it is exported and its symbol name must remain stable.
+@_cdecl("SWTFailureBreakpoint")
```
grynspan, line 150: Nitpick: the style we're using for C/C++ symbols is `swt_camelCase()`. We should either make this name match that style (`swt_failureBreakpoint()`) or change our style to be consistent.

grynspan (changes requested):
C naming nitpick.

stmontgomery:
@swift-ci please test

grynspan (approved):

grynspan:
:shipit:

stmontgomery:
@swift-ci please test

PR #64, opened 2023-10-16
Update link to `#fileID` documentation that has moved to the stdlib.

grynspan:
`#fileID` is no longer a literal but is now a macro in the stdlib, so its documentation has moved. Update our documentation to match.

### DIFF
--- a/Sources/Testing/SourceAttribution/SourceLocation.swift
+++ b/Sources/Testing/SourceAttribution/SourceLocation.swift
@@ -16,7 +16,6 @@ public struct SourceLocation: Sendable {
   ///
   /// - ``moduleName``
   /// - ``fileName``
-  /// - [The Swift Programming Language &mdash; Literal Expressions](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/expressions/#Literal-Expression)
   public var fileID: String {
     didSet {
       precondition(!fileID.isEmpty)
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):
Seems fine but I wonder if we could avoid it by phrasing as "…in the documentation for `#fileID` in the standard…"

grynspan:
> Seems fine but I wonder if we could avoid it by phrasing as "…in the documentation for `#fileID` in the standard…"

Got an alternative? :)

Sources/Testing/SourceAttribution/SourceLocation.swift
```diff
@@ -32,11 +31,14 @@ public struct SourceLocation: Sendable {
   /// instance's ``fileID`` property is `"FoodTruck/WheelTests.swift"`, the
   /// file name is `"WheelTests.swift"`.
   ///
+  /// The structure of file IDs is described in the documentation for the
+  /// [`#fileID`](https://developer.apple.com/documentation/swift/fileID())
+  /// macro in the Swift standard library.
```
stmontgomery, line 36:
```suggestion
  /// The structure of file IDs is described in the documentation for
  /// [`#fileID`](https://developer.apple.com/documentation/swift/fileID())
  /// in the Swift standard library.
```

grynspan:
This is a comments-only change. Merging.

PR #68, opened 2023-10-17
Make `SkipInfo`, `SourceLocation`, `SourceContext`, `Test.ID`, and `Comment` conform to `Codable`

dennisweissmann:
This facilitates progress integrating the testing library with related tools by allowing `SkipInfo`, `SourceLocation`, `SourceContext`, `Test.ID`, and `Comment` (and eventually `Event`s generated by tests to be serialized.

### Motivation:

This is the second in a series of PRs (see #67 for the first one) to make types in the hierarchy of `Event` conform to `Codable`.

### Modifications:

I've used the compiler-generated conformances of `Codable` for these types.

### Result:

`SkipInfo`, `SourceLocation`, `SourceContext`, `Test.ID`, and `Comment` now conform to `Codable`.

Resolves: rdar://117053915

PR #67, opened 2023-10-17
Make `Test.Clock.Instant` conform to `Codable`

dennisweissmann:
This facilitates progress integrating the testing library with related tools by allowing `Test.Clock.Instant` (and eventually `Event`s generated by tests to be serialized.

### Motivation:

This is the first in a series of PRs to make types in the hierarchy of `Event` conform to `Codable`.

### Modifications:

I've added a custom `Codable` implementation for `Test.Clock.Instant` because due the rather complex setup of that type the compiler couldn't generate a default implementation.

I've also added a unit test verifying that what gets encoded is equal to what we gets decoded.

### Result:

`Test.Clock.Instant` now conforms to `Codable`.

Resolves: rdar://117053111

PR #73, opened 2023-10-22
Create a `Expectation.Snapshot` type that conforms to `Codable`

dennisweissmann:
This facilitates progress integrating the testing library with related tools by allowing Expectation (and eventually Events generated by tests to be serialized.

Motivation:

This is the fourth in a series of PRs (#67, #68, #69) to make types in the hierarchy of Event conform to Codable, or create Snapshot types for types that can't / shouldn't become Codable.

Modifications:

I created a Snapshot type that contains copies of all stored properties of `Expectation`.

Result:

Expectation itself does not conform to Codable but there's an equivalent `Expectation.Snapshot` type that is, and will get used in other snapshot types higher up the hierarchy of `Event`.

Resolves: rdar://117054841

### DIFF
--- a/Sources/Testing/Expectations/Expectation.swift
+++ b/Sources/Testing/Expectations/Expectation.swift
@@ -69,9 +69,9 @@ extension Expectation {
   /// A serializable type describing an expectation that has been evaluated.
   @_spi(ExperimentalSnapshotting)
   public struct Snapshot: Sendable, Codable {
-    /// The source code of the expression evaluated by this expectation, if
-    /// available at compile time.
-    public var sourceCode: String?
+    /// The source code description (as a String) of the expression evaluated by
+    /// this expectation, if available at compile time.
+    public var sourceCodeDescription: String?
 
     /// A description of the error mismatch that occurred, if any.
     ///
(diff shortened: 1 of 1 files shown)


dennisweissmann:
@swift-ci please test

Sources/Testing/Expectations/Expectation.swift
```diff
@@ -62,3 +62,65 @@ public struct ExpectationFailedError: Error {
...
+  public struct Snapshot: Sendable, Codable {
+    /// The source code of the expression evaluated by this expectation, if
+    /// available at compile time.
+    public var sourceCode: String?
+
+    /// A description of the error mismatch that occurred, if any.
+    ///
+    /// If this expectation passed, the value of this property is `nil` because no
+    /// error mismatch occurred.
+    @_spi(ExperimentalEventHandling)
```
grynspan, line 80: I'm not sure what to do about the `@_spi` annotations on these stored properties given that this is a snapshot we'd want to immediately encode to JSON or something. @stmontgomery What do you think?

stmontgomery: I'm not too concerned by this, and it matches the `@_spi` declarations of the original properties.

Sources/Testing/Expectations/Expectation.swift
```diff
@@ -62,3 +62,65 @@ public struct ExpectationFailedError: Error {
...
+
+// MARK: - Snapshot
+
+extension Expectation {
+  /// A serializable type describing an expectation that has been evaluated.
+  @_spi(ExperimentalSnapshotting)
+  public struct Snapshot: Sendable, Codable {
+    /// The source code of the expression evaluated by this expectation, if
+    /// available at compile time.
+    public var sourceCode: String?
```
stmontgomery, line 74: We have discussed _eventually_ making our `SourceCode` type conform to `Codable` itself, but we don't feel that type's design is sufficiently nailed down to do that right now. Since it's possible/likely that we will eventually make `SourceCode` codable itself, whenever we do that we'll probably want to use the name `sourceCode` for that property here.

Given that, should we name this `String`-typed property sourceCode**Description**, or something other than plain `sourceCode`?

dennisweissmann:
@swift-ci please test

stmontgomery (approved):

PR #75, opened 2023-10-23
Fix warnings that pointers aren't sendable when using the Swift 5.10 toolchain.

grynspan:
As of Swift 5.10, pointers (e.g. `UnsafeMutableRawPointer`) are no longer considered sendable because the compiler and runtime can't reason about them, so we're starting to get warnings in the error-to-backtrace code where we're using them as keys in a sendable dictionary.

This PR switches to using `ObjectIdentifier` which _is_ sendable. (If that changes in the future, we can switch to `UInt` instead, but strong typing is a good thing in Swift.)


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #79, opened 2023-10-24
Create a `Issue.[Kind.]Snapshot` type that conforms to `Codable`

dennisweissmann:
This facilitates progress integrating the testing library with related tools by allowing Issue information (and eventually Event information) generated by tests to be serialized.

Motivation:

This is the fifth in a series of PRs (#67, #68, #69, #73) to make types in the hierarchy of Event conform to Codable, or create Snapshot types for types that can't / shouldn't become Codable.

Modifications:

I created two snapshot types that contain copies of all stored properties of Issue and Issue.Kind. I've also added a test that encodes and decodes various Issues / IssueSnapshots.

Result:

Issue and Issue.Kind themselves do not conform to Codable but there are equivalent Issue.Snapshot and Issue.Kind.Snapshot types that are.

Resolves: rdar://117054687

### DIFF
--- a/Sources/Testing/Expectations/Expectation.swift
+++ b/Sources/Testing/Expectations/Expectation.swift
@@ -113,7 +113,7 @@ extension Expectation {
 
     /// Creates a snapshot expectation from a real ``Expectation``.
     /// - Parameter expectation: The real expectation.
-    init(expectation: Expectation) {
+    init(snapshotting expectation: Expectation) {
       self.sourceCodeDescription = String(describing: expectation.sourceCode)
       self.mismatchedErrorDescription = expectation.mismatchedErrorDescription
       self.expandedExpressionDescription = expectation.expandedExpressionDescription
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -179,3 +179,225 @@ extension Issue.Kind: CustomStringConvertible {
     }
   }
 }
+
+// MARK: - Codable
+
+extension Issue {
+  /// A serializable type describing a failure or warning which occurred during a test.
+  @_spi(ExperimentalSnapshotting)
+  public struct Snapshot: Sendable, Codable {
+    /// The kind of issue this value represents.
+    public var kind: Kind.Snapshot
+
+    /// Any comments provided by the developer and associated with this issue.
+    ///
+    /// If no comment was supplied when the issue occurred, the value of this
+    /// property is the empty array.
+    public var comments: [Comment]
+
+    /// A ``SourceContext`` indicating where and how this issue occurred.
+    public var sourceContext: SourceContext
+
+    /// Whether or not this issue is known to occur.
+    public var isKnown = false
+
+    /// Initialize an issue instance with the specified details.
+    ///
+    /// - Parameter issue: The original issue that gets snapshotted.
+    init(snapshotting issue: Issue) {
+      self.kind = Issue.Kind.Snapshot(snapshotting: issue.kind)
+      self.comments = issue.comments
+      self.sourceContext = issue.sourceContext
+      self.isKnown = issue.isKnown
+    }
+  }
+}
+
+extension Issue.Kind {
+  /// Serializable kinds of issues which may be recorded.
+  @_spi(ExperimentalSnapshotting)
+  public enum Snapshot: Sendable, Codable {
+    /// An issue which occurred unconditionally, for example by using
+    /// ``Issue/record(_:fileID:filePath:line:column:)``.
+    case unconditional
+
+    /// An issue due to a failed expectation, such as those produced by
+    /// ``expect(_:_:sourceLocation:)``.
+    ///
+    /// - Parameters:
+    ///   - expectation: The expectation that failed.
+    case expectationFailed(_ expectation: Expectation.Snapshot)
+
+    /// An issue due to a confirmation being confirmed the wrong number of
+    /// times.
+    ///
+    /// - Parameters:
+    ///   - actual: The number of times ``Confirmation/confirm(count:)`` was
+    ///     actually called.
+    ///   - expected: The expected number of times
+    ///     ``Confirmation/confirm(count:)`` should have been called.
+    ///
+    /// This issue can occur when calling
+    /// ``confirmation(_:expectedCount:fileID:filePath:line:column:_:)`` when
+    /// the confirmation passed to these functions' `body` closures is confirmed
+    /// too few or too many times.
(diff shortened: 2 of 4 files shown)


Sources/Testing/Issues/Issue.swift
```diff
@@ -179,3 +179,217 @@ extension Issue.Kind: CustomStringConvertible {
...
+    case knownIssueNotRecorded
+
+    /// An issue occurred due to misuse of the testing library.
+    case apiMisused
+
+    /// An issue due to a failure in the underlying system, not due to a failure
+    /// within the tests being run.
+    case system
+
+    init(kind: Issue.Kind) {
```
grynspan, line 275: Simple casting initializers like this one should have unlabelled arguments, so: `init(_ kind: Issue.Kind)`

Sources/Testing/Issues/Issue.swift
```diff
@@ -179,3 +179,217 @@ extension Issue.Kind: CustomStringConvertible {
...
+      case .knownIssueNotRecorded:
+          .knownIssueNotRecorded
+      case .apiMisused:
+          .apiMisused
+      case .system:
+          .system
+      }
+    }
+
+    private enum _CodingKeys: CodingKey {
```
grynspan, line 296: Please document.

Sources/Testing/Issues/Issue.swift
```diff
@@ -179,3 +179,217 @@ extension Issue.Kind: CustomStringConvertible {
...
+    /// This issue can occur when calling
+    /// ``confirmation(_:expectedCount:fileID:filePath:line:column:_:)`` when
+    /// the confirmation passed to these functions' `body` closures is confirmed
+    /// too few or too many times.
+    case confirmationMiscounted(actual: Int, expected: Int)
+
+    /// An issue due to an `Error` being thrown by a test function and caught by
+    /// the testing library.
+    ///
+    /// - Parameters:
```
grynspan, line 249: Please ensure documentation matches code. (Since this documentation was copied verbatim from the original type, we can probably slim it down and don't need as much detail. We can refer readers back to the original type for more info.)

Sources/Testing/Issues/Issue.swift
```diff
@@ -179,3 +179,217 @@ extension Issue.Kind: CustomStringConvertible {
...
 }
+
+// MARK: - Codable
+
+extension Issue {
+  /// A serializable type describing a failure or warning which occurred during a test.
+  @_spi(ExperimentalSnapshotting)
+  public struct Snapshot: Sendable, Codable {
+    /// The kind of issue this value represents.
+    public var kind: Kind.Snapshot
```
grynspan, line 190: In terms of layering and design, it might be better to make `Kind.Snapshot` `Snapshot.Kind`, i.e. it's just the `Kind` of the issue snapshot. 🤔

Sources/Testing/Issues/Issue.swift
```diff
@@ -179,3 +179,217 @@ extension Issue.Kind: CustomStringConvertible {
...
+    /// A ``SourceContext`` indicating where and how this issue occurred.
+    public var sourceContext: SourceContext
+
+    /// Whether or not this issue is known to occur.
+    public var isKnown = false
+
+    /// Initialize an issue instance with the specified details.
+    ///
+    ///
+    init(issue: Issue) {
```
grynspan, line 207: Simple casting initializers like this one should have unlabelled arguments, so: `init(_ issue: Issue)`.

Or, for this and the `Kind.Snapshot` initializer, we could use `snapshotting` as the label so that it's clear it's a destructive conversion. That might be better?

Sources/Testing/Issues/Issue.swift
dennisweissmann: I don't feel strongly but then Snapshot would only be a namespace (empty enum), right? We've used the pattern I've used here for Expectation as well.

grynspan: No, `Issue.Snapshot` would still be a fully-formed nominal type, it's just we wouldn't add a separate `Kind.Snapshot`, it'd be `Kind` under `Issue.Snapshot`.

This is probably not critical to the overall design, but I want to give it a bit of thought either way.

Sources/Testing/Expectations/Expectation.swift
```diff
@@ -113,7 +113,7 @@ extension Expectation {
 
     /// Creates a snapshot expectation from a real ``Expectation``.
     /// - Parameter expectation: The real expectation.
-    init(expectation: Expectation) {
+    init(snapshotting expectation: Expectation) {
```
grynspan, line 116: Okay, if we _are_ using `snapshotting`, we should update `Expectation.snapshot.init(expectation:)` to match.

Sources/Testing/Issues/Issue.swift
dennisweissmann: Oh, I misunderstood, sorry. I don't feel strongly about that either.

dennisweissmann: Actually, I think it makes more sense to me have all "original" types have `Snapshot` types nested, instead of re-creating that same type hierarchy somewhere else.

Sources/Testing/Expectations/Expectation.swift
dennisweissmann: That's what that is? I think?

grynspan: I must have drank decaf this morning… sorry!

dennisweissmann: No problem at all, I was just a bit confused :)

dennisweissmann:
@swift-ci please test

stmontgomery (approved):

Sources/Testing/Issues/Issue.swift
```diff
@@ -179,3 +179,225 @@ extension Issue.Kind: CustomStringConvertible {
...
+        case expected
+      }
+
+      /// The keys used to encode``Issue.Kind.errorCaught``.
+      enum _ErrorCaughtKeys: CodingKey {
+        case error
+      }
+
+      /// The keys used to encode ``Issue.Kind.timeLimitExceeds``.
+      enum _TimeLimitExceededKeys: CodingKey {
```
grynspan, line 327: Can this share code with `Clock.Instant`?

dennisweissmann:
@swift-ci please test

Sources/Testing/Issues/Issue.swift
dennisweissmann: I'll explore that in a follow-up PR.

PR #77, opened 2023-10-23
Add an _ersatz_ filtering mechanism to `XCTestScaffold`.

grynspan:
Because SwiftPM is not (yet) aware of swift-testing, `swift test --filter` does not have any effect on swift-testing tests. This PR adds temporary support for an environment variable, `"SWT_SELECTED_TEST_IDS"`, to stand in for `--filter` until such time as SwiftPM can be modified to support swift-testing.

Note that the supported value of this environment variable is (intentionally) constrained to just test IDs right now. Matching display names or other metadata about tests is certainly possible, but it's beyond the scope of today's PR.

PR #80, opened 2023-10-24
Add the ability to filter tests by tag using an envvar.

grynspan:
This PR is a follow-on to #77. It adds the ability to filter tests to just the ones that contain a given tag using an environment variable. As before, `swift test --filter` is unaware of swift-testing, so we need a temporary solution until such time as `swift test` is taught about us.

To try it out:

```
SWT_SELECTED_TAGS='trait' swift test
```

Will run only those tests with the `"trait"` tag.

If both `"SWT_SELECTED_TAGS""` and `"SWT_SELECTED_TEST_IDS"` are specified, a test must have (at least one of) the given tags _and_ be selected by ID. Whether or not this is the right way to combine these environment variables is, to be polite, probably uninteresting since this is not meant to be a permanent solution.

This change also implements `==` and `hash(into:)` on `Tag` as they were previously synthesized which would cause them to consider the internal `sourceCode` property, which was not intentional.

### DIFF
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -159,16 +159,26 @@ extension Runner.Plan {
     // them, in which case it will be .recordIssue().
     var testGraph = Graph<String, Test?>()
     var actionGraph = Graph<String, Action>(value: .run)
-    for test in tests where _isTestIncluded(test, using: configuration.testFilter) {
+    for test in tests {
       let idComponents = test.id.keyPathRepresentation
       testGraph.insertValue(test, at: idComponents)
       actionGraph.insertValue(.run, at: idComponents, intermediateValue: .run)
     }
 
     // Ensure the trait lists are complete for all nested tests. (Make sure to
-    // do this before we start calling prepare(for:) or we'll miss the
-    // recursively-added ones.)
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

briancroom (approved):

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -159,17 +159,27 @@ extension Runner.Plan {
     // them, in which case it will be .recordIssue().
     var testGraph = Graph<String, Test?>()
     var actionGraph = Graph<String, Action>(value: .run)
-    for test in tests where _isTestIncluded(test, using: configuration.testFilter) {
+    for test in tests {
       let idComponents = test.id.keyPathRepresentation
       testGraph.insertValue(test, at: idComponents)
       actionGraph.insertValue(.run, at: idComponents, intermediateValue: .run)
```
stmontgomery, line 165: Moving the `where ...` clause here will cause the `actionGraph` to always be updated, even if some of those tests are later filtered out. Does something else later on cause the corresponding nodes in that graph to be pruned?

grynspan: Yes, immediately below we prune them out (after applying traits recursively) on line 173.

PR #81, opened 2023-10-24
Add a "main" entry point function for use by SPM.

grynspan:
This PR adds a new symbol, `swiftPMEntryPoint()`, that can be called by Swift Package Manager to trigger running tests. This function shares most of its implementation with `XCTestScaffold`, so that function's implementation has been factored out into a common function both call.


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -0,0 +1,82 @@
...
+/// - Returns: The result of invoking the testing library. The type of this
+///   value is subject to change.
+///
+/// This function examines the command-line arguments to the current process
+/// and then invokes available tests in the current process.
+///
+/// - Warning: This function is used by Swift Package Manager. Do not call it
+///   directly.
+@_spi(SwiftPackageManagerSupport)
+public func __swiftPMEntryPoint() async -> CInt {
```
stmontgomery, line 25: Thus far, all double-underscored `public` APIs I've seen have been for use by macro expanded code. I wonder if this should only have one underscore, to preserve that convention?

Or hmm… maybe it should just have no underscores at all, since it is already `@_spi`? I could understand including underscores with its former name (`__main`) to avoid confusion, but now that it has a more specific name that might no longer be necessary.

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -0,0 +1,82 @@
...
+      $exitCode.withLock { exitCode in
+        exitCode = EXIT_FAILURE
+      }
+    }
+  }
+  return exitCode
+}
+#endif
+
+/// The common implementation of `__main()` and
```
stmontgomery, line 45:
```suggestion
/// The common implementation of `__swiftPMEntryPoint()` and
```
or whatever final name you go with

Sources/Testing/Running/EntryPoint.swift
grynspan: Yeah I can drop the underscores since it's SPI. That's a good point. I didn't have the SPI annotation at first.

Tests/TestingMacrosTests/TestSupport/Scaffolding.swift
stmontgomery, line 1: Is it premature to delete these two scaffold files in this PR? I presume there will be at least a brief window during which we'll be running CI without a toolchain which adopts this, right?

grynspan:
@swift-ci please test

Tests/TestingMacrosTests/TestSupport/Scaffolding.swift
grynspan: Already added one back. We don't need both since they both end up running all tests.

stmontgomery: I thought both files were required because they run the tests within each of our two test targets, respectively?

grynspan: Discussed offline.

stmontgomery (approved):

briancroom (approved):

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -0,0 +1,192 @@
...
+///     the standard error stream.
+func runTests(identifiedBy testIDs: [Test.ID]? = nil, taggedWith tags: Set<Tag>? = nil, eventHandler: @escaping Event.Handler = { _, _ in }) async {
+  let eventRecorder = Event.Recorder(options: .forStandardError) { string in
+    let stderr = swt_stderr()
+    fputs(string, stderr)
+    fflush(stderr)
+  }
+
+  var configuration = Configuration()
+  configuration.isParallelizationEnabled = false
```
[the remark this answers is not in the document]
grynspan: Yes! Baby steps.

PR #83, opened 2023-10-25
Generalize time types.

grynspan:
This PR generalizes the storage of timestamps and durations using a new internal-only `TimeValue` type. This type represents any "seconds and subseconds" value such as a `timespec`, `SuspendingClock.Instant`, `Duration`, etc.

This type allows us to simplify the story for storing, handling, encoding, and decoding time values. Whereas previously we had a lot of conditionalized code like:

```swift
#if SWT_TARGET_OS_APPLE
doSomething(uptime)
#else
doSomethingSimilar(suspending)
#endif
```

We can now generally use the same code path for both.


dennisweissmann (approved):

grynspan:
@swift-ci please test

PR #85, opened 2023-10-27
Allow converting from TimeValue to timespec

stmontgomery:
This extends the `TimeValue` type added in #83 by allowing conversion to `timespec`. It then adopts that in a couple places to ensure consistency.


dennisweissmann (approved):

stmontgomery:
@swift-ci please test

briancroom (approved):

PR #86, opened 2023-10-28
Add the ability for a trait to run code before and after a test

dennisweissmann:
Add the ability for a trait to run code before and after a test.

### Motivation:

As laid out in the [vision document](https://github.com/apple/swift-testing/blob/main/Documentation/Vision.md#trait-extensibility), we want traits to become extensible, and be able to invoke code *before* and *after* a test's body is invoked.

This PR provides the foundation that will not just give test authors a lot more power and flexibility in what they can achieve with traits, but will allow us (in a future PR) to re-implement e.g. the timeout behavior to be less special-cased.

### Modifications:

I added a new protocol `CustomExecutionTrait` which inherits from `Trait`, that has a single additional requirement: an `execute` function that provides a function that eventually executes the test body in its argument so that trait authors can invoke it in whatever way they need, ie. have the ability to execute code before and after they invoke that function.

An example of such a trait is

```swift
struct DoSomethingBeforeAndAfterTrait: CustomExecutionTrait & TestTrait {

  func execute(_ function: () async throws -> Void, for test: Test) async throws {
    something()
    try await function()
    somethingElse()
  }
}
```

(or see `CustomTraitTest`).

### Result:

Test authors now have a much more flexible system to write traits, allowing them to execute custom code before and after a test body is run.

### DIFF
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -114,6 +114,51 @@ extension Runner {
     }
   }
 
+  /// Execute the ``CustomExecutionTrait/execute(_:for:testCase:)`` functions
+  /// associated with the test in a plan step.
+  ///
+  /// - Parameters:
+  ///   - step: The step being performed.
+  ///   - testCase: The test case, if applicable, for which to execute the
+  ///     custom trait.
+  ///   - body: A function to execute from within the
+  ///     ``CustomExecutionTrait/execute(_:for:testCase:)`` functions of each
+  ///     trait applied to `step.test`.
+  ///
+  /// - Throws: Whatever is thrown by `body` or by any of the
+  ///   ``CustomExecutionTrait/execute(_:for:testCase:)`` functions.
+  private func _executeTraits(
+    for step: Plan.Step,
+    testCase: Test.Case?,
+    _ body: @escaping @Sendable () async throws -> Void
+  ) async throws {
+    // If the test does not have any traits, exit early to avoid unnecessary
+    // heap allocations below.
+    if step.test.traits.isEmpty {
+      return try await body()
+    }
+
+    if case .skip = step.action {
+      return try await body()
+    }
+
+    // Construct a recursive function that invokes each trait's ``execute(_:for:testCase:)``
+    // function. The order of the sequence is reversed so that the last trait is
+    // the one that invokes body, then the second-to-last invokes the last, etc.
+    // and ultimately the first trait is the first one to be invoked.
+    let executeAllTraits = step.test.traits.lazy
+      .reversed()
+      .compactMap { $0 as? any CustomExecutionTrait }
+      .compactMap { $0.execute(_:for:testCase:) }
+      .reduce(body) { executeAllTraits, traitExecutor in
+        {
+          try await traitExecutor(executeAllTraits, step.test, testCase)
+        }
+      }
+
+    try await executeAllTraits()
+  }
+
   /// Run this test.
   ///
   /// - Parameters:
@@ -171,10 +216,14 @@ extension Runner {
       }
     }
 
-    if let step = stepGraph.value, case .run = step.action, let testCases = step.test.testCases {
+    if let step = stepGraph.value, case .run = step.action {
       try await Test.withCurrent(step.test) {
         try await _withErrorHandling(for: step, sourceLocation: step.test.sourceLocation) {
-          try await _runTestCases(testCases, within: step)
+          try await _executeTraits(for: step, testCase: nil) {
+            if let testCases = step.test.testCases {
+              try await _runTestCases(testCases, within: step)
+            }
+          }
         }
       }
     }
@@ -257,7 +306,9 @@ extension Runner {
       let sourceLocation = step.test.sourceLocation
       try await _withErrorHandling(for: step, sourceLocation: sourceLocation) {
         try await withTimeLimit(for: step.test, configuration: configuration) {
-          try await testCase.body()
+          try await _executeTraits(for: step, testCase: testCase) {
+            try await testCase.body()
+          }
         } timeoutHandler: { timeLimit in
           Issue.record(
             .timeLimitExceeded(timeLimitComponents: timeLimit),
--- a/Sources/Testing/Traits/Trait.swift
+++ b/Sources/Testing/Traits/Trait.swift
@@ -98,3 +98,38 @@ extension SuiteTrait {
     false
   }
 }
+
+/// A protocol extending ``Trait`` that offers an additional customization point
+/// for trait authors to execute code before and after each test function (if
+/// added to the traits of a test function), or before and after each test suite
+/// (if added to the traits of a test suite).
+@_spi(ExperimentalTraits)
+public protocol CustomExecutionTrait: Trait {
+
+  /// Execute a function with the effects of this trait applied.
+  ///
+  /// - Parameters:
+  ///   - function: The function to perform. If `test` represents a test suite,
+  ///     this function encapsulates running all the tests in that suite. If
+  ///     `test` represents a test function, this function is the body of that
+  ///     test function (including all cases if it is parameterized.)
+  ///   - test: The test under which `function` is being performed.
+  ///   - testCase: The test case, if any, under which `function` is being
+  ///     performed. This is `nil` when invoked on a suite.
+  ///
+  /// - Throws: Whatever is thrown by `function`, or an error preventing the
+  ///   trait from running correctly.
+  ///
+  /// This function is called for each ``CustomExecutionTrait`` on a test suite
+  /// or test function and allows additional work to be performed before and
+  /// after the test runs.
+  ///
+  /// This function is invoked once for the test it is applied to, and then once
+  /// for each test case in that test, if applicable.
+  ///
+  /// Issues recorded by this function are recorded against `test`.
+  ///
+  /// - Note: If a test function or test suite is skipped, this function does
+  ///   not get invoked by the runner.
+  @Sendable func execute(_ function: @escaping @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws
+}
--- a/Tests/TestingTests/Traits/CustomExecutionTraitTests.swift
+++ b/Tests/TestingTests/Traits/CustomExecutionTraitTests.swift
@@ -1,0 +1,78 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+@testable @_spi(ExperimentalTestRunning) @_spi(ExperimentalEventHandling) @_spi(ExperimentalTraits) import Testing
+
+private struct CustomTrait: CustomExecutionTrait, TestTrait {
+    var before: Confirmation
+    var after: Confirmation
+    func execute(_ function: @escaping @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws {
+        before()
+        defer {
+            after()
+        }
+        try await function()
+    }
+}
+
+private struct CustomThrowingErrorTrait: CustomExecutionTrait, TestTrait {
+    fileprivate struct CustomTraitError: Error {}
+
+    func execute(_ function: @escaping @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws {
+        throw CustomTraitError()
+    }
+}
+
+struct CustomTraitTest {
+    @Test("Execute code before and after a non-parameterized test.")
+    func executeCodeBeforeAndAfterNonParameterizedTest() async {
+        // `expectedCount` is 2 because we run it both for the test and the test case
+        await confirmation("Code was run before the test", expectedCount: 2) { before in
+            await confirmation("Code was run after the test", expectedCount: 2) { after in
+                await Test(CustomTrait(before: before, after: after)) {
+                    // do nothing
+                }.run()
+            }
+        }
+    }
+
+    @Test("Execute code before and after a parameterized test.")
+    func executeCodeBeforeAndAfterParameterizedTest() async {
+        // `expectedCount` is 3 because we run it both for the test and each test case
+        await confirmation("Code was run before the test", expectedCount: 3) { before in
+            await confirmation("Code was run after the test", expectedCount: 3) { after in
+                await Test(CustomTrait(before: before, after: after), arguments: ["Hello", "World"]) { _ in
+                    // do nothing
+                }.run()
+            }
+        }
+    }
+
+    @Test("Custom execution trait throws an error")
+    func customExecutionTraitThrowsAnError() async throws {
+        var configuration = Configuration()
+        await confirmation("Error thrown", expectedCount: 1) { errorThrownConfirmation in
+            configuration.eventHandler = { event, _ in
+                guard case let .issueRecorded(issue) = event.kind,
+                      case let .errorCaught(error) = issue.kind else {
+                    return
+                }
+
+                #expect(error is CustomThrowingErrorTrait.CustomTraitError)
+                errorThrownConfirmation()
+            }
+
+            await Test(CustomThrowingErrorTrait()) {
+                // Make sure this does not get reached
+                Issue.record("Expected trait to fail the test. Should not have reached test body.")
+            }.run(configuration: configuration)
+        }
+    }
+}


dennisweissmann:
@swift-ci please test

Sources/Testing/Running/Runner.swift
```diff
@@ -114,6 +114,54 @@ extension Runner {
...
+    guard let test = step?.test else {
+      return try await body()
+    }
+
+    if case .skip = step?.action {
+      return try await body()
+    }
+
+    // Construct a recursive function that invokes each trait's ``execute(_:for:)``
+    // function. The order of the sequence is reversed so that the last trait is
```
stmontgomery, line 147: I'm curious about the behavior of **reversing** the traits here and whether that's the right decision. I think it will also mean that traits applied to containing suite types and inherited would also have their execute methods invoked _after_ traits directly applied to a test function, and conceptually that feels a bit wrong to me. Does anyone have usage examples in mind for the ordering?

(Either way, we should document what we decide here in the DocC docs for this protocol method.)

Sources/Testing/Traits/Trait.swift
```diff
@@ -64,6 +64,45 @@ public protocol SuiteTrait: Trait {
...
+  /// This method is called for each trait on a test suite or test function and
+  /// allows additional work to be performed before or after the test runs.
+  /// Issues recorded by this function are recorded against `test`.
+  ///
+  /// - Note: If a test function or test suite is skipped, this function does
+  ///   not get invoked by the runner.
+  ///
+  /// The default implementation of this function invokes `function` and
+  /// returns its result or rethrows any error it throws.
+  func execute(_ function: () async throws -> Void, for test: Test) async throws
```
stmontgomery, line 94: The API being proposed here will only be run once for each `Test`, and not once for each `Test.Case`, i.e. not once for every combination of arguments passed to a **parameterized** test.

I think it's very possible/likely some use cases will require performing the trait's custom execution logic once per test _case_, so we should probably figure out some way to support that via the API. I wonder if adding one additional parameter `testCase: Test.Case?` would accomplish that? (And adjusting the `Runner` implementation accordingly.)

stmontgomery, line 94: Should we include `@Sendable` on the `function` parameter?

grynspan: Adding a `Test.Case` parameter necessitates making this protocol SPI.

Sources/Testing/Running/Runner.swift
grynspan: Reversing the order means we invoke them in the order they're specified. It's counter-intuitive but it's a side effect of the nested calls. See the comment here.

Sources/Testing/Traits/Trait.swift
grynspan: Possibly `@escaping` too, since the trait implementation might need to pass it around.

Tests/TestingTests/Traits/CustomTraitTests.swift
```diff
@@ -16,3 +16,40 @@ import Testing
...
+  }
+
+  fileprivate func setCounter(_ newValue: Int) {
+    _counter = newValue
+  }
+}
+
+private let sharedState = SharedState()
+
+private struct CustomCountTrait: CustomExecutionTrait & TestTrait {
```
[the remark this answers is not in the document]
grynspan: For those watching at home: this is a style change only and `&` and `,` have the same effect in this position.

Sources/Testing/Traits/Trait.swift
stmontgomery: #191 has now been merged, elevating `Test.Case` to public API (removing `@_spi`), so I think can now safely add a `testCase: Test.Case?` parameter to this method.

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,43 @@ public protocol SuiteTrait: Trait {
...
+  ///     this function encapsulates running all the tests in that suite. If
+  ///     `test` represents a test function, this function is the body of that
+  ///     test function (including all cases if it is parameterized.)
+  ///   - test: The test under which `function` is being performed.
+  ///   - testCase: The testCase under which `function` is being performed.
+  ///
+  /// - Throws: Whatever is thrown by `function`, or an error preventing the
+  ///   trait from running correctly.
+  ///
+  /// This method is called for each trait on a test suite or test function and
```
grynspan, line 103: Let's make sure this comment documents when the function is called: once per test it is applied to, then once per test case for that test (if applicable.)

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,43 @@ public protocol SuiteTrait: Trait {
...
+  /// - Note: If a test function or test suite is skipped, this function does
+  ///   not get invoked by the runner.
+  ///
+  /// The default implementation of this function invokes `function` and
+  /// returns its result or rethrows any error it throws.
+  @Sendable func execute(_ function: @escaping @Sendable () async throws -> Void, test: Test, testCase: Test.Case?) async throws
+}
+
+extension CustomExecutionTrait {
+  func execute(_ function: @escaping @Sendable () async throws -> Void, test: Test, testCase: Test.Case?) async throws {
```
stmontgomery, line 116: I'm not so sure we need a default implementation for this method. Since this requirement is within a protocol dedicated to this purpose, and not part of the `Trait` protocol directly, any type which conforms to this protocol ought to be required to explicitly implement the function.

Note, though: If we _did_ decide to include this default implementation, we would need to mark this extension method `public` otherwise clients won't have access to it. (You might not see this in unit tests since they typically use `@testable import`.)

Sources/Testing/Running/Runner.swift
```diff
@@ -114,6 +114,50 @@ extension Runner {
...
+    for step: Plan.Step,
+    testCase: Test.Case?,
+    _ body: @escaping @Sendable () async throws -> Void
+  ) async throws {
+    if step.test.traits.isEmpty {
+      return try await body()
+    }
+
+    if case .skip = step.action {
+      return try await body()
```
grynspan, line 141: This `if` basically says "don't run `execute(_:test:testCase:)` if the test is to be skipped, which we should document in the documentation for `CustomExecutionTrait`.

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,43 @@ public protocol SuiteTrait: Trait {
...
+  /// This method is called for each trait on a test suite or test function and
+  /// allows additional work to be performed before or after the test runs.
+  /// Issues recorded by this function are recorded against `test`.
+  ///
+  /// - Note: If a test function or test suite is skipped, this function does
+  ///   not get invoked by the runner.
+  ///
+  /// The default implementation of this function invokes `function` and
+  /// returns its result or rethrows any error it throws.
+  @Sendable func execute(_ function: @escaping @Sendable () async throws -> Void, test: Test, testCase: Test.Case?) async throws
```
grynspan, line 112: In a previous iteration, the label for the test was `for` (i.e. `for test: Test`). I _think_ that's consistent with what we do elsewhere? Mind sanity-checking? Either label is fine here, but I like consistency. :)

Sources/Testing/Traits/Trait.swift
grynspan: This is a good point—we can remove this default implementation.

Tests/TestingTests/Traits/CustomTraitTests.swift
```diff
@@ -8,11 +8,75 @@
...
-// specifically intended to validate that a type conforming to `TestTrait` can
-// be declared without importing anything except the base.
-import Testing
+@testable @_spi(ExperimentalTestRunning) @_spi(ExperimentalEventHandling) import Testing
 
 // This is a "build-only" test which simply validates that we can successfully
 // declare a type conforming to `TestTrait` and use its default implementations.
 struct MyCustomTrait: TestTrait {}
+
+private struct CustomCountTrait: CustomExecutionTrait, TestTrait {
```
grynspan, line 17: Since this trait no longer _counts_ anything, let's rename it to something more appropriate. Since it's in a test, it doesn't need to be perfect API, but it should at least have a name that's internally consistent. :)

grynspan (changes requested):
Most of my comments are just about documentation and naming. I think the overall shape of the PR is good.

dennisweissmann:
@swift-ci please test

Sources/Testing/Running/Runner.swift
```diff
@@ -114,6 +114,52 @@ extension Runner {
...
+  ///   - step: The step being performed. If this argument is `nil`, `body` is
+  ///     immediately executed.
+  ///   - testCase: The test case, if applicable, for which to execute the
+  ///               custom trait.
+  ///   - body: A function to execute from within the
+  ///           ``CustomExecutionTrait/execute(_:for:testCase:)`` functions of
+  ///           each trait applied to `step.test`.
+  ///
+  /// - Throws: Whatever is thrown by `body` or by any of the
+  ///   `       `CustomExecutionTrait/execute(_:for:testCase:)`` functions.
```
stmontgomery, line 130: The documentation formatting looks off here with these backticks

Sources/Testing/Running/Runner.swift
```diff
@@ -114,6 +114,52 @@ extension Runner {
     }
   }
 
+  /// Execute the ``CustomExecutionTrait/execute(_:for:testCase:)`` functions
+  /// associated with the test in a plan step.
+  ///
+  /// - Parameters:
+  ///   - step: The step being performed. If this argument is `nil`, `body` is
+  ///     immediately executed.
```
stmontgomery, line 122: The `step` parameter is non-Optional, so should this description be updated?

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,41 @@ public protocol SuiteTrait: Trait {
...
+  ///
+  /// Issues recorded by this function are recorded against `test`. It is _not_
+  /// getting called for skipped tests.
+  ///
+  /// - Note: If a test function or test suite is skipped, this function does
+  ///   not get invoked by the runner.
+  ///
+  /// The default implementation of this function invokes `function` and
+  /// returns its result or rethrows any error it throws.
+  @Sendable func execute(_ function: @escaping @Sendable () async throws -> Void, for: Test, testCase: Test.Case?) async throws
```
stmontgomery, line 116: The `for:` parameter should include the `test` as its argument name, both for clarity and since the argument name is referred in the DocC above
```suggestion
  @Sendable func execute(_ function: @escaping @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws
```

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,41 @@ public protocol SuiteTrait: Trait {
...
+  /// - Throws: Whatever is thrown by `function`, or an error preventing the
+  ///   trait from running correctly.
+  ///
+  /// This function is called for each ``CustomExecutionTrait`` on a test suite
+  /// or test function and allows additional work to be performed before or
+  /// after the test runs.
+  /// It gets invoked once for each test, and then once for each test case.
+  ///
+  /// Issues recorded by this function are recorded against `test`. It is _not_
+  /// getting called for skipped tests.
```
stmontgomery, line 109: Wording nit
```suggestion
  /// Issues recorded by this function are recorded against `test`. This function
  /// will not be called for skipped tests.
```

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,41 @@ public protocol SuiteTrait: Trait {
...
+  ///   - test: The test under which `function` is being performed.
+  ///   - testCase: The testCase under which `function` is being performed.
+  ///
+  /// - Throws: Whatever is thrown by `function`, or an error preventing the
+  ///   trait from running correctly.
+  ///
+  /// This function is called for each ``CustomExecutionTrait`` on a test suite
+  /// or test function and allows additional work to be performed before or
+  /// after the test runs.
+  /// It gets invoked once for each test, and then once for each test case.
```
grynspan, line 106:
```suggestion
  /// This function is invoked once for the test it is applied to, and then once for each test case in that test, if applicable.
```

Did you mean for this to be a separate paragraph or a continuation of the previous one?

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,41 @@ public protocol SuiteTrait: Trait {
...
+
+  /// Execute a function with the effects of this trait applied.
+  ///
+  /// - Parameters:
+  ///   - function: The function to perform. If `test` represents a test suite,
+  ///     this function encapsulates running all the tests in that suite. If
+  ///     `test` represents a test function, this function is the body of that
+  ///     test function (including all cases if it is parameterized.)
+  ///   - test: The test under which `function` is being performed.
+  ///   - testCase: The testCase under which `function` is being performed.
```
stmontgomery, line 98: Can the wording about the `testCase` parameter include "if any", to clarify that it's Optional? And then add a sentence explaining what circumstances it would be nil (a suite)

Tests/TestingTests/Traits/CustomTraitTests.swift
```diff
@@ -8,11 +8,75 @@
...
+@testable @_spi(ExperimentalTestRunning) @_spi(ExperimentalEventHandling) import Testing
 
 // This is a "build-only" test which simply validates that we can successfully
 // declare a type conforming to `TestTrait` and use its default implementations.
 struct MyCustomTrait: TestTrait {}
+
+private struct CustomTrait: CustomExecutionTrait, TestTrait {
+  var before: Confirmation
+  var after: Confirmation
+  func execute(_ function: @escaping @Sendable () async throws -> Void, for: Test, testCase: Test.Case?) async throws {
```
grynspan, line 20:
```suggestion
  func execute(_ function: @escaping @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws {
```

Tests/TestingTests/Traits/CustomTraitTests.swift
```diff
@@ -8,11 +8,75 @@
...
+      after()
+    }
+    try await function()
+  }
+}
+
+private struct CustomThrowingErrorTrait: CustomExecutionTrait, TestTrait {
+  fileprivate struct CustomTraitError: Error {}
+
+  func execute(_ function: @escaping @Sendable () async throws -> Void, for: Test, testCase: Test.Case?) async throws {
```
grynspan, line 32:
```suggestion
  func execute(_ function: @escaping @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws {
```

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,41 @@ public protocol SuiteTrait: Trait {
...
+  /// It gets invoked once for each test, and then once for each test case.
+  ///
+  /// Issues recorded by this function are recorded against `test`. It is _not_
+  /// getting called for skipped tests.
+  ///
+  /// - Note: If a test function or test suite is skipped, this function does
+  ///   not get invoked by the runner.
+  ///
+  /// The default implementation of this function invokes `function` and
+  /// returns its result or rethrows any error it throws.
```
stmontgomery, line 115: We can delete this paragraph since there is no longer a default implementation provided

Sources/Testing/Traits/Trait.swift
grynspan, line 109:
```suggestion
  /// called for skipped tests.
```

But also, these two sentences are not related. I think the new comment about skipped tests belongs in the previous paragraph?

Sources/Testing/Running/Runner.swift
```diff
@@ -114,6 +114,51 @@ extension Runner {
     }
   }
 
+  /// Execute the ``CustomExecutionTrait/execute(_:for:testCase:)`` functions
+  /// associated with the test in a plan step.
+  ///
+  /// - Parameters:
+  ///   - step: The step being performed.
+  ///   - testCase: The test case, if applicable, for which to execute the
```
grynspan, line 122: Nitpick: the whitespace in the documentation for this symbol does not match house style. (Gotta love comments about whitespace.)

Tests/TestingTests/Traits/CustomTraitTests.swift
```diff
@@ -8,11 +8,75 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-// Note: Do NOT include `@_spi` on this import. The contents of this file are
-// specifically intended to validate that a type conforming to `TestTrait` can
-// be declared without importing anything except the base.
-import Testing
+@testable @_spi(ExperimentalTestRunning) @_spi(ExperimentalEventHandling) import Testing
```
stmontgomery, line 11: I had this comment here with the goal that we would maintain a test file which does _not_ include any `@_spi` attributes on an import of `Testing` module, because that's important to validate that a custom `Trait` type can be declared without needing SPIs. We had a bug in the past about this, and I'd like to ensure we don't lose our ability to test that.

Can we move your newly-added test to a distinct file, so that this `@_spi`-less import can be preserved

Sources/Testing/Traits/Trait.swift
grynspan, line 98:
```suggestion
  ///   - testCase: The test case under which `function` is being performed.
```

FranzBusch:
This seems really cool. I am wondering if we can also provide something that can run a test in a `withXXX` style method. We see those more and more in structured concurrency and they don't fit into the before + after pattern.

stmontgomery:
> I am wondering if we can also provide something that can run a test in a `withXXX` style method. We see those more and more in structured concurrency and they don't fit into the before + after pattern.

This new API should support that use case, if I understand you correctly. Or if you don't think it will, can you elaborate?

Sources/Testing/Traits/Trait.swift
dennisweissmann: Ah yes!

Sources/Testing/Traits/Trait.swift
dennisweissmann: A new paragraph, I inserted another newline.

Sources/Testing/Traits/Trait.swift
dennisweissmann: Absolutely, thanks for catching that!

Sources/Testing/Traits/Trait.swift
dennisweissmann: I removed that sentence since the Note below says pretty much the same thing.

Sources/Testing/Running/Runner.swift
dennisweissmann: Nah you're good, I very much like being consistent, sorry for not comparing against the other styles myself!

Tests/TestingTests/Traits/CustomTraitTests.swift
dennisweissmann: Oh, this PR has been open for too long - I wrongly assumed that I had added it and wasn't sure if it would be a valuable addition but yes, I'll re-add it and move my tests into another file.

dennisweissmann:
@swift-ci please test

Sources/Testing/Traits/Trait.swift
```diff
@@ -81,6 +81,40 @@ public protocol SuiteTrait: Trait {
   var isRecursive: Bool { get }
 }
 
+/// A protocol extending ``Trait`` that offers an additional customization point
```
grynspan, line 84: I want to do an editorial pass on this type's documentation after we promote it to API, but it's fine at this point.

grynspan (approved):

briancroom (approved):

dennisweissmann:
@swift-ci please test

PR #87, opened 2023-10-29
Adopt SE-0409 (Access Level-on-Imports)

stmontgomery:
This adopts [SE-0409 Access-level modifiers on import declarations](https://github.com/apple/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md) in this package.

### Motivation:

We are currently using `@_implementationOnly` to ensure imports of our internal `TestingInternals` module are not exposed to clients, but it produces warnings when building without library evolution enabled.

### Modifications:

- Enable the experimental `AccessLevelOnImport` feature. (It's technically no longer experimental since the feature was approved, but including this attempts to preserve compatibility with older toolchains.)
- Enable the upcoming `InternalImportsByDefault` feature.
- Update various `import` declarations as appropriate.
- Remove `@_implementationOnly` usages from test targets where they are unnecessary.

Resolves rdar://115311986

### DIFF
--- a/Sources/Testing/Events/Clock.swift
+++ b/Sources/Testing/Events/Clock.swift
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-@_implementationOnly import TestingInternals
+private import TestingInternals
 
 @_spi(ExperimentalEventHandling)
 extension Test {
--- a/Sources/Testing/Events/TimeValue.swift
+++ b/Sources/Testing/Events/TimeValue.swift
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-@_implementationOnly import TestingInternals
+internal import TestingInternals
 
 /// A container type representing a time value that is suitable for storage,
 /// conversion, encoding, and decoding.
--- a/Sources/Testing/Events/TimeValue.swift
+++ b/Sources/Testing/Events/TimeValue.swift
@@ -8,6 +8,8 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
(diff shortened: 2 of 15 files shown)


stmontgomery:
@swift-ci please test

dennisweissmann (approved):

Sources/Testing/Events/TimeValue.swift
```diff
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-@_implementationOnly import TestingInternals
+internal import TestingInternals
```
grynspan, line 11: If imports are `internal` by default, do we need this explicit `internal`? (And any others.)

Sources/Testing/Events/Clock.swift
```diff
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-@_implementationOnly import TestingInternals
+private import TestingInternals
```
grynspan, line 11: It's probably fine to always `internal` import TestingInternals.

Sources/Testing/Events/TimeValue.swift
stmontgomery: I was open to taking this suggestion, but just tried it and realized the compiler requires it:

> error: ambiguous implicit access level for import of 'TestingInternals'; it is imported as 'private' elsewhere

Sources/Testing/Events/Clock.swift
stmontgomery: I like having `private` on those which are eligible, as a statement of intent, mainly. I feel like clarifying the role that the `TestingInternals` (or other modules) play can be useful to understand the effect of an `import` within a particular file

stmontgomery:
@swift-ci please test

grynspan (approved):

PR #88, opened 2023-10-31
Warn when using `XCTestScaffold` when SwiftPM supports swift-testing.

grynspan:
This PR introduces a new compile-time check for `SWIFT_PM_SUPPORTS_SWIFT_TESTING` around `XCTestScaffold.runTests(hostedBy:)`. If that compiler condition is set, then `XCTestScaffold.runTests(hostedBy:)` is deprecated and calling it emits a warning message but does not run any tests.

Right now, nothing sets that condition, so this is dead code, but the expectation here is that SwiftPM _will_ set it in a future update, at which point developers who opt into experimental support for swift-testing won't end up running their tests multiple times.

Since `XCTestScaffold` is intended to be transient code, and since this code path is currently unreachable, I have not included new unit tests. Code has been tested at-desk with a branch of SwiftPM that sets the aforementioned condition.

Examples of new output from calling the function:

### macOS

Screenshot 2023-10-31 at 10 33 24 AM

### Ubuntu

Screenshot 2023-10-31 at 10 43 59 AM

### DIFF
--- a/Sources/Testing/Running/XCTestScaffold.swift
+++ b/Sources/Testing/Running/XCTestScaffold.swift
@@ -167,7 +167,20 @@ public enum XCTestScaffold: Sendable {
   /// ## See Also
   ///
   /// - <doc:TemporaryGettingStarted>
+#if SWIFT_PM_SUPPORTS_SWIFT_TESTING
+  @available(*, deprecated, message: "This version of Swift Package Manager supports running swift-testing tests directly.")
+#endif
   public static func runAllTests(hostedBy testCase: XCTestCase) async {
+#if SWIFT_PM_SUPPORTS_SWIFT_TESTING
+    let message = warning("This version of Swift Package Manager supports running swift-testing tests directly.", options: .forStandardError)
+#if SWT_TARGET_OS_APPLE
+    let stderr = swt_stderr()
+    fputs(message, stderr)
+    fflush(stderr)
+#else
+    print(message)
+#endif
+#else
     let testCase = UncheckedSendable(rawValue: testCase)
 #if SWT_TARGET_OS_APPLE
     let isProcessLaunchedByXcode = Environment.variable(named: "XCTestSessionIdentifier") != nil
@@ -218,6 +231,7 @@ public enum XCTestScaffold: Sendable {
       }
 #endif
     }
+#endif
   }
 }
 #endif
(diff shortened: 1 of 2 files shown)


Sources/Testing/Running/XCTestScaffold.swift
```diff
@@ -167,7 +167,20 @@ public enum XCTestScaffold: Sendable {
   /// ## See Also
   ///
   /// - <doc:TemporaryGettingStarted>
+#if SWIFT_PM_SUPPORTS_SWIFT_TESTING
+  @available(*, deprecated, message: "This version of Swift Package Manager supports running swift-testing tests directly.")
+#endif
   public static func runAllTests(hostedBy testCase: XCTestCase) async {
+#if SWIFT_PM_SUPPORTS_SWIFT_TESTING
+    let message = warning("This version of Swift Package Manager supports running swift-testing tests directly.", options: .forStandardError)
```
stmontgomery, line 175: Can the message also state that because SwiftPM has support built-in, `XCTestScaffold` itself will _not_ run tests? Just to clarify that SwiftPM will be taking precedence and explain the effect of this warning.

Also, less important but maybe something along the lines of "Usage of `XCTestScaffold` can safely be removed."?

grynspan: If you have specific phrasing, I'll take it.

I am leaving out the "remove this" part for now because this is going to be fragile between toolchain versions. I'll add it when we get to a point where we can assume everybody has a recent-enough toolchain.

grynspan:
@swift-ci please test

Sources/Testing/Running/XCTestScaffold.swift
stmontgomery:
```suggestion
    let message = warning("Ignoring XCTestScaffold, since this version of Swift Package Manager runs tests directly.", options: .forStandardError)
```

One thing I like about this phrasing is that it not only says SwiftPM _supports_ running tests, but actually confirms that it _will_ run them (or perhaps already has, by this point). In other words it avoids confusion about whether the user needs to adopt something to enable the new SwiftPM behavior; they get it for free.

grynspan: How about:

> This version of Swift Package Manager supports running swift-testing tests directly. Ignoring XCTestScaffold.runTests(hostedBy:).

grynspan:
@swift-ci please test

stmontgomery (approved):

PR #90, opened 2023-10-31
Remove the use of `std::span`.

grynspan:
This PR switches from `std::span` to raw array iteration. This change is necessary because swift-ci's C++ stdlib is somewhat older and does not include this type when building on Linux. See: https://ci.swift.org/job/swift-package-manager-self-hosted-Linux-smoke-test/2992/consoleText

Resolves rdar://117215039.


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #91, opened 2023-10-31
Do not rely on `constinit std::vector`.

grynspan:
This PR removes our reliance on `constinit std::vector`. This is a STL feature available in recent C++ versions, but it is not available in older macOS SDKs and has caused at least one CI failure.

Instead, just do the ol' `dispatch_once` dance.

Resolves rdar://117215036.


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #93, opened 2023-11-01
Remove `setenv()` interface and `Configuration`'s environment variables.

grynspan:
This PR removes two chunks of environment-variable-based code:

1. An interface for setting environment variables (`setenv()`). The testing library does not ever need to set environment variables; only the testing library's own tests ever do so.
2. All environment variables used by `Configuration` to specify defaults. The codebase is now mature enough to allow a caller to set up an instance of `Configuration` directly and these environment variables are no longer needed.

This PR also removes tests that test the above code. It adds tests for `Environment` (since there _are_ still some environment variables we read) that test its functionality directly.


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

briancroom (approved):

stmontgomery (approved):

PR #95, opened 2023-11-01
Fix a crash on Linux when invoking unavailable code.

grynspan:
On at least some toolchains, we crash on Linux when touching a synthesized test container type. This is happening because the types are annotated `unavailable`, and we end up calling into the Swift runtime function `_diagnoseUnavailableCodeReached()`. See https://ci.swift.org/job/swift-package-manager-self-hosted-Linux-smoke-test/3000/consoleText

The types are marked `unavailable` so they can't be used accidentally, but they have fairly esoteric names, so we can just remove the annotation and we should be safe.

Darwin is not affected, so the change is conditional on platform.


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #98, opened 2023-11-06
Remove a dependency on XCTest.

grynspan:
This PR removes a static dependency on XCTest in swift-testing. The dependency exists because a single Swift function cannot be _both_ an XCTest test method _and_ a swift-testing test function, so we have a runtime check in our generated thunk that records an issue if a developer manages to produce such code.

This change removes a reference to `XCTestCase` on non-Darwin platforms. Instead, on those platforms we dynamically look up the `XCTest.XCTestCase` class by name at runtime (similar to how we find it on Darwin.)

The generic function that performs this check is now generic over `AnyObject` rather than `XCTestCase`. There is a nominal runtime cost to taking this code path for all reference types, however it is not expected to be a problem except for pathologically deep class hierarchies (which we'd expect would suffer performance issues elsewhere in Swift/the runtime anyway.)

### DIFF
--- a/Sources/Testing/Support/Additions/TypeAdditions.swift
+++ b/Sources/Testing/Support/Additions/TypeAdditions.swift
@@ -67,3 +67,22 @@ func isImportedFromC(_ type: Any.Type) -> Bool {
   let endIndex = mangledTypeName.index(mangledTypeName.startIndex, offsetBy: 2)
   return mangledTypeName[..<endIndex] == "So"
 }
+
+/// Check if a class is a subclass (or equal to) another class.
+///
+/// - Parameters:
+///   - subclass: The (possible) subclass to check.
+///   - superclass The (possible) superclass to check.
+///
+/// - Returns: Whether `subclass` is a subclass of, or is equal to,
+///   `superclass`.
+func isClass(_ subclass: AnyClass, subclassOf superclass: AnyClass) -> Bool {
+  if subclass == superclass {
+    true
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

dennisweissmann (approved):

Sources/Testing/Support/Additions/TypeAdditions.swift
```diff
@@ -67,3 +67,22 @@ func isImportedFromC(_ type: Any.Type) -> Bool {
...
+
+/// Check if a class is a subclass (or equal to) another class.
+///
+/// - Parameters:
+///   - subclass: The (possible) subclass to check.
+///   - superclass The (possible) superclass to check.
+///
+/// - Returns: Whether `subclass` is a subclass of, or is equal to,
+///   `superclass`.
+func isClass(_ subclass: AnyClass, subclassOf superclass: AnyClass) -> Bool {
```
stmontgomery, line 79: Feels like it'd read much more nicely as an extension on `AnyClass` like `func isSubclass(of superclass: Self) -> Bool`

stmontgomery (approved):
Thanks! Good to shed this dependency

Sources/Testing/Support/Additions/TypeAdditions.swift
grynspan: Would be nice. :)

grynspan:
@swift-ci please test

PR #100, opened 2023-11-08
Add `includeHiddenTests` parameter to `setTestFilter(toMatch:)` APIs.

ejvaughan:
### Motivation:

When constructing a `Configuration` instance, the **public** `setTestFilter(toMatch:)` API can be used to specify a particular set of test IDs that should be executed. Today, the implementation includes tests that are annotated with the internal `.hidden` trait. This PR changes the implementation to instead _exclude_ these hidden tests.

### Modifications:

I added an internal variant of the `setTestFilter(toMatch:)` API that has an additional parameter, `includeHiddenTests`. If `false` is passed as an argument to this parameter, a test annotated with the `.hidden` trait will _not_ be included, even if its ID is present in the set of IDs (or if it's a hidden descendant of a non-hidden suite whose ID is present in the set).

### Result:

I added a unit test to verify this new functionality.

### DIFF
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -128,17 +128,38 @@ public struct Configuration: Sendable {
   ///
   /// By default, all tests are run and no filter is set.
   public mutating func setTestFilter(toMatch selection: Set<Test.ID>?) {
-    setTestFilter(toMatch: selection.map(Test.ID.Selection.init))
+    setTestFilter(toMatch: selection, includeHiddenTests: false)
   }
-  
+
+    /// The granularity to enforce test filtering.
+    /// - Parameters:
+    ///   - selection: A set of test IDs to be filtered. If `nil`, the current
+    ///     selection is cleared.
+    ///   - includeHiddenTests: If false, a test annotated with the `.hidden` trait will not be included, even if its ID is present in `selection`.
+    ///
+    /// By default, all tests are run and no filter is set.
+    mutating func setTestFilter(toMatch selection: Set<Test.ID>?, includeHiddenTests: Bool) {
+      setTestFilter(toMatch: selection.map(Test.ID.Selection.init), includeHiddenTests: includeHiddenTests)
+    }
+
   /// The granularity to enforce test filtering.
   ///
   /// - Parameters:
   ///   - selection: A selection of test IDs to be filtered. If `nil`, the
   ///     current selection is cleared.
+  ///   - includeHiddenTests: If false, a test annotated with the `.hidden` trait will not be included, even if its ID is present in `selection`.
   ///
   /// By default, all tests are run and no filter is set.
-  mutating func setTestFilter(toMatch selection: Test.ID.Selection?) {
-    testFilter = selection.map { $0.contains }
+  mutating func setTestFilter(toMatch selection: Test.ID.Selection?, includeHiddenTests: Bool) {
+    if let selection {
+      testFilter = { test in
+        if includeHiddenTests || !test.isHidden {
+          return selection.contains(test)
+        }
+        return false
+      }
+    } else {
+      testFilter = nil
+    }
   }
 }
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -131,16 +131,16 @@ public struct Configuration: Sendable {
     setTestFilter(toMatch: selection, includeHiddenTests: false)
   }
 
-    /// The granularity to enforce test filtering.
-    /// - Parameters:
-    ///   - selection: A set of test IDs to be filtered. If `nil`, the current
-    ///     selection is cleared.
-    ///   - includeHiddenTests: If false, a test annotated with the `.hidden` trait will not be included, even if its ID is present in `selection`.
-    ///
-    /// By default, all tests are run and no filter is set.
-    mutating func setTestFilter(toMatch selection: Set<Test.ID>?, includeHiddenTests: Bool) {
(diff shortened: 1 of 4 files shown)


Tests/TestingTests/RunnerTests.swift
```diff
@@ -316,6 +316,24 @@ final class RunnerTests: XCTestCase {
...
+      Test.ID(type: S.self).child(named: "f()")
+    ]
+
+    var configuration = Configuration()
+    configuration.setTestFilter(toMatch: selectedTestIDs, includeHiddenTests: false)
+
+    let runner = await Runner(configuration: configuration)
+    let plan = runner.plan
+
+    XCTAssert(plan.steps.count == 0)
```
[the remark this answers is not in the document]
stmontgomery: I was able to reproduce that compiler crash. It seems worth filing as a separate issue, but in the mean time it appears to go away if you move this local `struct S` type declaration _out_ of the test method body, into a parent scope of the code:

```swift
  @Suite(.hidden) struct S {
    @Test(.hidden) func f() {}
  }
```

I am guessing something about using the `@Suite` or `@Test` macros in this scenario is causing an error, but that workaround allowed me to use `XCTAssertEqual`. And perhaps that's a lead on identifying the root cause of the crash.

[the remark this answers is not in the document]
grynspan: Please file a GitHub issue (or a radar if appropriate) tracking the compiler crash. We can discuss off GitHub how best to gather relevant diagnostic info.

Sources/Testing/Running/Configuration.swift
```diff
@@ -128,17 +128,38 @@ public struct Configuration: Sendable {
   ///
   /// By default, all tests are run and no filter is set.
   public mutating func setTestFilter(toMatch selection: Set<Test.ID>?) {
-    setTestFilter(toMatch: selection.map(Test.ID.Selection.init))
+    setTestFilter(toMatch: selection, includeHiddenTests: false)
   }
-  
+
+  /// The granularity to enforce test filtering.
```
grynspan, line 134: This brief doesn't seem to make sense in context? I realize you copied it from the existing method—we should probably fix it. :)

Also, and I know this sounds like nitpicking: we need a blank line after the brief before `- Parameters:` to ensure DocC correctly parses this comment.

Sources/Testing/Running/Configuration.swift
```diff
@@ -128,17 +128,38 @@ public struct Configuration: Sendable {
...
   ///
   /// - Parameters:
   ///   - selection: A selection of test IDs to be filtered. If `nil`, the
   ///     current selection is cleared.
+  ///   - includeHiddenTests: If false, a test annotated with the `.hidden` trait will not be included, even if its ID is present in `selection`.
   ///
   /// By default, all tests are run and no filter is set.
-  mutating func setTestFilter(toMatch selection: Test.ID.Selection?) {
-    testFilter = selection.map { $0.contains }
+  mutating func setTestFilter(toMatch selection: Test.ID.Selection?, includeHiddenTests: Bool) {
```
grynspan, line 153: Do we need _both_ new overloads for this internal code? Are we using both?

Sources/Testing/Running/Configuration.swift
```diff
@@ -128,17 +128,38 @@ public struct Configuration: Sendable {
   ///
   /// By default, all tests are run and no filter is set.
   public mutating func setTestFilter(toMatch selection: Set<Test.ID>?) {
```
grynspan, line 130: After this change, is there any code coverage of this overload?

Tests/TestingTests/RunnerTests.swift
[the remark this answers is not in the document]
ejvaughan: Filed a radar for the compiler crash (rdar://118130329).

Sources/Testing/Running/Configuration.swift
ejvaughan: We don't. I'll eliminate the redundancy.

Sources/Testing/Running/Configuration.swift
ejvaughan: This method has one callsite, in `XCTestScaffold.runAllTests(hostedBy:)`. I chose not to modify this method since it's public API, and we don't want to expose the concept of hidden tests to clients since that's internal to the project.

Sources/Testing/Running/Configuration.swift
```diff
@@ -121,28 +121,18 @@ public struct Configuration: Sendable {
...
-  /// The granularity to enforce test filtering.
   /// - Parameters:
   ///   - selection: A set of test IDs to be filtered. If `nil`, the current
   ///     selection is cleared.
-  ///   - includeHiddenTests: If false, a test annotated with the `.hidden` trait will not be included, even if its ID is present in `selection`.
   ///
   /// By default, all tests are run and no filter is set.
-  mutating func setTestFilter(toMatch selection: Set<Test.ID>?, includeHiddenTests: Bool) {
-    setTestFilter(toMatch: selection.map(Test.ID.Selection.init), includeHiddenTests: includeHiddenTests)
+  public mutating func setTestFilter(toMatch selection: Set<Test.ID>?) {
```
grynspan, line 131: Is there any code coverage of this function now?

Sources/Testing/Running/Configuration.swift
ejvaughan: Done.

stmontgomery (approved):

ejvaughan:
@swift-ci please test

PR #102, opened 2023-11-08
Lower Apple Platform requirement in Package.swift

### LINKED ISSUE #101: Lower the platform requirement
### Description

The current platform requirement makes it hard for other existing package to adopt swift-testing.
Consider lowering the requirement and mark related API with `@avaiable` mark.

### Expected behavior

A package with iOS 13 support can have a dependency on this package. 

### Actual behavior

A package with iOS 13 support can not have a dependency on this package. 

### Steps to reproduce

_No response_

_No response_

_No response_

Kyle-Ye:
Lower Apple Platform requirement in Package.swift

### Motivation:

Close #101 

### Modifications:

Lower Apple Platform requirement in Package.swift

### Result:

A Package with iOS 13 support can at least have a dependency on it.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -17,10 +17,11 @@ let package = Package(
   name: "swift-testing",
 
   platforms: [
-    .macOS(.v13),
-    .iOS(.v16),
-    .watchOS(.v9),
-    .tvOS(.v16),
+    .macOS(.v10_15),
+    .iOS(.v13),
+    .watchOS(.v6),
+    .tvOS(.v13),
+    .macCatalyst(.v13),
     .visionOS(.v1),
   ],
 
--- a/Sources/GitStatus/main.swift
+++ b/Sources/GitStatus/main.swift
@@ -56,10 +56,17 @@ func _runGit(passing arguments: String..., readingUpToCount maxOutputCount: Int)
   defer {
     process.terminate()
   }
-  guard let output = try? stdoutPipe.fileHandleForReading.read(upToCount: maxOutputCount) else {
-    return nil
+  if #available(macOS 10.15.4, *) {
+    guard let output = try? stdoutPipe.fileHandleForReading.read(upToCount: maxOutputCount) else {
+      return nil
+    }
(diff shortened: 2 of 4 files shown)


grynspan:
@swift-ci please test

Package.swift
```diff
@@ -17,10 +17,11 @@ let package = Package(
...
   platforms: [
-    .macOS(.v13),
-    .iOS(.v16),
-    .watchOS(.v9),
-    .tvOS(.v16),
+    .macOS(.v10_15),
+    .iOS(.v13),
+    .watchOS(.v6),
+    .tvOS(.v13),
+    .macCatalyst(.v13),
```
stmontgomery, line 24: @Kyle-Ye, would you like to update these versions to match those I proposed in my [reply](https://github.com/apple/swift-testing/issues/101#issuecomment-1808673624) to #101?

Kyle-Ye: Sorry, I still do not get the point. Could you elaborate on why we need macOS 13 instead of macOS 10.15 and watchOS 7 instead of watchOS 6? See https://github.com/apple/swift-testing/issues/101#issuecomment-1809525632.

Kyle-Ye: Specifically, I maintain a package which needs to support the above platform version. The current swift-testing configuration will block me from migrating to use swift-testing.

stmontgomery:
Per my recent [reply](https://github.com/apple/swift-testing/issues/101#issuecomment-1814841731), I think we can proceed with this.

But first: I'm running CI again to confirm a theory that this will fail on macOS because I landed a change in #104 that requires newer OS versions after this PR was posted. I have a fix incoming for that, however, so we can land that first and then proceed with landing this.

stmontgomery:
@swift-ci please test

stmontgomery (approved):
Approving, but waiting to merge until CI failures are resolved

stmontgomery:
CI is failing in the way I would expect, now. Fixing that in #121

stmontgomery:
@swift-ci please test

PR #103, opened 2023-11-08
 Use `reinterpret_cast` instead of a C-style cast in Discovery.cpp.

grynspan:
As described: this file is C++ and should prefer C++-style casts where possible. My mistake.


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #104, opened 2023-11-09
Lower the 'TestCases' protocol to internal

stmontgomery:
Lower the access level of the `TestCases` protocol from `@_spi public` to `internal` (implicitly).

### Motivation:

Due to some limitations in Swift's type system, we currently need a protocol named `TestCases` as a workaround to type-erase our sequence of test cases so that it can be constrained to `Sendable` (see [comment](https://github.com/apple/swift-testing/blob/b9a03b12b01afe238894904eaf56ba118093e960/Sources/Testing/Test.Case.swift#L81)). But because the `Test.testCases` is `public`, that protocol also needs to match its access level. Since this protocol is just a workaround, we'd rather not expose the protocol and only expose a property returning a `Sequence` of test case instances on `Test`.

### Modifications:

- Add `Test._testCases` as a `private` stored property of type `(any TestCases)?`.
- Change `Test.testCases` to a computed property which conditionally down-casts `self._testCases`.
- Lower `TestCases` protocol to `internal`.

### Result:

The `TestCases` protocol is no longer exposed, and the type of the computed `Test.testCases` property is `any Sequence<Test.Case>`.

### Checklist:

### DIFF
--- a/Tests/TestingTests/MiscellaneousTests.swift
+++ b/Tests/TestingTests/MiscellaneousTests.swift
@@ -433,7 +433,6 @@ struct MiscellaneousTests {
     #expect(!monomorphicTestFunction.isParameterized)
     let monomorphicTestFunctionTestCases = try #require(monomorphicTestFunction.testCases)
     #expect(monomorphicTestFunctionTestCases.underestimatedCount == 1)
-    #expect(!monomorphicTestFunctionTestCases.isParameterized)
     let monomorphicTestFunctionParameters = try #require(monomorphicTestFunction.parameters)
     #expect(monomorphicTestFunctionParameters.isEmpty)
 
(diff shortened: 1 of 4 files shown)


stmontgomery:
@swift-ci please test

Tests/TestingTests/MiscellaneousTests.swift
```diff
@@ -433,15 +433,13 @@ struct MiscellaneousTests {
     #expect(!monomorphicTestFunction.isParameterized)
     let monomorphicTestFunctionTestCases = try #require(monomorphicTestFunction.testCases)
     #expect(monomorphicTestFunctionTestCases.underestimatedCount == 1)
-    #expect(!monomorphicTestFunctionTestCases.isParameterized)
```
grynspan, line 436: Why was this expectation removed from the tests?

stmontgomery: This is calling the `isParameterized` property from the `TestCases` protocol, but since this PR no longer exposes the property of that protocol type this expectation no longer makes sense.

A few lines above, we still have an expectation which calls the same-named `isParameterized` property on `Test`, whose implementation is derived from the former.

stmontgomery:
@swift-ci please test

Tests/TestingTests/MiscellaneousTests.swift
grynspan: Fair enough!

grynspan (approved):

PR #105, opened 2023-11-09
Remove the 'Test.Case.index' property

stmontgomery:
Remove the `Test.Case.index` property.

### Motivation:

The `Test.Case` type includes an `Int` property named `index`, whose value is assigned using a simple counter while iterating all argument combinations for a parameterized test function. This property does not always have a stable value between runs: this is most obvious when the arguments passed to the `@Test` macro are modified, but even given the same arguments, a `Collection` (such as `Dictionary`) may iterate its elements in a different order each time. Also, the semantics of this property can be unclear when it comes to parameterized tests accepting 2 or more arguments.

This PR removes this property due to these problems. In a future PR, my goal is to introduce a stronger, stable representation of the argument(s) of each `Test.Case`.

### Modifications:

- Remove the `Test.Case.index` property and its supporting code in `Test.Case.Generator`.

### Checklist:

### DIFF
--- a/Sources/Testing/Test.Case.Generator.swift
+++ b/Sources/Testing/Test.Case.Generator.swift
@@ -31,11 +31,10 @@ extension Test.Case {
     /// A closure that maps an element from `_sequence` to a test case instance.
     ///
     /// - Parameters:
-    ///   - index: The zero-based index of the element.
     ///   - element: The element from `_sequence`.
     ///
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

grynspan (approved):

grynspan:
We do need a way to identify test cases uniquely, so please file an issue to track adding something.

PR #106, opened 2023-11-09
Add a mutable subscript to `Graph` when `V` is optional.

grynspan:
This PR modifies the subscript operator of the `Graph` pseudo-collection type such that, when `V` is optional, it is settable (with intermediate values defaulting to `nil`). This change simplifies some code that uses these subscripts.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Graph.swift
+++ b/Sources/Testing/Support/Graph.swift
@@ -79,7 +79,7 @@ struct Graph<K, V> where K: Hashable {
   ///   present.
   ///
   /// - Complexity: O(*n*) where *n* is the number of elements in `keyPath`.
-  subscript(keyPath: some Collection<K>) -> V? {
+  @_disfavoredOverload subscript(keyPath: some Collection<K>) -> V? {
     subgraph(at: keyPath)?.value
   }
 
@@ -93,7 +93,7 @@ struct Graph<K, V> where K: Hashable {
   ///   present.
   ///
   /// - Complexity: O(*n*) where *n* is the number of elements in `keyPath`.
-  subscript(keyPath: K...) -> V? {
+  @_disfavoredOverload subscript(keyPath: K...) -> V? {
     self[keyPath]
   }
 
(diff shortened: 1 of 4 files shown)


Sources/Testing/Support/Graph.swift
```diff
@@ -223,6 +223,55 @@ extension Graph {
     self.init(value: nil, children: children)
   }
 
+  /// Get or set the leaf value at the node identified by the specified sequence
+  /// of keys.
```
stmontgomery, line 227: I've seen this documented in [other](https://developer.apple.com/documentation/swift/dictionary/subscript(_:)-8rfql) mutable subscripts as "**Accesses** the value … for reading and writing", using the verb 'access'.

Sources/Testing/Support/Graph.swift
```diff
@@ -236,9 +285,10 @@ extension Graph {
   /// intermediate to it, the graph is not modified. To add a value when none
   /// previously exists, use ``insertValue(_:at:intermediateValue:)``.
   ///
-  /// - Complexity: O(*m* + *n*) where *n* is the number of elements in
-  ///   `keyPath` and *m* is the number of children at the penultimate node in
-  ///   `keyPath`.
+  /// - Complexity: To get a value, O(*n*) where *n* is the number of elements
+  ///   in `keyPath`. To set a value, O(*m* + *n*) where *n* is the number of
+  ///   elements in `keyPath` and *m* is the number of children at the
+  ///   penultimate node in `keyPath`.
```
stmontgomery, line 291: Why include the "To **get** a value..." sentence in this `updateValue` method, if it's focused on setting a value?

grynspan: This was a pasting error. :)

grynspan:
@swift-ci please test

stmontgomery (approved):

Sources/Testing/Support/Graph.swift
grynspan: The grammar gets a bit hard to parse if I change it because "for reading and writing" ends up quite far from "Accesses" so I'm just gonna leave it as-is.

PR #108, opened 2023-11-09
Implement `swift test --xunit-output`.

grynspan:
This PR implements the `--xunit-output` option when using SwiftPM. Note that internally, the code refers to JUnit rather than xUnit: that's because the format produced by XCTest and SwiftPM today does not _not_ conform to the xUnit schema(s) described at https://www.xunit.org, but rather are a variant of the (schema-less?) JUnit format.

This change renames `Event.Recorder` to `Event.ConsoleOutputRecorder` and adds `Event.JUnitXMLRecorder`. It adds a protocol, `EventRecorder` (remember that protocols cannot be nested in types, so no dot) to which both of these types conform, although that's a formality only as we are not doing anything generic over recorder instances.

If the user specifies `--xunit-output` when calling into `swift test`, we open a file (using good ol' `fopen()`.) Our first official use of a non-copyable type is here (hooray!) as we use one to box the resulting file handle and ensure we call `fclose()` after tests finish running. We do **not** use Foundation's `FileHandle` for this purpose because, although it has the right ownership semantics, it does not expose API for creating _new_ files, only for opening existing ones.

`fopen()` can of course fail, so we now need to potentially throw a C error code. I've added an internal `CError` type to represent them. We're not using `POSIXError` here because it's a Foundation dependency, but also because it's possible for `errno` to be set to a value that is not representable as an instance of `POSIXErrorCode`, which is a closed enumeration.

> [!NOTE]
> Without changes (TBD) in swift-package-manager, using `--xunit-output` when running _both_ swift-testing and XCTest tests will result in one library overwriting the other's output. The work to resolve that will need to take place on the SwiftPM side.

### Checklist:

Resolves rdar://118199250.

### DIFF
--- a/Sources/Testing/Events/Clock.swift
+++ b/Sources/Testing/Events/Clock.swift
@@ -101,7 +101,7 @@ extension Test.Clock.Instant {
   ///
   /// - Returns: The number of nanoseconds between `self` and `other`. If
   ///   `other` is ordered before this instance, the result is negative.
-  func nanoseconds(until other: Test.Clock.Instant) -> Int64 {
+  func nanoseconds(until other: Self) -> Int64 {
     if other < self {
       return -other.nanoseconds(until: self)
     }
--- a/Sources/Testing/Events/Clock.swift
+++ b/Sources/Testing/Events/Clock.swift
@@ -96,6 +96,9 @@ extension Test.Clock.Instant {
 
   /// Get the number of nanoseconds from this instance to another.
   ///
+  /// - Parameters:
+  ///   - other: The later instant.
+  ///
   /// - Returns: The number of nanoseconds between `self` and `other`. If
   ///   `other` is ordered before this instance, the result is negative.
   func nanoseconds(until other: Test.Clock.Instant) -> Int64 {
--- a/Sources/Testing/Events/Clock.swift
+++ b/Sources/Testing/Events/Clock.swift
@@ -93,6 +93,19 @@ extension Test.Clock.Instant {
   public var durationSince1970: Duration {
     Duration(wall)
   }
+
+  /// Get the number of nanoseconds from this instance to another.
+  ///
+  /// - Returns: The number of nanoseconds between `self` and `other`. If
+  ///   `other` is ordered before this instance, the result is negative.
+  func nanoseconds(until other: Test.Clock.Instant) -> Int64 {
+    if other < self {
+      return -other.nanoseconds(until: self)
+    }
+    let otherNanoseconds = (other.suspending.seconds * 1_000_000_000) + (other.suspending.attoseconds / 1_000_000_000)
+    let selfNanoseconds = (suspending.seconds * 1_000_000_000) + (suspending.attoseconds / 1_000_000_000)
+    return otherNanoseconds - selfNanoseconds
+  }
 }
 #endif
 
@@ -209,9 +222,7 @@ extension Test.Clock.Instant {
   ///   up to millisecond accuracy.
   func descriptionOfDuration(to other: Test.Clock.Instant) -> String {
 #if SWT_TARGET_OS_APPLE
-    let otherNanoseconds = (other.suspending.seconds * 1_000_000_000) + (other.suspending.attoseconds / 1_000_000_000)
-    let selfNanoseconds = (suspending.seconds * 1_000_000_000) + (suspending.attoseconds / 1_000_000_000)
-    let (seconds, nanosecondsRemaining) = (otherNanoseconds - selfNanoseconds).quotientAndRemainder(dividingBy: 1_000_000_000)
+    let (seconds, nanosecondsRemaining) = nanoseconds(until: other).quotientAndRemainder(dividingBy: 1_000_000_000)
     return String(describing: TimeValue((seconds, nanosecondsRemaining * 1_000_000_000)))
 #else
     return String(describing: TimeValue(Duration(other.suspending) - Duration(suspending)))
--- a/Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
@@ -14,7 +14,7 @@ extension Event {
   ///
   /// The format of the output is not meant to be machine-readable and is
   /// subject to change.
-  public struct Recorder: Sendable {
+  public struct ConsoleOutputRecorder: Sendable {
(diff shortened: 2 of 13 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Events/Recorder/EventRecorder.swift
```diff
@@ -0,0 +1,23 @@
...
+  /// Record the specified event by generating a representation of it in this
+  /// instance's output format and writing it to this instance's destination.
+  ///
+  /// - Parameters:
+  ///   - event: The event to record.
+  ///   - context: The context associated with the event.
+  ///
+  /// - Returns: Whether any output was produced and written to this instance's
+  ///   destination.
+  func record(_ event: borrowing Event, in context: borrowing Event.Context) -> Bool
```
stmontgomery, line 22: `@discardableResult`? Or can protocol methods not include that attribute?

Tests/TestingTests/EventRecorderTests.swift
```diff
@@ -202,6 +207,33 @@ struct EventRecorderTests {
     #expect(match.output.2 == 4)
   }
 #endif
+
+  @Test("JUnitXMLRecorder outputs valid XML")
+  func junitXMLIsValid() async throws {
```
stmontgomery, line 212: Feel like we ought to split the ConsoleOutput and JUnit recorder types into their own test files & types, since I could imagine the new JUnit one gaining more tests over time

Tests/TestingTests/SwiftPMTests.swift
```diff
@@ -83,6 +91,28 @@ struct SwiftPMTests {
...
   }
 
+  @Test("--xunit-output argument")
+  func xunitOutput() throws {
+    // Test that a bad path produces an error.
+    #expect(throws: CError.self) {
+      _ = try configurationForSwiftPMEntryPoint(withArguments: ["PATH", "--xunit-output", "/nonexistent/path/we/cannot/write/to"])
+    }
+
+    // Test that a file is opened when requested. Testing of the actual output
```
stmontgomery, line 101: This feels like it's really two tests. Maybe create a nested suite type and decompose these into two `@Test` funcs within?

Sources/Testing/Support/CError.swift
```diff
@@ -0,0 +1,27 @@
...
+///
+/// This type is necessary because Foundation's `POSIXError` is not available in
+/// all contexts.
+///
+/// This type is not part of the public interface of the testing library.
+struct CError: Error, RawRepresentable {
+  var rawValue: CInt
+}
+
+extension CError: CustomStringConvertible {
```
stmontgomery, line 23:
```suggestion
// MARK: - CustomStringConvertible

extension CError: CustomStringConvertible {
```

Sources/Testing/Events/Clock.swift
```diff
@@ -93,6 +93,22 @@ extension Test.Clock.Instant {
...
   }
+
+  /// Get the number of nanoseconds from this instance to another.
+  ///
+  /// - Parameters:
+  ///   - other: The later instant.
+  ///
+  /// - Returns: The number of nanoseconds between `self` and `other`. If
+  ///   `other` is ordered before this instance, the result is negative.
+  func nanoseconds(until other: Test.Clock.Instant) -> Int64 {
```
stmontgomery, line 104: Can we add (or do we already have) a unit test for `nanoseconds(until:)`?

Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
```diff
@@ -14,7 +14,7 @@ extension Event {
   ///
   /// The format of the output is not meant to be machine-readable and is
   /// subject to change.
```
stmontgomery, line 16: Consider mentioning via a "see also" comment the JUnitXML recorder type as a machine-readable alternative (since this explicitly describes ConsoleOutputRecorder as non-machine readable)

Sources/Testing/Events/Recorder/EventRecorder.swift
grynspan: Doesn't inherit from the protocol, no. I tried.

Tests/TestingTests/EventRecorderTests.swift
grynspan: Indeed, but I'll save that for when we write those tests.

Sources/Testing/Support/CError.swift
grynspan: Hoisted by my own petard.

Sources/Testing/Events/Clock.swift
grynspan: We have indirect test coverage for it already, yeah, but I can add an explicit parameterized test.

Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
grynspan: Sure.

grynspan: Will do.

Sources/Testing/Events/Clock.swift
grynspan: We have indirect code coverage, but I can add a parameterized test for it directly.

Tests/TestingTests/SwiftPMTests.swift
grynspan: Done.

Sources/Testing/Events/Recorder/EventRecorder.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+@_spi(ExperimentalEventHandling)
+public protocol EventRecorder: Sendable {
```
stmontgomery, line 12: Can we document the protocol itself?

grynspan: Of course. Oops.

grynspan:
@swift-ci please test

stmontgomery (approved):

PR #109, opened 2023-11-10
Add an `index` property to `Test.ParameterInfo`

stmontgomery:
Add an `index` property to `Test.ParameterInfo` containing the index of the parameter within its associated test function's parameter list.

### Motivation:

Parameterized test functions can include multiple parameters, and like any Swift function, the names of those parameters may be repeated. When looking at one `Test.ParameterInfo` parameter in isolation, it can be useful to know what index the parameter is, to distinguish it from other parameters which may have the same name.

### Modifications:

- Add `index` to `Test.Parameter`
- Adjust callers to derive the index from macro-expanded code based on position in array.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Macro.swift
+++ b/Sources/Testing/Test+Macro.swift
@@ -172,6 +172,19 @@ extension Test {
   }
 }
 
+extension [Test.__ParameterInfo] {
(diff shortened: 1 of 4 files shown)


stmontgomery:
@swift-ci please test

dennisweissmann (approved):

PR #110, opened 2023-11-13
Testing: fix Windows specific code for application data

compnerd:
The parameter is expected to be `KNOWNFOLDERID` which is a `Optional<UnsafePointer<GUID>>`.  Additionally, correct the spelling of `NULL` to `nil`.  This allows the target to build on Windows.

### DIFF
--- a/Sources/Testing/Traits/Tag.Color+Loading.swift
+++ b/Sources/Testing/Traits/Tag.Color+Loading.swift
@@ -36,7 +36,8 @@ private var _homeDirectoryPath: String? {
 /// The path to the current user's App Data directory, if known.
 private var _appDataDirectoryPath: String? {
   var appDataDirectoryPath: PWSTR? = nil
-  if S_OK == SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, NULL, &appDataDirectoryPath), let appDataDirectoryPath {
+  var FOLDERID_LocalAppData = FOLDERID_LocalAppData
+  if S_OK == SHGetKnownFolderPath(&FOLDERID_LocalAppData, 0, nil, &appDataDirectoryPath), let appDataDirectoryPath {
     defer {
       CoTaskMemFree(appDataDirectoryPath)
     }


Sources/Testing/Traits/Tag.Color+Loading.swift
```diff
@@ -36,13 +36,15 @@ private var _homeDirectoryPath: String? {
 /// The path to the current user's App Data directory, if known.
 private var _appDataDirectoryPath: String? {
   var appDataDirectoryPath: PWSTR? = nil
-  if S_OK == SHGetKnownFolderPath(FOLDERID_LocalAppData, 0, NULL, &appDataDirectoryPath), let appDataDirectoryPath {
-    defer {
-      CoTaskMemFree(appDataDirectoryPath)
+  return withUnsafePointeR(to: FOLDERID_LocalAppData) {
```
grynspan, line 39: Well _now_ who spelled it incorrectly. ;)

If we need a pointer to the folder ID, can we just use `&`? (If not, please label the closure argument to match house style.)

grynspan (changes requested):
See above.

Sources/Testing/Traits/Tag.Color+Loading.swift
compnerd: Haha, that is what happens when you test and commit across different environments!

grynspan: We could copy it to a local:

```swift
var FOLDERID_LocalAppData = FOLDERID_LocalAppData
if S_OK == SHGetKnownFolderPath(&FOLDERID_LocalAppData, 0, nil, &appDataDirectoryPath), let appDataDirectoryPath {
```

compnerd: That sounds good to me.

compnerd:
@swift-ci please test

grynspan (approved):

PR #111, opened 2023-11-13
git: ignore vim swap files

compnerd:
These files should not be added to the repository and will be added if they are not ignored.


grynspan:
@swift-ci please test

grynspan (approved):

PR #112, opened 2023-11-13
Represent each parameterized test function argument as a separate element in `Test.Case.arguments`

stmontgomery:
Represent each parameterized test function argument as a separate element in `Test.Case.arguments`, including tuple types which have been de-structured into discrete function parameters.

### Motivation:

This is an enhancement meant to represent arguments of parameterized test functions in `Test.Case` instances in a more useful way, and facilitate progress towards allowing (re-)running specific argument(s) of such functions.

### Modifications:

- Introduce a struct `Test.Case.Argument` which represents one argument passed to one parameter of a parameterized test function. This includes each argument's associated `Test.ParameterInfo`.
- Change the `Test.Case.arguments` property to be an array of the new struct type.
- Enhance `Test.Case.Generator` to generate `Test.Case` instances using the new argument structure—including de-structuring tuples into their constituent elements when needed. It accomplishes this more efficiently now using some specialized generic overloads.
- Remove `Test.Case.arguments(pairedWith:)` function, since it's no longer necessary given the `arguments` property already pairs these values.
- Introduce a convenience utility for testing purposes only, which handles errors thrown in `eventHandler` closures.

### Checklist:


stmontgomery:
@swift-ci please test

Sources/Testing/Test.Case.Generator.swift
```diff
@@ -93,34 +117,124 @@ extension Test.Case {
...
+    /// dictionary of argument values.
+    ///
+    /// - Parameters:
+    ///   - dictionary: A dictionary of argument values for which test cases
+    ///     should be generated.
+    ///   - parameters: The parameters of the test function for which test cases
+    ///     should be generated.
+    ///   - testFunction: The test function to which each generated test case
+    ///     passes an argument value from `dictionary`.
+    init<Key, Value>(
```
grynspan, line 222: Please document why this overload is useful (perf optimization for a perceived common case.)

Sources/Testing/Test.Case.Generator.swift
```diff
@@ -93,34 +117,124 @@ extension Test.Case {
...
     ///     passes an argument value from `zippedCollections`.
     init<C1, C2>(
       arguments zippedCollections: Zip2Sequence<C1, C2>,
+      parameters: [Test.ParameterInfo],
       testFunction: @escaping @Sendable ((C1.Element, C2.Element)) async throws -> Void
-    ) where S == Zip2Sequence<C1, C2> {
-      self.init(sequence: zippedCollections) { element in
-        Test.Case(arguments: [element]) {
-          try await testFunction(element)
+    ) where S == Zip2Sequence<C1, C2>, C1: Collection, C2: Collection {
```
grynspan, line 208: Is the compiler forcing you to specify that `C1` and `C2` are collections? It shouldn't have done so previously?

stmontgomery: This is effectively a drive-by fix to add missing constraints I believe we should have had all along. Compiler did not force me to do it, but I discovered it while working on this

stmontgomery:
@swift-ci please test

Sources/Testing/Test+Macro.swift
```diff
@@ -369,8 +370,49 @@ extension Test {
...
-    let caseGenerator = Case.Generator(arguments: collection1, collection2, testFunction: testFunction)
-    return Self(name: testFunctionName, displayName: displayName, traits: traits, sourceLocation: sourceLocation, containingType: containingType, xcTestCompatibleSelector: xcTestCompatibleSelector, testCases: caseGenerator, parameters: paramTuples.parameters)
+    let parameters = paramTuples.parameters
+    let caseGenerator = Case.Generator(arguments: collection1, collection2, parameters: parameters, testFunction: testFunction)
+    return Self(name: testFunctionName, displayName: displayName, traits: traits, sourceLocation: sourceLocation, containingType: containingType, xcTestCompatibleSelector: xcTestCompatibleSelector, testCases: caseGenerator, parameters: parameters)
+  }
+
+  /// Create an instance of ``Test`` for a parameterized function.
+  ///
+  /// - Warning: This function is used to implement the `@Test` macro. Do not
```
grynspan, line 380:
```suggestion
    /// This initializer overload is specialized for sequences of 2-tuples to
     /// efficiently de-structure their elements when appropriate.
       /// - Warning: This function is used to implement the `@Test` macro. Do not
```

Sources/Testing/Test+Macro.swift
```diff
@@ -369,8 +370,49 @@ extension Test {
...
   ) -> Self where C1: Collection & Sendable, C1.Element: Sendable, C2: Collection & Sendable, C2.Element: Sendable {
-    let caseGenerator = Case.Generator(arguments: collection1, collection2, testFunction: testFunction)
-    return Self(name: testFunctionName, displayName: displayName, traits: traits, sourceLocation: sourceLocation, containingType: containingType, xcTestCompatibleSelector: xcTestCompatibleSelector, testCases: caseGenerator, parameters: paramTuples.parameters)
+    let parameters = paramTuples.parameters
+    let caseGenerator = Case.Generator(arguments: collection1, collection2, parameters: parameters, testFunction: testFunction)
+    return Self(name: testFunctionName, displayName: displayName, traits: traits, sourceLocation: sourceLocation, containingType: containingType, xcTestCompatibleSelector: xcTestCompatibleSelector, testCases: caseGenerator, parameters: parameters)
+  }
+
+  /// Create an instance of ``Test`` for a parameterized function.
+  ///
```
grynspan, line 379:
```suggestion
  ///
  /// This initializer overload is specialized for sequences of 2-tuples to
  /// efficiently de-structure their elements when appropriate.
```

Sources/Testing/Test+Macro.swift
```diff
@@ -369,8 +370,49 @@ extension Test {
...
+    parameters paramTuples: [__ParameterInfo],
+    testFunction: @escaping @Sendable ((E1, E2)) async throws -> Void
+  ) -> Self where C: Collection & Sendable, C.Element == (E1, E2), E1: Sendable, E2: Sendable {
+    let parameters = paramTuples.parameters
+    let caseGenerator = Case.Generator(arguments: collection, parameters: parameters, testFunction: testFunction)
+    return Self(name: testFunctionName, displayName: displayName, traits: traits, sourceLocation: sourceLocation, containingType: containingType, xcTestCompatibleSelector: xcTestCompatibleSelector, testCases: caseGenerator, parameters: parameters)
+  }
+
+  /// Create an instance of ``Test`` for a parameterized function.
+  ///
```
grynspan, line 399:
```suggestion
  ///
  /// This initializer overload is specialized for collections of 2-tuples to
  /// efficiently de-structure their elements when appropriate.
```

Sources/Testing/Test.Case.Generator.swift
```diff
@@ -93,34 +117,134 @@ extension Test.Case {
...
+    ///     passes an argument value from `dictionary`.
+    ///
+    /// This initializer overload is specialized for dictionary collections, to
+    /// efficiently de-structure their elements (which are known to be 2-tuples)
+    /// when appropriate.
+    init<Key, Value>(
+      arguments dictionary: Dictionary<Key, Value>,
+      parameters: [Test.ParameterInfo],
+      testFunction: @escaping @Sendable ((Key, Value)) async throws -> Void
+    ) where S == Dictionary<Key, Value> {
```
grynspan, line 236: Dictionaries are sequences of 2-tuples already, so is this overload needed in _addition_ to the 2-tuple overload? Seems like the decomposition of a dictionary here is exactly the same as the decomposition of a tuple (either into a single tuple argument or into _n_ desugared arguments.)

Sources/Testing/Test.Case.Generator.swift
```diff
@@ -72,15 +72,39 @@ extension Test.Case {
...
+    @_disfavoredOverload
     init(
       arguments collection: S,
+      parameters: [Test.ParameterInfo],
       testFunction: @escaping @Sendable (S.Element) async throws -> Void
     ) where S: Collection {
-      self.init(sequence: collection) { element in
-        Test.Case(arguments: [element]) {
-          try await testFunction(element)
+      if parameters.count > 1 {
```
grynspan, line 90: This change is going to cause us to front-load mirroring rather than performing it lazily, which may have a negative impact on performance when the argument to a test is a sequence of tuples (but not a dictionary). Are we sure we are okay with taking that hit?

Tests/TestingTests/TestSupport/TestingAdditions.swift
```diff
@@ -290,6 +290,29 @@ extension Test.ID.Selection {
...
+  /// - Parameters:
+  ///   - eventHandler: The throwing ``Event/Handler`` to set.
+  ///
+  /// Errors thrown by `eventHandler` are caught and recorded as an ``Issue``.
+  ///
+  /// This is meant for testing the testing library itself. In production tests,
+  /// event handlers should not typically need to be throwing, but if they do,
+  /// the event handler should implement its own error-handling logic since
+  /// recording an ``Issue`` would be inappropriate.
+  mutating func setEventHandler(_ eventHandler: @escaping @Sendable (_ event: borrowing Event, _ context: borrowing Event.Context) throws -> Void) {
```
grynspan, line 305: Was this change intentional? Should the error-handling logic be part of the overall interface instead?

Sources/Testing/Test.Case.Generator.swift
stmontgomery: The other `init` overload I added below, the one for `Collection` values whose `Element` are 2-tuples, is meant to handle that. That means this overload (which now has `@_disfavoredOverload`) will only be used for collections of tuples of 3+ elements. With some further variadic generic improvements which are [in progress](https://github.com/apple/swift/pull/67594), I believe we may be able to solve this for all tuple lengths. 

So yeah I do think this is justified, and the specialized generic overloads help minimize its impact in common cases.

Sources/Testing/Test.Case.Generator.swift
stmontgomery: I thought it was redundant also but the compiler treats `Dictionary` differently because its `Element` type is `(key: Key, value: Value)` and not `(Key, Value)`. The tuple labels make it not identical, in other words.

Tests/TestingTests/TestSupport/TestingAdditions.swift
stmontgomery: I tried to explain my rationale for why I don't think it should be built-in in the last paragraph of the DocC comment above. In particular, it doesn't seem appropriate for a "production" event handler to record `Issue`s itself, so I think it's reasonable to say that event handler closures must catch & handle errors on their own. It's just that we're a special case here because we're using a nested event handler so we _can_ safely record `Issue`s from our "inner" event handler.

stmontgomery:
@swift-ci please test

Sources/Testing/Test.Case.Generator.swift
grynspan: Got it!

Sources/Testing/Test.Case.Generator.swift
grynspan: Boooo-urns. Okay. Can you add a comment about that then please?

grynspan: (And add it in Test+Macro.swift too? Actually… I wonder if we've specialized `__function()` enough at this point that we could (in a different PR!) refactor this type and move the logic into that file, reducing redundancy?

Tests/TestingTests/TestSupport/TestingAdditions.swift
grynspan: 🤔 I see what you mean. I do wish we had the ability to make the SPI event handler throwing as well, but as you say the outermost runner isn't going to have anywhere to record issues to.

Not for this PR, but… we _could_ set up the default state to `fatalError()` while allowing nested handlers to propagate errors outward, if we wanted to.

Sources/Testing/Test.Case.Generator.swift
stmontgomery: Sure, will note that in docs

stmontgomery:
@swift-ci please test

grynspan (approved):

PR #113, opened 2023-11-13
build: ignore `TestingMacrosTest` on Windows

compnerd:
This relies on symbolic renaming which is not available on Windows. This allows us to build the test suite on Windows.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -100,7 +100,7 @@ let package = Package(
   cxxLanguageStandard: .cxx20
 )
 
-// https://github.com/apple/swift-package-manager/issues/6367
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

Package.swift
```diff
@@ -108,6 +100,20 @@ let package = Package(
   cxxLanguageStandard: .cxx20
 )
 
+// https://github.com/apple/swift-package-manager/issues/6367
```
grynspan, line 103:
```suggestion
// BUG: swift-package-manager-#6367
```

stmontgomery (approved):

PR #114, opened 2023-11-13
Support: fix `Environment.variable(named:)` on Windows

compnerd:
The return value being `0` may or may not indicate.  We need to additionally check `GetLastError` to see if `ERROR_ENVVAR_NOT_FOUND` is set to indicate the missing variable.  Otherwise, the variable is just an empty string.  This repairs the 2 failing tests on Windows.

Thanks to @grynspan for the fix!

### DIFF
--- a/Sources/Testing/Support/Environment.swift
+++ b/Sources/Testing/Support/Environment.swift
@@ -41,10 +41,19 @@ enum Environment {
     name.withCString(encodedAs: UTF16.self) { name in
       func getVariable(maxCount: Int) -> String? {
         withUnsafeTemporaryAllocation(of: wchar_t.self, capacity: maxCount) { buffer in
+          SetLastError(DWORD(ERROR_SUCCESS))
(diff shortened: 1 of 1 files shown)


briancroom (approved):

Sources/Testing/Support/Environment.swift
```diff
@@ -41,10 +41,19 @@ enum Environment {
...
+          if count == 0 {
+            return switch GetLastError() {
+            case DWORD(ERROR_SUCCESS):
+              // Empty String
+              ""
+            case DWORD(ERROR_ENVVAR_NOT_FOUND):
+              // The environment variable wasn't set.
+              nil
+            case let errorCode:
+              fatalError("unexpected error code: \(errorCode)")
```
grynspan, line 55: We should explain _where_ the error occurred, i.e. `"when getting environment variable \"\(name)\""`

compnerd: Done, added that to the message.

grynspan (approved):

grynspan:
@swift-ci please test

PR #115, opened 2023-11-14
Simplify how we filter out hidden tests.

grynspan:
Right now, we're checking if tests are hidden in multiple places during execution. This is somewhat inefficient, but it's also error-prone because it means that we can create new code paths that _don't_ correctly check for hidden tests.

This PR consolidates the checks for hidden tests to `Configuration.testFilter` itself. With this change, hidden tests are _always_ filtered out by that filtering function regardless of what value it's set to. The only escape hatch—which we need for our own unit tests only—is to explicitly call the internal-only `Configuration.setTestFilter(toMatch:includeHiddenTests:)` and pass `true`.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -118,38 +118,48 @@ public struct Configuration: Sendable {
   /// - Returns: A Boolean value representing if the test satisfied the filter.
   public typealias TestFilter = @Sendable (Test) -> Bool
 
+  /// Storage for ``testFilter``.
+  private var _testFilter: TestFilter = { !$0.isHidden }
+
   /// The test filter to which tests should be filtered when run.
-  public var testFilter: TestFilter?
+  public var testFilter: TestFilter {
+    get {
+      _testFilter
+    }
+    set {
+      // By default, the test filter should always filter out hidden tests. This
+      // is the appropriate behavior for external clients of this SPI. If the
+      // testing library needs to enable hidden tests in its own test targets,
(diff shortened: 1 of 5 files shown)


Sources/Testing/Running/Configuration.swift
```diff
@@ -118,38 +118,49 @@ public struct Configuration: Sendable {
...
   ///
   /// - Parameters:
-  ///   - selection: A set of test IDs to be filtered. If `nil`, the current
-  ///     selection is cleared.
+  ///   - selection: A set of test IDs to be filtered.
   ///
   /// By default, all tests are run and no filter is set.
-  public mutating func setTestFilter(toMatch selection: Set<Test.ID>?) {
-    setTestFilter(toMatch: selection.map(Test.ID.Selection.init), includeHiddenTests: false)
+  public mutating func setTestFilter(toMatch selection: Set<Test.ID>) {
```
grynspan, line 146: By making the argument here non-optional, we simplify the logic quite a bit. If a caller wants to reset the test filter, they can set it to `{ _ in true }`, but in practice I don't think mutating a test filter in-flight is a concern for us.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

stmontgomery (approved):

dennisweissmann (approved):

Sources/Testing/Running/Configuration.swift
suzannaratcliff: okay!

suzannaratcliff (approved):

PR #117, opened 2023-11-14
Simplify loading tag colors from the home folder.

grynspan:
This PR simplifies the code used to load tags from the home folder. In particular, it adds `CodingKeyRepresentable` conformance to `Tag` so that tags can be used directly as keys in a JSON-encoded dictionary without us having to jump through hoops to convert them from strings.

### DIFF
--- a/Sources/Testing/Running/EntryPoint.swift
+++ b/Sources/Testing/Running/EntryPoint.swift
@@ -260,7 +260,9 @@ extension [Event.ConsoleOutputRecorder.Option] {
 #endif
 
     // Load tag colors from user/package preferences on disk.
-    result += tagColorOptions()
+    if let tagColors = try? loadTagColors() {
+      result.append(.useTagColors(tagColors))
+    }
 
     return result
   }
--- a/Sources/Testing/Traits/Tag.Color+Loading.swift
+++ b/Sources/Testing/Traits/Tag.Color+Loading.swift
@@ -77,26 +77,22 @@ var swiftTestingDirectoryPath: String {
 ///   - swiftTestingDirectoryPath: The `".swift-testing"` directory from which
 ///     tag color data should be read.
 ///
-/// - Returns: A sequence of zero or more ``Event/Recorder/Option`` values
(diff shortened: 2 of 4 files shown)


grynspan:
@swift-ci please test

briancroom (approved):

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -260,7 +260,9 @@ extension [Event.ConsoleOutputRecorder.Option] {
 #endif
 
     // Load tag colors from user/package preferences on disk.
-    result += tagColorOptions()
+    if let tagColors = try? loadTagColors() {
```
stmontgomery, line 263: If there's an error thrown here should we log at least, while proceeding to run tests? (I see this is not really a behavior change from before this PR, but might be nice, still)

grynspan: _Maybe_. I don't want to leak a dependency on Foundation out of that file, but we'd need to check for `CocoaError(.fileNotFound)` (or however it's spelt.)

stmontgomery: Oh I figured it could just handle _any_ error, not necessarily specific ones. Tag colors aren't mission critical, I wouldn't say, so any error that arises could just be noted but we continue to run tests

grynspan: That wouldn't work because it throws an error if the user hasn't created a "tag-colors.json" file, which is obviously not a failure mode. Hence we'd need to special-case file-not-found errors.

stmontgomery: Maybe separating out the "is there a colors file" and "parse said file" logic would make that easier? Seems like the 
former could be non-throwing

grynspan: That would require us to leak the Foundation dependency out in the form of returning an instance of `Data?` which, again, we want to avoid.

Since, as you said, tag colors aren't mission critical, we're probably overthinking it. :)

PR #118, opened 2023-11-14
Reference each parameterized test argument via a unique ID and allow running specific arguments

stmontgomery:
Introduce support for referencing each argument of a parameterized test via a unique ID, and specifying cases to run based on those IDs.

### Motivation:

This facilitates allowing (re-)running specific argument(s) of parameterized test functions.

### Modifications:

- Add an `id` property to `Test.Case.Argument` containing the unique ID of a particular argument passed to a parameterized test.
- Add a `CustomTestArgument` protocol, which types can optionally conform to in order to provide a stable, unique ID for test arguments.
- Add a `Test.Case.ID` consisting of an array of argument IDs, referencing the value of `Test.Case.Argument.id` for each argument.
- Add conformance to `Identifiable` protocol to `Test.Case`.
- Add `Test.ID.CaseSelection` and `Test.Case.ID.Selection` types which compose together to allow specifying test case to run, stored via a new `Configuration.selectedTestCases` SPI property.
- Modify `Runner` and its related types to support filtering out non-selected test cases.
- Add `testCaseID` SPI property to `Event`.

### Checklist:

Resolves rdar://117535243


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/Parameterization/Test.Case.ID.Selection.swift
```diff
@@ -0,0 +1,61 @@
...
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+extension Test.Case.ID {
+  /// A selection of ``Test/Case/ID`` instances.
+  struct Selection: Sendable {
```
grynspan, line 13: Should we make `Selection` generic over some `T: Sendable & Identifiable & ?`?

Sources/Testing/Parameterization/Test.ID.CaseSelection.swift
```diff
@@ -0,0 +1,67 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+extension Test.ID {
+  /// A type representing a selection of test cases grouped by their associated
+  /// test.
+  struct CaseSelection: Sendable {
```
grynspan, line 14: I'm a bit confused why we need this _and_ `Test.Case.ID.Selection`. Can we simplify things?

Sources/Testing/Parameterization/Test.Case.ID.swift
```diff
@@ -0,0 +1,44 @@
...
+extension Test.Case: Identifiable {
+  /// The ID of a test case.
+  ///
+  /// Instances of this type are considered unique within the scope of a given
+  /// parameterized test function. They are not necessarily unique across two
+  /// different ``Test`` instances.
+  public struct ID: Sendable, Equatable, Hashable {
+    /// The IDs of the arguments of this instance's associated ``Test/Case``, in
+    /// the order they appear in ``Test/Case/arguments``.
+    public var argumentIDs: [String]
```
grynspan, line 21: Does this behave correctly for monomorphic test functions (where the set of arguments is the empty array and the test has a single case)?

Sources/Testing/Parameterization/CustomTestArgument.swift
```diff
@@ -0,0 +1,73 @@
...
+  /// `String(describing:)`. The resulting string may not be stable or unique.
+  /// If the type of the argument conforms to `Identifiable` and its associated
+  /// `ID` type is `String`, the value of calling its `id` property is used
+  /// instead.
+  ///
+  /// It is possible that neither of the approaches for obtaining an ID detailed
+  /// above provide sufficient stability or uniqueness. If the type of the
+  /// argument is made to conform to ``CustomTestArgument``, then the ID
+  /// returned by this method is used.
+  func argumentID(in context: Test.Case.Argument.Context) -> String
```
grynspan, line 37: We have several `Context` types already. Seems like a rethink might be appropriate here.

Sources/Testing/Running/Configuration.swift
```diff
@@ -162,4 +162,21 @@ public struct Configuration: Sendable {
       testFilter = selection.contains
     }
   }
+
+  /// The selected test cases to run.
+  private(set) var selectedTestCases = Test.ID.CaseSelection()
```
grynspan, line 167: Should we consider modelling this as `testCaseFilter` instead?

Sources/Testing/Parameterization/Test.Case.ID.Selection.swift
stmontgomery: Are you asking whether we should make `Test.ID.Selection` generic, or `Test.Case.ID.Selection`? Note that `Test.ID.Selection` has different semantics: its "contains" operation considers ancestors and descendants

Sources/Testing/Parameterization/Test.Case.ID.swift
stmontgomery: Yeah, it's represented as a "trivial" `Test.Case.ID` containing an empty `argumentIDs` array.

Sources/Testing/Parameterization/Test.Case.ID.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+
+@_spi(ExperimentalParameterizedTesting)
+extension Test.Case: Identifiable {
+  /// The ID of a test case.
+  ///
+  /// Instances of this type are considered unique within the scope of a given
+  /// parameterized test function. They are not necessarily unique across two
+  /// different ``Test`` instances.
+  public struct ID: Sendable, Equatable, Hashable {
```
grynspan, line 18: Seems to me that a test case's ID should include the ID of the test, otherwise the test cases here are going to compare equal by ID:

```swift
@Test("is negative", arguments: [0, 1, 2]) func isNegative(...)
@Test("is positive", arguments: [0, 1, 2]) func isPositive(...)
```

Sources/Testing/Parameterization/Test.Case.ID.Selection.swift
grynspan: I'm suggesting they're all the same thing.

Sources/Testing/Parameterization/Test.Case.ID.swift
stmontgomery: I made note of this in the symbol docs for this `ID` type: the scope of uniqueness is only within a test case's associated test—it's not intended to be globally unique across any arbitrary test and I don't think it needs to be. The [documentation](https://developer.apple.com/documentation/swift/identifiable) for the `Identifiable` protocol allows adopters to define their own semantics around this (**emphasis** mine):

> `Identifiable` leaves the duration and scope of the identity unspecified. Identities can have any of the following characteristics:
> - Guaranteed always unique, like UUIDs.
> - Persistently unique per environment, like database record keys.
> - Unique for the lifetime of a process, like global incrementing integers.
> - Unique for the lifetime of an object, like object identifiers.
> - **Unique within the current collection, like collection indices.**

Sources/Testing/Parameterization/CustomTestArgument.swift
stmontgomery: The concept of "a context" to provide additional info is pretty broad, and these types we have serve different purposes. I nested this one to keep it fairly out-of-the-way, so I'm not too concerned by having multiple of these for different roles.

briancroom (approved):

Sources/Testing/Running/Configuration.swift
```diff
@@ -162,4 +162,33 @@ public struct Configuration: Sendable {
...
+  /// A function that handles filtering test cases.
+  ///
+  /// - Parameters:
+  ///   - testCase: The test case to be filtered.
+  ///   - test: The test which `testCase` is associated with.
+  ///
+  /// - Returns: A Boolean value representing if the test case satisfied the
+  ///   filter.
+  @_spi(ExperimentalParameterizedTesting)
+  public typealias TestCaseFilter = @Sendable (_ testCase: Test.Case, _ test: Test) -> Bool
```
grynspan, line 175: Should this take a single argument of type `Event.Context` (or whatever we called it)?

Sources/Testing/Parameterization/Test.Case.ID.swift
grynspan: Right, I am aware that `Identifiable` leaves that up to the implementation, I'm just suggesting that it might actually be important to us to track the test ID here too.

Sources/Testing/Running/Configuration.swift
stmontgomery: I don't think they're the same thing semantically: We only form an `Event.Context` once we have an `Event`, but here we're deciding whether to include a test case, which happens before we begin creating events for it. Plus this aligns more with `TestFilter` typealias which accepts a `Test`.

Sources/Testing/Parameterization/Test.Case.ID.swift
stmontgomery: Open to doing that but not sure it's something we need currently, and it introduces a little overhead so I'd prefer to wait until we know a specific scenario which justifies it. The places where these show up are generally paired with a `Test` (or `Test.ID`) already, e.g. in events (or their snapshots).

stmontgomery:
@swift-ci please test

Sources/Testing/Parameterization/CustomTestArgument.swift
```diff
@@ -0,0 +1,151 @@
...
+
+/// Create an encodable wrapper for a value which conforms to
+/// ``CustomTestArgumentEncodable``.
+///
+/// - Parameters:
+///   - value: The value which implements custom test argument encoding logic.
+///   - context: The context in which the custom test argument was used.
+///
+/// - Returns: An encodable wrapper for the specified value.
+private func _customArgumentWrapper(for value: some CustomTestArgumentEncodable, in context: Test.Case.Argument.Context) -> some Encodable {
```
grynspan, line 137: I think this function may be redundant? Can't you just call the constructor for `_CustomArgumentWrapper` directly?

Sources/Testing/Parameterization/CustomTestArgument.swift
```diff
@@ -0,0 +1,151 @@
...
+  /// - Throws: Any error encountered while attempting to encode `value`.
+  ///
+  /// This function is not part of the public interface of the testing library.
+  ///
+  /// ## See Also
+  ///
+  /// - ``CustomTestArgumentEncodable``
+  init?(identifying value: some Sendable, in context: Test.Case.Argument.Context) throws {
+#if canImport(Foundation)
+    guard Configuration.current?.isTestArgumentEncodingEnabled ?? false else {
```
grynspan, line 69: `Configuration` will be `nil` on detached tasks and non-Swift-owned threads.

Sources/Testing/Parameterization/CustomTestArgument.swift
stmontgomery: Tried that first but the compiler doesn't open the existential in an initializer, it seems

grynspan: You can make it a local function in the one place that calls it, I think. No need for a whole symbol to open an existential. (We do that in a few other spots.)

Sources/Testing/Parameterization/CustomTestArgument.swift
stmontgomery: Understood, I don't think that will be a practical concern in the contexts we care about here

stmontgomery:
@swift-ci please test

Sources/Testing/Parameterization/CustomTestArgument.swift
```diff
@@ -0,0 +1,154 @@
...
+  /// ## See Also
+  ///
+  /// - ``CustomTestArgumentEncodable``
+  init?(identifying value: some Sendable, parameter: Test.ParameterInfo) throws {
+#if canImport(Foundation)
+    guard Configuration.current?.isTestArgumentEncodingEnabled ?? false else {
+      return nil
+    }
+
+    func _customArgumentWrapper(for value: some CustomTestArgumentEncodable) -> some Encodable {
```
grynspan, line 73: You can shuffle this down right into the `if` branch that uses it. And drop the leading underscore.

Sources/Testing/Parameterization/CustomTestArgument.swift
```diff
@@ -0,0 +1,154 @@
...
+    return .init(try encoder.encode(value))
+  }
+#endif
+}
+
+/// A encodable type which wraps a ``CustomTestArgumentEncodable`` value.
+private struct _CustomArgumentWrapper<T>: Encodable where T: CustomTestArgumentEncodable {
+  /// The value this instance wraps, which implements custom test argument
+  /// encoding logic.
+  var value: T
```
grynspan, line 128: `RawRepresentable` might be appropriate here, but it's not critical.

Sources/Testing/Parameterization/CustomTestArgument.swift
```diff
@@ -0,0 +1,154 @@
...
+}
+
+extension Encoder {
+  /// The test parameter which the test argument being encoded was passed to, if
+  /// any.
+  ///
+  /// The value of this property is non-`nil` when this encoder is being used to
+  /// encode an argument passed to a parameterized test function.
+  @_spi(ExperimentalParameterizedTesting)
+  public var parameter: Test.ParameterInfo? {
```
grynspan, line 151: Since we're extending a stdlib type here, we should pick a less generic property name. How about `testParameterInfo`?

Sources/Testing/Parameterization/Test.Case.ID.swift
```diff
@@ -0,0 +1,64 @@
...
+    /// The IDs of the arguments of this instance's associated ``Test/Case``, in
+    /// the order they appear in ``Test/Case/arguments``.
+    ///
+    /// The value of this property is `nil` if _any_ of the associated test
+    /// case's arguments has a `nil` ID.
+    public var argumentIDs: [Argument.ID]?
```
grynspan, line 24: What's the difference between a `nil` array and an empty array here?

Sources/Testing/Parameterization/Test.Case.swift
```diff
@@ -55,13 +78,16 @@ extension Test {
...
+    /// - Throws: Any error encountered attempting to encode test arguments.
     init(
       values: [any Sendable],
       parameters: [ParameterInfo],
       body: @escaping @Sendable () async throws -> Void
-    ) {
-      let arguments = zip(values, parameters).map { value, parameter in
-        Argument(value: value, parameter: parameter)
+    ) throws {
+      let arguments = try zip(values, parameters).map { value, parameter in
```
grynspan, line 88: If an error occurs here, it might not be a test failure. For instance, if a particular input is of a codable type but is not safely encodable in this context (e.g. XPC-codable-only types?), we shouldn't fail the test, we should just fail to identify the argument (thus it's not re-runnable independently, but it's still _testable_.)

Sources/Testing/Running/Configuration.swift
```diff
@@ -162,4 +162,33 @@ public struct Configuration: Sendable {
...
+  /// test's associated test cases to determine if they should run. If the
+  /// result of invoking this property is `nil`, then all of the test cases of
+  /// the test will be run (i.e. no filter will be applied).
+  ///
+  /// ## See Also
+  ///
+  /// - ``testFilter-swift.property``
+  /// - ``setTestFilter(toMatch:)``
+  @_spi(ExperimentalParameterizedTesting)
+  public var testCaseFilter: @Sendable (_ test: Test) -> TestCaseFilter? = { _ in nil }
```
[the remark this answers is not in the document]
grynspan: I'm still confused about this part.

Sources/Testing/Parameterization/CustomTestArgument.swift
stmontgomery: I'll go with `testParameter`, since elsewhere we just call it "parameter" (despite its formal type including the word "Info")

Sources/Testing/Parameterization/CustomTestArgument.swift
stmontgomery: Done!

Sources/Testing/Parameterization/CustomTestArgument.swift
stmontgomery: Can't because I am using [SE-0380 (If and switch expressions)](https://github.com/apple/swift-evolution/blob/main/proposals/0380-if-switch-expressions.md). Dropped the leading underscore

Sources/Testing/Parameterization/Test.Case.swift
stmontgomery: Ok I can buy that. I do want to capture and reflect the error back to the user somehow, so they know that there was an error during encoding that will prevent them from re-running, but we can track that as a separate enhancement.

Sources/Testing/Running/Configuration.swift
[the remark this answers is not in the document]
stmontgomery: I structured it this way for efficiency: For the common case where a particular test does not wish to filter to only run _specific_ test cases, `Runner` will not need to repeatedly invoke the same filter closure for each of that test's cases. It can instead invoke this outer `testCaseFilter` once for each `Test`, and then only if it returned a non-`nil` `TestCaseFilter` closure does it need to repeatedly invoke the returned closure to decide whether each test _case_ should run.

In other words it allows a client to more efficiently indicate, "Do not perform any filtering of the test cases of [TestX]".

stmontgomery:
@swift-ci please test

Sources/Testing/Parameterization/CustomTestArgument.swift
stmontgomery: I have adjusted the code further to change where we make this determination so any problems here should be even less likely now, and match how `Event` handles deciding which `Configuration` to use

Sources/Testing/Running/Configuration.swift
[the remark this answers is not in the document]
grynspan: Seems like that can be accomplished by just passing both the test and test case into the closure, no? Let the caller worry about the nuances like that since they'll be caller-specific.

Sources/Testing/Events/Event.swift
```diff
@@ -158,11 +168,23 @@ public struct Event: Sendable {
     instant: Test.Clock.Instant = .now,
     configuration: Configuration? = nil
   ) {
-    // Create both the event and its associated context here at same point, to
-    // ensure their task local-derived values are the same.
-    let event = Event(kind, testID: test?.id, instant: instant)
     let context = Event.Context(test: test, testCase: testCase)
-    event._post(in: context, configuration: configuration)
+
+    withConfiguration(configuration) { configuration in
```
grynspan, line 173: This really seems like the wrong way to handle this since the configuration may well not be available or knowable here, and you're using it _just_ for `isTestArgumentEncodingEnabled`. I'd expect that this property would generally be `true`, and having it be unexpectedly `false` (because e.g. an event is posted from a detached task) has spooky effects.

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Configuration.swift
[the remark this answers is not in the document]
stmontgomery: Ok I think this is acceptable, since in the circumstances where this is used, we will typically _also_ be filtering to only run the specific tests which contain those cases

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

PR #121, opened 2023-11-16
Avoid runtime casts to parameterized protocol types, which require newer OS versions

stmontgomery:
This fixes a build failure seen when attempting to build for macOS/iOS versions older than 13.0/16.0, respectively, which was introduced in #104. This introduces a type-erasing wrapper, which fixes the failure and also improves the `Test.testCases` derived property by allowing it to use `some` instead of `any` and further constraining the return type to `Sendable`.

### Checklist:


stmontgomery:
@swift-ci please test

dennisweissmann (approved):

briancroom (approved):

PR #122, opened 2023-11-17
Remove synthesized `@available(*, unavailable)` attributes from test types.

grynspan:
This PR removes the `@available(*, unavailable)` attribute we add to synthesized `__TestContainer`-conforming types. Although we don't intend for these symbols to be usable outside the testing library, when building a test executable instead of a .xctest bundle on Darwin, this attribute causes us to crash with an error cascade:

```
Fatal error: Unavailable code reached
Fatal error: Unavailable code reached
Fatal error: Unavailable code reached
Fatal error: Unavailable code reached
Fatal error: Unavailable code reached
Fatal error: Unavailable code reached
error: Exited with unexpected signal code 5
error: ExitCode(rawValue: 1)
```

Darwin now behaves the same as Linux/etc. here, which avoids the fatal error.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #124, opened 2023-11-17
Remove the `_TestCases` and `Cases` types.

grynspan:
This PR removes a chunk of infrastructure that we use to type-erase instances of `Test.Case.Generator<S>`. What we _really_ want to do is express `any Sequence<Test.Case> & Sendable`, however compiler work is required to make that possible ([96960993](rdar://96960993)).

In lieu of that compiler work, and instead of rolling our own type-erasing box, we can just use [`AnySequence<Test.Case>`](https://github.com/apple/swift/blob/main/stdlib/public/core/ExistentialCollection.swift#L1246) for storage. Combine it with `UncheckedSendable` (since we know our sequence really is sendable and the implementation of `AnySequence` is safely trivial) and we get what we want without needing to build it ourselves.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #127, opened 2023-11-18
The details/disclosure arrow used for comments should be a `Symbol`.

grynspan:
Internal bookkeeping: we use a Unicode/SF Symbols symbol to represent "details" about an event such as its associated comments. We should represent it as a `Symbol` rather than as a bare Unicode character, to match how we represent other symbols.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

dennisweissmann (approved):

PR #128, opened 2023-11-18
`swift test` should warn when `--filter` excludes all tests (_à la_ XCTest.)

grynspan:
When using XCTest and `swift test --filter`, if all tests are filtered out a diagnostic of the form:

> warning: No matching test cases were run

Is emitted by SwiftPM. We should have a similar warning we emit from swift-testing in this scenario.

### Checklist:


grynspan:
Draft for now because we may end up needing to implement the relevant logic in a single location in SwiftPM so that the message reflects the _overall_ output, not per-library output. TBD.

grynspan:
No longer required due to #536 and https://github.com/swiftlang/swift-package-manager/pull/7777.

PR #129, opened 2023-11-20
Tweaks to some feature conditionals.

grynspan:
This PR replaces `SWT_NO_TAG_COLORS` with `SWT_NO_FILE_IO` and applies that conditional elsewhere in the package where we perform file I/O. It also fixes `!SWT_NO_UTC_CLOCK` gating availability of `Test.Clock.Instant.nanoseconds(until:)`, which was not intentional.

### Checklist:


grynspan:
@swift-ci please test

PR #131, opened 2023-11-21
(Tests only) Conditionalize imports of Foundation and always import TestingInternals as private.

grynspan:
This PR conditionalizes imports of Foundation and always import TestingInternals as private within our own test target. This is necessary because some environments may not have Foundation available during testing, and because the lack of `private` on `import TestingInternals` leads to ambiguity and a build error when our targets are statically linked together.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -43,10 +43,7 @@ let package = Package(
         "TestingInternals",
         "TestingMacros",
       ],
-      swiftSettings: .packageSettings + [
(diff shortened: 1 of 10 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #133, opened 2023-11-23
Do not evaluate unavailable test function parameters.

grynspan:
This PR causes swift-testing to capture the arguments to parameterized tests as closures (returning sequences) rather than as sequences. These closures are then lazily evaluated during test execution. This has two broad effects:

1. The cost of evaluating these properties is deferred to after the start of a test instead of during the planning stage (a good thing, probably!); and
2. If a test's parameters are unavailable on the current system (due to appropriate `@available` attributes), they are not evaluated, avoiding a crash. This crash would not normally happen in pure Swift code, but can happen when using swift-testing because we go "off the rails" to discover test content at runtime (see Discovery.cpp).

Resolves rdar://118518234.

### Checklist:

### DIFF
--- a/Sources/Testing/Test.Case.Generator.swift
+++ b/Sources/Testing/Test.Case.Generator.swift
@@ -19,13 +19,17 @@ extension Test.Case {
   ///     ([103416861](rdar://103416861))
   /// }
   struct Generator<S>: Sendable where S: Sequence & Sendable, S.Element: Sendable {
-    /// The underlying sequence of argument values.
+    /// A closure that produces the underlying sequence of argument values.
     ///
-    /// The sequence _must_ be iterable multiple times. Hence, initializers
-    /// accept only _collections_, not sequences. The constraint here is only to
-    /// `Sequence` to allow the storage of computed sequences over collections
-    /// (such as `CartesianProduct` or `Zip2Sequence`) that are safe to iterate
-    /// multiple times.
+    /// The resulting sequence _must_ be iterable multiple times. Hence,
+    /// initializers accept only _collections_, not sequences. The constraint
+    /// here is only to `Sequence` to allow the storage of computed sequences
+    /// over collections (such as `CartesianProduct` or `Zip2Sequence`) that are
+    /// safe to iterate multiple times.
+    ///
+    /// This property is a closure rather than an instance of `S` so that it can
+    /// be lazily evaluated rather than requiring evaluation as soon as the
+    /// owning instance of ``Test`` is initialized.
     private var _sequence: @Sendable () async -> S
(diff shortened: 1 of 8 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
TODO: ensure `unavailable`, `swift`, and `deprecated` all work correctly. (For `unavailable` at least, we can drop the parameters entirely.)

grynspan:
It's not possible to construct a test function with arguments of `unavailable` type, so no changes are necessary to guard them.

grynspan:
Arguments with `deprecated` types work fine, however they generate a compiler diagnostic, e.g.:

```swift
@Suite(.hidden) struct DeprecatedArguments {
  @available(*, deprecated)
  struct Arg: Sendable {}

  @available(*, deprecated)
  @Test(.hidden, arguments: [Arg(), Arg(), Arg()]) // ⚠️ 'Arg' is deprecated
  func f(arg: Arg) {}
}
```

I filed a bug report (rdar://118832534) internally at Apple to track a solution.

grynspan:
If availability is gated on language version instead, this becomes an error:

```swift
@Suite(.hidden) struct DeprecatedArguments {
  @available(swift 999) // ℹ️ 1. 'Arg' was introduced in Swift 999
  struct Arg: Sendable {}

  @available(swift 999)
  @Test(.hidden, arguments: [Arg(), Arg(), Arg()]) // 🛑 'Arg' is unavailable
  func f(arg: Arg) {}
}
```

Same basic problem as with `deprecated`. Workaround is to gate the test with `#if swift(>=999)` instead of `@available`.

Sources/Testing/Test.Case.Generator.swift
```diff
@@ -26,7 +26,7 @@ extension Test.Case {
     /// `Sequence` to allow the storage of computed sequences over collections
     /// (such as `CartesianProduct` or `Zip2Sequence`) that are safe to iterate
     /// multiple times.
-    private var _sequence: S
+    private var _sequence: @Sendable () async -> S
```
stmontgomery, line 29: I think this, and the modified parameters of the initializers below, might need to have the DocC updated to reflect that it's a function which returns the Sequence now. And maybe this could explain the reason why it's a closure

stmontgomery (approved):

Sources/Testing/Test.Case.Generator.swift
grynspan: Done.

grynspan:
@swift-ci please test

PR #132, opened 2023-11-23
Import SwiftSyntax as `public`.

grynspan:
We have adopted the upcoming [access levels on imports](https://github.com/apple/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md) feature in swift-testing. We use some types from the SwiftSyntax module in our macro target (as part of our conformances to the various macro protocols.)

We saw some build failures locally:

```
[...]/Sources/TestingMacros/SuiteDeclarationMacro.swift:28:22: error: static method cannot be declared public because its generic parameter uses an internal type
public static func expansion(
                   ^
 [...]/.build/checkouts/swift-syntax/Sources/SwiftSyntax/generated/SyntaxBaseNodes.swift:22:17: note: type declared here
public protocol DeclSyntaxProtocol: SyntaxProtocol {}
                ^
```

We can mark our `import SwiftSyntax` statements as `public` to avoid this issue. This change is only necessary in the macros target (not our tests or the GitStatus plugin), but must occur in every file in that target that imports SwiftSyntax.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #134, opened 2023-11-28
Import SwiftSyntaxMacros as `public`.

grynspan:
This PR marks all imports of the SwiftSyntaxMacros module as `public`. This change is necessary to resolve build failures of the same flavour as #132.

### Checklist:

### DIFF
--- a/Tests/TestingMacrosTests/ConditionMacroTests.swift
+++ b/Tests/TestingMacrosTests/ConditionMacroTests.swift
@@ -16,7 +16,7 @@ import SwiftOperators
 import SwiftParser
 public import SwiftSyntax
 import SwiftSyntaxBuilder
-import SwiftSyntaxMacros
+public import SwiftSyntaxMacros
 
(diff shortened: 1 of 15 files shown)


Tests/TestingMacrosTests/ConditionMacroTests.swift
```diff
@@ -16,7 +16,7 @@ import SwiftOperators
 import SwiftParser
 public import SwiftSyntax
 import SwiftSyntaxBuilder
-import SwiftSyntaxMacros
+public import SwiftSyntaxMacros
```
stmontgomery, line 19: I know why we need `public import` in `TestingMacros`, but why do we need it in any files in `TestingMacrosTests`?

grynspan: Static linkage on some platforms may cause conflicts, so best to just be consistent.

PR #136, opened 2023-11-28
Fix access levels of import statements to ensure compatbility with latest 'main' snapshot toolchains

stmontgomery:
This PR modifies the access levels of `import` declarations in our macro target and its test target, so that they compile cleanly with the latest (2023-11-27) main Swift toolchain. It also removes `public` access level from the code expanded by the `@Test` and `@Suite` attached macros, because these cause new diagnostics in that newer toolchain since the expanded code references declarations in the `Testing` module and typically that module is not imported in client code using `public`.

### Checklist:


dennisweissmann (approved):

stmontgomery:
@swift-ci please test

grynspan (approved):

PR #137, opened 2023-11-29
Move test discovery code out of Test+Macros.swift.

grynspan:
This PR moves test discovery code from Test+Macros.swift to a new Test+Discovery.swift file and tweaks the signature of the supporting C++ function a bit to make it easier to read when used in Swift.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

dennisweissmann (approved):

Sources/TestingInternals/include/Discovery.h
```diff
@@ -49,7 +49,11 @@ typedef bool (* SWTTypeNameFilter)(const char *typeName, void *_Null_unspecified
 /// it is present in an image's metadata table multiple times, or if it is an
 /// Objective-C class implemented in Swift.) Callers are responsible for
 /// deduping type metadata pointers passed to `body`.
-SWT_EXTERN void swt_enumerateTypes(SWTTypeNameFilter _Nullable nameFilter, SWTTypeEnumerator body, void *_Null_unspecified context);
+SWT_EXTERN void swt_enumerateTypes(
+  void *_Null_unspecified context,
+  SWTTypeEnumerator body,
+  SWTTypeNameFilter _Nullable nameFilter
+) SWT_SWIFT_NAME(swt_enumerateTypes(_:_:withNamesMatching:));
```
stmontgomery, line 56: No biggie but it seems like if you are customizing its Swift name, you could drop the `swt_` prefix since in Swift it will import qualified with the module name. Maybe that is a pattern we ought to establish and apply more broadly
```suggestion
) SWT_SWIFT_NAME(enumerateTypes(_:_:withNamesMatching:));
```

grynspan: I intentionally left it that way since it's still a C++ function and I don't want to lose sight of that. If we ever need to call this in another place, I intend to abstract it with a Swift function so it can directly take Swift closures instead of C functions plus a context pointer.

PR #140, opened 2023-11-29
[WIP] Remove "`swt_`" and "`SWT`" prefixes from C/C++ symbols in Swift.

grynspan:
This PR adopts `SWT_SWIFT_NAME()` across TestingInternals so that symbols imported into Swift from that module do not have to be called with the "SWT" prefix.

### DIFF
--- a/Sources/TestingInternals/include/Discovery.h
+++ b/Sources/TestingInternals/include/Discovery.h
@@ -23,7 +23,7 @@ SWT_ASSUME_NONNULL_BEGIN
 ///   - typeMetadata: A type metadata pointer that can be bitcast to `Any.Type`.
 ///   - context: An arbitrary pointer passed by the caller to
 ///     `swt_enumerateTypes()`.
-typedef void (* SWTTypeEnumerator)(void *typeMetadata, void *_Null_unspecified context);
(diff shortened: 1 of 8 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

Sources/TestingInternals/include/Discovery.h
```diff
@@ -53,7 +53,7 @@ SWT_EXTERN void swt_enumerateTypes(
   void *_Null_unspecified context,
   SWTTypeEnumerator body,
   SWTTypeNameFilter _Nullable nameFilter
-) SWT_SWIFT_NAME(swt_enumerateTypes(_:_:withNamesMatching:));
+) SWT_SWIFT_NAME(enumerateTypes(_:_:withNamesMatching:));
```
stmontgomery, line 56: Sort of unrelated, but now this is making me wonder why we don't give the first parameter a label like `context:`, to help explain its purpose?

stmontgomery (approved):

grynspan:
Doesn't look like we can reliably apply this change across platforms, so wontfix.

PR #141, opened 2023-11-29
Rename the SwiftPM entry point.

grynspan:
This PR renames the SwiftPM entry point function such that it does not need to be SPI. By marking it SPI, we unintentionally made it very difficult to use with a toolchain that does not include private module information.

The PR continues to export the old symbol name as used by SwiftPM; once we've updated SwiftPM to match, we will remove the old function.

### Checklist:


grynspan:
@swift-ci please test

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -68,10 +66,21 @@ public func swiftPMEntryPoint() async -> CInt {
...
+
+/// Call `__swiftPMEntryPoint()`.
+///
+/// This function is an old alias of `__swiftPMEntryPoint()`. It will be removed
+/// in the near future.
+///
+/// - Warning: This function is used by Swift Package Manager. Do not call it
+///   directly.
 @_spi(SwiftPackageManagerSupport)
 public func swiftPMEntryPoint() async -> Never {
```
stmontgomery, line 82: deprecate too?

stmontgomery (approved):

Sources/Testing/Running/EntryPoint.swift
grynspan: No, it'll just produce noise right now.

PR #144, opened 2023-11-29
Make events and associated types move-only.

grynspan:
This PR makes `Event`, `Event.Context`, `Event.ConsoleOutputRecorder`, and `Event.JUnitXMLRecorder` move-only (`~Copyable`.)

These types are fairly expensive to create and copy. By making them move-only, we teach the compiler to help us avoid making expensive copies.

### Checklist:


grynspan:
@swift-ci please test

PR #145, opened 2023-11-29
Split up `ConsoleOutputRecorder` into two parts.

grynspan:
This PR splits `Event.ConsoleOutputRecorder` such that it only contains the code necessary to produce _console output_ (that is, output destined for a terminal with ANSI escape codes, etc.)

The logic to produce human-readable text is moved to a separate type, `Event.HumanReadableOutputRecorder`. This type has no knowledge of where the text it generates will be rendered or consumed.

By splitting this type into two pieces, we (or external developers) can readily create new event-recording mechanisms that use the same text we do, but which translate that text to different representations such as HTML or `AttributedString`.


grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please clean test Linux

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test Linux

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -0,0 +1,378 @@
...
+extension Event {
+  /// A type which handles ``Event`` instances and outputs representations of
+  /// them as human-readable messages.
+  ///
+  /// This type can be used compositionally to produce output in other
+  /// human-readable formats such as rich text or HTML.
+  ///
+  /// The format of the output is not meant to be machine-readable and is
+  /// subject to change. For machine-readable output, use ``JUnitXMLRecorder``.
+  public struct HumanReadableOutputRecorder: Sendable {
```
stmontgomery, line 20: I'd like to propose we establish a distinct `@_spi` group for this, such as `@_spi(ExperimentalEventRecording)`. I realize `Event` is already under an SPI group but this is carving out a new subsystem within that domain and it seems appropriate to distinguish them. Also "event _handling_" doesn't feel quite fitting for what this is doing.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -0,0 +1,378 @@
...
+      var testData = Graph<String, TestData?>()
+    }
+
+    /// This event recorder's mutable context about events it has received,
+    /// which may be used to inform how subsequent events are written.
+    @Locked private var _context = _Context()
+
+    /// Initialize a new human-readable event recorder.
+    ///
+    /// - Parameters:
```
stmontgomery, line 67: Do zero-param declarations actually need `- Parameters:` ?

Tests/TestingTests/EventRecorderTests.swift
stmontgomery, line 1: Should we rename this file too?

Sources/Testing/Events/Recorder/Event.Symbol.swift
```diff
@@ -0,0 +1,171 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+extension Event {
+  /// An enumeration describing the symbols used as prefixes when recording an
+  /// event.
+  public enum Symbol: Sendable {
```
stmontgomery, line 14: If you do adopt the suggestion elsewhere about a new `@_spi` group, I imagine it ought to be applied here, too

stmontgomery, line 14: The name "Symbol", on its own in this nested context, feels a little unclear. Can we consider alternatives? I suspect people won't need to type the name often, and will typically just receive an instance of this value, so maybe a longer name like `Event.RecorderMessageSymbol` would be acceptable?

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: That's fine with me.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: Copy-and-paste error. :)

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -0,0 +1,378 @@
...
+  /// Record the specified event by generating zero or more messages that
+  /// describe it.
+  ///
+  /// - Parameters:
+  ///   - event: The event to record.
+  ///   - eventContext: The context associated with the event.
+  ///
+  /// - Returns: An array of zero or more messages that can be displayed to the
+  ///   user.
+  @discardableResult public func record(_ event: borrowing Event, in eventContext: borrowing Event.Context) -> [Message] {
```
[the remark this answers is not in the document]
grynspan: The only difference would be that the tag color dots logic has been hoisted out. Otherwise yes, it's producing the same output.

Tests/TestingTests/EventRecorderTests.swift
grynspan: Eh, it's fine since it covers a few different types of event recorder at this point.

Sources/Testing/Events/Recorder/Event.Symbol.swift
grynspan: It's the symbol to use for an event, no?

I think putting it in the new `@_spi` category plus being nested in `Event` is a pretty good hint.

grynspan:
@swift-ci please test

stmontgomery (approved):

grynspan:
@swift-ci please test

PR #148, opened 2023-12-03
Introduce snapshot types for Runner.Plan and its supporting types

stmontgomery:
This makes several of the types which represent runner plans serializable, to facilitate integration with tools.

### Modifications:

- Add the following snapshot types:
  - `Runner.Plan.Snapshot`
  - `Runner.Plan.Step.Snapshot`
  - `Runner.Plan.Action.Snapshot`
- Add direct conformance by `SkipInfo` to `Equatable` and `Codable` since it is a trivial type.

### Checklist:

Resolves rdar://91900185

### DIFF
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -55,7 +55,7 @@ extension Runner {
       /// The test to be passed to an instance of ``Runner``.
       public var test: Test
 
-      /// The action to perform with ``Test/Plan/Step/test``.
+      /// The action to perform with ``test``.
       public var action: Action
     }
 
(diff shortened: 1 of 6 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -282,3 +282,132 @@ extension Runner.Plan {
...
+    public var steps: some Collection<Step.Snapshot> {
+      _stepGraph.compactMap(\.value)
+    }
+  }
+}
+
+extension Runner.Plan.Snapshot: Codable {
+  /// The coding keys used for serializing a
+  /// ``Runner/Plan-swift.struct/Snapshot`` instance.
+  private enum _CodingKeys: CodingKey {
```
grynspan, line 320: Would it help if we made `Graph` conditionally `Codable`? It ought to be trivial to do so.

grynspan (approved):

Sources/Testing/Running/Runner.Plan.swift
stmontgomery: Yeah I'm not opposed to that as a general enhancement, but for this usage in particular it felt reasonable for the serialized representation of this to just be a simpler, array-like data data structure. It feels a bit more conservative and safer for us to encode it that way, since it reserves our ability to evolve `Graph`'s implementation further without affecting serialization. So for now I plan to proceed as-is, but we can keep thinking about it.

PR #153, opened 2023-12-05
Replace `setTestFilter` with `makeTestFilter`.

grynspan:
`setTestFilter` is too rigid because it replaces the existing test filter and that makes composition harder than it needs to be. Let's replace it with a function that returns a test filter which can then be composed:

```swift
let idFilter = makeTestFilter(matching: myTestIDs)
func canHaveAPony(_ test: Test) -> Bool { ... }
configuration.testFilter = { test in
  idFilter(test) && canHaveAPony(test)
}
```

Resolves rdar://119205417.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -118,7 +118,7 @@ public struct Configuration: Sendable {
   /// - Returns: A Boolean value representing if the test satisfied the filter.
   public typealias TestFilter = @Sendable (_ test: Test) -> Bool
 
-  /// Storage for ``testFilter``.
+  /// Storage for ``testFilter-swift.property``.
   private var _testFilter: TestFilter = { !$0.isHidden }
 
   /// The test filter to which tests should be filtered when run.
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci please test

Sources/Testing/Running/Configuration.swift
```diff
@@ -180,3 +168,36 @@ public struct Configuration: Sendable {
   @_spi(ExperimentalParameterizedTesting)
   public var testCaseFilter: TestCaseFilter = { _, _ in true }
 }
+
+// MARK: - Test filter factory functions
+
+/// Filter tests to run to those specified via a set of test IDs.
```
stmontgomery, line 174: Can we update the wording of these two factory functions? Now, they don't _perform_ the filtering, but only generate a filter

stmontgomery (approved):
LGTM aside from documentation suggestion

Sources/Testing/Running/Configuration.swift
grynspan: Oh, sure.

grynspan:
@swift-ci please test

suzannaratcliff (approved):

PR #159, opened 2023-12-12
Mark `XCTestScaffold` to-be-deprecated.

grynspan:
Currently, the documentation for `XCTestScaffold` has a "warning" callout to indicate that it is a temporary interface. This PR replaces the callout with a formal deprecation attribute so that the type and its members show as deprecated in our documentation. They are marked as deprecated in Swift 100000.0, which is a "magic" version number used in Apple's SDK to indicate "to-be-deprecated" status.

This change does not cause new diagnostics to be emitted at compile-time; it affects documentation only.

#### Before

Screenshot 2023-12-12 at 6 47 40 PM

#### After

Screenshot 2023-12-12 at 6 51 36 PM

And, when referenced from another article:

Screenshot 2023-12-12 at 6 46 11 PM

### Checklist:

### DIFF
--- a/Sources/Testing/Running/XCTestScaffold.swift
+++ b/Sources/Testing/Running/XCTestScaffold.swift
@@ -86,13 +86,14 @@ extension XCTIssue {
 /// A type providing temporary tools for integrating the testing library and
 /// the XCTest framework.
 ///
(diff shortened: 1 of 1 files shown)


stmontgomery (approved):

grynspan:
@swift-ci please test

PR #161, opened 2023-12-14
Expand property accesses in `#expect()` and `#require()`.

grynspan:
Currently, swift-testing is able to expand member function accesses like `foo.bar()` into `foo` and `bar()` so that it can provide the values of both subexpressions on expectation failure. This PR enables the same sort of functionality for property accesses (`foo.isBar`).

This PR also ensures that optional chaining in either member function calls or property accesses is observed so that an expression like `foo?.bar()` will not fail to compile.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):
Nice, thanks!

PR #163, opened 2023-12-19
Report an error when attempting to use --filter or --skip in unsupported environments

stmontgomery:
This modifies the SwiftPM entry point argument parsing logic to ensure that if a user passes the `--filter` or `--skip` flags on an OS version which does not support those features, an error is reported.

### Motivation:

Currently, we silently ignore these flags on unsupported OS versions but that can cause problems if the user was only expecting certain tests to run, so this ensures that we either honor the user's request or report a meaningful failure message.

### Modifications:

- Implement the `else` branch of `if #available(...)` conditions which guard the implementations of these CLI flags.
- Adjust tests accordingly.

### Checklist:

Resolves rdar://119837856


stmontgomery:
@swift-ci please test

briancroom (approved):

PR #169, opened 2024-01-04
Remove redundant `Codable` implementation from `Tag`.

grynspan:
The default synthesized implementation of `Codable` is equivalent to what we've manually implemented for `Tag`, so we can delete the manual implementation.

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #170, opened 2024-01-04
Make sure events are borrowed when snapshotting.

grynspan:
`Event.Snapshot.init(snapshotting:)` needs to borrow its argument (since we want to explicitly avoid copying `Event` instances.)

### Checklist:


dennisweissmann (approved):

stmontgomery (commented):
Thanks!

PR #177, opened 2024-01-10
Don't shadow `VersionTupleSyntax.components`.

grynspan:
Since we first wrote the code in VersionTupleSyntaxAdditions.swift, a `components` property has been added to `VersionTupleSyntax`. We shouldn't shadow it with our own property (and we should call it in order to more efficiently and correctly get version info than we do now.)

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #178, opened 2024-01-10
Remove `Test.underestimatedCaseCount` SPI

stmontgomery:
This removes the `Test.underestimatedCaseCount` SPI property.

### Motivation:

We originally thought this property might be needed or useful but it hasn't been necessary. Nowadays, `Test` has another property named `isParameterized` which can be used to more directly query whether a test is parameterized, and there are no usages within the testing library's own implementation.

### Checklist:


stmontgomery:
@swift-ci please test

grynspan (approved):

briancroom (approved):

PR #180, opened 2024-01-11
Optionally capture source code representations for all traits.

grynspan:
The current macro expansion implementation for `@Test` and `@Suite` checks for known patterns corresponding to _tags_ (i.e. the `.tags()` trait) and injects additional code to capture the source code of each tag.

This change generalizes that logic by adding a requirement to the `Trait` protocol that the macro expansion calls for _all_ traits. It passes the source code of the trait expression as well as the source code of each argument to the trait (if it's a function call.) The macro expansion thus no longer needs to recognize the `Tag.List` trait type specifically.

The source code representations are captured as autoclosures to avoid evaluating them if not needed. The default implementation of the requirement ignores them, so for a typical trait there is no effect. `Tag.List` implements the requirement and maps the argument source code representations to its tags.

This change also removes the overload of `.tags()` that takes a sequence as it makes expressions like `.tags(foo)` ambiguous.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
Rebased.

grynspan:
Rebased.

grynspan:
@swift-ci please test

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,22 @@ public protocol Trait: Sendable {
...
+  /// - Parameters:
+  ///   - sourceCode: The source code representation of the entire trait.
+  ///   - arguments: Source code representations of each argument to the trait,
+  ///     if any.
+  ///
+  /// - Returns: A modified copy of `self`.
+  ///
+  /// The default implementation of this method does nothing and returns `self`
+  /// unmodified.
+  func addingSourceCode(
```
stmontgomery, line 56: This is intended for use only by the macros, right? How about we double underscore-prefix it, but keep it `public`, to reinforce that and avoid potentially misleading users into thinking this is something they might want to use?

grynspan: Discussed offline. It's experimental API, since third parties could write traits that want access to their source code representations, but we can't use `@_spi` in this position reliably so we'll mark it with an underscore and document that it is subject to change.

grynspan:
@swift-ci please test

stmontgomery (approved):

grynspan:
@swift-ci please test macOS

PR #184, opened 2024-01-16
Console output should prefix messages with "Suite" instead of "Test" for suites

stmontgomery:
This changes the console output to prefix messages with the word "Suite" instead of "Test" for events representing suites.

### Motivation:

It's helpful to users to distinguish between test functions (aka "tests") and suite types (aka "suites") in the human-readable console output shown while running tests, to clarify the role each item plays.

### Modifications:

- Change logic in `Event.HumanReadableOutputRecorder` to check for suites and use the appropriate word.
- Add new test and adjust others.

### Result:

Console output now shows "Suite" for suites.

### Checklist:

Resolves rdar://120962053

### DIFF
--- a/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
@@ -142,6 +142,14 @@ extension Event.HumanReadableOutputRecorder {
   }
 }
 
+extension Test {
+  /// A title for this test (either "Test" or "Suite"), capitalized and suitable
+  /// for use as the leading word of a human-readable message string.
+  fileprivate var capitalizedTitle: String {
+    isSuite ? "Suite" : "Test"
+  }
+}
+
 extension Test.Case {
   /// The arguments of this test case, formatted for presentation, prefixed by
   /// their corresponding parameter label when available.
@@ -224,7 +232,7 @@ extension Event.HumanReadableOutputRecorder {
       return [
         Message(
           symbol: .default,
-          stringValue: "Test \(testName) started."
+          stringValue: "\(test.capitalizedTitle) \(testName) started."
         )
       ]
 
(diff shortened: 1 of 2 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -142,6 +142,14 @@ extension Event.HumanReadableOutputRecorder {
   }
 }
 
+extension Test {
```
grynspan, line 145: I'd prefer to make this just a function local to the file rather than extension because the extension implies something a bit stronger than just "this is the string this recorder uses" to me.

grynspan (changes requested):
Nitpick, but otherwise looks good to me.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: Also "title" seems like the wrong noun, but that's probably a bit _too_ nitpicky.

stmontgomery:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
stmontgomery: Open to suggestions but this is just private to the file so I'm not too concerned by the word "title"

grynspan (approved):

stmontgomery:
There are known issues impacting CI currently so merging based on local testing.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: I got nothin'.

PR #185, opened 2024-01-16
`Locked` shouldn't be a property wrapper.

grynspan:
Property wrappers, when used as static properties, are inherently concurrency-unsafe (because they must be `var`, not `let`.) Discussed with the core Swift folks; this PR removes property wrapper "conformance" from `Locked`, adds `RawRepresentable` to replace it, and modifies call sites as appropriate.

Resolves rdar://121054518.

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Confirmation.swift
+++ b/Sources/Testing/Issues/Confirmation.swift
@@ -14,8 +14,7 @@ public struct Confirmation: Sendable {
   ///
   /// This property is fileprivate because it may be mutated asynchronously and
   /// callers may be tempted to use it in ways that result in data races.
-  @Locked
-  fileprivate var count = 0
+  fileprivate var count = Locked(rawValue: 0)
 
   /// Confirm this confirmation.
   ///
@@ -26,7 +25,7 @@ public struct Confirmation: Sendable {
   /// directly.
   public func confirm(count: Int = 1) {
     precondition(count > 0)
-    $count.add(count)
+    self.count.add(count)
   }
 }
 
@@ -101,7 +100,7 @@ public func confirmation<R>(
 ) async rethrows -> R {
   let confirmation = Confirmation()
   defer {
-    let actualCount = confirmation.count
+    let actualCount = confirmation.count.rawValue
     if actualCount != expectedCount {
       Issue.record(
         .confirmationMiscounted(actual: actualCount, expected: expectedCount),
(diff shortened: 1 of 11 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please clean test

Sources/Testing/Support/Locked.swift
```diff
@@ -77,14 +76,10 @@ struct Locked<T>: @unchecked Sendable where T: Sendable {
     }
   }
 
-  var wrappedValue: T {
+  var rawValue: T {
```
stmontgomery, line 79: Should it now conform to `RawRepresentable`?

Sources/Testing/Support/Locked.swift
```diff
@@ -25,7 +25,6 @@ private import TestingInternals
 /// - Bug: The state protected by this type should instead be protected using
 ///     actor isolation, but actor-isolated functions cannot be called from
 ///     synchronous functions. ([83888717](rdar://83888717))
-@propertyWrapper
 struct Locked<T>: @unchecked Sendable where T: Sendable {
```
stmontgomery, line 28: Do we still think the single word "Locked" is a good name for this type, given that it's no longer a property wrapper?

grynspan: I think it's fine, because you'd pronounce it `Locked<Int>` or similar. I'm open to alternatives (`LockedValue<Int>` sounds silly to me though.)

Sources/Testing/Support/Locked.swift
grynspan: Oops! I meant it to. Fixed.

grynspan:
@swift-ci please test

Sources/Testing/Issues/Confirmation.swift
```diff
@@ -14,8 +14,7 @@ public struct Confirmation: Sendable {
   ///
   /// This property is fileprivate because it may be mutated asynchronously and
   /// callers may be tempted to use it in ways that result in data races.
-  @Locked
-  fileprivate var count = 0
+  fileprivate var count = Locked(rawValue: 0)
```
stmontgomery, line 17: Can this, and all other remaining `var` variables containing `Locked` types, be `let` now? (I see at least one other, above.)

grynspan: Yes, they should all be `let` now. Good eye.

grynspan:
CI failures are unrelated to this PR and are being tracked elsewhere; changes have been verified at desk on Linux and macOS.

PR #187, opened 2024-01-17
Remove `public` keyword from most swift-syntax imports in the macro target.

grynspan:
This PR removes the explicit `public` keyword from imports of SwiftSyntax and SwiftSyntaxMacros in all files except those that actually declare macro implementations. If compiling using Swift 5.10 or earlier, the `public` keyword is preserved because earlier compilers don't correctly handle a scenario where a module is imported publicly in one file and internally in another.

This change removes a number of warnings of the form:

```
.../swift-testing/Sources/TestingMacros/Support/AttributeDiscovery.swift:11:8: warning: public import of 'SwiftSyntax' was not used in public declarations or inlinable code
public import SwiftSyntax
~~~~~~~^
.../swift-testing/Sources/TestingMacros/Support/AttributeDiscovery.swift:12:8: warning: public import of 'SwiftSyntaxMacros' was not used in public declarations or inlinable code
public import SwiftSyntaxMacros
~~~~~~~^
```

This change also tweaks the declarations of `ExpectMacro` and `RequireMacro` as they indirectly conform to `ExpressionMacro` in a way that may confuse some versions of the Swift compiler.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/Additions/DeclGroupSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/DeclGroupSyntaxAdditions.swift
@@ -8,7 +8,11 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
+#if swift(>=5.11)
+import SwiftSyntax
+#else
 public import SwiftSyntax
+#endif
 
 extension DeclGroupSyntax {
   /// The type declared or extended by this instance.
--- a/Package.swift
+++ b/Package.swift
@@ -131,17 +131,6 @@ extension Array where Element == PackageDescription.SwiftSetting {
       .enableExperimentalFeature("StrictConcurrency"),
       .enableUpcomingFeature("ExistentialAny"),
 
-      // FIXME: In a certain range of Swift toolchains, this feature causes
-      // warnings to be emitted such as:
-      //
-      // ```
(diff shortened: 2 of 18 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/TestingMacros/Support/Additions/DeclGroupSyntaxAdditions.swift
```diff
@@ -8,7 +8,11 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
+#if swift(>=5.11)
+import SwiftSyntax
+#else
 public import SwiftSyntax
+#endif
```
stmontgomery, line 15: Are you sure we actually need these `#if swift(>=5.11)` guards? My understanding was 5.10 was not really affected, only an earlier range of 5.11-series toolchains were

grynspan: Yes, I checked: if you use the latest 5.10 toolchain, you will get an error here without the `public` keyword.

grynspan:
@swift-ci please test

stmontgomery (approved):

PR #188, opened 2024-01-17
Work around SwiftPM build regression affecting packages with C++ targets

stmontgomery:
This is a workaround to the SwiftPM issue being fixed here:

https://github.com/apple/swift-package-manager/pull/7267

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -78,7 +78,12 @@ let package = Package(
     // "Support" targets: These contain C family code and are used exclusively
     // by other targets above, not directly included in product libraries.
     .target(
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

briancroom (approved):

grynspan (changes requested):
Given that a fix PR is open on SwiftPM and presumably will be merged shortly, I don't think we need to apply a workaround here. We can wait a couple more days.

stmontgomery:
No longer relevant, the SwiftPM bug was fixed!

PR #190, opened 2024-01-18
Fix a failing time-sensitive test.

grynspan:
One of the tests for our `.timeLimit()` trait uses `Clock.measure {}` to determine how long it takes to execute, but that's unsafe because it will include time spent waiting for the Swift runtime and the OS to schedule the child task in which the test runs. That time may be significant when tests are running in parallel.

This PR modifies that test to explicitly measure the start of the test when the test function is entered and the end of the test when the local test run completes. This avoids accidentally measuring any initial wait time.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
CI shows the test passing now.

Remaining failure on Linux is fixed by https://github.com/apple/swift-package-manager/pull/7273.
Remaining failure on macOS is fixed by https://github.com/apple/swift-package-manager/pull/7267.

stmontgomery (approved):

PR #191, opened 2024-01-19
Remove the 'ExperimentalParameterizedTesting' `@_spi` group from several parameterized testing-related declarations

stmontgomery:
This removes the `@_spi(ExperimentalParameterizedTesting)` attribute from several declarations related to parameterized tests. Note that some of these still have _other_ `@_spi` attributes, which prevent them from being fully exposed.

### Motivation:

The parameterized test feature has become more mature, after a series of recent PRs which have refined its functionality and added test coverage. It now feels appropriate to lift some of the `@_spi` attributes which were artificially limiting access to certain APIs — in particular, those for inspecting the currently-executing test case (if any) and controlling how arguments are handled.

### Modifications:

This removes the `@_spi(ExperimentalParameterizedTesting)` attribute from several declarations. However, since some of those still have _other_ SPI attributes, not all of those modified are becoming fully exposed.

### Result:

The following declarations become fully exposed, without any remaining `@_spi` attributes:

- `Test.isParameterized`
- `Test.parameters`
- `Test.Case`
- `Test.Case.current`
- `Test.Case.Argument`
- `Test.Parameter`
- `CustomTestArgumentEncodable`

While auditing these, I opted to _keep_ the `@_spi(ExperimentalParameterizedTesting)` on one declaration worth calling out in particular: `Test.testCases`. This accessor allows accessing all of the individual cases of a parameterized test, and in general it's not something test authors should need to access because the testing library manages running these.

### Checklist:

Resolves rdar://118575309

### DIFF
--- a/Sources/Testing/Test.swift
+++ b/Sources/Testing/Test.swift
@@ -89,7 +89,6 @@ public struct Test: Sendable {
   }
 
   /// Whether or not this test is parameterized.
-  @_spi(ExperimentalParameterizedTesting)
   public var isParameterized: Bool {
     guard let parameterCount = parameters?.count else {
       return false
@@ -103,7 +102,6 @@ public struct Test: Sendable {
   /// an array of values describing its parameters, which may be empty if the
   /// test function is non-parameterized. If this instance represents a test
   /// suite, the value of this property is `nil`.
-  @_spi(ExperimentalParameterizedTesting)
   public var parameters: [Parameter]?
 
   /// Whether or not this instance is a test suite containing other tests.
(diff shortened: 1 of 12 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/Test.swift
```diff
@@ -89,7 +89,6 @@ public struct Test: Sendable {
   }
 
   /// Whether or not this test is parameterized.
-  @_spi(ExperimentalParameterizedTesting)
   public var isParameterized: Bool {
     guard let parameterCount = parameters?.count else {
```
grynspan, line 93: While you're in here, you could change this to `isEmpty == false`?

grynspan (approved):

Sources/Testing/Test.swift
stmontgomery: Done

stmontgomery:
@swift-ci please test

PR #194, opened 2024-01-23
Rename `Test.ParameterInfo` to `Test.Parameter`

stmontgomery:
### Motivation:

In preparation for exposing some parameterized test-related declarations as public API (#191), we'd like to simplify the name of the `Test.ParameterInfo` type to just `Test.Parameter`. This makes certain property names like `Test.parameters` more consistent with Swift API design idioms.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Macro.swift
+++ b/Sources/Testing/Test+Macro.swift
@@ -149,7 +149,7 @@ extension Test {
   ///
   /// - Warning: This type alias is used to implement the `@Test` macro. Do not
   ///   use it directly.
-  public typealias __ParameterInfo = (firstName: String, secondName: String?)
+  public typealias __Parameter = (firstName: String, secondName: String?)
 
   /// Create an instance of ``Test`` for a function.
   ///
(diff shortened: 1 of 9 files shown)


Sources/Testing/Test+Macro.swift
```diff
@@ -149,7 +149,7 @@ extension Test {
   ///
   /// - Warning: This type alias is used to implement the `@Test` macro. Do not
   ///   use it directly.
-  public typealias __ParameterInfo = (firstName: String, secondName: String?)
+  public typealias __Parameter = (firstName: String, secondName: String?)
```
grynspan, line 152: We determined this shouldn't be the same type, right?

grynspan (approved):

Sources/Testing/Test+Macro.swift
stmontgomery: Right, it makes the macro expansion simpler, avoids needing to expose `init`, and also we derive the `index` value from the position in this array

PR #195, opened 2024-01-25
Fix access levels used when re-exporting from cross-module overlay.

grynspan:
As with other uses of `public import`, we need to consider both Swift 5.10 and 5.11.

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

stmontgomery (approved):

PR #196, opened 2024-01-26
Capture discrete subexpressions in a `SourceCode` instance.

grynspan:
The `SourceCode` type currently represents the source code, as written by a developer, of some expression such as the body of a `#expect()` macro invocation. This PR enhances `SourceCode` to let it capture, at runtime, information about its various subexpressions and to propagate that information to an expectation or expectation snapshot.

This change allows us to (in a future PR) provide more verbose descriptions of expectation failures when invoking `swift test --verbose`.

Resolves rdar://121382419.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

dennisweissmann (approved):

Sources/Testing/Traits/Trait.swift
```diff
@@ -42,25 +42,22 @@ public protocol Trait: Sendable {
...
   /// unmodified.
   ///
   /// - Warning: This protocol requirement is experimental. It is public due to
   ///   technical limitations in Swift. It may be modified or removed in a
   ///   future update to the testing library.
-  func _addingSourceCode(
-    _ sourceCode: @autoclosure () -> SourceCode,
-    arguments: @autoclosure () -> [(label: String?, sourceCode: SourceCode)]
-  ) -> Self
+  func _capturingExpression(_ expression: @autoclosure () -> Expression) -> Self
```
stmontgomery, line 60: Nice to see these parameters consolidated!

Sources/Testing/Expectations/Expectation.swift
```diff
@@ -10,9 +10,9 @@
 
 /// A type describing an expectation that has been evaluated.
 public struct Expectation: Sendable {
-  /// The source code of the expression evaluated by this expectation, if
-  /// available at compile time.
-  public var sourceCode: SourceCode?
+  /// The expression evaluated by this expectation, if available at compile
+  /// time.
+  public var evaluatedExpression: Expression?
```
stmontgomery, line 15: The description of this property now feels a bit odd: Since `Expression` now contains both the source code of an expression and the runtime value (if evaluated) of said expression, the "if available at compile time" may be confusing since one of those (the runtime value) can't possibly be known at compile time.

Can this wording be clarified? (And any edits propagated down to the corresponding `Snapshot` property below)

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -0,0 +1,261 @@
...
+
+    /// The expression represents a function call.
+    ///
+    /// - Parameters:
+    ///   - value: The value on which the function was called, if any.
+    ///   - functionName: The name of the function that was called.
+    ///   - arguments: The arguments passed to the function.
+    indirect case functionCall(value: Expression?, functionName: String, arguments: [FunctionCallArgument])
+
+    /// The expression represets a property access.
```
stmontgomery, line 64: typo
```suggestion
    /// The expression represents a property access.
```

Sources/Testing/Expectations/Expectation.swift
stmontgomery, line 15: Could the `evaluatedExpression` property become non-`Optional` now?

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -0,0 +1,261 @@
...
+  @_spi(ExperimentalSourceCodeCapturing)
+  public var runtimeValueDescription: String?
+
+  /// The fully-qualified name of the type of this subexpression's runtime
+  /// value.
+  ///
+  /// If the runtime value of this subexpression has not been evaluated, the
+  /// value of this property is `nil`.
+  @_spi(ExperimentalSourceCodeCapturing)
+  public var fullyQualifiedTypeNameOfRuntimeValue: String?
```
stmontgomery, line 119: If there is a need to decompose this to (for example) omit showing the module name, would you suggest splitting on period character? Is there any benefit to exposing a tuple or other structured representation of this?

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -0,0 +1,261 @@
...
+        value: value?.capturingRuntimeValues(firstValue),
+        functionName: functionName,
+        arguments: zip(arguments, additionalValuesArray).map { argument, value in
+          .init(label: argument.label, value: argument.value.capturingRuntimeValues(value))
+        }
+      )
+    case let .propertyAccess(value, keyPath):
+      result.kind = .propertyAccess(
+        value: value.capturingRuntimeValues(firstValue),
+        keyPath: keyPath.capturingRuntimeValues(additionalValuesArray.first ?? nil)
```
stmontgomery, line 164: This is probably a pre-existing concern, but I wonder about error handling here, if the number of runtime values does not match the number of source code sub-expressions. What do you think about at least having a "FIXME" comment or similar noting that it could happen and may need to be addressed at some point? My only worry is tha usage of `zip` (for example) could lead to something getting silently dropped

Sources/Testing/SourceAttribution/Expression+Macro.swift
```diff
@@ -0,0 +1,78 @@
...
+  ///     any.
+  ///   - functionName: The name of the member function.
+  ///   - argumentLabel: Optionally, the argument label.
+  ///   - argumentValue: The value of the argument to the function.
+  ///
+  /// - Returns: A new instance of ``Expression``.
+  ///
+  /// - Warning: This function is used to implement the `@Test`, `@Suite`,
+  ///   `#expect()` and `#require()` macros. Do not call it directly.
+  public static func __functionCall(_ value: String?, _ functionName: String, _ arguments: (label: String?, value: String)...) -> Self {
```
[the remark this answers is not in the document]
grynspan: Yes, it's a typo, but it's an existing typo, so I'm going to leave it for now. I was planning to rename these symbols later anyway.

Sources/Testing/SourceAttribution/Expression.swift
grynspan: I don't know what you're specifically concerned about so I am unsure what I'd put in a FIXME. Can you clarify?

Sources/Testing/SourceAttribution/Expression.swift
grynspan: Yes, you'd split on the first period character. We don't try to represent type names as a tuple/structure elsewhere, so I wasn't planning to do so here.

Sources/Testing/Expectations/Expectation.swift
grynspan: Maybe—I'm honestly not sure if there are outstanding code paths that don't produce a value here… I'll investigate.

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -0,0 +1,261 @@
...
+/// converted to strings representing their source code (captured at compile
+/// time) using `String.init(describing:)`.
+///
+/// If parsing is needed, use the swift-syntax package to convert an instance of
+/// this type to an instance of `ExprSyntax` using a Swift expression such as:
+///
+/// ```swift
+/// let swiftSyntaxExpr: ExprSyntax = "\(testExpr)"
+/// ```
+public struct Expression: Sendable {
```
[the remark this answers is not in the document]
grynspan: We'd previously exposed `SourceCode`, hence exposing this renamed type.

If we're not confident in the shape of the interface, we can make it SPI.

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -0,0 +1,261 @@
...
+  /// Initialize an instance of this type containing the specified source code.
+  ///
+  /// - Parameters:
+  ///   - sourceCode: The source code of the expression being described.
+  ///
+  /// To get the string value of an instance of ``Expression``, pass it to
+  /// `String.init(describing:)`.
+  ///
+  /// This initializer does not attempt to parse `sourceCode`.
+  public init(_ sourceCode: String) {
```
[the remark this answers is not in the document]
grynspan: If we make the type SPI, it's moot and I'm not worried. If it remains API, I do think it's useful but since we're not doing any parsing… perhaps minimally so. We could make it non-public if you feel strongly.

Sources/Testing/SourceAttribution/Expression.swift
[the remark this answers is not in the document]
grynspan: Oh, it needs to be API so that `_capturingExpression()` can see it.

Sources/Testing/Expectations/Expectation.swift
grynspan: We can remove the subordinate clause ("if …")?

Sources/Testing/SourceAttribution/Expression.swift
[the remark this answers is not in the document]
grynspan: And all the `__check()` overloads.

Sources/Testing/SourceAttribution/Expression+Macro.swift
```diff
@@ -0,0 +1,78 @@
...
+  ///
+  /// - Parameters:
+  ///   - syntaxNode: The complete syntax node (expression, declaration,
+  ///     statement, etc.)
+  ///
+  /// - Returns: A new instance of ``Expression``.
+  ///
+  /// - Warning: This function is used to implement the `@Test`, `@Suite`,
+  ///   `#expect()` and `#require()` macros. Do not call it directly.
+  public static func __fromSyntaxNode(_ syntaxNode: String) -> Self {
```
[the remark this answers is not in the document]
grynspan: I'm planning to revise the symbol names in this file in a subsequent PR.

Sources/Testing/Expectations/Expectation.swift
grynspan: Removed.

grynspan: Works!

grynspan:
@swift-ci please test

Sources/Testing/SourceAttribution/Expression.swift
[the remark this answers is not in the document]
stmontgomery: Can we leave it public but make it `@_spi(ExperimentalSourceCodeCapturing)`?

Sources/Testing/SourceAttribution/Expression.swift
stmontgomery: Just the possibility that the number of variadic generic parameters passed to this function might not equal the number of subexpressions. If they mismatch, the `zip` function will use the shorter value, and could cause some data to be lost

stmontgomery (changes requested):
Latest changes look great! My only remaining request is to add `@_spi` to the `public init` I mentioned, if you agree

stmontgomery (approved):

Sources/Testing/SourceAttribution/Expression.swift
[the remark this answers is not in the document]
grynspan: Done.

Sources/Testing/SourceAttribution/Expression.swift
grynspan: The number of arguments vs. values will be equal in that case, because the source code will just look like:

```swift
f(a, b, c)
```

And the set of values will be `[a, b, c]`. We have no knowledge of the _signature_ of `f()` in this spot, so we have no knowledge of potential variadic sequences. We just see the function call expression.

grynspan: If we come up with a concrete problem, then yes let's document it and try to guard against it. :)

grynspan:
@swift-ci please test

PR #198, opened 2024-01-31
Work around a bug with variadic generics when mixed with C strings.

grynspan:
There's an apparent bug (rdar://122011759) when using variadic generics and passing them a string that needs to be converted to a C string. For example:

```swift
let file = try #require(fopen("/test/file", "wb"))
```

The compiler generates invalid code for this call that results in the two C string arguments to `fopen()` having the same address and containing only the trailing NUL character.

I have only reproduced the issue with functions imported from C; I cannot reproduce it at this time with functions declared in Swift.

This change works around the issue by providing explicit overloads of `__checkFunctionCall()` for functions taking 1, 2, 3, or 4 arguments. Tests calling functions that take more arguments can be refactored, e.g.:

```swift
let filePtr = f("a", "b", "c", "d", "e") // ...
let file = try #require(filePtr)
```

Resolves rdar://122014061.

### Checklist:

### DIFF
--- a/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
+++ b/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
@@ -194,6 +194,116 @@ public func __checkFunctionCall<T, each U>(
   )
 }
 
+#if !SWT_FIXED_122011759
+/// Check that an expectation has passed after a condition has been evaluated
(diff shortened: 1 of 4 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #199, opened 2024-01-31
Distinguish string literals in `Expression`.

grynspan:
This PR adds a new case to `Expression.Kind` for string literals and modifies the surrounding code to support it and plumb it through where necessary. It also starts the process of making expression capture recursive (although we don't meaningfully support arbitrary levels of recursion here at this time.)

The purpose of the change is to allow us to distinguish tags that were specified in source by string literal from tags specified by some symbol name. This will allow us to build out tag filtering such that you can filter by tags with value `"x"` or filter by tags _named_ `x`. Filtering by tag is not supported in SwiftPM yet, but it's something we want to do in the near future and this is part of the groundwork for it.

Resolves rdar://121692183.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/Expression+Macro.swift
+++ b/Sources/Testing/SourceAttribution/Expression+Macro.swift
@@ -23,6 +23,21 @@ extension Expression {
     Self(kind: .generic(syntaxNode))
   }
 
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #201, opened 2024-02-02
Add a document describing how to tag a release.

grynspan:
This PR adds a new Markdown document describing how to tag a release of swift-testing.

### Checklist:

### DIFF
--- a/Documentation/Releases.md
+++ b/Documentation/Releases.md
@@ -43,13 +43,13 @@ git checkout -b release/x.y.z
 The package manifest files (Package.swift _and_ Package@swift-5.11.swift) must
 be updated so that the release can be used as a package dependency:
 
-0. Take note of any availability definitions (passed to the Swift compiler with
+1. Take note of any availability definitions (passed to the Swift compiler with
    the `-define-availability` option.) Search for these definitions in the
    package source and replace them with their values. For example, replace all
    `_distantFuture` uses with `macOS 99.0, iOS 99.0, watchOS 99.0, tvOS 99.0`.
-0. Delete any unsafe flags from `var packageSettings` as well as elsewhere in
+1. Delete any unsafe flags from `var packageSettings` as well as elsewhere in
    the package manifest files.
-0. Open the "Documentation/Testing.docc/TemporaryGettingStarted.md" file and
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

Documentation/Releases.md
```diff
@@ -0,0 +1,103 @@
...
+git pull
+git checkout -b release/x.y.z
+```
+
+## Preparing the repository's contents
+
+The package manifest files (Package.swift _and_ Package@swift-5.11.swift) must
+be updated so that the release can be used as a package dependency:
+
+0. Take note of any availability definitions (passed to the Swift compiler with
```
stmontgomery, line 46: The number 0. seems to cause GitHub to actually use zero as the first number in the list, is that what you intended?

Documentation/Releases.md
```diff
@@ -0,0 +1,103 @@
...
+   the `-define-availability` option.) Search for these definitions in the
+   package source and replace them with their values. For example, replace all
+   `_distantFuture` uses with `macOS 99.0, iOS 99.0, watchOS 99.0, tvOS 99.0`.
+0. Delete any unsafe flags from `var packageSettings` as well as elsewhere in
+   the package manifest files.
```
stmontgomery, line 51: I think we ought to include a step _after_ adjusting the package's code for release but _before_ pushing the tag where we perform a brief qualification from a client's perspective. Essentially this would involve creating a throwaway client package, declaring swift-testing as a dependency (probably via local file path instead of tag, since the new tag wouldn't be pushed yet) and ensuring the client can build successfully.

The code changes we make for a tag could potentially render swift-testing unable to build successfully, or prevent clients from building, so we ought to at least perform some basic qualification I think. My focus is on simply building, not necessarily running swift-testing's unit test suite (although that would be nice too, eventually).

stmontgomery (approved):

Documentation/Releases.md
grynspan: Can add something like that. :)

stmontgomery (approved):

grynspan:
@siwft-ci please test

grynspan:
@swift-ci please test

PR #204, opened 2024-02-05
Do not use `String.init(validatingUTF8: [CChar])`.

grynspan:
The overload of `String.init(validatingUTF8:)` that takes an array of `CChar` values rather than a pointer is not really meant to be used. The overload exists for the benefit of the type checker so that future compilers can warn when a developer misuses the implicit array-to-pointer conversion.

This PR changes our wrapper around `sysctlbyname()` to avoid producing a temporary `[CChar]`, thus dodging the problem.

See also https://github.com/apple/swift/pull/71224.

Resolves rdar://122198850.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Versions.swift
+++ b/Sources/Testing/Support/Versions.swift
@@ -19,8 +19,8 @@ private import TestingInternals
 /// This value is not part of the public interface of the testing library.
 let operatingSystemVersion: String = {
 #if !SWT_NO_SYSCTL && SWT_TARGET_OS_APPLE
-  let productVersion = sysctlbyname("kern.osproductversion").flatMap(String.init(validatingUTF8:)) ?? ""
-  let buildNumber = sysctlbyname("kern.osversion").flatMap(String.init(validatingUTF8:)) ?? ""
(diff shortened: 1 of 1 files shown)


suzannaratcliff (approved):

Sources/Testing/Support/Versions.swift
```diff
@@ -141,30 +141,25 @@ let swiftStandardLibraryVersion: String = {
...
-/// This function does not convert the result to a `String` since not all
-/// `sysctl()` values are strings. To convert the result to a string, use
-/// `String(validatingUTF8:)`.
+/// - Returns: A string containing the requested value interpreted as a UTF-8
+///   string, or `nil` if the value could not be read or could not be
+///   interpreted as UTF-8.
 ///
 /// This function is not part of the public interface of the testing library.
-func sysctlbyname(_ name: String) -> [CChar]? {
+func sysctlbyname(_ name: String, as _: String.Type) -> String? {
```
stmontgomery, line 154: Worth making private?
```suggestion
private func _sysctlbyname(_ name: String, as _: String.Type) -> String? {
```

stmontgomery (approved):

PR #205, opened 2024-02-06
Add support for `swift test --verbose`.

grynspan:
This PR adds the ability to detect the `--verbose` flag and alter console output accordingly. When `--verbose` is specified, swift-testing will now emit more version information (when not specified, we'll only log the library version) and will emit a breakdown of subexpressions for failed expectations.

### Without `--verbose`:

Screenshot 2024-02-06 at 2 37 58 PM

### With `--verbose`:

Screenshot 2024-02-06 at 2 37 43 PM

Resolves rdar://121382539.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
@@ -81,7 +81,7 @@ extension Event {
       ///
       /// When specified, additional output is provided. The exact nature of the
       /// additional output is implementation-defined and subject to change.
-      case recordVerboseOutput
+      case useVerboseOutput
     }
 
     /// The options for this event recorder.
@@ -269,7 +269,7 @@ extension Event.ConsoleOutputRecorder {
   /// - Returns: Whether any output was produced and written to this instance's
   ///   destination.
   @discardableResult public func record(_ event: borrowing Event, in context: borrowing Event.Context) -> Bool {
-    let verbose = options.contains(.recordVerboseOutput)
+    let verbose = options.contains(.useVerboseOutput)
     let messages = _humanReadableOutputRecorder.record(event, in: context, verbosely: verbose)
     for message in messages {
       let symbol = message.symbol?.stringValue(options: options) ?? " "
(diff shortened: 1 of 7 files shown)


grynspan:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
```diff
@@ -76,6 +76,12 @@ extension Event {
       /// This option is ignored unless ``useANSIEscapeCodes`` is also
       /// specified.
       case useTagColors(_ tagColors: [Tag: Tag.Color])
+
+      /// Record verbose output.
+      ///
+      /// When specified, additional output is provided. The exact nature of the
+      /// additional output is implementation-defined and subject to change.
+      case recordVerboseOutput
```
stmontgomery, line 84: The other options here have "use" as their leading word, so I'm wondering why this one uses "record". They all affect the behavior of the recording action

Sources/Testing/Expectations/ExpectationChecking+Macro.swift
```diff
@@ -991,6 +992,7 @@ public func __checkClosureCall<R>(
     }
     mismatchExplanationValue = explanation
   } catch {
+    expression = expression.capturingRuntimeValues(error)
```
stmontgomery, line 995: Can you explain the effect of this? I wonder what will happen if the expression is comprised of multiple sub-expressions, won't the error be attributed to the first one even if it was a subsequent sub-expr which threw the error?

I'm mainly trying to get a sense of what the presentation of these errors will be, so an example or two would be great

Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
grynspan: I couldn't come up with better phrasing. Any ideas?

stmontgomery: `useVerboseOutput`

Sources/Testing/Expectations/ExpectationChecking+Macro.swift
stmontgomery: Ok I see, thanks

stmontgomery (commented):
Only feedback is a minor one about the Option case name

Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
grynspan: Sure, will change.

grynspan: (We could also probably just drop `use` from all the cases as redundant.)

grynspan:
@swift-ci please test

stmontgomery (approved):

PR #208, opened 2024-02-08
Implement a test iteration feature.

grynspan:
This PR introduces a test iteration feature similar to the one supported by XCTest. Developers setting up an instance of `Configuration` can specify how many times to repeat a test plan during a test run as well as under what conditions iteration should continue or stop.

A new property, `iterationPolicy`, is added to `Configuration`, with two properties of its own as described above and a couple of convenience factory methods. When a `Runner` is run, its configuration's iteration policy is consulted and the entire test run is repeated as needed. New event kinds have been added for iteration start and end.

SwiftPM integration to come in a later PR in that repository.

Resolves rdar://88665403.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -20,6 +20,80 @@ public struct Configuration: Sendable {
   /// Whether or not to parallelize the execution of tests and test cases.
   public var isParallelizationEnabled = true
 
+  /// A type describing whether or not, and how, to iterate a test plan
+  /// repeatedly.
+  ///
+  /// When a ``Runner`` is run, it will run all tests in its corresponding
+  /// ``Runner/Plan`` according to the policy described by its
+  /// ``Configuration/iterationPolicy-swift.property`` property. For instance,
+  /// if that property is set to:
+  ///
+  /// ```swift
+  /// .repeating(.untilIssueRecorded, count: 10)
+  /// ```
+  ///
+  /// The entire test plan will be run repeatedly, up to 10 times. If an issue
+  /// is recorded, the current iteration will complete, but no further
+  /// iterations will be attempted.
+  ///
+  /// If the value of an instance's ``count`` property is `1`, the value of its
+  /// ``continuationCondition-swift.property`` property has no effect.
+  public struct IterationPolicy: Sendable {
+    /// An enumeration describing the conditions under which test iterations
+    /// should continue.
+    public enum ContinuationCondition: Sendable {
+      /// The test plan should continue iterating until an unknown issue is
+      /// recorded.
+      ///
+      /// When this continuation condition is used and an issue is recorded, the
+      /// current iteration will complete, but no further iterations will be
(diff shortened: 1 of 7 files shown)


briancroom (approved):

grynspan:
@swift-ci please test

Sources/Testing/Running/Configuration.swift
```diff
@@ -20,6 +20,80 @@ public struct Configuration: Sendable {
...
+  /// .repeating(.untilIssueRecorded, count: 10)
+  /// ```
+  ///
+  /// The entire test plan will be run repeatedly, up to 10 times. If an issue
+  /// is recorded, the current iteration will complete, but no further
+  /// iterations will be attempted.
+  ///
+  /// If the value of an instance's ``count`` property is `1`, the value of its
+  /// ``continuationCondition-swift.property`` property has no effect.
+  public struct IterationPolicy: Sendable {
```
stmontgomery, line 41: Bikeshedding a little, I'd like to propose we call this type "RepetitionPolicy", and refer to the general functionality with the base word "repetition", but still use the word "iteration" anywhere we need to refer to one individual unit of the repetition. So to me, that means these `Configuration` declarations would be phrased in terms of "repetition", while the additions to `Event` would continue using the word "iteration". Thoughts?

Sources/Testing/Running/Configuration.swift
```diff
@@ -20,6 +20,80 @@ public struct Configuration: Sendable {
...
+    /// If the value of this property is `nil`, a test plan will be run
+    /// ``count`` times regardless of whether or not issues are encountered
+    /// while running.
+    public var continuationCondition: ContinuationCondition?
+
+    /// The maximum number of times the test run should iterate.
+    ///
+    /// - Precondition: The value of this property must be greater than or equal
+    ///   to `1`.
+    public var count: Int {
```
stmontgomery, line 69: Can the name of this property include the word "maximum", e.g. `maximumRepetitionCount`? Just because, per the continuation conditions, it may not actually repeat the precise number of times this property returns

stmontgomery (approved):

Sources/Testing/Running/Configuration.swift
grynspan: Sure, I can change it.

Sources/Testing/Running/Configuration.swift
grynspan: I went back and forth on it but settled on "iteration" as a consistent term of art here. I'm not sure why we would want to say "repetition" here but "iteration" there—seems like it might be confusing and could imply two separate concepts. Do you feel strongly about changing it?

grynspan: It's been a few days since I looked at this code and ugh, I am not consistent at all between the two terms. I'll try to adopt what you suggest here.

grynspan:
@swift-ci please test

stmontgomery (approved):

PR #209, opened 2024-02-08
[WIP] Experimental `NSException` support for `#expect(throws:)`.

grynspan:
Experiment. Do not merge.

This PR would introduce a new overload of `#expect(throws:)` that takes `NSException` or a subclass thereof and checks if one is thrown within its body closure.

The bulk of the code as-written can be extended to a hypothetical CxxStdlib cross-import overlay to allow for checking that a C++ exception was thrown (rather than an `NSException` instance.) _Any_ value can be thrown in C++, not just instances of `std::exception`, but we would probably want to limit the API surface to just `std::exception` and related types.

⚠️ The effect of throwing an exception through a Swift frame is undefined, so this functionality could probably never be merged, but it's an interesting experiment and might prove useful if C++ interop adds exception-handling support in the future.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

PR #210, opened 2024-02-09
[WIP] Enable support for the `@_section` experimental attribute.

grynspan:
This PR enables support for `@_used` and `@_section` to allow embedding test data directly into its own section in a binary. This means we don't have to look up synthesized test types at runtime.

Limitations:

1. Test clients must enable support for the experimental `"SymbolLinkageMarkers"` feature.
1. The `@_section` attribute currently doesn't support application to constants of function type, thinking they are non-constant.
1. We will need to make a change to the Swift runtime on Linux/Windows to add the new section to `MetadataSections` (or alternatively, the experimental feature needs to provide a different way to discover sections at runtime.) See https://github.com/apple/swift/pull/71509.

:warning: Do not merge: this is a proof-of-concept and work-in-progress.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/TestDeclarationMacro.swift
+++ b/Sources/TestingMacros/TestDeclarationMacro.swift
@@ -444,17 +444,15 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
     result.append(DeclSyntax(thunkDecl))
 
     // Create the expression that returns the Test instance for the function.
-    var testsBody: CodeBlockItemListSyntax = """
-    return [
-      .__function(
-        named: \(literal: functionDecl.completeName),
-        in: \(typealiasExpr),
-        xcTestCompatibleSelector: \(selectorExpr ?? "nil"),
-        \(raw: attributeInfo.functionArgumentList(in: context)),
-        parameters: \(raw: functionDecl.testFunctionParameterList),
-        testFunction: \(thunkDecl.name)
-      )
-    ]
+    var testBody: CodeBlockItemListSyntax = """
+    return .__function(
+      named: \(literal: functionDecl.completeName),
+      in: \(typealiasExpr),
+      xcTestCompatibleSelector: \(selectorExpr ?? "nil"),
+      \(raw: attributeInfo.functionArgumentList(in: context)),
+      parameters: \(raw: functionDecl.testFunctionParameterList),
+      testFunction: \(thunkDecl.name)
+    )
     """
 
     // If this function has arguments, then it can only be referenced (let alone
@@ -471,29 +469,37 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
       result.append(
         """
         @available(*, deprecated, message: "This property is an implementation detail of the testing library. Do not use it directly.")
-        private \(_staticKeyword(for: typeName)) nonisolated func \(unavailableTestName)() async -> [Testing.Test] {
-          [
-            .__function(
-              named: \(literal: functionDecl.completeName),
-              in: \(typealiasExpr),
-              xcTestCompatibleSelector: \(selectorExpr ?? "nil"),
-              \(raw: attributeInfo.functionArgumentList(in: context)),
-              testFunction: {}
-            )
-          ]
+        private \(_staticKeyword(for: typeName)) nonisolated func \(unavailableTestName)() async -> Testing.Test {
+          .__function(
+            named: \(literal: functionDecl.completeName),
+            in: \(typealiasExpr),
+            xcTestCompatibleSelector: \(selectorExpr ?? "nil"),
+            \(raw: attributeInfo.functionArgumentList(in: context)),
+            testFunction: {}
+          )
         }
         """
       )
 
       // Add availability guards if needed. If none are needed, the extra thunk
       // is unused.
-      testsBody = createSyntaxNode(
+      testBody = createSyntaxNode(
         guardingForAvailabilityOf: functionDecl,
-        beforePerforming: testsBody,
+        beforePerforming: testBody,
         orExitingWith: "return await \(unavailableTestName)()",
         in: context
       )
     }
+    
+    let testGetterFunctionName = context.makeUniqueName(thunking: functionDecl)
+    result.append(
+      """
+      @available(*, deprecated)
+      @Sendable private \(_staticKeyword(for: typeName)) func \(testGetterFunctionName)() async -> Testing.Test {
+        \(raw: testBody)
+      }
+      """
+    )
 
     // The emitted type must be public or the compiler can optimize it away
     // (since it is not actually used anywhere that the compiler can see.)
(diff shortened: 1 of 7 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@compnerd says the Windows section needs the suffix `"$B"`.

Sources/TestingMacros/TestDeclarationMacro.swift
```diff
@@ -527,13 +533,33 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
...
+//    result.append(
+//      """
+//      #if hasFeature(SymbolLinkageMarkers)
+//      @_used
+//        #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS) || os(visionOS)
+//        @_section("__DATA_CONST,__swift5_tests")
+//        #elseif os(Linux)
+//        @_section("swift5_tests")
+//        #elseif os(Windows)
+//        @_section(".sw5test")
```
compnerd, line 556:
```suggestion
//        @_section("sw5test$B")
```

Sources/TestingInternals/Discovery.cpp
```diff
@@ -158,20 +163,89 @@ struct SWTTypeMetadataRecord {
...
-extern "C" const char sectionEnd __asm("section$end$__TEXT$__swift5_types");
+extern "C" const char typesSectionBegin __asm("section$start$__TEXT$__swift5_types");
+extern "C" const char typesSectionEnd __asm("section$end$__TEXT$__swift5_types");
+
+extern "C" const char testsSectionBegin __asm("section$start$__DATA_CONST$__swift5_tests");
+extern "C" const char testsSectionEnd __asm("section$end$__DATA_CONST$__swift5_tests");
+
+// Ensure the tests section is actually emitted.
+__attribute__((used)) __attribute__((section("__DATA_CONST,__swift5_tests")))
+constinit const SWTTestGetter ensureTestSectionIsEmitted = std::make_pair(nullptr, nullptr);
```
compnerd, line 231: This seems like it would be limited to MachO? Should this region also have a restriction? Or do you intend to re-declare these for ELF, COFF, and WASM?

Sources/TestingInternals/Discovery.cpp
```diff
@@ -248,105 +322,108 @@ static SWTMachHeaderList getMachHeaders(void) {
...
-        continue;
-      }
+  enumerateImages([=] (const SWTMetadataSections& sections) {
+    // If this image has a tests section, then it should be used instead of the
+    // image's type metadata section.
+    if (sections.version >= SWT_METADATA_SECTION_MINIMUM_VERSION_WITH_TESTS
+        && sections.swift5_tests.start != 0
+        && sections.swift5_tests.length > 0) {
+      return; // continue
+    }
```
compnerd, line 379: Personally, I think that splitting this up might be a bit nicer:

```c++
if (sections.version < SWT_METADATA_SECTION_MINIMUM_VERSION_WITH_TESTS)
  return;

const auto &swift5_tests = sections.swift5_tests;
if (swift5_tests.start == 0 || swift5_tests.length <= 0)
  return;
```

Sources/TestingInternals/Discovery.cpp
```diff
@@ -248,105 +322,108 @@ static SWTMachHeaderList getMachHeaders(void) {
...
+          // parameters, which is not something we can know abstractly.
           continue;
         }
+
+        // If the caller supplied a name filtering function, check that the type's
+        // name passes. This will be more expensive than the checks above, but
+        // should be cheaper than realizing the metadata.
+        if (nameFilter) {
+          const char *typeName = contextDescriptor->getName();
+          bool nameOK = typeName && (* nameFilter)(typeName, context);
```
compnerd, line 402: Nit: extra space.

```suggestion
          bool nameOK = typeName && (*nameFilter)(typeName, context);
```

Sources/TestingInternals/Discovery.cpp
```diff
@@ -248,105 +322,108 @@ static SWTMachHeaderList getMachHeaders(void) {
...
+          if (!nameOK) {
+            continue;
+          }
```
compnerd, line 405: Does the `nameOK` add anything to clarity or does it make sense to remove it?

```c++
if (!typeName || !(*nameFilter)(typeName, context))
  continue;
```

Sources/TestingMacros/TestDeclarationMacro.swift
grynspan: @kubamracek Do you know if `@_section` requires the leading period or not?

Sources/TestingInternals/Discovery.cpp
grynspan: You can ignore this section; it's intentionally only functional with Mach-O. Linux and Windows should always use the code path below that uses `swift_enumerateAllMetadataSections()`.

Embedded Swift is an open question right now; we're only just starting to look at what's feasible. Presumably, solutions for Embedded Swift will look similar to this one, but will need to be arch- or platform-specific.

Sources/TestingInternals/Discovery.cpp
grynspan: I can do that, sure.

Sources/TestingInternals/Discovery.cpp
grynspan: House style (not that it comes up much.) Well… my style, anyway, because I am old and `fp()` was, historically, not a valid way to invoke a function pointer in C… anyway, I put an onion on my belt, which was the style at the time…

Sources/TestingInternals/Discovery.cpp
grynspan: This code's not new; I found it easier to follow the logic here by splitting it apart this way. Compare your suggestion here: https://github.com/apple/swift-testing/pull/210#discussion_r1494989225

Sources/TestingInternals/Discovery.cpp
grynspan: Well, not exactly that, as we want to fall back to the path that looks at type metadata. But I'll split it up regardless.

grynspan:
@swift-ci please test

Sources/TestingMacros/TestDeclarationMacro.swift
grynspan: Okay, I think I got it right then.

grynspan:
@swift-ci please test

grynspan:
Linux failure is expected right now; the new `swift5_tests` section needs a newer toolchain.

grynspan:
Looks like Windows has a compile-time complaint about the constness of C function pointers.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

PR #211, opened 2024-02-09
Ensure `nil` is emitted for expectation failures when unwrapping opti…

grynspan:
…onals.

When `try #require(x)` is used where `x` is some optional expression, we expect that on failure its value (`nil`) is included in the failure message so that it is clear  _what_ has failed. For example, given the following test:

```swift
@Test func f() throws {
  let x: Int? = nil
  let y = try #require(x)
  #expect(x == y)
}
```

The following output _was_ produced, which is unclear:

> ✘ Test f() recorded an issue at ...: Expectation failed: x

What it _should_ produce is instead:

> ✘ Test f() recorded an issue at ...: Expectation failed: x → nil

This PR does that, yay! It's a bit complicated because the logic involved in capturing values is… well, a bit byzantine, and worthy of future cleanup.

As an aside, `try #require(x as? T)` is special-cased because `(x as? T) → nil` looks odd in test output. We thought about extending this logic for expectations based on boolean expressions, but `(x == y) → false` seemed even weirder, since the initial expression was already human-readable as boolean.

Resolves rdar://122663252.

### Result:

_[After your change, what will change.]_

### Checklist:

### DIFF
--- a/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
+++ b/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
@@ -688,10 +688,16 @@ public func __checkValue<T>(
   isRequired: Bool,
   sourceLocation: SourceLocation
 ) -> Result<T, any Error> {
+  // The double-optional below is because capturingRuntimeValue() takes optional
+  // values and interprets nil as "no value available". Rather, if optionalValue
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

briancroom (approved):

PR #214, opened 2024-02-12
Change `var` to `let` to suppress a warning.

grynspan:
This line needs a tweak:

```swift
case var .generic(sourceCode), var .stringLiteral(sourceCode, _):
```

We are no longer mutating `sourceCode` in this case statement, so the variables can be constant (`let`.)

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #215, opened 2024-02-12
Work around an issue in the toolchain calling `Result.get()`.

grynspan:
There's currently a miscompile of some sort calling `Result.get()` using a main-branch toolchain. (I'm guessing it's related to SE-0413 adoption, but that's speculation only.) ([122797397](rdar://122797397))

This change works around the issue by switching over `Result` instead of calling `get()` in the one spot where we do so.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #216, opened 2024-02-12
Expand on `Test.current` and why it is optional.

grynspan:
The current documentation for [`Test.current`](https://swiftpackageindex.com/apple/swift-testing/main/documentation/testing/test/current) doesn't explain why it's an optional property. If no test is running, or if the current thread is detached from the task that's running a test, then the value is `nil`. Our documentation should reflect that.

(Same for `Test.Case.current`.)

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

briancroom (approved):

PR #217, opened 2024-02-13
Add a `.serial` trait that forces a test to run its content serially.

grynspan:
As of right now, when parallelization is enabled (the default for `swift test`, opt-in for `XCTestScaffold`), all tests scramble to run at the same time. This is generally overall a good thing for performance, because the total time spent running tests is measurably lower than if they ran serially.

However, some tests, such as those that touch shared global state, really need to run serially. This PR introduces a new trait, `.serial` AKA `SerialTrait`, that allows those tests to opt out of parallelization. Note that this is not a global effect: the tests may run in parallel with other unrelated tests, but their test cases will run serially relative to each other, and in the case of suites any contained tests will run serially (and the effect is recursive, so a parameterized test function contained in a suite that is contained in a serialized suite will run its cases serially.)

To avoid teaching the runner about another specialized trait, I wanted to modify the existing `Trait.prepare(for:)` function to take an `inout` `Runner.Plan.Action` value that a trait could then modify. However, `Runner.Plan.Action`'s design is not ready to become API, so instead I have introduced a new protocol `SPIAwareTrait` that refines `Trait` and adds the appropriate requirement.

This change means that `Runner.Plan.Action.run` now has an associated value, because whether or not a test runs parallelized is necessarily part of the step for that test. I've made the associated value an options-y structure so that we can extend it later.

> [!NOTE]
> It's very hard to write unit tests that validate that `.serial` does what it's designed to do; I'm open to writing additional tests if we can think up some that can be implemented in a reliable way.

Resolves rdar://95629402.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -413,6 +413,9 @@ extension Runner.Plan.Action {
   @_spi(ExperimentalSnapshotting)
   public enum Snapshot: Sendable, Codable {
     /// The test should be run.
+    ///
+    /// - Parameters:
+    ///   - options: Options to apply to this action when it is run.
     case run(options: RunOptions)
 
     /// The test should be skipped.
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -14,8 +14,29 @@ extension Runner {
   public struct Plan: Sendable {
     /// The action to perform for a test in this plan.
     public enum Action: Sendable {
+      /// A type describing options to apply to actions of case
+      /// ``Runner/Plan/Action/run(options:)`` when they are run.
+      public struct RunOptions: Sendable, Codable {
+        /// Whether or not this step should be run in parallel with other tests.
+        ///
+        /// By default, all steps in a runner plan are run in parallel if the
+        /// ``Configuration/isParallelizationEnabled`` property of the
+        /// configuration passed during initialization has a value of `true`.
+        ///
+        /// Traits such as ``Trait/serial`` applied to individual tests may
+        /// affect whether or not that test is parallelized.
+        ///
+        /// ## See Also
+        ///
+        /// - ``SerialTrait``
+        public var isParallelizationEnabled: Bool
+      }
+
       /// The test should be run.
-      case run
+      ///
+      /// - Parameters:
+      ///   - options: Options to apply to this action when it is run.
+      case run(options: RunOptions)
 
       /// The test should be skipped.
       ///
@@ -44,6 +65,19 @@ extension Runner {
           // case.
           return true
         }
+      }
+
+      /// Whether or not this action enables parallelization.
+      ///
+      /// If this action is of case ``run(options:)``, the value of this
+      /// property equals the value of its associated
+      /// ``RunOptions/isParallelizationEnabled`` property. Otherwise, the value
+      /// of this property is `nil`.
+      var isParallelizationEnabled: Bool? {
+        if case let .run(options) = self {
+          return options.isParallelizationEnabled
+        }
+        return nil
       }
     }
 
@@ -140,12 +174,13 @@ extension Runner.Plan {
     // Convert the list of test into a graph of steps. The actions for these
     // steps will all be .run() *unless* an error was thrown while examining
     // them, in which case it will be .recordIssue().
+    let runAction = Action.run(options: .init(isParallelizationEnabled: configuration.isParallelizationEnabled))
     var testGraph = Graph<String, Test?>()
-    var actionGraph = Graph<String, Action>(value: .run)
+    var actionGraph = Graph<String, Action>(value: runAction)
     for test in tests {
       let idComponents = test.id.keyPathRepresentation
       testGraph.insertValue(test, at: idComponents)
-      actionGraph.insertValue(.run, at: idComponents, intermediateValue: .run)
+      actionGraph.insertValue(runAction, at: idComponents, intermediateValue: runAction)
     }
 
     // Ensure the trait lists are complete for all nested tests. (Make sure to
@@ -171,7 +206,7 @@ extension Runner.Plan {
         return
       }
 
-      var action = Action.run
+      var action = runAction
       var firstCaughtError: (any Error)?
 
       // Walk all the traits and tell each to prepare to run the test.
@@ -181,7 +216,11 @@ extension Runner.Plan {
       // `SkipInfo`, the error should not be recorded.
       for trait in test.traits {
         do {
-          try await trait.prepare(for: test)
+          if let trait = trait as? any SPIAwareTrait {
+            try await trait.prepare(for: test, action: &action)
+          } else {
+            try await trait.prepare(for: test)
+          }
         } catch let error as SkipInfo {
           action = .skip(error)
           break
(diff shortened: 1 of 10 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

briancroom (approved):

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -374,7 +413,7 @@ extension Runner.Plan.Action {
   @_spi(ExperimentalSnapshotting)
   public enum Snapshot: Sendable, Codable {
     /// The test should be run.
-    case run
+    case run(options: RunOptions)
```
stmontgomery, line 416: `- Parameters:` ...

Sources/Testing/Testing.docc/Parallelization.md
```diff
@@ -0,0 +1,63 @@
...
+Parallelization can be disabled on a per-function or per-suite basis using the
+``Trait/serial`` trait:
+
+```swift
+@Test(.serial, arguments: Food.allCases) func prepare(food: Food) {
+  // This function will be invoked serially, once per food, because it has the
+  // .serial trait.
+}
+
+@Suite(.serial) {
```
stmontgomery, line 33: Missing `struct` or similar

grynspan: lol

Sources/Testing/Running/Runner.swift
```diff
@@ -114,12 +114,55 @@ extension Runner {
...
+  ///   - step: The plan step that controls parallelization. If `nil`, or if its
+  ///   ``Runner/Plan/Step/action`` property is not of case
+  ///   ``Runner/Plan/Action/run(options:)``, the
+  ///   ``Configuration/isParallelizationEnabled`` property of this runner's
+  ///   ``configuration`` property is used instead to determine if
+  ///   parallelization is enabled.
+  ///   - body: The function to invoke.
+  ///
+  /// - Throws: Whatever is thrown by `body`.
+  private func _forEach<E>(
```
stmontgomery, line 131: Would this be more readable as an instance method in an extension of `Sequence`?

Sources/Testing/Testing.docc/Parallelization.md
```diff
@@ -0,0 +1,63 @@
+# Parallelizing test execution
+
+<!--
+This source file is part of the Swift.org open source project
+
+Copyright (c) 2023 Apple Inc. and the Swift project authors
```
stmontgomery, line 6:
```suggestion
Copyright (c) 2024 Apple Inc. and the Swift project authors
```

Sources/Testing/Testing.docc/Parallelization.md
```diff
@@ -0,0 +1,63 @@
...
+trait causes that suite to run its contained test functions and sub-suites
+serially instead of in parallel.
+
+This trait is recursively applied: if it is applied to a suite, any
+parameterized tests or test suites contained in that suite are also serialized
+(as are any tests contained in those suites, and so on.)
+
+This trait does not affect the execution of a test relative to its peers or to
+unrelated tests. This trait has no effect if test parallelization is globally
+disabled (by, for example, passing `--no-parallel` to the `swift test` command.)
```
stmontgomery, line 53: Would also be good to contrast this with the approach of applying a global actor, such as `@MainActor`, to tests, and explain the nuances of each

Sources/Testing/Traits/SPIAwareTrait.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+
+/// A protocol describing traits that can be added to a test function or to a
+/// test suite and that make use of SPI symbols in the testing library.
+///
+/// This protocol refines ``Trait`` in various ways that require the use of SPI.
+/// Ideally, such requirements will be promoted to API when their design
+/// stabilizes.
+@_spi(ExperimentalTestRunning)
+public protocol SPIAwareTrait: Trait {
```
stmontgomery, line 18: Could we make this non-`@_spi public` for now, just `internal`? 
```suggestion
protocol InternalTrait: Trait {
```

Sources/Testing/Testing.docc/Parallelization.md
```diff
@@ -0,0 +1,63 @@
...
+    // containing suite has the .serial trait.
+  }
+}
+```
+
+When added to a parameterized test function, this trait causes that test to run
+its cases serially instead of in parallel. When applied to a non-parameterized
+test function, this trait has no effect. When applied to a test suite, this
+trait causes that suite to run its contained test functions and sub-suites
+serially instead of in parallel.
```
stmontgomery, line 45: It might be worth including an example code snippet of a suite type containing two non-parameterized test functions, just to reinforce this statement

Sources/Testing/Traits/SPIAwareTrait.swift
grynspan: Well my thinking was that developers may want to create custom trait types, and might want to be able to leverage the same SPI we can. I can make this internal-only, but then we're the only ones who can experiment with the interfaces in this protocol.

Sources/Testing/Testing.docc/Parallelization.md
grynspan: Agreed; I'm going to save that for a separate PR though in which I'll flesh out this document.

Sources/Testing/Traits/SerialTrait.swift
```diff
@@ -0,0 +1,58 @@
...
+
+// MARK: -
+
+extension Trait where Self == SerialTrait {
+  /// A trait that serializes the test to which it is applied.
+  ///
+  /// ## See Also
+  ///
+  /// - ``SerialTrait``
+  public static var serial: Self {
```
stmontgomery, line 55: My primary comment about this PR (which introduces some _really_ useful new functionality!) is the way the trait is exposed as API. I strongly suspect that over time, we are going to want to offer more than one way to control parallelization with varying levels of granularity/scope. A few examples:

- Some test codebases might be configured to disable parallelization _by default_, but allow selected tests to opt-in to parallelization. This kind of workflow would allow incremental adoption of parallelization over time. 
- Some common testing workflows involve running more than one test target, and the supervising tool may choose to run those targets in parallel with each other. From the testing library's perspective, it may only be aware of a single process—the one running the current test target—so that is all it can _directly_ control. But a test author might need to tell the supervising tool whether it is safe to parallelize other tests in other processes at the same time as this process. If tests touch system-global state, such as the file system, this can be important.
- A user might want the test functions in a suite to _not_ be parallelized relative to each other, but might want the test cases of each of any parameterized test functions to still be parallelized with respect to the other cases — or vice-versa.

As I've been considering how we would design APIs for these use cases eventually, it feels like we'd want a slightly more generic trait which allows specifying a parallelization "style" or strategy. The above examples _could_ look something like this:

- `.parallelized(true)`: Basic opt-in to parallelization.
- `.parallelized(.serial)`: Conceptually the same as `.serial` in this PR.
- `.parallelized(.exclusivelySerial)`: A stronger form of serialization, which expands the scope to all test targets/processes.
- `.parallelized(.byCase)`: Enable parallellization of test cases, but not of test functions.

These are just examples, and the exact spellings of these things are made up for this discussion. But my thinking is that it may serve us well to provide a more generalized trait like `.parallelized` (or a different spelling) which allows _configuring_ parallelization, instead of having `SerialTrait` be a top-level trait of its own.

Sources/Testing/Running/Runner.swift
grynspan: Possibly, although that would require hoisting it out and probably into a separate file SequenceAdditions.swift, whereas it's really an implementation detail of `Runner`.

grynspan:
@swift-ci please test

stmontgomery (approved):

PR #218, opened 2024-02-13
Support re-running parameterized tests with RawRepresentable arguments automatically

stmontgomery:
This augments our heuristics for determining whether an argument to a parameterized test can be re-run, by allowing `RawRepresentable`-conforming types whose associated `RawValue` type is `Encodable`.

### Motivation:

A specific, common use case of this is to allow enum types imported from C or Objective-C to automatically be able to be re-run, which will work with these changes because such types are imported as `RawRepresentable`-conforming types with a numeric `RawValue` type.

### Modifications:

- Look for `RawRepresentable` types w/encodable `RawValue` in `Test.Case.Argument.ID.init(identifying:parameter:)`.
- Add a new section to the Parameterized Testing article detailing the known protocols and their precedence.
- Add some unit tests.

### Checklist:

Resolves rdar://122321338

### DIFF
--- a/Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
+++ b/Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
@@ -15,24 +15,9 @@ private import Foundation
 /// A protocol for customizing how arguments passed to parameterized tests are
 /// encoded, which is used to match against when running specific arguments.
 ///
-/// By default, the testing library checks whether a test argument conforms to
-/// `Encodable` and encodes it using `encode(to:)` if it does. `Encodable` is
-/// sufficient for many types, but for some types the output of that function
-/// may contain large data payloads that cause poor performance, may not be
-/// stable and unique, or may otherwise be deemed unsuitable for testing.
+/// ## See Also
 ///
-/// If the type of the argument does not conform to `Encodable` but it does
-/// conform to `Identifiable` and its associated `ID` type conforms to
-/// `Encodable`, the value of its `id` property is encoded with `encode(to:)`
-/// and is used as the encoded representation.
-///
-/// If neither of the approaches for encoded representation described above is
-/// sufficient, a type can be made to conform to
-/// ``CustomTestArgumentEncodable``. The testing library will then call
-/// ``encodeTestArgument(to:)`` on values of that type instead of `encode(to:)`.
-///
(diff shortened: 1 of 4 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
```diff
@@ -15,24 +15,9 @@ private import Foundation
...
-/// conform to `Identifiable` and its associated `ID` type conforms to
-/// `Encodable`, the value of its `id` property is encoded with `encode(to:)`
-/// and is used as the encoded representation.
-///
-/// If neither of the approaches for encoded representation described above is
-/// sufficient, a type can be made to conform to
-/// ``CustomTestArgumentEncodable``. The testing library will then call
-/// ``encodeTestArgument(to:)`` on values of that type instead of `encode(to:)`.
-///
-/// A type that conforms to this protocol is not required to conform to either
```
grynspan, line 34: I think we still want this note on the type docs.

grynspan (changes requested):

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
stmontgomery: Are you talking specifically about this final paragraph,
> A type that conforms to this protocol is not required to conform to either `Encodable` or `Decodable`.

or more than that? I can add part of this back, but a lot of it is now covered more comprehensively by the additions to `ParameterizedTesting.md`

stmontgomery:
@swift-ci please test

briancroom (approved):

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
grynspan: Yes, just that line. I think it's an important bit of information about the protocol itself and belongs in the protocol documentation (even if we end up repeating it in the prose docs.) 😄

stmontgomery: Sounds good, I have restored that line in my revised docs for this protocol

grynspan (approved):

PR #219, opened 2024-02-14
Reduce the sizes of `Event.Kind`, `Issue.Kind`, and `Runner.Plan.Action`.

grynspan:
Right now, all cases in the `Event.Kind`, `Issue.Kind`, and `Runner.Plan.Action` enums are direct, i.e. stored inline in instances of those enums. That requires that the enums be at least as large as their largest associated values, which turns out to be a fair bit:

| Type | Size (Bytes) |
|-|-|
| `Event` | 464 |
| `Event.Kind` | 345 |
| `Issue` | 225 |
| `Issue.Kind` | 160 |
| `Runner.Plan.Action` | 225 |

Nearly half a kilobyte for an event is excessive. We can reduce the space needed by storing the associated values of these enums _indirectly_ (using the `indirect` keyword.) There are cases of both enums that don't have any associated values, so let's just make _all_ associated values indirect. We end up with:

| Type | Size (Bytes) |
|-|-|
| `Event` | 120 |
| `Event.Kind` | 8 |
| `Issue` | 73 |
| `Issue.Kind` | 8 |
| `Runner.Plan.Action` | 8 |

The tradeoff is that indirect storage is basically a pointer dereference, so we lose locality—but the frequency with which events and issues are raised doesn't make the performance cost worth worrying about.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Event.swift
+++ b/Sources/Testing/Events/Event.swift
@@ -329,13 +329,13 @@ extension Event.Kind {
     /// To enable events of this kind, set
     /// ``Configuration/deliverExpectationCheckedEvents`` to `true` before
     /// running tests.
-    case expectationChecked(_ expectation: Expectation.Snapshot)
+    indirect case expectationChecked(_ expectation: Expectation.Snapshot)
 
     /// An issue was recorded.
     ///
     /// - Parameters:
     ///   - issue: The issue which was recorded.
-    case issueRecorded(_ issue: Issue.Snapshot)
+    indirect case issueRecorded(_ issue: Issue.Snapshot)
 
     /// A test ended.
     case testEnded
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

dennisweissmann (approved):

briancroom:
For my education, how did you measure the sizes to produce those tables?

grynspan:
> For my education, how did you measure the sizes to produce those tables?

```swift
@Test func f() {
  func p<T>(_ t: T.Type) {
    print(T.self, ": ", MemoryLayout<T>.size)
  }
  p(Event.self)
  p(Event.Kind.self)
  p(Issue.self)
  p(Issue.Kind.self)
}
```

Then prettified them with Markdown.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

PR #220, opened 2024-02-15
Migrate file I/O code to a dedicated `FileHandle` type.

grynspan:
This PR migrates the bulk of our file I/O code into a single file/type, `FileHandle`. Previously, the code we used to touch the file system and `stderr` was spread across several Swift files. This has started to become a bit difficult to maintain, especially when you consider platform-specific differences (I'm looking at you, Windows…)

So I've moved the code into a single type that wraps a C `FILE *` handle. The type is move-only and sendable, and abstracts away most of the platform-specific code we had around files. Access to the underlying file handle is done via a scoped-access "with" function.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/EntryPoint.swift
+++ b/Sources/Testing/Running/EntryPoint.swift
@@ -41,7 +41,7 @@ private import TestingInternals
         oldEventHandler(event, context)
       }
 
-      var options: [Event.ConsoleOutputRecorder.Option] = .forStandardError
+      var options: [Event.ConsoleOutputRecorder.Option] = .for(.stderr)
       if args.contains("--verbose") {
         options.append(.useVerboseOutput)
       }
@@ -49,9 +49,10 @@ private import TestingInternals
       await runTests(options: options, configuration: configuration)
     }
   } catch {
-    let stderr = swt_stderr()
-    fputs(String(describing: error), stderr)
-    fflush(stderr)
+    FileHandle.stderr.withUnsafeCFILEHandle { stderr in
+      fputs(String(describing: error), stderr)
+      fflush(stderr)
+    }
 
     exitCode.withLock { exitCode in
       exitCode = EXIT_FAILURE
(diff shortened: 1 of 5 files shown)


grynspan:
@swift-ci please test

Tests/TestingTests/Support/FileHandleTests.swift
```diff
@@ -0,0 +1,146 @@
...
+#endif
+
+  @Test("Can write to a file")
+  func canWrite() throws {
+    // NOTE: we are not trying to test mkstemp() here. We are trying to test the
+    // capacity of FileHandle to open a file for writing, and need a temporary
+    // file to write to.
+    let fileName = "can_write_to_file_\(UInt64.random(in: 0 ..< .max))"
+#if os(Windows)
+    let cPath = try #require(_tempnam("c:\\tmp", fileName))
```
compnerd, line 59: This path is not really guaranteed to exist. Computing a temporary path is pretty complicated, so might be better to use `GetTempPathW` here.

grynspan: This code is borrowed directly from Microsoft's documentation, but I can change it to just use `tmpnam_s()` instead.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test linux

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Support/FileHandle.swift
```diff
@@ -0,0 +1,224 @@
...
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+internal import TestingInternals
+
+#if !SWT_NO_FILE_IO
```
compnerd, line 13: A newline after this to match the `#endif` would be nice.

Sources/Testing/Support/FileHandle.swift
```diff
@@ -0,0 +1,224 @@
...
+/// A type representing a file handle.
+///
+/// Instances of this type abstract away the C `FILE *` type for Swift.
+///
+/// Whether or not a particular C file handle is safe to use concurrently is
+/// platform- and file-specific. The testing library assumes that the standard
```
compnerd, line 19: A space after `file-` would be nice to be clearer that `specific` applies to both `file` and `platform`.

compnerd (approved):

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -260,14 +233,14 @@ func runTests(options: [Event.ConsoleOutputRecorder.Option], configuration: Conf
 
 extension [Event.ConsoleOutputRecorder.Option] {
   /// The set of options to use when writing to the standard error stream.
-  static var forStandardError: Self {
+  static func `for`(_ fileHandle: borrowing FileHandle) -> Self {
```
stmontgomery, line 236: Why not make this an `init`?

Sources/Testing/Support/FileHandle.swift
grynspan: This is grammatically correct and I will die on this hill!

Sources/Testing/Running/EntryPoint.swift
grynspan: Custom initializers on array types are somewhat discouraged, and it reads better this way IMHO.

grynspan:
@swift-ci please test

Sources/Testing/Support/FileHandle.swift
grynspan: Fixed it by deleting the one at the end instead.

stmontgomery (approved):

grynspan:
This passed on the last run and the macOS buildbot is hanging right now. The only change is deleting a newline, so I'm going to (irresponsibly) merge this PR before the build completes.

PR #221, opened 2024-02-15
Make Expression.expandedDescription(depth:includingTypeNames:includingParenthesesIfNeeded:) -> String SPI

dennisweissmann:
Expose `Expression.expandedDescription(depth:includingTypeNames:includingParenthesesIfNeeded:) -> String SPI` which is currently IPI as SPI so that it can be used from outside the Testing module.

### Motivation:
If you want to inspect issues from outside the Testing library, it would be useful to have that method available since it provides a nice summary of the expression that failed a test.

### Modifications:

Added `@_spi(ExperimentalSourceCodeCapturing) public` to `Expression.expandedDescription(depth:includingTypeNames:includingParenthesesIfNeeded:) -> String SPI`.

### Result:

`Expression.expandedDescription(depth:includingTypeNames:includingParenthesesIfNeeded:) -> String SPI` is now accessible from outside the Testing module.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/Expression.swift
+++ b/Sources/Testing/SourceAttribution/Expression.swift
@@ -207,7 +207,8 @@ public struct Expression: Sendable {
   ///     information this instance contains.)
   ///
   /// - Returns: A string describing this instance.
-  func expandedDescription(depth: Int = 0, includingTypeNames: Bool = false, includingParenthesesIfNeeded: Bool = true) -> String {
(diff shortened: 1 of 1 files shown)


grynspan:
Okay for now, however we probably will want to revisit and provide a proper SPI interface into this functionality in the future.

grynspan (approved):

grynspan:
@swift-ci please test

briancroom (approved):

PR #223, opened 2024-02-16
Rethink symbolic vs. string literal tags.

grynspan:
## Introduction

Today, every tag wraps a string, which means that symbolic tags can be represented in two different ways:

```swift
extension Tag {
  static var critical: Self { "unimportant" }
}
```

The above tag can be referred to as `.critical` or as `"unimportant"` when added to a test, and the two names are equivalent.

The raw string value is necessary because tags need some sort of _identity_. If a tag were just `Tag()` with no associated metadata, then any two tags would compare equal, and there would be no way to distinguish them.

## Problem

String values are important for Swift Package Manager because when working at the command line, all you have to help you out are (about) 102 keys and the `--help` option. If tags map to and from strings, then you can write something like `swift test --filter-by-tag '.critical'` and have it "just work." But because a tag may have two distinct names, we need to make decisions like:

1. Should a test tagged `"unimportant"` be run if a developer asks to run tests with the `.critical` tag?
1. If a `.critical` tag is defined in two different contexts (say, two `fileprivate` declarations in different files), should they be treated as equivalent even if they have different raw string values?
1. Should `let critical: Tag = ...` at file scope be treated as equivalent to `.critical` (a static member of `Tag`)?
1. Are any of these possible equivalences transitive? If `.critical` and `critical` are the same, is `critical` the same as `"unimportant"`?
1. What happens if somebody declares a tag with a string value that's only known at runtime (e.g. taken from an environment variable)?
1. If a tag is declared symbolically in another module (e.g. `let experimental: Tag = "Swift Experimental Feature"`) how do we find out its string representation and consistently recognize it as equivalent if we don't have access to the source code of the other module?

We don't have good answers to these questions, and we've found that any solution we try to design ends up unsound or, at best, inexplicable to the end user.

## Solution

This PR refactors `Tag` pretty significantly. Gone is `RawRepresentable` conformance. In its place are two disjoint families of tags: tags declared as static members of `Tag`, and tags declared as compile-time constant string literals.

### Static members of Tag

Tags declared as static members use a new `@Tag` macro (name subject to change) to indicate at compile-time that they are tags that may be applied to tests:

```swift
extension Tag {
  @Tag static var critical: Tag
}

@Test(.tags(.critical)) func f() { ... }
```

> [!NOTE]
> The exact shape of this macro may change in the future, especially with recently landed changes to swift-syntax to allow gathering the lexical context of a macro. We may in the future change the above to something like:
>
> ```swift
> extension Tag {
>   static var critical = #tag
> }
>
> @Test(.tags(.critical)) func f() { ... }
> ```
>
> But that's a future direction we are unable to implement at this time, so best not to dwell on it.

Note that tags declared this way _do not have raw string values associated with them_. The ambiguity of `.critical` versus `"unimportant"` above is eliminated. We also disallow applying this macro to symbols that are _not_ members of `Tag`, so a file-scoped `critical` tag is not possible, and that ambiguity is eliminated too.

### String literals

Tags declared as string literals behave more or less the same as before:

```swift
@Test(.tags("unimportant")) func f() { ... }
```

### Expressing tags at the command line

We define a trivial transformation function to/from tag and string: if the first character of a string is a period, then the string is interpreted as the name of a static member tag; otherwise, it is interpreted as the value of a string literal tag. A leading slash is dropped, but indicates that the tag is a string literal tag, so `".oddChoice"` can be "escaped" and correctly interpreted. This format is easy to represent at the command line and easy to understand/explain:

```sh
# Run tests with the symbolic tag .critical or the string tag "unimportant"
swift test --filter-by-tags .critical unimportant
```

### Macro knowledge of tags

One unfortunate caveat here is that the compiler, by way of the testing library's macro target, needs to have knowledge of tags. We'd really prefer it if the macro target had as little knowledge of individual traits as possible, but given the integration tags have (or rather, should have) with tools like Swift Package Manager, it seems unavoidable that other parts of the swift-testing stack also need to understand them.

Macro targets are limited to examining syntax and do not have type information. In the future, we could explore changes to the language and compiler to allow better compile-time context for the testing library's macros and to improve the overall experience for developers when using tags.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/AddingTags.md
+++ b/Sources/Testing/Testing.docc/AddingTags.md
@@ -26,25 +26,84 @@ source files, and even test targets.
 
 ## Adding tags
 
-To add a tag to a test, use the ``Trait/tags(_:)`` trait. These traits take
-sequences of tags as arguments, and those tags are then applied to the
+To add a tag to a test, use the ``Trait/tags(_:)-505n9`` trait. These traits
+take sequences of tags as arguments, and those tags are then applied to the
 corresponding test at runtime. If they are applied to a test suite, then all
 tests in that suite inherit those tags.
 
-Tags themselves are instances of ``Tag`` and can be expressed as string literals
-or as named constants declared elsewhere:
+The testing library does not assign any semantic meaning to any tags, nor does
+the presence or absence of tags affect how the testing library runs tests.
+
+Tags themselves are instances of ``Tag`` and can be expressed as string
+literals directly in a test suite or test function's declaration. Tags can also
+be expressed as named constants declared as static members of ``Tag``. To
+declare a named constant tag, use the ``Tag()`` macro:
 
 ```swift
 extension Tag {
-  static let legallyRequired = Tag(...)
+  @Tag static var legallyRequired: Self
 }
 
-@Test("Vendor's license is valid", .tags("critical", .legallyRequired))
+@Test("Vendor's license is valid", .tags("critical"), .tags(.legallyRequired))
 func licenseValid() { ... }
 ```
 
-The testing library does not assign any semantic meaning to any tags, nor does
-the presence or absence of tags affect how the testing library runs tests.
+If two tags with the same name (`legallyRequired` in the above example) are
+declared in different files, modules, or other contexts, the testing library
+treats them as equivalent.
+
+If it is important for a tag to be distinguished from similar tags declared
+elsewhere in a package or project (or its dependencies), use
+ [reverse-DNS naming](https://en.wikipedia.org/wiki/Reverse_domain_name_notation)
+to create a unique Swift symbol name for your tag:
+
+```swift
+extension Tag {
+  enum com_example_foodtruck {}
+}
+
+extension Tag.com_example_foodtruck {
+  @Tag static var extraSpecial: Self
+}
+
+@Test
+  "Extra Special Sauce recipe is secret",
+  .tags(.com_example_foodtruck.extraSpecial)
+)
+func secretSauce() { ... }
+```
+
+### Where tags can be declared
+
+Tags must always be declared as members of ``Tag`` in an extension to that type
+or in a type nested within ``Tag``. Redeclaring a tag under a second name has no
+effect and the additional name will not be recognized by the testing library.
+The following example is unsupported:
+
+```swift
+extension Tag {
+  @Tag static var legallyRequired: Self // ✅ OK: declaring a new tag
+
+  static var requiredByLaw: Self { // ❌ ERROR: this tag name will not be
+                                   // recognized at runtime
+    legallyRequired
+  }
+}
+```
+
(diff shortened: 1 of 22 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

briancroom (approved):

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Traits/Tag.swift
```diff
@@ -10,56 +10,158 @@
...
+      if stringLiteral.first == "." || stringLiteral.first == #"\"# {
+        #"\\#(stringLiteral)"#
+      } else {
+        stringLiteral
+      }
+    }
+  }
+
+  public func encode(to encoder: any Encoder) throws {
+    try _codableStringValue.encode(to: encoder)
```
stmontgomery, line 125: This is encoding/decoding `Tag` as a single-value, right? I wonder if we may find that limiting, and would find it more future-proof to enclose it in a keyed container

grynspan: It is. If, in the future, we need more flexibility, we can adapt these methods to support multiple formats. Better for now to keep the encoding simple.

(Also, since we're currently using tags as keys in tag-colors.json, they need to be stringy, though that particular part of the library is easily forgotten.)

Sources/TestingMacros/Support/TagConstraints.swift
```diff
@@ -0,0 +1,89 @@
...
+  // Find tags that are in an unsupported format (only .member and "literal"
+  // are allowed.)
+  for traitExpr in traitExprs {
+    // At this time, we are only looking for .tags() traits in this function.
+    guard let functionCallExpr = traitExpr.as(FunctionCallExprSyntax.self),
+          let calledExpr = functionCallExpr.calledExpression.as(MemberAccessExprSyntax.self) else {
+      continue
+    }
+
+    // Check for .tag() traits.
```
stmontgomery, line 36:
```suggestion
    // Check for .tags() traits.
```

Sources/Testing/Testing.docc/AddingTags.md
```diff
@@ -26,25 +26,84 @@ source files, and even test targets.
...
+}
+
+extension Tag.com_example_foodtruck {
+@Tag static var extraSpecial: Self
+}
+
+@Test(
+"Extra Special Sauce recipe is secret",
+.tags(.com_example_foodtruck.extraSpecial)
+)
```
stmontgomery, line 72: Is the indentation expected here?

stmontgomery (approved):

Sources/Testing/Testing.docc/AddingTags.md
grynspan: No, looks like it got messed up at some point. Will correct.

grynspan:
@swift-ci please test

PR #226, opened 2024-02-18
Convert uses of `@unchecked Sendable` to checked `Sendable`.

grynspan:
`@unchecked Sendable` disables (most) concurrency checking on types to which it is applied. We have a couple of types, `FileHandle` and `Locked`, that use it because they are internally concurrency-safe but use lower-level types that cannot prove their safety to the concurrency checker at compile time.

This PR changes those types to use our dedicated `UncheckedSendable` wrapper to wrap their corresponding unsendable stored properties. This means that the types themselves can be `Sendable` rather than `@unchecked Sendable`, so we can still get concurrency checking for their code and storage that doesn't directly touch the unchecked stored properties.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/FileHandle.swift
+++ b/Sources/Testing/Support/FileHandle.swift
@@ -21,9 +21,9 @@ internal import TestingInternals
 /// platforms.
 ///
 /// This type is not part of the public interface of the testing library.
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci please test

PR #227, opened 2024-02-18
Fix build when `SWT_NO_ENVIRONMENT_VARIABLES` is set.

grynspan:
The build is broken when `SWT_NO_ENVIRONMENT_VARIABLES` is set. This PR fixes that by adjusting call sites that use `Environment._environment` (since `Locked` is no longer a property wrapper and we moved `setVariable(_:named:)` to the test target.)

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #230, opened 2024-02-21
Add an SPI interface for excluding tests when running.

grynspan:
We have an interface for filtering tests so only a selection of them run. This PR adds an equivalent, yet inverted, interface for excluding tests that should _not_ run.

Resolves rdar://123389027.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

briancroom (approved):

PR #231, opened 2024-02-22
Allow filtering tests by arbitrary conditions.

grynspan:
This PR replaces the `Configuration.TestFilter` function typealias with a nominal type (a structure) that allows us to apply arbitrary predicate functions to tests.

A simple predicate function is insufficient because it does not track enough information about a test and its position in a test graph. For instance, to only run tests tagged `"hello"`, we might naïvely write a test filter function like:

```swift
configuration.testFilter = { test in
	test.tags.contains("hello")
}
```

But this will result in the test's _containing_ suite (or, recursively, suites) to not be included in the test plan. The test function will run, but events will not be generated for its enclosing scopes. This makes such functions appear to be global/free functions, which they are not. Badness ensues if you're writing a test harness or CI plugin that expects events for containing types to be delivered before and after member test functions.

This is generally solved when filtering by known test ID because we have an internal structure, `Test.ID.Selection`, that we use to fuzzy-match test IDs. However, that type requires us to know the set of matching test IDs upfront, whereas filtering by tag cannot be performed until _after_ we know the set of tests we might want to run.

The new type contains, in effect, two internal abstractions: one that takes a test ID selection and uses it to trivially filter a test graph, and one that takes a function, applies it to each test, and uses it to construct a test ID selection (that it then uses to filter the test graph). There's probably a more algorithmically efficient way to do this, but I've been unable to come up with something that still produces the expected results for complex test graphs and filtering conditions.

Resolves rdar://119634169.

### Checklist:


grynspan:
WIP: needs unit test coverage.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

briancroom (approved):

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -0,0 +1,295 @@
...
+  }
+
+  /// Initialize this instance to operate based on a set of tags.
+  ///
+  /// - Parameters:
+  ///   - tags: The set of tags to either include or exclude.
+  ///   - anyOf: Whether tests must have _any_ of the tags in `tags` (as opposed
+  ///     to all of them.)
+  ///   - membership: How to interpret the result when predicating tests.
+  init(tags: some Collection<Tag>, anyOf: Bool, membership: Membership) {
```
stmontgomery, line 131: I feel like it might be worth exposing this `init`, which accepts the `anyOf:` and `membership:` parameters, directly. Otherwise, a tool adopting this would need to selectively call one of the more specific `init` overloads below using logic that would seem cleaner if they could just pass these values directly.

I realize, as it's currently written this suggestion would require elevating the `Membership` enum to `public`, but It seems like instead of doing that, this parameter could be represented as a plain `Bool` type? I.e.

```swift
public init(tags: some Collection<Tag>, anyOf: Bool, including: Bool) {
  // ...convert `including` to `Membership` first, then call the existing init...
}
```

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -0,0 +1,295 @@
...
+    if !includeHiddenTests {
+      result = result.mapValues { test in
+        (test?.isHidden == true) ? nil : test
+      }
+    }
+
+    return result
+  }
+
+  /// Combine this test filter with another one.
```
stmontgomery, line 261: nit: these `combining` functions might make sense in their own `extension`, separate from `apply()` above

stmontgomery (approved):

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -0,0 +1,295 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
```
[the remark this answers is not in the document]
grynspan: Oops.

Sources/Testing/Running/Configuration.TestFilter.swift
grynspan: I intentionally did not include such an overload because it raises the aesthetic question of whether `including` or `excluding` should be `true` or `false`, and how to properly represent `anyOf` vs. `allOf` (presumably another enum.)

Since this is a tools-only interface, we can always improve upon it later!

grynspan:
@swift-ci please test

PR #234, opened 2024-02-26
Guard the use of `FILE *` in C++ with the `SWT_NO_FILE_IO` condition.

grynspan:
This PR just uses the same `SWT_NO_FILE_IO` condition to guard uses of `FILE *` in our C++ internal target. This will aid porting to embedded platforms that do not support standard C I/O.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #235, opened 2024-02-26
Ensure that type names used in extensions are correctly parsed.

grynspan:
We use the internal runtime function `_typeName(_:qualified:)` to get fully qualified type names in a few places. If a type is referenced from an extension to that type in another module, the resulting string is prefixed with something like `"(extension in OtherModule):"` which messes with test plan graphing and our general expectations about the names of types.

This PR normalizes how we get fully-qualified type names so that we always go through a particular function that is aware of this prefix and knows how to remove it.

The extra string is emitted [here](https://github.com/apple/swift/blob/b0f3338d255d1d19ae0feedef132f1da4b5d5c11/lib/Demangling/NodePrinter.cpp#L1474) by the demangler.

### Checklist:

### DIFF
--- a/Sources/Testing/Parameterization/TypeInfo.swift
+++ b/Sources/Testing/Parameterization/TypeInfo.swift
@@ -33,8 +33,8 @@ public struct TypeInfo: Sendable {
   /// - Parameters:
   ///   - type: The type which this instance should describe.
   init(describing type: Any.Type) {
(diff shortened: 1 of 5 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

MattSanford (approved):

PR #236, opened 2024-02-27
Rename `Package@swift-5.11.swift` to `Package@swift-6.0.swift`

MaxDesiatov:
### Motivation:

This package currently can't be built with `main` branch of SwiftPM, since version 5.11 no longer exists.

### Modifications:

 The package manifest should refer to 6.0 instead.

### Result:

This package can be built with Swift 6.0.

### DIFF
--- a/Package@swift-6.0.swift
+++ b/Package@swift-6.0.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.11
+// swift-tools-version: 6.0
 
 //
 // This source file is part of the Swift.org open source project
--- a/Documentation/Releases.md
+++ b/Documentation/Releases.md
@@ -40,7 +40,7 @@ git checkout -b release/x.y.z
 
 ## Preparing the repository's contents
 
-The package manifest files (Package.swift _and_ Package@swift-5.11.swift) must
+The package manifest files (Package.swift _and_ Package@swift-6.0.swift) must
 be updated so that the release can be used as a package dependency:
 
 1. Take note of any availability definitions (passed to the Swift compiler with


Package@swift-6.0.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.11
+// swift-tools-version: 6.0
```
grynspan, line 1: We need to keep a -5.11 manifest around for folks who don't have a bleeding-edge toolchain installed.

grynspan (changes requested):
There are other files in the repository that refer to Swift 5.11 that should also be updated. As well, Documentation/Releases.md specifically needs to be updated to refer to the correct file.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test Linux

Package@swift-6.0.swift
grynspan: Thinking about that further, if we _don't_ keep a 5.11 manifest around, I think the effect is that SPM 5.11 will use the 5.10 manifest, right? So the only real consequence is that Git information will be unavailable, which is fine.

grynspan:
@swift-ci please test

grynspan (approved):
The main toolchain used by CI has been updated, so other PRs are starting to fail. I've tested this change locally on macOS, Ubuntu, and Windows with the last-available 5.11 toolchains and they are functional (though they miss `GitInformation`.) So we're okay to merge this now.

PR #239, opened 2024-02-27
Avoid direct calls to `fwrite()` and `fputs()`.

grynspan:
This PR was originally going to resolve what looked like a bug on Windows re: Unicode support, but testing shows that it actually isn't needed. However, the overall code changes make `FileHandle` a bit nicer/less-redundant to use, so I'm opening a PR with them anyway. 🚀

~~Ensure terminal output on Windows is interpreted as Unicode.~~

~~Right now, we use `fputs()` to write to `stderr` on all platforms. On Linux and Darwin, this is fine because we can generally assume terminals use UTF-8 there (technically this is an unsafe assumption, but in practice it hasn't been an issue.)~~

~~On Windows, the console may not be configured with the UTF-8 pseudo-codepage (65001) and may instead still be using a locale-specific codepage like Latin-1 (1252).~~

~~Microsoft specifically [recommends](https://learn.microsoft.com/en-us/windows/console/console-code-pages) using the `"w"` functions when reading from or writing to the console. By doing so, we can ensure our output is correctly interpreted as Unicode.~~

~~Microsoft also [documents](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/fputs-fputws?view=msvc-170) that `fputs()` doesn't work with streams opened as Unicode, but `fputws()` works with ANSI and Unicode streams, so that's another reason to switch.~~

_[One line description of your change]_

### Checklist:

### DIFF
--- a/Sources/Testing/Running/EntryPoint.swift
+++ b/Sources/Testing/Running/EntryPoint.swift
@@ -53,10 +53,7 @@ private import TestingInternals
     }
   } catch {
 #if !SWT_NO_FILE_IO
-    FileHandle.stderr.withUnsafeCFILEHandle { stderr in
-      fputs(String(describing: error), stderr)
-      fflush(stderr)
-    }
+    try? FileHandle.stderr.write(String(describing: error))
 #endif
 
     exitCode.withLock { exitCode in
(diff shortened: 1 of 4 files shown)


grynspan:
@swift-ci please test

compnerd (approved):
Thanks, I really like this cleanup!

briancroom (approved):

stmontgomery (approved):

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test Linux

grynspan:
The Linux CI failure appears to be infrastructural. Tests pass at desk on all platforms, so I'm proceeding with the merge.

grynspan:
Follow-up: the Swift stdlib explicitly sets the output code page to UTF-8 (see `_swift_stdlib_configure_console_mode()`), so that explains that.

PR #241, opened 2024-02-28
Don't use `@Tag` in the library target.

grynspan:
We've identified a few build scenarios where the macro target is built _after_ the library target, resulting in the library target failing to build because it uses the `@Tag` macro to declare the predefined colour tags. To resolve, we'll explicitly declare the values for those tags.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #242, opened 2024-02-29
`swift test --list` should write to `stdout` rather than calling `print()`.

grynspan:
I'm seeing some data getting dropped when calling `print()` while handling `swift test --list`, specifically when the list of tests is short. I suspect that `print()` is buffering in a way that's not entirely compatible with the piping that SwiftPM performs.

This issue is specific to Darwin and how it handles standard I/O buffering. I don't have a great explanation as to _why_ though, since `print()` ultimately calls `fwrite(stdout)` which should buffer the same way as `fputs(stdout)`. I do know that Linux and Windows _aren't_ affected. We have the same problem already in `XCTestScaffold` which is why that type _also_ special-cases Darwin when avoiding `print()`.

Repro case is to create a new package with `swift package init --enable-experimental-swift-testing` and then run `swift test list`: you should see `moduleName.example()` in the output, but from what I can see it isn't present. Other packages with more tests don't display the same issue.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/EntryPoint.swift
+++ b/Sources/Testing/Running/EntryPoint.swift
@@ -27,7 +27,11 @@ private import TestingInternals
     let args = CommandLine.arguments()
     if args.count == 2 && args[1] == "--list-tests" {
       for testID in await listTestsForSwiftPM(Test.all) {
+#if SWT_TARGET_OS_APPLE && !SWT_NO_FILE_IO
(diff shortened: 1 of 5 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

stmontgomery (approved):

dennisweissmann (approved):

PR #244, opened 2024-03-01
Allow combining `TestFilter` instances with `OR`-like semantics.

grynspan:
This PR extends the `combining(with:)` and `combine(with:)` functions of `Configuration.TestFilter` to allow specifying whether combinations use `AND`- or `OR`-like semantics.

Because of how test graphs are constructed, these operations aren't trivially equivalent to the Swift standard library `&&` and `||` operators defined for boolean operands, but they have the same overall effect.

This PR introduces a new enumeration in `Configuration.TestFilter` that covers the two operators; I had at one point in development declared public/SPI operator functions so that you could write `f1.combine(with: f2, using: &&)`, but the operator functions had the signature `(Test?, Test?) -> Test?` which while correct for our use cases would probably _not_ be correct in the general context of comparing two instances of `Test?` and could be accidentally misused. An enumeration, while slightly less _Swifty!™_, is unambiguous.

We do not currently make use of `OR`-like semantics anywhere in the testing library, but adding them opens up new possibilities for tools like Swift Package Manager in the future.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Configuration.TestFilter.swift
+++ b/Sources/Testing/Running/Configuration.TestFilter.swift
@@ -54,10 +54,12 @@ extension Configuration {
       /// - Parameters:
       ///   - lhs: The first test filter's kind.
       ///   - rhs: The second test filter's kind.
+      ///   - op: The operator to apply when combining the results of the two
+      ///     filters.
       ///
       /// The result of a test filter with this kind is the combination of the
-      /// results of its subfilters (as if with `&&`.)
-      indirect case combined(_ lhs: Kind, _ rhs: Kind)
+      /// results of its subfilters using `operator`.
+      indirect case combined(_ lhs: Kind, _ rhs: Kind, _ op: CombinationOperator)
     }
 
     /// The kind of test filter.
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci please test

Tests/TestingTests/PlanTests.swift
```diff
@@ -337,6 +337,38 @@ struct PlanTests {
     #expect(planTests.contains(testB))
   }
 
+  @Test("Combining test filters with ||")
```
stmontgomery, line 340:
```suggestion
  @Test("Combining test filters with .or")
```

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -54,10 +54,12 @@ extension Configuration {
       /// - Parameters:
       ///   - lhs: The first test filter's kind.
       ///   - rhs: The second test filter's kind.
+      ///   - op: The operator to apply when combining the results of the two
+      ///     filters.
       ///
       /// The result of a test filter with this kind is the combination of the
-      /// results of its subfilters (as if with `&&`.)
-      indirect case combined(_ lhs: Kind, _ rhs: Kind)
+      /// results of its subfilters using `operator`.
```
stmontgomery, line 61:
```suggestion
      /// results of its subfilters using `op`.
```

stmontgomery (approved):

grynspan:
@swift-ci please test

briancroom:
Very nice, thank you!

briancroom (approved):

PR #246, opened 2024-03-03
Replace unsafe '-define-availability' flags with experimental AvailabilityMacro feature

stmontgomery:
This replaces the `-define-availability` flags in this package's `Package.swift` with the pseudo-experimental `AvailabilityMacro=...` feature.

### Motivation:

We want to use availability macros throughout swift-testing's codebase to simplify things and reduce the possibility of errors or inconsistencies. So far, this has only been possible by passing the `-define-availability` frontend compiler flag, but passing flags this way is considered unsafe and cannot be used in release versions of the package. This means that each time we prepare a new tag for release, we have needed to perform a manual step of replacing each availability macro with its actual versions, and we'd prefer to not need to do this step.

### Modifications:

- Replace all `-define-availability` settings with `AvailabilityMacro=...` experimental feature.
- Remove a now-obsolete step from our release preparation procedure documentation.

### Checklist:

### DIFF
--- a/Documentation/Releases.md
+++ b/Documentation/Releases.md
@@ -43,10 +43,6 @@ git checkout -b release/x.y.z
 The package manifest files (Package.swift _and_ Package@swift-6.0.swift) must
 be updated so that the release can be used as a package dependency:
 
-1. Take note of any availability definitions (passed to the Swift compiler with
-   the `-define-availability` option.) Search for these definitions in the
-   package source and replace them with their values. For example, replace all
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

briancroom (approved):

grynspan (approved):
~~This is an obscure (and probably not permanent) compiler feature, so let's be sure to document what we're doing. A link to the PR that introduced the feature is probably sufficient.~~

Which you already did! Oops.

PR #243, opened 2024-03-01
Capture substructure of arguments and evaluated subexpressions of expectations

stmontgomery:
This introduces logic and SPI for capturing the substructure of arguments and evaluated subexpressions of failed expectations.

### Motivation:

This facilitates more progress towards actionable failure information by allowing more details to be shown in test results.

### Modifications:

- Introduce an SPI `Expression.Value` type which represents the details of a particular value.
- Consolidate some existing properties here, including `Expression.runtimeValueDescription` and `Expression.runtimeTypeInfo`.
- Adopt this as well in `Test.Case.Argument.Snapshot`, consolidating analogous properties like `valueDescription` and `valueTypeInfo`.
- Add new tests and adjust existing ones.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

Resolves rdar://123767150

PR #247, opened 2024-03-04
'Expression.Value' initializer does not assign 'label' property

stmontgomery:
Fix an oversight in the changes made in #243.

### Modifications:

- Fix an oversight in the changes made in #243 where the `label` property of `Expression.Value` was not assigned in its initializer.
- Add some unit tests to catch this.

### Checklist:


stmontgomery:
@swift-ci please test

briancroom (approved):

grynspan (approved):

PR #248, opened 2024-03-04
Add additional warnings for unimplemented platform-specific stuff.

grynspan:
We don't have a _huge_ amount of platform-specific code, but what we do have tends to look like:

```swift
func spinWidget() -> Widget {
#if SWT_TARGET_OS_APPLE // shorthand for os(macOS) || os(iOS) || ...
  return apple_impl()
#elseif os(Linux)
  return linux_impl()
#elseif os(Windows)
  return WindowsImplExW()
#endif
}
```

But this makes it harder to add new platform support because there's no `#else` branch to take for such a platform. Lots of platforms out there are POSIX-like and can often share code with either Darwin or Linux (or both), but we don't want to make that assumption without a human engineer double-checking. So we have a convention of adding to functions like this one an `#else` case of the general form:

```swift
#else
#warning("Platform-specific implementation missing: Widget spinning unavailable")
  return defaultWidget
#endif
```

This `#else` case allows other platforms to build successfully while still producing helpful diagnostics that can be fixed over time.

This PR adds additional warnings where platform-specific code may be unimplemented.

Note: I didn't really touch `FileHandle` because its `#else` paths already assume support for the C standard library. If a new platform does not support C standard file I/O via the functions in [stdio.h](https://en.cppreference.com/w/c/io), whoever is adding support for that platform should define `SWT_NO_FILE_IO` in the package manifest.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/EntryPoint.swift
+++ b/Sources/Testing/Running/EntryPoint.swift
@@ -334,6 +334,9 @@ extension [Event.ConsoleOutputRecorder.Option] {
     // Windows does not set the "TERM" variable, so assume it supports 256-color
     // ANSI escape codes.
     true
+#else
+#warning("Platform-specific implementation missing: terminal colors unavailable")
+    return false
 #endif
   }
 #endif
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #261, opened 2024-03-06
Add support for true color in Terminal output.

grynspan:
Ubuntu and Windows both support true color output in their respective Terminal applications, so we can support it as well. (It's admittedly a subtle change from 256-color support.)

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
@@ -156,10 +156,7 @@ extension Event.Symbol {
     var symbolCharacter = String(unicodeCharacter)
 #if os(macOS) || (os(iOS) && targetEnvironment(macCatalyst))
     if options.contains(.useSFSymbols) {
-      symbolCharacter = String(sfSymbolCharacter)
-      if useColorANSIEscapeCodes {
-        symbolCharacter += " "
-      }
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
```diff
@@ -148,17 +148,19 @@ extension Event.Symbol {
   /// - Returns: A string representation of `self` appropriate for writing to
   ///   a stream.
   fileprivate func stringValue(options: Set<Event.ConsoleOutputRecorder.Option>) -> String {
+    var useColorANSIEscapeCodes = false
```
SeanROlszewski, line 151: If you wanted to you could use an if expression here and avoid making this mutable. 

```swift
let useColorANSIEscapeCodes = if let colorBitDepth = options.colorBitDepth {
  colorBitDepth >= 4
} else {
  false
}

SeanROlszewski (approved):

Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
grynspan: Sure, but this is one line shorter. :D

PR #264, opened 2024-03-07
Introduce a snapshot type for Error values

stmontgomery:
This introduces a new serializable "snapshot" type for representing `Error` values.

### Motivation:

The `Issue.Kind.errorCaught(_ error: any Error)` enum case stores an `Error` value, but when taking a snapshot of an `Issue`, we currently only capture a textual description of the error because `Error` is not `Codable`. Sometimes, though, is it is useful to have a real `Error` value on the receiving (i.e. deserializing) end of the snapshot, even if it is not identical to the original error, for the purpose of propagating that error or representing it via mechanisms which expect an `Error` value.

### Modifications:

- Add `ErrorSnapshot` as SPI.
- Replace the `String` associated value describing an error in `Issue.Kind.Snapshot.errorCaught(_:)` with `ErrorSnapshot`.
- Add a derived `error` property to `Issue.Snapshot`, returning the `ErrorSnapshot` (if any).

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/ErrorSnapshot.swift
+++ b/Sources/Testing/Issues/ErrorSnapshot.swift
@@ -1,0 +1,42 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// A serializable snapshot of an `Error` value.
+///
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

SeanROlszewski (approved):
Lgtm Stuart, thanks!

grynspan:
Looks like a good start. I think we can improve this further in subsequent PRs:

1. On Darwin, check if the error is an instance of `NSCFError` and encode its domain/code/userInfo (or viable subset of userInfo) and then on decoding, reconstitute back to the same `NSCFError`. This would improve fidelity on Darwin.
2. On all platforms, see if we can sniff out `Codable` conformance and use it to provide more output than just the error description (which for a typical Swift error might not be very descriptive at all.)

PR #279, opened 2024-03-11
Adopt `lexicalContext` from swift-syntax-6.0.0.

grynspan:
This PR conditionally adopts the new `lexicalContext` member of `MacroExpansionContext` added in https://github.com/apple/swift-syntax/pull/1554. If the SwiftSyntax600 pseudo-module is available, then this member should be available for use and can be used to perform additional diagnostics for tests and to get the names of their containing types.

With this PR, if built against an older toolchain (5.11 or earlier), the old hacky "is there leading whitespace?" mechanism is still used.

A future PR will recursively perform suite-level diagnostics on the lexical contexts containing tests and suites, so that a test cannot be (easily) inserted into a type that cannot be used as a suite.

Resolves rdar://109439578.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/SuiteDeclarationMacro.swift
+++ b/Sources/TestingMacros/SuiteDeclarationMacro.swift
@@ -21,7 +21,9 @@ public struct SuiteDeclarationMacro: MemberMacro, PeerMacro, Sendable {
     providingMembersOf declaration: some DeclGroupSyntax,
     in context: some MacroExpansionContext
   ) throws -> [DeclSyntax] {
-    _diagnoseIssues(with: declaration, suiteAttribute: node, in: context)
+    guard _diagnoseIssues(with: declaration, suiteAttribute: node, in: context) else {
+      return []
+    }
     return _createTestContainerDecls(for: declaration, suiteAttribute: node, in: context)
   }
 
@@ -33,7 +35,7 @@ public struct SuiteDeclarationMacro: MemberMacro, PeerMacro, Sendable {
     // The peer macro expansion of this macro is only used to diagnose misuses
     // on symbols that are not decl groups.
     if declaration.asProtocol((any DeclGroupSyntax).self) == nil {
-      _diagnoseIssues(with: declaration, suiteAttribute: node, in: context)
+      _ = _diagnoseIssues(with: declaration, suiteAttribute: node, in: context)
     }
     return []
   }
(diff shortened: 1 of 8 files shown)


grynspan:
@swift-ci please test

Package.swift
```diff
@@ -33,7 +33,7 @@ let package = Package(
   ],
 
   dependencies: [
-    .package(url: "https://github.com/apple/swift-syntax.git", from: "509.0.0"),
```
grynspan, line 36: This is temporary for the sake of CI; I'll switch it back after it passes.

Package.swift
```diff
@@ -33,7 +33,7 @@ let package = Package(
...
+    .package(url: "https://github.com/apple/swift-syntax.git", branch: "main"),
```
grynspan, line 36: This is temporary for the sake of CI; I'll switch it back after it passes.

stmontgomery: For CI, does this change also need to be applied (temporarily) to `Package@swift-6.0.swift`?

grynspan: 🤔 possibly. Will do that too.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/TestingMacros/TestDeclarationMacro.swift
```diff
@@ -144,6 +166,8 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
         }
       }
     }
+
+    return !diagnostics.map(\.severity).contains(.error)
```
dennisweissmann, line 170: also lazy?

Sources/TestingMacros/SuiteDeclarationMacro.swift
```diff
@@ -115,6 +125,8 @@ public struct SuiteDeclarationMacro: MemberMacro, PeerMacro, Sendable {
         diagnostics.append(.availabilityAttributeNotSupported(noasyncAttribute, on: declaration, whenUsing: suiteAttribute))
       }
     }
+
+    return !diagnostics.map(\.severity).contains(.error)
```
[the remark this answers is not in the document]
grynspan: We tend to avoid `lazy` in the macros target for clarity reasons but yeah, in this case since we're just looking for any matching element, it makes sense here.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
Holding on to this PR for the moment as I'd like @DougGregor's review as well.

Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
```diff
@@ -87,3 +87,24 @@ func diagnoseIssuesWithTags(in traitExprs: [ExprSyntax], addedTo attribute: Attr
...
+/// - Returns: An array of zero or more diagnostic messages related to the
+///   lexical context containing `decl`.
+func diagnoseIssuesWithLexicalContext(
+  containing decl: some DeclSyntaxProtocol,
+  attribute: AttributeSyntax,
+  in context: some MacroExpansionContext
+) -> [DiagnosticMessage] {
+  context.lexicalContext
+    .filter { !$0.isProtocol((any DeclGroupSyntax).self) }
+    .map { .containingNodeUnsupported($0, whenUsing: attribute) }
```
DougGregor, line 108: So this will allow `@Test` at module scope and within type context, but not when nested in (e.g.) closures or other functions. That seems like a reasonable set of restrictions

DougGregor (approved):

Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
grynspan: Precisely!

grynspan:
@swift-ci please test

PR #283, opened 2024-03-12
Make event recorder types non-copyable.

grynspan:
This PR makes the event recorder types in swift-testing non-copyable. They can still be captured by an event handler without issue, but cannot be moved around. An event recorder carries with it a large amount of data (the statistics around recorded events that are used to inform later events) and copying them, even inadvertently, can be very expensive.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #286, opened 2024-03-12
Add experimental support for event streaming using JSON.

grynspan:
This PR adds a new command-line argument to the SwiftPM entry point that allows callers to stream snapshots of events to a file or named pipe. Note that this argument has not (yet) been added to `swift test`, so to use it you must first call `swift build --build-tests`, then invoke the executable build product directly.

This functionality would be useful for clients like VSCode that want to gather detailed, structured output as it is generated. The existing XML-based output is of course structured, but is only available and parseable after the test process finishes, and may fail to decode if the test process crashes.

~~The output format consists of JSON blobs representing each event along with its associated test and test case (where applicable.) The events are encoded to JSON, then the size of the JSON is transmitted as a 64-bit little-endian integer followed by the JSON itself. The lengths are binary and UTF-8 is assumed for the JSON. (We considered using [JSONL](https://jsonlines.org/), but it's significantly harder to decode when using e.g. standard C file streams, which makes it harder for us to test.)~~ We're usin' JSONL.

Any writeable file path can be passed. A regular file can be used, however most platforms will not allow reading from a regular file while it is open for writing. Instead, on platforms where it is supported, a named pipe can be used. On Darwin and Linux, you can use [`mkfifo()`](https://pubs.opengroup.org/onlinepubs/9699919799/functions/mkfifo.html). On Windows, you can use [`CreateNamedPipeA()`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createnamedpipea). Bash and Windows PowerShell have equivalents that can be used instead of C functions if needed.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/EntryPoint.swift
+++ b/Sources/Testing/Running/EntryPoint.swift
@@ -9,6 +9,9 @@
 //
 
 private import TestingInternals
+#if canImport(Foundation)
+private import Foundation
+#endif
 
 /// The entry point to the testing library used by Swift Package Manager.
 ///
@@ -165,6 +168,19 @@ func configurationForSwiftPMEntryPoint(withArguments args: [String]) throws -> C
       oldEventHandler(event, context)
     }
   }
+
+#if canImport(Foundation)
+  // Event stream output (experimental)
+  if let eventOutputIndex = args.firstIndex(of: "--experimental-event-stream-output"), eventOutputIndex < args.endIndex {
+    let eventStreamOutputPath = args[args.index(after: eventOutputIndex)]
+    let eventHandler = try _eventHandlerForStreamingEvents(toFileAtPath: eventStreamOutputPath)
+    let oldEventHandler = configuration.eventHandler
+    configuration.eventHandler = { event, context in
+      eventHandler(event, context)
+      oldEventHandler(event, context)
+    }
+  }
+#endif
 #endif
 
   // Filtering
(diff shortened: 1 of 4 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

adam-fowler:
Oh and wow there's masses of detail in there, should cover most situations 😄

grynspan:
If `mkfifo` is sufficient, I'd use it instead of a socket simply because it's less niche-case code in the testing library. We have affordances for file I/O already, but not for socket I/O. (Also I haven't touched a WinSock API in about 20 years and that sounds like the sort of platform-specific nightmare code I'd rather avoid here if possible.)

Adam pinged me outside GitHub and let me know this PR should work for him either way, so we can move forward with it as-is and, if sockets become critical in the future, revisit at that point.

grynspan:
@swift-ci please test

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -244,6 +260,75 @@ func runTests(options: [Event.ConsoleOutputRecorder.Option], configuration: Conf
...
+  let file = try FileHandle(forWritingAtPath: path)
+
+  return { event, context in
+    let snapshot = EventAndContextSnapshot(
+      event: Event.Snapshot(snapshotting: event),
+      test: context.test.map { Test.Snapshot(snapshotting: $0) },
+      testCase: context.testCase.map { Test.Case.Snapshot(snapshotting: $0) }
+    )
+    if let snapshotJSON = try? JSONEncoder().encode(snapshot) {
+      try? file.withLock {
```
stmontgomery, line 319: For these `try?` usages could we consider some fallback logic which encodes an explanation of the error, or something to help convey to a supporting tool that an error occurred and data is missing in the stream?

If not for this PR, then maybe a "FIXME" comment to note it for later improvement.

grynspan: For the JSON failure, no, because the `try?` can't really fail for these types, so it'd just be dead code. And for the file calls, if they throw an error, that means there's nowhere to write the diagnostic data.

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -244,6 +260,75 @@ func runTests(options: [Event.ConsoleOutputRecorder.Option], configuration: Conf
...
+/// decode the JSON produced for this type in implementation-specific ways.
+struct EventAndContextSnapshot {
+  /// A snapshot of the event.
+  var event: Event.Snapshot
+
+  /// A snapshot of the test associated with the event, if any.
+  var test: Test.Snapshot?
+
+  /// A snapshot of the test case associated with the event, if any.
+  var testCase: Test.Case.Snapshot?
```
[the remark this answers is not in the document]
grynspan: This is, as you've noted, an experimental interface that I'm putting together to help bootstrap VSCode plugin integration (also experimental.) What you're describing hasn't been requested by anybody at this point so it'd be a hard sell.

If it turns out we need something like that to make the VSCode plugin work, we can revisit/revise.

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -166,6 +169,19 @@ func configurationForSwiftPMEntryPoint(withArguments args: [String]) throws -> C
       oldEventHandler(event, context)
     }
   }
+
+#if canImport(Foundation)
+  // Event stream output (experimental)
+  if let eventOutputIndex = args.firstIndex(of: "--experimental-event-stream-output"), eventOutputIndex < args.endIndex {
```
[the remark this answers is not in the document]
grynspan: See `--xunit-output` which already exists as SwiftPM API.

If/when this flag ceases to be experimental, we'd be renaming it anyway and are free to come up with a better name following community feedback.

Sources/Testing/Running/EntryPoint.swift
stmontgomery: Why is it not possible for the JSON encoding to potentially fail for these types?

grynspan: Because the types we're encoding have no failure modes.

stmontgomery (approved):
I think this is a really helpful step forward! My large comment yesterday was just meant to give some thoughts on future directions and optimization opportunities, but this looks good.

I do think that we should keep this "experimental" until such time as we get more confidence in the structure and contents of the `Snapshot` types, since elevating this to non-experimental will imply some stability in those values.

Sources/Testing/Running/EntryPoint.swift
stmontgomery: Certainly would not consider this specific thing a blocker, but I do think we ought to eventually incorporate some way for this stream to convey errors which occurred during a test run. This is conceptually different than conveying `Issue`s recorded _by_ tests—what I'm describing is more akin to an infrastructure failure. 

That's the sort of scenario I had in mind here: what if our `Snapshot` types fail to encode due to a problem with their contents, such as un-serializable data values supplied at runtime by a test? Or more generally: if something about running tests goes wrong, how can we represent that in this stream?

briancroom: It's worth noting that the `Runner.run()` function is not `throw`ing, and we do have `Issue.Kind.system`, which is intended to represent that sort of problem when recorded with no associated test or case. We may still conclude that it would be useful to have some additional category of error reporting possible, but I'm not sure we have enough data on hand right now to design it properly.

briancroom: Perhaps we ought to just crash upon an encoding failure, though? It’d be a very nasty bug to investigate if we ended up silently dropping events in the stream due to some future regression in the encoding implementation.

briancroom (approved):

Sources/Testing/Running/EntryPoint.swift
grynspan: Crashing the test process is a catastrophic failure and we only ever reach for `fatalError()` and friends when there's a bug in our code that prevents forward progress. It seems heavy-handed to crash the process because a pipe got closed early or something.

briancroom: Oh, I was suggesting that only for a failure from JSONEncoder, not for a file I/O problem.

briancroom: Consider also: it seems to me that a dropped event represents data loss in this context, and silent data loss seems worse than noisy data loss?

grynspan: If data is being lost silently, then our only options are to remain silent or to cause _more_ data loss with early termination.

grynspan:
@swift-ci please test

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -243,6 +259,99 @@ func runTests(options: [Event.ConsoleOutputRecorder.Option], configuration: Conf
...
+      test: context.test.map { Test.Snapshot(snapshotting: $0) },
+      testCase: context.testCase.map { Test.Case.Snapshot(snapshotting: $0) }
+    )
+    if var snapshotJSON = try? JSONEncoder().encode(snapshot) {
+      func isASCIINewline(_ byte: UInt8) -> Bool {
+        byte == 10 || byte == 13
+      }
+
+#if DEBUG
+      // We don't actually JSONEncoder() to produce output containing newline
```
briancroom, line 325: Typo / missing word? "expect"?

grynspan: I swear I typed it.

grynspan:
@swift-ci please test

PR #288, opened 2024-03-13
Drop remaining 5.9 support.

grynspan:
This PR updates the package manifest to require Swift 5.10 or newer. We previously had a tools version of 5.9 so that content from the package would be indexed by Swift Package Index, but they have updated their systems to use 5.10 so this is no longer required.

As a reminder, Swift 5.10 support is temporary for the benefit of developers who cannot rely on a nightly main-branch toolchain. **Swift 5.10 support will be removed in a future update.**

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/DefiningTests.md
+++ b/Sources/Testing/Testing.docc/DefiningTests.md
@@ -87,7 +87,7 @@ a test is unable to run due to limited availability:
 
 ```swift
 @available(macOS 11.0, *)
-@available(swift, introduced: 5.9, message: "Requires Swift 5.9 features to run")
+@available(swift, introduced: 8.0, message: "Requires Swift 8.0 features to run")
 @Test func foodTruckExists() { ... }
 ```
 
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

dennisweissmann (approved):

Sources/Testing/Testing.docc/DefiningTests.md
```diff
@@ -87,7 +87,7 @@ a test is unable to run due to limited availability:
 
 ```swift
 @available(macOS 11.0, *)
-@available(swift, introduced: 5.9, message: "Requires Swift 5.9 features to run")
+@available(swift, introduced: 8.0, message: "Requires Swift 8.0 features to run")
```
Kyle-Ye, line 90: Is "8.0" intentional? My first thought is "6.0".

Also why bother to updating the example here? The original "5.9" seems fine to me.

grynspan: 8.0 is intentional as an arbitrary Swift version that doesn't exist. 5.9 doesn't make sense because all swift-testing tests require at least 5.10 to run, so availability limited to 5.9 doesn't do anything.

PR #290, opened 2024-03-13
Add a `comment` property to the `.bug()` trait.

grynspan:
This PR adds an optional `comment` property to the `.bug()` trait so that developers can associate additional notes that may not be obvious from just a bug's identifier. Since the testing library can't go off crawling arbitrary bug trackers looking for metadata, it makes sense to include basic metadata inline. For example, since this PR is resolving rdar://124101140 whose description in Apple's Radar system is "Add an optional Comment parameter to `.bug` trait", an associated test could have this trait added to it:

```swift
.bug("rdar://124101140", "Add an optional Comment parameter to `.bug` trait")
```

This PR also adds `Codable` conformance to `Bug` so that it can be encoded and decoded when needed.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/AssociatingBugs.md
+++ b/Sources/Testing/Testing.docc/AssociatingBugs.md
@@ -25,8 +25,8 @@ specific bugs with tests that reproduce them or verify they are fixed.
 
 ## Associating a bug with a test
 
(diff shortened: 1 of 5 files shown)


grynspan:
@swift-ci please test

dennisweissmann (approved):

grynspan:
@swift-ci please test

briancroom (approved):

PR #293, opened 2024-03-15
Only load tag colors from disk if color output is enabled.

grynspan:
If color output is disabled (`NO_COLOR`, no ANSI escape code support, or writing to a regular file), don't bother reading tag colors from disk.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #294, opened 2024-03-15
Include test names in verbose output.

grynspan:
This PR inserts test names (as opposed to test _display_ names) into verbose output. Without this change, a test with a display name is identified solely by its display name. With this change, when the developer passes `--verbose` to `swift test`, the test function or suite type name is included after the display name. For tests/suites without display names, there is no change in behavior.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
@@ -210,7 +210,7 @@ extension Event.HumanReadableOutputRecorder {
     let testName = if let test {
       if let displayName = test.displayName {
         if verbose {
-          "\"\(displayName)\" → \(test.name)"
+          "\"\(displayName)\" (aka '\(test.name)')"
         } else {
           "\"\(displayName)\""
         }
--- a/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
@@ -207,13 +207,18 @@ extension Event.HumanReadableOutputRecorder {
     verbosely verbose: Bool = false
   ) -> [Message] {
     let test = eventContext.test
-    var testName: String
-    if let displayName = test?.displayName {
-      testName = "\"\(displayName)\""
-    } else if let test {
-      testName = test.name
+    let testName = if let test {
+      if let displayName = test.displayName {
+        if verbose {
+          "\"\(displayName)\" → \(test.name)"
+        } else {
+          "\"\(displayName)\""
+        }
+      } else {
+        test.name
+      }
     } else {
-      testName = "«unknown»"
+      "«unknown»"
     }
     let instant = event.instant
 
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -207,13 +207,18 @@ extension Event.HumanReadableOutputRecorder {
...
     let test = eventContext.test
-    var testName: String
-    if let displayName = test?.displayName {
-      testName = "\"\(displayName)\""
-    } else if let test {
-      testName = test.name
+    let testName = if let test {
+      if let displayName = test.displayName {
+        if verbose {
+          "\"\(displayName)\" → \(test.name)"
```
stmontgomery, line 213: We have logic elsewhere in this file for including type names of arguments when in verbose mode, and for those we surround the type name in parentheses (e.g. `<arg> (<type name>)`). Using parentheses for this might be better. We primarily use the arrow character to indicate an evaluated value, but this seems different.

stmontgomery (approved):
Awesome improvement, just had one formatting suggestion

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: Either is obviously implementable, but for test _functions_ parentheses look weird because they have argument lists:

```
Fun test times with "My great test" (myGreatTest())
```

And I wanted whatever formatting I picked to be consistent between suites and tests. What do you think?

stmontgomery: Ok how about for test names, we format them verbosely as:
```suggestion
          "\"\(displayName)\" (aka \(test.name))"
```
And for better consistency with that, we can change how type names of arguments are formatted in verbose mode, to be:
```diff
-"\(labeledArgument) (\(typeInfo.qualifiedName))"
+"\(labeledArgument) (of type \(typeInfo.qualifiedName))"
```

"(aka ...)" has a precedent in certain Swift compiler diagnostics I believe, for providing the base name of a typealias'ed type.

grynspan: Can you find prior art for "aka" so we can refer to it here, in case somebody asks about it later?

For _argument types_, a colon would be consistent with how Swift expresses them, so e.g.

```
\(labeledArgument) :\(typeInfo.qualifiedName)
```

stmontgomery: > Can you find prior art for "aka" so we can refer to it here, in case somebody asks about it later?

Closest thing I found is [this](https://github.com/apple/swift/blob/main/include/swift/AST/DiagnosticEngine.h#L445), and there are a bunch of examples in compiler unit tests. Looks like it actually surrounds the "aka" type in single quotes, so

```swift
"\"\(displayName)\" (aka '\(test.name)')"
```

> For _argument types_, a colon would be consistent with how Swift expresses them, so e.g.
> 
> ```
> \(labeledArgument) :\(typeInfo.qualifiedName)
> ```

Sure, sounds good

dennisweissmann (approved):

grynspan:
@swift-ci please test

PR #295, opened 2024-03-16
Begin including SPI documentation in local builds, and fix all newly-revealed DocC issues

stmontgomery:
Begin including documentation for `@_spi` declarations in local development (non-distribution) builds, and fix all newly-revealed DocC issues in those SPI declarations.

I also took this opportunity to simplify the mechanism for conditionalizing certain package settings for distribution builds, and fix a few other documentation omissions and inconsistencies I noticed.

### Checklist:

Resolves rdar://118416838

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -65,15 +65,7 @@ let package = Package(
         .product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
         .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
       ],
-      swiftSettings: .packageSettings + [
-        // The only target which needs the ability to import this macro
-        // implementation target's module is its unit test target. Users of the
-        // macros this target implements use them via their declarations in the
-        // Testing module. This target's module is never distributed to users,
-        // but as an additional guard against accidental misuse, this specifies
-        // the unit test target as the only allowable client.
-        .unsafeFlags(["-Xfrontend", "-allowable-client", "-Xfrontend", "TestingMacrosTests"]),
-      ]
+      swiftSettings: .packageSettings
     ),
 
     // "Support" targets: These contain C family code and are used exclusively
@@ -110,12 +102,20 @@ package.targets.append(contentsOf: [
 ])
 #endif
 
+/// Whether this package is building for distribution.
+///
+/// This can be used to conditionalize certain settings which are not intended
+/// for use when building the package for distribution to end users.
+/// Non-distribution (i.e. development-only) builds may enable things
+/// like stricter compiler features which are unsupported or inappropriate for
+/// client builds.
+let isBuildingForDistribution = false
+
 extension Array where Element == PackageDescription.SwiftSetting {
   /// Settings intended to be applied to every Swift target in this package.
   /// Analogous to project-level build settings in an Xcode project.
   static var packageSettings: Self {
-    availabilityMacroSettings + [
-      .unsafeFlags(["-require-explicit-sendable"]),
+    var settings = availabilityMacroSettings + [
       .enableExperimentalFeature("StrictConcurrency"),
       .enableUpcomingFeature("ExistentialAny"),
 
@@ -124,6 +124,12 @@ extension Array where Element == PackageDescription.SwiftSetting {
 
       .define("SWT_TARGET_OS_APPLE", .when(platforms: [.macOS, .iOS, .macCatalyst, .watchOS, .tvOS, .visionOS])),
     ]
+
+    if !isBuildingForDistribution {
+      settings.append(contentsOf: developmentOnlySettings)
+    }
+
+    return settings
   }
 
   /// Settings which define commonly-used OS availability macros.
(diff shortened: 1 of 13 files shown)


stmontgomery:
@swift-ci please test

Package.swift
```diff
@@ -65,15 +65,7 @@ let package = Package(
...
         .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
       ],
-      swiftSettings: .packageSettings + [
-        // The only target which needs the ability to import this macro
-        // implementation target's module is its unit test target. Users of the
-        // macros this target implements use them via their declarations in the
-        // Testing module. This target's module is never distributed to users,
-        // but as an additional guard against accidental misuse, this specifies
-        // the unit test target as the only allowable client.
-        .unsafeFlags(["-Xfrontend", "-allowable-client", "-Xfrontend", "TestingMacrosTests"]),
```
stmontgomery, line 75: Even though I still like the idea of this setting, in practice it is not giving us any benefit or protection since currently it must be commented-out for distribution builds (which are the usage scenario it was meant for). Having it here causes us an additional manual step each time we cut a release, and there's little practical value in conditionalizing it to only non-distribution builds IMO.

dennisweissmann (approved):

Sources/Testing/Issues/Issue.swift
```diff
@@ -49,12 +49,10 @@ public struct Issue: Sendable {
...
     /// - Parameters:
     ///   - timeLimitComponents: The time limit reached by the test.
-    ///
-    /// @Comment {
-    ///   - Bug: The associated value of this enumeration case should be an
-    ///     instance of `Duration`, but the testing library's deployment target
-    ///     predates the introduction of that type.
-    /// }
+    //
+    // - Bug: The associated value of this enumeration case should be an
```
stmontgomery, line 53: I confirmed that the rendered DocC still shows the documentation block comments above, despite having 2-slash `//` comments in between.

grynspan:
Please hold on to this PR.

Package.swift
```diff
@@ -110,12 +102,20 @@ package.targets.append(contentsOf: [
...
 #endif
 
+/// Whether this package is building for distribution.
+///
+/// This can be used to conditionalize certain settings which are not intended
+/// for use when building the package for distribution to end users.
+/// Non-distribution (i.e. development-only) builds may enable things
+/// like stricter compiler features which are unsupported or inappropriate for
+/// client builds.
+let isBuildingForDistribution = false
```
grynspan, line 112: What if instead we checked for an environment variable and built SPI when it's set?

```swift
let buildSPIDocumentation: Bool = {
  if let boolValue = Context.environment["SWT_SPI_DOCUMENTATION_ENABLED"].flatMap(Bool.init) {
    return boolValue
  } else if let intValue = Context.environment["SWT_SPI_DOCUMENTATION_ENABLED"].flatMap(UInt64.init) {
    return intValue != 0
  }
  return false // disabled by default
}()
```

(That's a slightly simplified version of what our own `Environment` type does.)

grynspan (changes requested):
I think we can avoid having to manually toggle `isBuildingForDistribution`. See my comment above about an environment variable.

Package.swift
stmontgomery: We could do this, although a workflow I commonly use for viewing docs is the "Product > Build Documentation" command in recent versions of Xcode, and I was hoping for that to get this flag too. I don't know of a way to apply an env var in that scenario (when using the GUI Xcode app, not its xcodebuild CLI).

stmontgomery: The constant I added here would still be needed, or at least it would still be helpful, even if we conditionalized the SPI docs flag based on an env var, because the constant is what (now) decides whether all the unsafeFlags get included

grynspan: Honestly we can also just drop the unsafe flags outright.

grynspan: > although a workflow I commonly use for viewing docs is the "Product > Build Documentation" command in recent versions of Xcode

We can ask the DocC folks if they can help with this.

Kyle-Ye: > We could do this, although a workflow I commonly use for viewing docs is the "Product > Build Documentation" command in recent versions of Xcode, and I was hoping for that to get this flag too. I don't know of a way to apply an env var in that scenario (when using the GUI Xcode app, not its xcodebuild CLI).

+1 for some similar scenario I experienced. And I believe it is a Xcode/SwiftPM issue instead of a DocC issue.

Xcode only gives us the ability to set env value for executable target when they are running. I could not find a easy solution to setup the env value I use in my Package.swift file.

Currently I just use a small macOS [app](https://github.com/Kyle-Ye/MenuHelper) I wrote to trigger `export XX=YY && open ./ -a Xocde` for me in Finder.

Otherwise we'll have to quit Xcode and manually export the env value in shell and then open Xcode in the shell to make Package.swift access the env we set.

image

As a person who maintains some Swift packages and likes to add environment variable condition config in package, it is really troublesome at present if we do not use any other 3rd party tools.

grynspan:
What state is this PR in? Do we want to salvage it?

PR #299, opened 2024-03-18
Remove unused local error type from `Runner._withErrorHandling(...)`

stmontgomery:
This removes a bit of dead code from `Runner._withErrorHandling(...)`: the local `AlreadyHandled` error type and a `catch` clause for that type. This mechanism was formerly used in the implementation but is no longer. I also changed this function to non-throwing, for related reasons.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -65,7 +65,7 @@ extension Runner {
   ///
   /// This function encapsulates the standard error handling performed by
   /// ``Runner`` when running a test or test case.
-  private func _withErrorHandling(sourceLocation: SourceLocation, _ body: () async throws -> Void) async throws -> Void {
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

dennisweissmann (approved):

stmontgomery:
> We're clearly not using it now, but when did we _stop_? Is that a bug?

Earlier, we had logic which allowed a Runner to be configured to terminate the current Task upon the first recorded Issue, and that mechanism relied on this error type. We later removed that mechanism altogether but this vestige still remains. I think it is safe to remove at this point and not a bug.

grynspan:
Oh right. I forgot about that. Okay, yeah, all good now.

briancroom (approved):

PR #284, opened 2024-03-12
Make `Event.Context` non-copyable.

grynspan:
`Event.Context` contains a significant amount of data (a copy of the current test, current test case, and all parameterized inputs to the test.) It should not be possible to accidentally make a copy of this data.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #306, opened 2024-03-20
Revert "Make `Event.Context` non-copyable. (#284)"

grynspan:
This reverts commit f89573d37c325f2afd5a72090c08dd4a128c8c4f.

This revert is necessary because we realized after making this change that it will be common to send events off to some off-task, serial executor such as a dispatch queue. Doing so necessitates a copy, since that queue will need to execute independently of whatever thread created the event being handled. 😢

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #309, opened 2024-03-22
Do not assume `UnsafeRawBufferPointer` provides contiguous storage.

grynspan:
`UnsafeRawBufferPointer` only gained an implementation of `withContiguousStorageIfAvailable(_:)` with [SE-0333](https://github.com/apple/swift-evolution/blob/main/proposals/0333-with-memory-rebound.md) and https://github.com/apple/swift/pull/41288. Because swift-testing can be used on older versions of Darwin that don't have this implementation, one of our unit tests now crashes there when calling
`FileHandle.write(_: some Sequence<UInt8>)`. This PR resolves the crash by providing a concretely-typed overload that takes an instance of `UnsafeRawBufferPointer`.

This code path is technically dead right now (it survives from an intermediate implementation of #286) but because "writing bytes to a file" remains a useful tool we'll probably need in the future, I'm not ripping it out just yet.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/FileHandle.swift
+++ b/Sources/Testing/Support/FileHandle.swift
@@ -243,6 +243,23 @@ extension FileHandle {
     }
   }
 
+  /// Write a sequence of bytes to this file handle.
+  ///
+  /// - Parameters:
+  ///   - bytes: The bytes to write. This untyped buffer is interpreted as a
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci please test

briancroom (approved):

PR #310, opened 2024-03-22
Adopt `TypeInfo` in `Test`.

grynspan:
This PR adopts `TypeInfo` in `Test` so that we're not carrying around a direct type reference. This will allow us to synthesize instances of `Test` at runtime that describe types for which we don't have the metatype pointer. It may also enable future Embedded Swift work (where `Any.Type` is unavailable.)

This change updates `TypeInfo` to track either a metatype (where available) _or_ a qualified/unqualified name pair. If you start with a metatype, you keep it around until you need to encode the instance, at which point we encode strings, and subsequent decoding gives you back a `TypeInfo` that's string-backed. This should improve memory usage and storage requirements because otherwise every instance of `Test` in memory will need to keep strings around and they are unlikely to be interned/uniqued in the process.

### Checklist:

### DIFF
--- a/Sources/Testing/Parameterization/TypeInfo.swift
+++ b/Sources/Testing/Parameterization/TypeInfo.swift
@@ -12,20 +12,108 @@
 /// parameter of a test function.
 @_spi(ForToolsIntegrationOnly)
 public struct TypeInfo: Sendable {
+  /// An enumeration defining backing storage for an instance of ``TypeInfo``.
+  private enum _Kind: Sendable {
+    /// The type info represents a concrete metatype.
+    ///
+    /// - Parameters:
+    ///   - type: The concrete metatype.
+    case type(_ type: Any.Type)
+
+    /// The type info represents a metatype, but a reference to that metatype is
+    /// not available at runtime.
+    ///
+    /// - Parameters:
+    ///   - fullyQualified: The fully-qualified name of the type.
+    ///   - unqualified: The unqualified name of the type.
+    case nameOnly(fullyQualified: String, unqualified: String)
+  }
+
+  /// The kind of type info.
+  private var _kind: _Kind
+
   /// The complete name of this type, with the names of all referenced types
   /// fully-qualified by their module names when possible.
-  public var qualifiedName: String
+  ///
+  /// The value of this property is equal to ``fullyQualifiedName``, but is
+  /// split into components. For instance, given the following declaration in
+  /// the `Example` module:
+  ///
+  /// ```swift
+  /// struct A {
+  ///   struct B {}
+  /// }
+  /// ```
+  ///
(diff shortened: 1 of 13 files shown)


grynspan:
@swift-ci please test

Sources/Testing/Test+Discovery.swift
```diff
@@ -28,7 +28,7 @@ extension Test {
   /// All available ``Test`` instances in the process, according to the runtime.
   ///
   /// The order of values in this sequence is unspecified.
-  static var all: some Sequence<Test> {
+  static var all: some Collection<Test> {
```
briancroom, line 31: What motivates this change?

Sources/Testing/Parameterization/TypeInfo.swift
```diff
@@ -55,14 +142,55 @@ extension TypeInfo: CustomStringConvertible, CustomDebugStringConvertible {
...
+    }
+  }
+}
 
 // MARK: - Codable
 
-extension TypeInfo: Codable {}
+extension TypeInfo: Codable {
+  /// A simplified version of ``TypeInfo`` suitable for encoding and decoding.
+  fileprivate struct EncodedForm {
```
briancroom, line 175: Interesting pattern!

dennisweissmann (approved):

Sources/Testing/Parameterization/TypeInfo.swift
```diff
@@ -12,38 +12,125 @@
...
+  /// If this instance was created from a type name, or if it was previously
+  /// encoded and decoded, the value of this property is `nil`.
+  public var type: Any.Type? {
+    if case let .type(type) = _kind {
+      return type
+    }
+    return nil
+  }
+
+  init(fullyQualifiedName: String, unqualifiedName: String) {
```
briancroom, line 115: As of this PR, this initializer is only being used in the `Decodable` implementation, right?

briancroom (approved):

Sources/Testing/Parameterization/TypeInfo.swift
grynspan: It's also used in our tests. :)

Sources/Testing/Test+Discovery.swift
grynspan: Sorry, that was from a partial implementation and I didn't mean to include it. Will remove.

Sources/Testing/Parameterization/TypeInfo.swift
grynspan: Useful for when you don't want to deal with encoding/decoding containers directly.

grynspan:
@swift-ci please test

PR #311, opened 2024-03-22
Synthesize `Test` instances for all suite types.

grynspan:
This PR extends test discovery to synthesize `Test` instances for all suite types (that is, types containing other suites or test functions but which do not have the `@Suite` attribute explicitly applied to them.)

This ensures that deep type hierarchies like the following are fully represented in test plans:

```swift
struct AllTests {
  struct FoodTruckTests {
    @Suite struct MechanicalTests {
      struct BatteryTests {
      	@Test func isCharged() { ... }
      }
    }
  }
}
```

Resolves rdar://125241067.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Discovery.swift
+++ b/Sources/Testing/Test+Discovery.swift
@@ -103,11 +103,15 @@ extension Test {
       }
       let suiteID = ID(typeInfo: suiteTypeInfo)
       if tests[suiteID] == nil {
-        // If the real test is hidden, so shall the synthesized test be hidden.
-        // Copy the exact traits from the real test in case they someday carry
-        // any interesting metadata.
-        let traits = test.traits.compactMap { $0 as? HiddenTrait }
-        tests[suiteID] = Test(traits: traits, sourceLocation: test.sourceLocation, containingTypeInfo: suiteTypeInfo)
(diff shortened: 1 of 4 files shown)


grynspan:
I've XFAILed a test for now as we're not 100% sure what to do with it. We'll take a look after the weekend.

grynspan:
@swift-ci please test

Sources/Testing/Test+Discovery.swift
```diff
@@ -103,11 +103,15 @@ extension Test {
       }
       let suiteID = ID(typeInfo: suiteTypeInfo)
       if tests[suiteID] == nil {
-        // If the real test is hidden, so shall the synthesized test be hidden.
-        // Copy the exact traits from the real test in case they someday carry
-        // any interesting metadata.
-        let traits = test.traits.compactMap { $0 as? HiddenTrait }
```
briancroom, line 109: Is this no longer needed?

briancroom (approved):

Sources/Testing/Test+Discovery.swift
grynspan: Yes, it's redundant as we already recursively apply traits like this one during planning.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

PR #316, opened 2024-03-25
Fix issues building when using Xcode Command Line Tools (no XCTest.)

grynspan:
This PR removes some stray dependencies on XCTest that prevent building the testing library and its test target when XCTest is not available on Darwin. This can occur if the Xcode Command Line Tools are installed instead of the full IDE.

Automated testing for this change is difficult because it relies on a build environment that is not supported in CI (namely the presence of the CL tools but not Xcode nor XCTest.framework.) I have manually tested the change against swift-testing's own test target.

For more information about this change, see https://github.com/apple/swift-package-manager/pull/7426. That PR affects the Swift main-branch toolchain (Swift 6), not Swift 5.10 or earlier.

Resolves rdar://125371899.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/XCTestScaffold.swift
+++ b/Sources/Testing/Running/XCTestScaffold.swift
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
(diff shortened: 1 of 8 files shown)


grynspan:
@swift-ci please test

briancroom (approved):

PR #319, opened 2024-03-27
Some tweaks to the markup for `Test.ID`.

grynspan:
As described. Affects SPI only.

### Checklist:


grynspan:
Self-reviewing. Docs only so no CI needed.

PR #321, opened 2024-03-29
Extend existing Snapshot types and introduce new ones for Event.Context

dennisweissmann:
This PR extends some of the existing Snapshot types (mostly introduces new computed properties from their "original" types) and introduces a new one for `Event.Context`

### Motivation:

Components outside the test process itself sometimes need access to (a serialized form) of the event context (ie. the test and test case information).

### Modifications:

I added a new Snapshot type for `Event.Context`, extended various other existing snapshot types to have more properties (to be more in line with their "original" types), and changed the initializers of existing Snapshot types to `borrow` their argument (where possible).

### Result:

`Event.Context` has a serializable variant, and snapshot types are generally a bit more similar to the types they represent.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Event.swift
+++ b/Sources/Testing/Events/Event.swift
@@ -278,6 +278,7 @@ extension Event {
 
 extension Event {
   /// A serializable event that occurred during testing.
+  @_spi(ForToolsIntegrationOnly)
   public struct Snapshot: Sendable, Codable {
 
     /// The kind of event.
@@ -444,6 +445,39 @@ extension Event.Kind {
       case .runEnded:
         self = .runEnded
       }
+    }
+  }
+}
+
+extension Event.Context {
+
+  /// A serializable type which provides context about a posted ``Event``.
+  ///
+  @_spi(ForToolsIntegrationOnly)
+  public struct Snapshot: Sendable, Codable {
+    /// A snapshot of the test for which this instance's associated ``Event``
+    /// occurred, if any.
+    ///
+    /// If an event occurred independently of any test, or if the running test
+    /// cannot be determined, the value of this property is `nil`.
+    public var test: Test.Snapshot?
+
+    /// A snapshot of the test case for which this instance's associated
+    /// ``Event`` occurred, if any.
+    ///
+    /// The test case indicates which element in the iterated sequence is
+    /// associated with this event. For non-parameterized tests, a single test
+    /// case is synthesized. For test suite types (as opposed to test
+    /// functions), the value of this property is `nil`.
+    public var testCase: Test.Case.Snapshot?
+
+    /// Initialize a new instance of this type.
+    ///
+    /// - Parameters:
+    ///   - context: The context to snapshot.
+    public init(snapshotting context: borrowing Event.Context) {
+      test = context.test.map { Test.Snapshot(snapshotting: $0) }
+      testCase = context.testCase.map { Test.Case.Snapshot(snapshotting: $0) }
     }
   }
 }
--- a/Sources/Testing/Running/EntryPoint.swift
+++ b/Sources/Testing/Running/EntryPoint.swift
@@ -282,11 +282,8 @@ struct EventAndContextSnapshot {
   /// A snapshot of the event.
   var event: Event.Snapshot
 
-  /// A snapshot of the test associated with the event, if any.
-  var test: Test.Snapshot?
-
-  /// A snapshot of the test case associated with the event, if any.
-  var testCase: Test.Case.Snapshot?
+  /// A snapshot of the event context.
+  var eventContext: Event.Context.Snapshot
 }
 
 extension EventAndContextSnapshot: Codable {}
@@ -323,8 +320,7 @@ private func _eventHandlerForStreamingEvents(toFileAtPath path: String) throws -
   return { event, context in
     let snapshot = EventAndContextSnapshot(
       event: Event.Snapshot(snapshotting: event),
-      test: context.test.map { Test.Snapshot(snapshotting: $0) },
-      testCase: context.testCase.map { Test.Case.Snapshot(snapshotting: $0) }
+      eventContext: Event.Context.Snapshot(snapshotting: context)
     )
     if var snapshotJSON = try? JSONEncoder().encode(snapshot) {
       func isASCIINewline(_ byte: UInt8) -> Bool {
--- a/Sources/Testing/Test.swift
+++ b/Sources/Testing/Test.swift
@@ -177,6 +177,20 @@ extension Test {
   /// A serializable snapshot of a ``Test`` instance.
   @_spi(ForToolsIntegrationOnly)
   public struct Snapshot: Sendable, Codable, Identifiable {
+
+    private enum CodingKeys: String, CodingKey {
+      case id
+      case name
+      case displayName
+      case sourceLocation
+      case testCases
+      case parameters
+      case comments
+      case tags
+      case associatedBugs
+      case _timeLimit = "timeLimit"
+    }
+
     /// The ID of this test.
     public var id: Test.ID
 
@@ -189,8 +203,6 @@ extension Test {
 
     /// The customized display name of this test, if any.
     public var displayName: String?
-
-    // FIXME: Include traits as well.
 
     /// The source location of this test.
     public var sourceLocation: SourceLocation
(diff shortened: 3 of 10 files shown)


dennisweissmann:
@swift-ci please test

Sources/Testing/Running/Configuration.swift
```diff
@@ -203,3 +203,61 @@ public struct Configuration: Sendable {
   /// The test case filter to which test cases should be filtered when run.
   public var testCaseFilter: TestCaseFilter = { _, _ in true }
 }
+
+extension Configuration.RepetitionPolicy {
```
grynspan, line 207: `RepetitionPolicy` does not need a snapshot type. It can just be `Codable` directly as it is pure data.

Sources/Testing/Running/Configuration.swift
```diff
@@ -203,3 +203,61 @@ public struct Configuration: Sendable {
...
+    }
+  }
+}
+
+extension Configuration.RepetitionPolicy.ContinuationCondition {
+
+  /// An enumeration describing the conditions under which test iterations
+  /// should continue.
+  @_spi(ForToolsIntegrationOnly)
+  public enum Snapshot: Sendable, Codable {
```
grynspan, line 241: `ContinuationCondition` does not need a snapshot type. It can just be `Codable` directly as it is pure data.

Sources/Testing/Test.swift
```diff
@@ -210,17 +208,46 @@ extension Test {
     /// - ``Test/parameters``
     public var parameters: [Parameter]?
 
+    /// We can't store arbitrary traits on snapshots because they're not
```
grynspan, line 211: I'd like to see if we can come up with some general-purpose solution for arbitrary traits, but this seems like the best we can do right now.

The documentation comments for these properties should actually document them (i.e. duplicate the comments on the non-snapshot properties.) This note about what we can and can't encode can be a simple comment above the properties.

Sources/Testing/Test.swift
```diff
@@ -210,17 +208,46 @@ extension Test {
...
+    /// generally encodable, but we can store tags.
+    public var tags: Set<Tag>
+
+    /// The set of bugs associated with this test.
+    ///
+    /// For information on how to associate a bug with a test, see the
+    /// documentation for ``Bug``.
+    public var associatedBugs: [Bug]
+
+    // This is not public because TimeValue is an internal type and we currently
```
grynspan, line 225: If we currently only need a property for the test suite, that means we're testing the test rather than the production code.

Sources/Testing/Traits/Comment.swift
```diff
@@ -150,3 +150,17 @@ extension Test {
...
+extension Test.Snapshot {
+
+  /// The complete set of comments about this test from all traits of a certain
+  /// type.
+  ///
+  /// - Parameters:
+  ///   - traitType: The type of ``Trait`` whose comments should be returned.
+  ///
+  /// - Returns: The comments found for the specified test trait type.
+  public func comments<T>(from traitType: T.Type) -> [Comment] where T: Trait {
```
grynspan, line 163: This function exists so that we can filter comments based on the traits that add them, but since trait types are lost in the snapshot, it can't function correctly. We shouldn't include it.

Sources/Testing/Events/Event.swift
```diff
@@ -447,3 +448,51 @@ extension Event.Kind {
...
+    public var testCase: Test.Case.Snapshot?
+
+    /// Initialize a new instance of this type.
+    ///
+    /// - Parameters:
+    ///   - test: The test for which this instance's associated event occurred,
+    ///     if any.
+    ///   - testCase: The test case for which this instance's associated event
+    ///     occurred, if any.
+    public init(
```
grynspan, line 481: Why do we need this in addition to `public init(snapshotting context: borrowing Context)`? Do we expect somebody would ever need to initialize an instance of the snapshot type without an instance of the type being snapshotted?

grynspan (changes requested):
See notes above; please also add test coverage for all new code.

Sources/Testing/Test.swift
dennisweissmann: Ah sorry, that comment is obsolete.

Sources/Testing/Events/Event.swift
dennisweissmann: Yes.

grynspan: Why? When?

dennisweissmann:
@swift-ci please test

Sources/Testing/Test.swift
```diff
@@ -210,17 +208,48 @@ extension Test {
...
+    public init(snapshotting test: borrowing Test) {
       id = test.id
       name = test.name
       displayName = test.displayName
       sourceLocation = test.sourceLocation
       testCases = test.testCases?.map(Test.Case.Snapshot.init)
       parameters = test.parameters
+      comments = test.comments
+      tags = test.tags
+      associatedBugs = test.traits.compactMap { $0 as? Bug }
```
grynspan, line 249:
```suggestion
      associatedBugs = test.associatedBugs
```

Sources/Testing/Test.swift
```diff
@@ -210,17 +208,48 @@ extension Test {
...
+    // generally encodable, but we can store tags.
+    public var tags: Set<Tag>
+
+    /// The set of bugs associated with this test.
+    ///
+    /// For information on how to associate a bug with a test, see the
+    /// documentation for ``Bug``.
+    public var associatedBugs: [Bug]
+
+    // Backing storage for ``Test.Snapshot.timeLimit``.
```
grynspan, line 227:
```suggestion
    // Backing storage for ``Test/Snapshot/timeLimit``.
```

Sources/Testing/Test.swift
```diff
@@ -210,17 +208,48 @@ extension Test {
     /// - ``Test/parameters``
     public var parameters: [Parameter]?
 
+    /// Serializable representation of the comments associated with this test.
```
grynspan, line 211: Perhaps just use the documentation comments from the original symbols? "Serialized representation" isn't technically correct.

Sources/Testing/Test.swift
```diff
@@ -210,17 +208,48 @@ extension Test {
...
       name = test.name
       displayName = test.displayName
       sourceLocation = test.sourceLocation
       testCases = test.testCases?.map(Test.Case.Snapshot.init)
       parameters = test.parameters
+      comments = test.comments
+      tags = test.tags
+      associatedBugs = test.traits.compactMap { $0 as? Bug }
+      if #available(_clockAPI, *) {
+        _timeLimit = test.timeLimit.map(TimeValue.init)
```
grynspan, line 251: Because `_timeLimit` is the stored property, we likely need a custom implementation of `CodingKeys` on this type in order to give it the right key in JSON (`"timeLimit"`, not `"_timeLimit"`.)

Sources/Testing/Events/Event.swift
```diff
@@ -447,3 +448,36 @@ extension Event.Kind {
     }
   }
 }
+
+extension Event.Context {
+
+  /// A serializable type which provides context about a posted ``Event``.
+  ///
+  @_spi(ForToolsIntegrationOnly)
+  public struct Snapshot: Sendable, Codable {
```
grynspan, line 457: Did you mean to adopt this in the experimental JSON output code too?

dennisweissmann: Not initially but I just did :)

Sources/Testing/Test.swift
dennisweissmann: Thanks!

dennisweissmann:
@swift-ci please test

grynspan (approved):

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -282,6 +282,9 @@ struct EventAndContextSnapshot {
   /// A snapshot of the event.
   var event: Event.Snapshot
 
+  /// A snapshot of the event context.
+  var eventContext: Event.Context.Snapshot
```
grynspan, line 286: We can replace `EventAndContextSnapshot` outright. :)

stmontgomery (approved):

dennisweissmann:
@swift-ci please test

PR #323, opened 2024-04-01
Make `Event.ConsoleOutputRecorder.Option[s]` a struct instead of an enum.

grynspan:
When I'd originally assembled the `Event.ConsoleOutputRecorder.Option` enum, I assumed that it would behave much like a set of flags (_à la_ C bitmask enums.) As it has evolved, it has taken on associated values and ceased to behave like an option set, so it really should be a struct. This PR changes it to a struct, renames it from `Option` to `Options`, and of course updates call sites.

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

briancroom (approved):

PR #325, opened 2024-04-03
Use `String(reflecting:)` instead of `_typeName(_:qualified:)` to get the fully-qualified names of types.

grynspan:
We were directly using the Swift standard library function `_typeName(_:qualified:)` to get the fully-qualified names of types, but `String(reflecting:)` does this for us and is supported API. This PR switches us over to the API.

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #307, opened 2024-03-21
Add experimental exit test support.

grynspan:
This PR adds a new kind of expectation called an "exit test", "death test", "expected crasher", etc. It can be invoked by writing:

```swift
#expect(exitsWith: .failure) {
  fatalError("How dare you!")
}
```

It works by spinning up a second process that re-runs the same test function. The parent process awaits the return of the child process, while the child process executes the closure passed to `#expect()`.

This is similar to calling `fork()`, however that function is fundamentally unusable on Darwin and unavailable on Windows, so I've pursued a different implementation here that works across macOS, Linux, and Windows.

There are some constraints to using exit tests that are documented, but which are not enforced at compile time. swift-syntax's `lexicalContext` should allow us to do general checks for (some of) these constraints at compile time.

Resolves #157.
(Sort of.)

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #326, opened 2024-04-03
Add a dedicated `Win32Error` type for errors in the Windows domain.

grynspan:
On Windows, errors produced by Win32 functions (`GetLastError()`) are in a separate domain from errors produced by C standard library functions (`errno`).

Although we're not currently making _much_ use of Win32 API that produces errors in this domain, we do use it in `Environment` and #307 uses it, so I've factored out the part of that PR that adds a dedicated error type.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/CError.swift
+++ b/Sources/Testing/Support/CError.swift
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-private import TestingInternals
+internal import TestingInternals
 
 /// A type representing an error from a C function such as `fopen()`.
 ///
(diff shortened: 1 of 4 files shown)


grynspan:
@compnerd If you have a chance, would you mind looking at this PR?

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

compnerd (approved):

Sources/Testing/Support/CError.swift
```diff
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-private import TestingInternals
+internal import TestingInternals
```
dennisweissmann, line 11: Should we remove the `internal`? There are only 2 other hits for `internal import`, all other usages are implicit.

dennisweissmann (approved):

Sources/Testing/Support/CError.swift
grynspan: It's necessary because this file uses a type from TestingInternals in an `internal` spot, and because other spots say `private`, we need to explicitly state that this import is `internal`.

PR #327, opened 2024-04-03
Specialize `FileHandle.write(_:)` when the input is `UnsafeBufferPointer<UInt8>`.

grynspan:
This PR adds a specialization of `FileHandle.write(_:)` for `UnsafeBufferPointer<UInt8>`. By splitting up the specialization that takes `some Sequence<UInt8>`, we avoid needing to call
`withContiguousStorageIfAvailable(_:)` on the buffer (which just yields `self`) and avoid the overhead of a generic call. The other overloads of `write(_:)` are also refactored to call the new overload too, so there's still one true way to call `fwrite()`.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test

PR #332, opened 2024-04-05
Diagnose when a test or suite is embedded in a non-final class.

grynspan:
The testing library does not support using non-final classes as test suites. Prior to swift-syntax-600, we don't have a way to detect the name of the class, so we cannot correctly construct the thunk for a test function (we can only say `Self`, not `TheClassName`, and that's not valid for non-final classes.)

As of swift-syntax-600, that constraint is lifted, but because test content isn't inherited by subclasses (which may be confusing), we're keeping the constraint in place for now.

This PR diagnoses when a test or suite is declared inside a non-final class when swift-syntax-600 is available:

Screenshot 2024-04-05 at 3 58 13 PM

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

dennisweissmann (approved):

PR #334, opened 2024-04-06
Use `lexicalContext` to improve diagnostics for tag declarations.

grynspan:
This PR uses the swift-syntax-600 `lexicalContext` property to enforce the compile-time requirement that tags declared with `@Tag` must be declared in an extension to `Tag` or in a type declared inside `Tag`. Fix-its are provided for a few scenarios. The new diagnostics are:

```swift
@Tag var x: Tag // 🛑 Attribute 'Tag' cannot be applied to a global variable
                // Declare in an extension to 'Tag' [Fix]
                // Remove attribute 'Tag' [Fix]

extension String {
  @Tag var x: Tag // 🛑 Attribute 'Tag' cannot be applied to a property except in an extension to 'Tag'
}

extension Tag {
  @Tag var x: Tag // 🛑 Attribute 'Tag' cannot be applied to an instance property
                  // Add 'static' [Fix]
}

extension Tag {
  @Tag var x: String // 🛑 Attribute 'Tag' cannot be applied to a property of type 'String'
                     // Change type to 'Tag' [Fix]
                     // Remove attribute 'Tag' [Fix]
}
```

As with #332 and #333, these problems are only diagnosed when using swift-syntax-600.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Package.swift
```diff
@@ -33,7 +33,7 @@ let package = Package(
   ],
 
   dependencies: [
-    .package(url: "https://github.com/apple/swift-syntax.git", from: "510.0.1"),
+    .package(url: "https://github.com/apple/swift-syntax.git", branch: "main"),
```
briancroom, line 36: This change would need to be reverted for a commit when tagging a release, right?

briancroom (commented):
Neat!

Package.swift
grynspan: I did not intend to include that in the PR. Will remove.

grynspan: Removed.

grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #335, opened 2024-04-08
Warn (eventually error) when a test or suite is contained within an extension.

grynspan:
This PR produces a warning diagnostic when a test or suite is contained _within_ an extension:

```swift
extension E {
  @Suite struct S {} // ⚠️ Attribute 'Suite' cannot be applied to a structure
                     // within an extension to type 'E'; this will be an error
                     // in the future
  @Test func f() {} // ⚠️ Attribute 'Test' cannot be applied to a function
                    // within an extension to type 'E'; this will be an error
                    // in the future
}
```

During macro expansion, if a test or suite is contained within an extension, only the extension declaration itself can be "seen" by the macros code (and then, only with swift-syntax-600.) The actual type declaration corresponding to the extension is not available. For most types, this is fine, but there are several scenarios that produce poor or fatal outcomes:

```swift
// Generic types cannot be realized, so tests contained in them cannot run, and
// there are no compile-time diagnostics indicating the cause of the problem:
struct S<T> {}
extension S {
  @Test func f() {}
}

// Protocols can be extended, but are not concrete types and the macro code
// fails to compile with confusing diagnostics:
protocol P {}
extension P {
  @Test func f() {} // 🛑 'any P' cannot be constructed because it has no
                    // accessible initializers
                    // 🛑 'any P' cannot be constructed because it has no
                    // accessible initializers (yes, twice!)
                    // 🛑 Type '$s12TestingTests1PPAAE1f4TestfMp_37__🟠$test_container__function__funcf__fMu_'
                    // cannot be nested in protocol extension of 'P'
                    // 🛑 Use of protocol 'P' as a type must be written 'any P'
}

// Suites with constrained availability are not (yet) supported because it is
// not possible to inspect their types at runtime on unsupported OS versions.
// Depending on platform and precise package or Xcode project configuration,
// a type with constrained availability may cause crashes at runtime:
@available(macOS 13, *)
struct S {}

extension S {
  @Test func f() {} // Will crash when run on macOS 12
}
```

It would be preferable to block the use of extensions with swift-testing until we can devise a solution to the above problems (e.g. having the macro expansion context provide stub type declarations for each extension in a lexical context.)

Because there are likely a number of packages that have adopted swift-testing already and are using extensions in this fashion, we've opted to introduce this diagnostic as a warning. We will, subject to community review, upgrade the warning to an error in a subsequent swift-testing tag.

As with previous PRs in this area, these diagnostics are only available when using **swift-syntax-600**.

### Checklist:

### DIFF
--- a/Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
+++ b/Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
@@ -116,6 +116,29 @@ struct TestDeclarationMacroTests {
     }
   }
 
+  @Test("Warning diagnostics emitted on API misuse",
+    arguments: [
+      // return types
+      "@Test func f() -> Int {}":
+        "The result of this function will be discarded during testing",
+      "@Test func f() -> Swift.String {}":
+        "The result of this function will be discarded during testing",
+      "@Test func f() -> Int? {}":
+        "The result of this function will be discarded during testing",
+      "@Test func f() -> (Int, Int) {}":
+        "The result of this function will be discarded during testing",
+    ]
+  )
+  func apiMisuseWarnings(input: String, expectedMessage: String) throws {
+    let (_, diagnostics) = try parse(input)
+
+    #expect(diagnostics.count > 0)
+    for diagnostic in diagnostics {
+      #expect(diagnostic.diagMessage.severity == .warning)
+      #expect(diagnostic.message == expectedMessage)
+    }
+  }
+
 #if canImport(SwiftSyntax600)
   @Test("Error diagnostics emitted for invalid lexical contexts",
     arguments: [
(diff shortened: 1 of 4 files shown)


Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
```diff
@@ -116,6 +116,29 @@ struct TestDeclarationMacroTests {
     }
   }
 
+  @Test("Warning diagnostics emitted on API misuse",
```
grynspan, line 119: Not new, just rearranged.

Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
```diff
@@ -146,22 +169,16 @@ struct TestDeclarationMacroTests {
       #expect(diagnostic.message == expectedMessage)
     }
   }
-#endif
 
-  @Test("Warning diagnostics emitted on API misuse",
```
grynspan, line 151: Not deleted, just rearranged.

grynspan:
@swift-ci please test

grynspan:
We are pursuing a different approach here.

PR #336, opened 2024-04-08
Apply suite diagnostics to the entire lexical context.

grynspan:
This PR applies all suite diagnostics\* to every decl group (declaring type or extension) in the lexical context during `@Suite` and `@Test` expansion. For example:

```swift
struct S1<T> {
  @Suite struct S2 { // 🛑 Attribute 'Suite' cannot be applied to a generic structure
    @Test func f() {} // 🛑 Attribute 'Test' cannot be applied to a generic function
  }
}
```

This PR is follow-up to #332 and #333.

\* Except for diagnostics specific to the `@Suite` attribute itself.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/DiagnosticMessage.swift
+++ b/Sources/TestingMacros/Support/DiagnosticMessage.swift
@@ -188,8 +188,11 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
   ///
   /// - Returns: A diagnostic message.
   static func genericDeclarationNotSupported(_ decl: some SyntaxProtocol, whenUsing attribute: AttributeSyntax, becauseOf genericClause: some SyntaxProtocol) -> Self {
-    Self(
-      syntax: (genericClause.root != decl.root) ? Syntax(decl) : Syntax(genericClause),
+    // Avoid using a syntax node from a lexical context (it won't have source
+    // location information.)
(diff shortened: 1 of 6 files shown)


Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -189,7 +189,7 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
   /// - Returns: A diagnostic message.
   static func genericDeclarationNotSupported(_ decl: some SyntaxProtocol, whenUsing attribute: AttributeSyntax, becauseOf genericClause: some SyntaxProtocol) -> Self {
     Self(
-      syntax: Syntax(genericClause),
+      syntax: (genericClause.root != decl.root) ? Syntax(decl) : Syntax(genericClause),
```
grynspan, line 192: Nodes from the lexical context are detached, so diagnostics applied to them don't align to any real source location. This check is only (currently) needed here since we use this diagnostic unmodified for both code paths and prefer that the diagnostic point to the generic clause rather than the affected symbol.

grynspan:
@swift-ci please test

dennisweissmann (approved):

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

PR #339, opened 2024-04-10
Avoid reassigning task-local runtime state in a Runner's event handler unless it's nested

stmontgomery:
This is a mitigation for a bug in which an expectation inside a `withTaskGroup` closure can cause a crash if it fails.

### Motivation:

It can sometimes be useful to perform expectations (e.g. `#expect`) within the body of the closure passed to `withTaskGroup` or similar APIs. However, doing so currently causes a crash if the expectation fails, illustrated by this small example:

```swift
@Test func example() async {
  await withTaskGroup(of: Void.self) { _ in
    #expect("abc" == "123") // 💥
  }
}
```

### Modifications:

This change avoids wrapping the event handler of a runner's configuration with one that reassigns a TaskLocal (potentially triggering the crash) unless that runner is running during the run operation of another runner. In other words, only mutate the TaskLocal for nested runners. Nesting, or running one runner in the context of another, is something which is only done in practice when performing tests of the testing library itself. Ordinary usage of the testing library only has one “outer“ runner.

### Result:

The above example code no longer crashes, verified by the new unit test.

### Checklist:

Resolves rdar://126132392

### DIFF
--- a/Tests/TestingTests/Runner.RuntimeStateTests.swift
+++ b/Tests/TestingTests/Runner.RuntimeStateTests.swift
@@ -25,4 +25,21 @@ struct Runner_RuntimeStateTests {
 
     await Test(name: "foo") {}.run(configuration: configuration)
   }
+
+  // This confirms that an event posted within the closure of `withTaskGroup`
+  // (and similar APIs) in regular tests (those which are not testing the
+  // testing library itself) are handled appropriately and do not crash.
+  @Test func eventPostingInTaskGroup() async {
+    // Enable delivery of "expectation checked" events, merely as a way to allow
+    // an event to be posted during the test below without causing any real
+    // issues to be recorded or otherwise confuse the testing harness.
+    var configuration = Configuration.current ?? .init()
+    configuration.deliverExpectationCheckedEvents = true
+
+    await Configuration.withCurrent(configuration) {
+      await withTaskGroup(of: Void.self) { group in
+        #expect(Bool(true))
+      }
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

dennisweissmann (approved):

briancroom (approved):

Tests/TestingTests/Runner.RuntimeStateTests.swift
```diff
@@ -25,4 +25,21 @@ struct Runner_RuntimeStateTests {
 
     await Test(name: "foo") {}.run(configuration: configuration)
   }
+
+  // This confirms that an event posted within the closure of `withTaskGroup`
+  // (and similar APIs) in regular tests (those which are not testing the
+  // testing library itself) are handled appropriately and do not crash.
+  @Test func eventPostingInTaskGroup() async {
```
grynspan, line 32: Can we have an additional test (maybe written with XCTest) that sanity-checks that the state is `nil` by default?

stmontgomery:
@swift-ci please test

Tests/TestingTests/Runner.RuntimeStateTests.swift
stmontgomery: Sure, added that

grynspan (approved):

ktoso (commented):
LGTM, thank you!

PR #347, opened 2024-04-12
Require that `.bug()` identifiers be parseable as URLs.

grynspan:
This PR constrains the valid values of `.bug()` identifiers. They must now be valid URLs or, as a fallback, valid unsigned integers. Arbitrary strings were never intended to be supported.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -58,6 +58,7 @@ let package = Package(
     .macro(
       name: "TestingMacros",
       dependencies: [
+        "TestingInternals",
         .product(name: "SwiftDiagnostics", package: "swift-syntax"),
         .product(name: "SwiftSyntax", package: "swift-syntax"),
         .product(name: "SwiftSyntaxBuilder", package: "swift-syntax"),
--- a/Package.swift
+++ b/Package.swift
@@ -81,7 +81,8 @@ let package = Package(
     // by other targets above, not directly included in product libraries.
     .target(
       name: "TestingInternals",
-      cxxSettings: .packageSettings
+      cxxSettings: .packageSettings,
+      linkerSettings: .packageSettings
     ),
 
     // Cross-module overlays (unsupported)
@@ -154,6 +155,16 @@ extension Array where Element == PackageDescription.CXXSetting {
   static var packageSettings: Self {
     [
       .define("_SWT_TESTING_LIBRARY_VERSION", to: #""unknown (Swift 5.10 toolchain)""#),
+    ]
+  }
+}
+
+extension Array where Element == PackageDescription.LinkerSetting {
+  /// Settings intended to be applied to every linked target in this package.
+  /// Analogous to project-level build settings in an Xcode project.
+  static var packageSettings: Self {
+    [
+      .linkedLibrary("curl", .when(platforms: [.macOS, .iOS, .macCatalyst, .watchOS, .tvOS, .visionOS, .linux]))
     ]
   }
 }
--- a/Package.swift
+++ b/Package.swift
@@ -58,7 +58,6 @@ let package = Package(
     .macro(
       name: "TestingMacros",
       dependencies: [
-        "TestingInternals",
         .product(name: "SwiftDiagnostics", package: "swift-syntax"),
         .product(name: "SwiftSyntax", package: "swift-syntax"),
         .product(name: "SwiftSyntaxBuilder", package: "swift-syntax"),
--- a/Package.swift
+++ b/Package.swift
@@ -81,10 +81,7 @@ let package = Package(
     // by other targets above, not directly included in product libraries.
     .target(
       name: "TestingInternals",
-      cxxSettings: .packageSettings,
-      linkerSettings: [
-        .linkedLibrary("curl", .when(platforms: [.macOS, .iOS, .macCatalyst, .watchOS, .tvOS, .visionOS, .linux]))
-      ]
+      cxxSettings: .packageSettings
     ),
 
     // Cross-module overlays (unsupported)
--- a/Package.swift
+++ b/Package.swift
@@ -82,7 +82,9 @@ let package = Package(
     .target(
       name: "TestingInternals",
       cxxSettings: .packageSettings,
-      linkerSettings: .packageSettings
+      linkerSettings: [
+        .linkedLibrary("curl", .when(platforms: [.macOS, .iOS, .macCatalyst, .watchOS, .tvOS, .visionOS, .linux]))
+      ]
     ),
 
     // Cross-module overlays (unsupported)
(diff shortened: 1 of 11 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@compnerd Mind sanity-checking the Win32 code here? Thanks!

Sources/TestingInternals/include/Includes.h
```diff
@@ -75,10 +75,19 @@
 #include <wasi/libc-environ.h>
 #endif
 
+#if !SWT_NO_CURL
```
compnerd, line 78: Should this be under `!defined(_WIN32)`?

Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
```diff
@@ -36,58 +36,171 @@ func diagnoseIssuesWithTags(in traitExprs: [ExprSyntax], addedTo attribute: Attr
...
+
+  func isURLStringValid(_ urlString: String) -> Bool {
+    guard urlString.allSatisfy(\.isASCII),
+          let colonIndex = urlString.firstIndex(of: ":") else {
+      // This can't be a valid URL as far as we're concerned. Exit early without
+      // properly parsing it.
+      return false
+    }
+
+#if SWT_TARGET_OS_APPLE || os(Linux)
```
compnerd, line 137: This might be good to replicate into the C side.

Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
```diff
@@ -36,58 +36,171 @@ func diagnoseIssuesWithTags(in traitExprs: [ExprSyntax], addedTo attribute: Attr
...
+#endif
+
+#elseif os(WASI)
+    // TODO: URL validation on WASI (this code runs on the host though)
+    return true
+#elseif os(Windows)
+    return urlString.withCString(encodedAs: UTF16.self) { urlString in
+      var components = URL_COMPONENTSW()
+      components.dwStructSize = DWORD(MemoryLayout.size(ofValue: components))
+      return withUnsafeTemporaryAllocation(of: wchar_t.self, capacity: 1) { buffer in
```
compnerd, line 185: I don't understand this bit - the capacity is 1, so this allocates 1 `wchar_t` right? not a `wchar_t *`?

grynspan: Yep. I need a `wchar_t` buffer, but I also only need to check if the scheme is empty, so I don't need more than one character.

Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
grynspan: We already use include guards for various headers, so it isn't an issue.

Sources/TestingInternals/include/Includes.h
grynspan: Nope, because Windows doesn't have this header anyway.

Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
compnerd: Might be nice to have a comment to that effect.

Package.swift
```diff
@@ -156,3 +158,13 @@ extension Array where Element == PackageDescription.CXXSetting {
     ]
   }
 }
+
+extension Array where Element == PackageDescription.LinkerSetting {
+  /// Settings intended to be applied to every linked target in this package.
+  /// Analogous to project-level build settings in an Xcode project.
+  static var packageSettings: Self {
```
stmontgomery, line 165: This property is named **package** settings but it's not actually applied to all of the targets currently, just one.

However, it seems like actually only the single target which uses curl needs this linkage. So how about instead of creating this new package-wide setting, we only add this linkage to the single target which requires it?

Sources/Testing/Testing.docc/BugIdentifiers.md
```diff
@@ -44,14 +47,15 @@ handling to detect:
...
+| `.bug(12345)` | Yes | None |
+| `.bug("12345")` | Yes | None |
+| `.bug("Things don't work")` | **No** | None |
+| `.bug("rdar:12345")` | Yes | Apple Radar |
+| `.bug("https://github.com/apple/swift/pull/12345")` | Yes | [GitHub Issues for the Swift project](https://github.com/apple/swift/issues) |
+| `.bug("https://bugs.webkit.org/show_bug.cgi?id=12345")` | Yes | [WebKit Bugzilla](https://bugs.webkit.org/) |
 <!--
-| `.bug("FB12345")` | Apple Feedback Assistant | // SEE ALSO: rdar://104582015
-| `.bug("#12345")` | GitHub Issues for the current repository (if hosted there) |
+| `.bug("FB12345")` | Yes | Apple Feedback Assistant | // SEE ALSO: rdar://104582015
```
stmontgomery, line 59: I'm a little confused why this is still considered valid. Does it parse as an integer?

grynspan: It's commented out as a hypothetical "future direction."

Package.swift
grynspan: Can do!

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/TestingInternals/include/Includes.h
compnerd: I'm not sure I understand this. The header is a curl header, and this header has a `_WIN32` path below.

Sources/TestingMacros/Support/Additions/EditorPlaceholderExprSyntaxAdditions.swift
```diff
@@ -0,0 +1,26 @@
...
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if compiler(>=5.11)
```
compnerd, line 11: This seems inverted? We want to use the ACL'ed import don't we?

grynspan: We are using ACL'ed imports. There's a difference in behaviour between the Swift 5.10 and Swift 5.11/6 compilers that we're working around here (you'll get various diagnostics if you don't do this while you've got ACL'ed imports enabled.)

Sources/TestingInternals/include/Includes.h
grynspan: If your Windows SDK happens to have the libcurl headers, we will end up including this header, but we just won't do anything with them. The same can be said of the other headers here. We added a `SWT_NO_CURL` check because not all Linux/POSIX-like variants will have the libcurl headers and some custom configuration may be needed.

compnerd (approved):

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

briancroom (approved):

PR #351, opened 2024-04-15
Only store Mach header addresses for images containing Swift data.

grynspan:
We currently try to do minimal work in our `objc_addLoadImageFunc()` callback on Darwin—we acquire a lock, add each reported Mach header to a list, relinquish the lock, and immediately return. We then look for appropriate section data on demand in `enumerateTypeMetadataSections()`. This is a problem if an image is unloaded after the callback to `objc_addLoadImageFunc()` returns but before we enumerate images. Images containing Swift do not support unloading, but pure C images might find themselves in this situation.

This PR checks for the presence of a `"__swift5_types"` section before adding an image to the list, so references to images that could be unloaded won't be held indefinitely.

Resolves rdar://124426864.

### Checklist:

### DIFF
--- a/Sources/TestingInternals/Discovery.cpp
+++ b/Sources/TestingInternals/Discovery.cpp
@@ -233,9 +233,18 @@ static SWTMachHeaderList getMachHeaders(void) {
     machHeaders->reserve(_dyld_image_count());
 
     objc_addLoadImageFunc([] (const mach_header *mh) {
-      os_unfair_lock_lock(&lock); {
-        machHeaders->push_back(reinterpret_cast<SWTMachHeaderList::value_type>(mh));
-      } os_unfair_lock_unlock(&lock);
+      auto mhn = reinterpret_cast<SWTMachHeaderList::value_type>(mh);
+
+      // Only store the mach header address if the image contains Swift data.
+      // Swift does not support unloading images, but images that do not contain
+      // Swift or Objective-C code may be unloaded at runtime and later crash
+      // the testing library when it calls enumerateTypeMetadataSections().
+      unsigned long size = 0;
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

grynspan:
@mikeash Please confirm this makes sense to you.

stmontgomery (approved):

dennisweissmann (approved):

Sources/TestingInternals/Discovery.cpp
```diff
@@ -233,9 +233,18 @@ static SWTMachHeaderList getMachHeaders(void) {
...
 
     objc_addLoadImageFunc([] (const mach_header *mh) {
-      os_unfair_lock_lock(&lock); {
-        machHeaders->push_back(reinterpret_cast<SWTMachHeaderList::value_type>(mh));
-      } os_unfair_lock_unlock(&lock);
+      auto mhn = reinterpret_cast<SWTMachHeaderList::value_type>(mh);
+
+      // Only store the mach header address if the image contains Swift data.
+      // Swift does not support unloading images, but images that do not contain
+      // Swift or Objective-C code may be unloaded at runtime and later crash
```
mikeash, line 240:
```suggestion
      // Swift code may be unloaded at runtime and later crash
```

This is very nitpicky, but ObjC can be unloaded if it's in a `MH_BUNDLE`.

grynspan: Booooo. Okay.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

briancroom (approved):

PR #352, opened 2024-04-15
Remove the new diagnostic added in #302.

grynspan:
In #302, we added a compile-time diagnostic (warning) for expressions like:

```swift
#expect(try await foo())
```

Because we figured that the lack of expression expansion for effectful expressions might be confusing. However, we've found that the diagnostic is significantly noisier than we'd like and the cons outweigh the pros. Hence, this PR removes that diagnostic.

Resolves rdar://126393932.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

briancroom (approved):

PR #355, opened 2024-04-16
Abstract away our dependency on Foundation for JSON encoding/decoding

grynspan:
This PR moves all our uses of `JSONEncoder` and `JSONDecoder` to a new namespace enum, `JSON`, and provides interfaces that don't use Foundation.

This change does _not_ remove our dependency on Foundation's JSON functionality, but it should allow us to replace that functionality in the future with JSON encoding/decoding logic from another library (or home-grown, if necessary.)

### Checklist:

### DIFF
--- a/Sources/Testing/Support/JSON.swift
+++ b/Sources/Testing/Support/JSON.swift
@@ -1,0 +1,65 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if canImport(Foundation)
+private import Foundation
+#endif
+
+enum JSON {
+  /// Encode a value as JSON.
+  ///
+  /// - Parameters:
+  ///   - value: The value to encode.
+  ///   - userInfo: Any user info to pass into the encoder during encoding.
+  ///   - body: A function to call.
+  ///
+  /// - Returns: Whatever is returned by `body`.
+  ///
+  /// - Throws: Whatever is thrown by `body` or by the encoding process.
+  static func withEncoding<R>(of value: some Encodable, userInfo: [CodingUserInfoKey: Any] = [:], _ body: (UnsafeRawBufferPointer) throws -> R) throws -> R {
+#if canImport(Foundation)
+    let encoder = JSONEncoder()
+
+    // Keys must be sorted to ensure deterministic matching of encoded data.
+    encoder.outputFormatting.insert(.sortedKeys)
+
+    // Set user info keys that clients want to use during encoding.
+    encoder.userInfo.merge(userInfo, uniquingKeysWith: { _, rhs in rhs})
+
+    let data = try encoder.encode(value)
+    return try data.withUnsafeBytes(body)
+#else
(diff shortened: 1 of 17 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
Linux failure is because we've removed some Foundation imports (which reexport Glibc, I guess.) Explicitly adding the necessary headers to Includes.h.

grynspan:
@swift-ci please test

briancroom (approved):

grynspan:
@swift-ci please test

Sources/Testing/Support/JSON.swift
```diff
@@ -0,0 +1,65 @@
...
+
+  /// Decode a value from JSON data.
+  ///
+  /// - Parameters:
+  ///   - jsonRepresentation: The JSON encoding of the value to decode.
+  ///
+  /// - Returns: An instance of `T` decoded from `jsonRepresentation`.
+  ///
+  /// - Throws: Whatever is thrown by the decoding process.
+  static func decode<T>(_ _: T.Type, from jsonRepresentation: UnsafeRawBufferPointer) throws -> T where T: Decodable {
```
stmontgomery, line 51: I know we don't need the type bound to a variable in the function body, but it seems like it'd be useful to document the parameter's purpose, and thus give it a name instead of `_`

grynspan:
@swift-ci please test

Sources/Testing/Support/JSON.swift
grynspan: I can name it, sure.

Sources/Testing/Traits/Tags/Tag.Color+Loading.swift
```diff
@@ -99,7 +99,9 @@ func loadTagColors(fromFileInDirectoryAtPath swiftTestingDirectoryPath: String =
   // nil is a valid decoded color value (representing "no color") that we can
   // use for merging tag color data from multiple sources, but it is not valid
   // as an actual tag color, so we have a step here that filters it.
-  return try JSONDecoder().decode([Tag: Tag.Color?].self, from: tagColorsData)
-    .compactMapValues { $0 }
+  return try tagColorsData.withUnsafeBytes { tagColorsData in
```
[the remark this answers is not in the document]
grynspan: Not in this file because it uses `URL` and `Data`—it goes away in the PR I've got lined up after this one.

Sources/Testing/Support/JSON.swift
```diff
@@ -0,0 +1,65 @@
...
+  ///
+  /// - Returns: Whatever is returned by `body`.
+  ///
+  /// - Throws: Whatever is thrown by `body` or by the encoding process.
+  static func withEncoding<R>(of value: some Encodable, userInfo: [CodingUserInfoKey: Any] = [:], _ body: (UnsafeRawBufferPointer) throws -> R) throws -> R {
+#if canImport(Foundation)
+    let encoder = JSONEncoder()
+
+    // Keys must be sorted to ensure deterministic matching of encoded data.
+    encoder.outputFormatting.insert(.sortedKeys)
```
stmontgomery, line 31: Most clients don't need sorted keys, although the usage in `CustomTestArgumentEncodable.swift` definitely does. Could we make that a defaulted `Bool` parameter?

grynspan: I made the Executive Decision™ here that we might as well just always sort keys. We could default it, but I don't believe there's any _problem_ with just leaving it on all the time.

stmontgomery (approved):

PR #357, opened 2024-04-16
Only block a single thread on Linux (and none on Windows) when waiting for exit tests.

grynspan:
This PR refines the implementation of exit tests on Linux so that multiple exit tests running concurrently only need to block a single thread rather than one thread per child process.

This PR also refines the implementation of exit tests on Windows to use `RegisterWaitForSingleObject()` for asynchronous notification of process termination rather than requiring a thread per child process.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/WaitFor.swift
+++ b/Sources/Testing/ExitTests/WaitFor.swift
@@ -43,6 +43,43 @@ private func _blockAndWait(for pid: pid_t) throws -> ExitCondition {
   }
 }
 
+#if os(Linux)
+/// A mapping of awaited child PIDs to their corresponding Swift continuations.
+private let _childProcessContinuations = Locked<[pid_t: CheckedContinuation<Void, Never>]>()
+
+/// The implementation of `_createWaitThread()`, run only once.
+private let _createWaitThread: Void = {
+  // Create the thread. We immediately detach it upon success to allow the
+  // system to reclaim its resources when done.
+
+  var thread = pthread_t()
+  _ = pthread_create(
+    &thread,
+    nil,
+    { _ in
+      // Run an infinite loop that waits for child processes to terminate and
+      // captures their exit statuses.
+      while true {
+        var siginfo = siginfo_t()
+        if 0 == waitid(P_ALL, 0, &siginfo, WEXITED | WNOWAIT) {
+          let continuation = _childProcessContinuations.withLock { childProcessContinuations in
+            childProcessContinuations.removeValue(forKey: siginfo.si_pid)
+          }
+          continuation?.resume()
+        }
+      }
+    },
+    nil
+  )
+}()
+
+/// Create a waiter thread that is responsible for waiting for child processes
+/// to exit.
+private func _createWaitThread() {
+  _createWaitThreadImpl
+}
+#endif
+
 /// Wait for a given PID to exit and report its status.
 ///
 /// - Parameters:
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

grynspan:
That's what I get for trying to brain-compile on three OSes.

grynspan:
@swift-ci please test

briancroom (approved):

compnerd (approved):
The Windows side looks good to me. I'll see if I can find some time to go over the Linux side later.

grynspan:
Thanks much @compnerd!

grynspan:
The Linux implementation works on Darwin as well, so if we don't want to take a libdispatch dependency, we could switch over to it without much fuss.

Sources/Testing/ExitTests/WaitFor.swift
```diff
@@ -43,6 +43,44 @@ private func _blockAndWait(for pid: pid_t) throws -> ExitCondition {
...
+    &thread,
+    nil,
+    { _ in
+      // Run an infinite loop that waits for child processes to terminate and
+      // captures their exit statuses.
+      while true {
+        var siginfo = siginfo_t()
+        if 0 == waitid(P_ALL, 0, &siginfo, WEXITED | WNOWAIT) {
+          let pid = swt_siginfo_t_si_pid(&siginfo)
+          let continuation = _childProcessContinuations.withLock { childProcessContinuations in
```
grynspan, line 66: There's a race condition here if `waitid(P_ALL)` fires for a PID before we've added it to the dictionary.

Sources/Testing/ExitTests/WaitFor.swift
```diff
@@ -65,40 +103,18 @@ func wait(for pid: pid_t) async throws -> ExitCondition {
...
-    // system to reclaim its resources when done.
-    var thread = pthread_t()
-    switch pthread_create(&thread, nil, body, context) {
-    case 0:
-      _ = pthread_detach(thread)
-    case let errorCode:
-      continuation.resume(throwing: CError(rawValue: errorCode))
+  await withCheckedContinuation { continuation in
+    let oldContinuation = _childProcessContinuations.withLock { childProcessContinuations in
+      childProcessContinuations.updateValue(continuation, forKey: pid)
```
grynspan, line 112: If we add a call to `waitpid(WNOHANG)` here before adding the continuation to the dictionary, then `waitid(P_ALL)` will be forced to wait to acquire the lock either _before_ the call to `waitpid(WNOHANG)` or _after_ the continuation is added.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
Probably need to watch for `ECHILD` and suspend (or just kill) the thread when encountered. Can then create a new thread the next time it's needed.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
Okay, I think I've noodled on this change enough.

Sources/Testing/ExitTests/WaitFor.swift
```diff
@@ -12,36 +12,123 @@
...
+
+/// The implementation of `_createWaitThread()`, run only once.
+private let _createWaitThreadImpl: Void = {
+  // The body of the thread's run loop.
+  func waitForAnyChild() {
+    // Listen for child process exit events. WNOWAIT means we don't perturb the
+    // state of a terminated (zombie) child process, allowing us to fetch the
+    // continuation (if available) before reaping.
+    var siginfo = siginfo_t()
+    if 0 == waitid(P_ALL, 0, &siginfo, WEXITED | WNOWAIT) {
```
iCharlesHu, line 66: Is it possible that `waitid` returns nonzero and sets `errno` to be not `ECHILD`? Since we are calling this method in a `while true` could this loop forever?

iCharlesHu (commented):
Overall looks great! My only "big picture" comment is -- should we support some sort of timeout? Since we are using `while true` loop when waiting, it has potential to run forever right?

Sources/Testing/ExitTests/WaitFor.swift
grynspan: Possible errors [per POSIX](https://pubs.opengroup.org/onlinepubs/7908799/xsh/waitid.html) are `ECHILD`, `EINTR`, and `EINVAL`. `ECHILD` is handled here; `EINTR` is an interrupt and the correct response is to loop. `EINVAL` is not possible in this circumstance as our inputs are valid. Linux does not AFAIK define any other errors for `waitid()`.

grynspan:
> Overall looks great! My only "big picture" comment is -- should we support some sort of timeout? Since we are using `while true` loop when waiting, it has potential to run forever right?

swift-testing has a `.timeLimit()` trait you can apply to a test, and tests time out after 1 minute by default. We don't otherwise provide API for fine-grained timeouts because we've found, historically, that they result in brittle tests that fail on your coworker's desk (they refuse to upgrade their 386, after all!) or in CI.

PR #358, opened 2024-04-17
Make Configuration.TestFilter a pure-data type

stmontgomery:
This modifies the structure of `Configuration.TestFilter` so that it's a pure data type and is easier to become serializable.

### Motivation:

For easier integration with tools it would be helpful for the stored properties of `Configuration` to be serializable, or be trivially convertible to/from serializable types at least.

`Configuration`'s nested `TestFilter` type currently includes a closure for filtering tests and is therefore cannot be serialized in its current form. This type doesn't really need to _store_ a closure, however; it only needs to store information about how to perform filtering (e.g. data about criteria to match against), and later that can be interpreted when applying filtering.

### Modifications:

- Change the `Configuration.TestFilter.Kind` enum so that its cases only store data, and no closures.
- Move the filtering logic derived from the pure data `Kind` type above to a new `Configuration.TestFilter.Operation` enum.

### Checklist:

Resolves rdar://126627664

### DIFF
--- a/Sources/Testing/Running/Configuration.TestFilter.swift
+++ b/Sources/Testing/Running/Configuration.TestFilter.swift
@@ -28,13 +28,12 @@ extension Configuration {
       case excluding
     }
 
-    /// An enumeration describing a predicate that a test filter may specify.
-    fileprivate enum Predicate: Sendable, Codable {
+    /// An enumeration describing the various kinds of test filter.
+    fileprivate enum Kind: Sendable, Codable {
       /// The test filter has no effect.
       ///
-      /// All tests are allowed when passed to a test filter with this
-      /// predicate.
-      case all
+      /// All tests are allowed when passed to a test filter with this kind.
+      case unfiltered
 
       /// The test filter contains a precomputed selection of test IDs.
       ///
@@ -59,21 +58,21 @@ extension Configuration {
       ///   - membership: How to interpret the result when predicating tests.
       case pattern(_ pattern: String, membership: Membership)
 
-      /// The test filter is a combination of other test predicates.
+      /// The test filter is a combination of other test filter kinds.
       ///
       /// - Parameters:
-      ///   - lhs: The first test filter's predicate.
-      ///   - rhs: The second test filter's predicate.
+      ///   - lhs: The first test filter's kind.
+      ///   - rhs: The second test filter's kind.
       ///   - op: The operator to apply when combining the results of the two
       ///     filters.
       ///
-      /// The result of a test filter with this predicate is the combination of
-      /// the results of its subfilters using `op`.
-      indirect case compound(_ lhs: Self, _ rhs: Self, _ op: CombinationOperator)
+      /// The result of a test filter with this kind is the combination of the
+      /// results of its subfilters using `op`.
+      indirect case combination(_ lhs: Self, _ rhs: Self, _ op: CombinationOperator)
     }
 
-    /// The predicate of this test filter.
-    private var _predicate: Predicate
+    /// The kind of test filter.
+    private var _kind: Kind
 
     /// Whether or not to include tests with the `.hidden` trait when filtering
     /// tests.
@@ -101,7 +100,7 @@ extension Configuration.TestFilter {
   /// This test filter allows all tests to run; it is the default test filter if
   /// another is not specified.
   public static var unfiltered: Self {
-    Self(_predicate: .all)
+    Self(_kind: .unfiltered)
   }
 
   /// Initialize this instance to filter tests to those specified by a set of
@@ -110,7 +109,7 @@ extension Configuration.TestFilter {
   /// - Parameters:
   ///   - testIDs: A set of test IDs to be filtered.
   public init(including testIDs: some Collection<Test.ID>) {
-    self.init(_predicate: .testIDs(Set(testIDs), membership: .including))
+    self.init(_kind: .testIDs(Set(testIDs), membership: .including))
   }
 
   /// Initialize this instance to filter tests to those _not_ specified by a set
@@ -119,7 +118,7 @@ extension Configuration.TestFilter {
   /// - Parameters:
   ///   - selection: A set of test IDs to be excluded.
   public init(excluding testIDs: some Collection<Test.ID>) {
-    self.init(_predicate: .testIDs(Set(testIDs), membership: .excluding))
+    self.init(_kind: .testIDs(Set(testIDs), membership: .excluding))
   }
 
   /// Initialize this instance to represent a pattern expression matched against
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -28,42 +28,52 @@ extension Configuration {
       case excluding
     }
 
-    /// An enumeration describing the various kinds of test filter.
-    fileprivate enum Kind: Sendable {
+    /// An enumeration describing a predicate that a test filter may specify.
+    fileprivate enum Predicate: Sendable, Codable {
```
grynspan, line 32: I think `Kind` here is probably still appropriate and is consistent with our other nested-enum types.

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -28,42 +28,52 @@ extension Configuration {
...
       /// The test filter has no effect.
       ///
-      /// All tests are allowed when passed to a test filter with this kind.
-      case unfiltered
+      /// All tests are allowed when passed to a test filter with this
+      /// predicate.
+      case all
```
grynspan, line 37: Is it necessary to rename this case? (It's not a problem _per se_ but it makes the diff more complicated.)

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -28,42 +28,52 @@ extension Configuration {
...
 
       /// The test filter contains a precomputed selection of test IDs.
       ///
       /// - Parameters:
       ///   - testIDs: The set of test IDs to predicate tests against.
       ///   - membership: How to interpret the result when predicating tests.
-      case precomputed(_ testIDs: Test.ID.Selection, membership: Membership)
+      case testIDs(_ testIDs: Set<Test.ID>, membership: Membership)
```
grynspan, line 44: By losing the precomputed selection and dropping down to a `Set`, we may be sacrificing a lot of performance. Can you confirm we're not recomputing the graph too frequently as a result of this change?

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -178,32 +181,92 @@ extension Configuration.TestFilter {
...
+}
 
-  /// Initialize this instance to represent a regular expression matched against
-  /// a test's ID.
+// MARK: - Operations
+
+extension Configuration.TestFilter {
+  /// An enumeration which represents filtering logic to be applied to a test
+  /// graph.
+  fileprivate enum Operation {
```
grynspan, line 193: Should be `Sendable`.

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -178,32 +181,92 @@ extension Configuration.TestFilter {
...
-  init(membership: Membership, matching regex: Regex<AnyRegexOutput>) {
-    let regex = UncheckedSendable(rawValue: regex)
-    self.init(membership: membership) { test in
-      let id = String(describing: test.id)
-      return id.contains(regex.rawValue)
+  /// - Throws: Any error encountered while converting the specified predicate
+  ///   into an operation. One example is the creation of a `Regex` from a
+  ///   `.pattern` predicate: if the pattern is not a valid regular expression,
+  ///   an error will be thrown.
+  init(_ predicate: Configuration.TestFilter.Predicate) throws {
```
grynspan, line 241: `func makeOperation()` on `TestFilter` maybe? 🤔 The initializer syntax would suggest it's a conversion/cast, but it's more of a factory function IMHO.

dennisweissmann (approved):

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -263,13 +326,13 @@ extension Configuration.TestFilter {
   ///   - testGraph: The test graph to filter.
   ///
   /// - Returns: A copy of `testGraph` with filtered tests replaced with `nil`.
-  func apply(to testGraph: Graph<String, Test?>) -> Graph<String, Test?> {
-    var result = _kind.apply(to: testGraph)
+  func apply(to testGraph: Graph<String, Test?>) throws -> Graph<String, Test?> {
+    var result = try Operation(_predicate).apply(to: testGraph)
```
grynspan, line 330: Do we ever need to separate `Operation` from `apply(to:)`? If not, can we just absorb the former into `apply(to:)`?

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -280,12 +343,23 @@ extension Configuration.TestFilter {
   }
 }
 
+extension Configuration.TestFilter {
+  /// A type describing an error encountered while performing test filtering.
+  fileprivate enum Error: Swift.Error {
```
grynspan, line 348: I recently added `SystemError`, can we use that? Or, perhaps this is a reason to coalesce our several error types into a single `TestingError` with a few different cases.

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: We still use `Test.ID.Selection` later at the point where we apply the filter, so I believe the perf characteristics of this are unchanged

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: Thanks! I expected `-require-explicit-sendable` to catch this, but it doesn't seem to have

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -190,7 +190,17 @@ extension Runner.Plan {
     // configuration. The action graph is not modified here: actions that lose
     // their corresponding tests are effectively filtered out by the call to
     // zip() near the end of the function.
-    testGraph = configuration.testFilter.apply(to: testGraph)
+    do {
+      testGraph = try configuration.testFilter.apply(to: testGraph)
+    } catch {
+      // FIXME: Handle this more gracefully, either by propagating the error
```
grynspan, line 196: We should get these FIXMEs sorted before merging.

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: You mean e.g. inline `Operation` enum into this function? I considered that but felt it would make the function body quite a bit more complicated. As it stands we're using access control to scope it pretty narrowly

grynspan: We could still split up the implementation into multiple functions/steps, but keep the _interface_ simple.

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

PR #360, opened 2024-04-17
Provide an ABI-friendly entry point function that takes/produces JSON.

grynspan:
This PR introduces an **experimental** function, intended to eventually become part of our stable ABI, named `swt_copyABIEntryPoint_v0()`. This function can be looked up at runtime and used to hook into the testing library when a test target is loaded into the current process. It takes a JSON-encoded structure as input and yields JSON-encoded events/event contexts via a callback.

The input is encoded from a type we've tentatively named `__CommandLineArguments_v0`. This type conforms to `Codable` and represents the parsed command-line arguments that the package can understand. Normally, swift-testing will construct an instance of this type from the command-line arguments to the current process, but a caller can pass in a different instance to the ABI entry point instead.

The output consists of `Event.Snapshot` and `Event.Context.Snapshot` values encoded in a single JSON object. The format of this JSON is identical to what is emitted when `--experimental-event-stream-output` is specified at the command line.

`__swiftPMEntryPoint()` has also been modified to use the same `__CommandLineArguments_v0` type as input and as much of the same implementation as possible.

Because trying to load a Swift function dynamically at runtime is Hard™, `swt_copyABIEntryPoint_v0()` is a C function that yields the "real" Swift function. This indirection also gives us the flexibility to produce a thick Swift closure rather than a thin function.

> [!WARNING]
> The signature of structure of `swt_copyABIEntryPoint_v0`, the structure of `__CommandLineArguments_v0`, how it encodes to/from JSON, and all other details about it are still subject to change. The hope is that it becomes a stable ABI surface _eventually_, but we're not there yet. :)

### Checklist:


grynspan:
@swift-ci please test

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -10,22 +10,121 @@
...
+///     If `nil`, a new instance is created from the command-line arguments to
+///     the current process.
+///   - eventHandler: An event handler
+private func _commonEntryPoint(passing args: __CommandLineArguments_v0?, eventHandler: Event.Handler?) async -> CInt {
   let exitCode = Locked(rawValue: EXIT_SUCCESS)
 
   do {
-    let args = CommandLine.arguments()
-    if args.count == 2 && args[1] == "--list-tests" {
+    let args = try args ?? parseCommandLineArguments(from̉: CommandLine.arguments())
```
dennisweissmann, line 126:
```suggestion
    let args = try args ?? parseCommandLineArguments(from: CommandLine.arguments())
```

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -10,22 +10,121 @@
...
+///
+/// This function can be used by tools that do not link directly to the testing
+/// library and wish to invoke tests in a binary that has been loaded into the
+/// current process. The function is emitted into the binary under the name
+/// `"swt_copyABIEntryPoint_v0"` and can be dynamically looked up at runtime
+/// using `dlsym()` or a platform equivalent.
+///
+/// - Warning: This function's signature and the structure of its JSON inputs
+///   and outputs have not been finalized yet.
+@_cdecl("swt_copyABIEntryPoint_v0")
```
dennisweissmann, line 57: This function needs to be surrounded by `#if SWIFT_PACKAGE`

grynspan:
Oh, right, `dlsym()` doesn't work that way on Linux…

grynspan:
@swift-ci please test

Sources/Testing/Running/EntryPoint.swift
grynspan: We don't support building swift-testing using other build systems like CMake, so I think that would be redundant. We can chat more later today when you get in.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please clean test

grynspan:
@swift-ci please clean test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
Oops, somehow pushed an outdated copy of Allocator.h.

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -10,22 +10,120 @@
...
+///
+/// - Returns: The result of invoking the testing library. The type of this
+///   value is subject to change.
+///
+/// This function examines the command-line arguments to the current process
+/// and then invokes available tests in the current process. The
+///
+/// - Warning: This function's signature and the structure of its JSON inputs
+///   and outputs have not been finalized yet.
+@_spi(ForToolsIntegrationOnly)
```
dennisweissmann, line 33: The type alias needs to be outside the `!SWT_NO_ABI_ENTRY_POINT` check, right?

grynspan: This was intentional—if there's no ABI entry point, then there's no defined type for the ABI entry point either. :) Presumably a caller looking to use this symbol will configure their copy of swift-testing so that `SWT_NO_ABI_ENTRY_POINT` is not set. Does that make sense?

dennisweissmann (approved):

Sources/Testing/Running/EntryPoint.swift
dennisweissmann: Yes, sorry, I was confusing the conditional (again)...

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -10,22 +10,120 @@
...
+///     created from the command-line arguments to the current process.
+///   - eventHandler: An event handler to which is passed a buffer to memory
+///     representing each event and its context, as with ``Event/Handler``, but
+///     encoded as JSON.
+///
+/// - Returns: The result of invoking the testing library. The type of this
+///   value is subject to change.
+///
+/// This function examines the command-line arguments to the current process
+/// and then invokes available tests in the current process. The
```
stmontgomery, line 29: Incomplete sentence

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -10,22 +10,120 @@
...
+/// The type of the entry point to the testing library used by tools that want
+/// to remain version-agnostic regarding the testing library.
+///
+/// - Parameters:
+///   - argumentsJSON: A buffer to memory representing the JSON encoding of an
+///     instance of `__CommandLineArguments_v0`. If `nil`, a new instance is
+///     created from the command-line arguments to the current process.
+///   - eventHandler: An event handler to which is passed a buffer to memory
+///     representing each event and its context, as with ``Event/Handler``, but
+///     encoded as JSON.
```
stmontgomery, line 23: Should this comment be amended given the recent change to only pass a single argument to the callback?

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -126,37 +214,133 @@ func listTestsForSwiftPM(_ tests: some Sequence<Test>) -> [String] {
...
-/// - Throws: If an argument is invalid, such as a malformed regular expression.
+/// - Parameters:
+///   - args: The command-line arguments to interpret.
 ///
 /// This function generally assumes that Swift Package Manager has already
 /// validated the passed arguments.
-func configurationForSwiftPMEntryPoint(withArguments args: [String]) throws -> Configuration {
-  var configuration = Configuration()
+@_spi(ForToolsIntegrationOnly)
+public func parseCommandLineArguments(from args: [String]) throws -> __CommandLineArguments_v0 {
```
stmontgomery, line 269: Do we actually need this to be `@_spi public` or could it be lower?

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -315,42 +496,69 @@ private func _eventHandlerForStreamingEvents(toFileAtPath path: String) throws -
...
+///
+/// The resulting event handler outputs data as JSON. For each event handled by
+/// the resulting event handler, a JSON object representing it and its
+/// associated context is created and is passed to `eventHandler`. These JSON
+/// objects are guaranteed not to contain any ASCII newline characters (`"\r"`
+/// or `"\n"`).
+///
+/// Note that `_eventHandlerForStreamingEvents(toFileAtPath:)` calls this
+/// function and performs additional postprocessing before writing JSON data.
+private func _eventHandlerForStreamingEvents(
```
stmontgomery, line 551: Is this function coupled to the specific ABI signature, and if so, should it be suffixed with "_v0" to match other things which are?

grynspan: This one's private, fortunately, so the name doesn't need to represent future functionality. In a hypothetical future where we introduce a "v1" variant of everything, we'd probably rename this one "_v0" at that point, yeah.

I can rename it now without harm, of course, just wasn't critical because nothing outside the file can see it. :)

Sources/Testing/Running/EntryPoint.swift
grynspan: I tried to phrase the comment correctly for a single argument, i.e. "**a** buffer to memory representing each event and its context" but I can rephrase it if you have a suggestion.

Sources/Testing/Running/EntryPoint.swift
grynspan: Weird.

Sources/TestingInternals/include/Allocator.h
```diff
@@ -0,0 +1,43 @@
...
+/// A type that acts as a C++ [Allocator](https://en.cppreference.com/w/cpp/named_req/Allocator)
+/// without using global `operator new` or `operator delete`.
+///
+/// This type is necessary because global `operator new` and `operator delete`
+/// can be overridden in developer-supplied code and cause deadlocks or crashes
+/// when subsequently used while holding a dyld- or libobjc-owned lock. Using
+/// `std::malloc()` and `std::free()` allows the use of C++ container types
+/// without this risk.
+template<typename T>
+struct SWTHeapAllocator {
```
grynspan, line 30: Not new, just moved.

Sources/TestingInternals/Discovery.cpp
```diff
@@ -177,27 +179,6 @@ static void enumerateTypeMetadataSections(const SectionEnumerator& body) {
...
-/// A type that acts as a C++ [Allocator](https://en.cppreference.com/w/cpp/named_req/Allocator)
-/// without using global `operator new` or `operator delete`.
-///
-/// This type is necessary because global `operator new` and `operator delete`
-/// can be overridden in developer-supplied code and cause deadlocks or crashes
-/// when subsequently used while holding a dyld- or libobjc-owned lock. Using
-/// `std::malloc()` and `std::free()` allows the use of C++ container types
-/// without this risk.
-template<typename T>
-struct SWTHeapAllocator {
```
grynspan, line 189: Not deleted, just hoisted to a header.

Sources/Testing/Running/EntryPoint.swift
stmontgomery: Yeah, I realized it was private, just thought it might help us understand/remember/reinforce the fact that it is currently v0-specific if it had that suffix now

stmontgomery (approved):

Sources/Testing/Running/EntryPoint.swift
stmontgomery: Looks like you did this, thanks 👍

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test macOS

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test macOS

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

PR #338, opened 2024-04-09
Adopt swift-syntax-600.

grynspan:
This PR changes our swift-syntax dependency from swift-syntax-510 to swift-syntax-600.

We rely on several upcoming Swift 6 features, including swift-syntax's `lexicalContext` property, in order to correctly build and run tests. This PR migrates us to the latest prerelease swift-syntax-600 package so that we can take advantage of said feature.

swift-testing clients that have an explicit or inherited dependency on swift-syntax-510 or earlier may fail to build as a result of this change. As a reminder, swift-testing is an experimental package and we do not guarantee compatibility with older versions of the toolchain or related packages.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #361, opened 2024-04-18
Block tests/suites in extensions to syntactic sugar types (`[T]`, `[T:U]`, `T?`)

grynspan:
This PR blocks tests and suites in extensions to types declared using syntactic sugar. For example, the following is invalid:

```swift
extension [Int] {
  @Test func f() {}
}
```

I doubt this is going to be a common pattern, but it's just as invalid as saying `extension Array<Int>` for the same reasons: if it successfully compiles, it produces metadata for a test that we cannot instantiate at runtime (because the type we can see in the metadata is still generic, not specialized.)

There is a guard symbol emitted by our macro already that triggers a diagnostic if a test or suite is in a generic type, but `extension [T]` is fully specialized and does not trigger the diagnostic, so it's important that we catch this issue and make sure some diagnostic is emitted.

With this change, when using swift-syntax-600, you should now see:

```swift
extension [Int] {
  @Test func f() {} // 🛑 Attribute 'Test' cannot be applied to a function
                    // within a generic extension to type '[Int]'
}
```

This PR also introduces an environment variable you can set when running `swift test` to opt into a different swift-syntax version, which makes testing with swift-syntax-600 easier. This diagnostic will be removed as redundant if/when #338 is merged.

Finally, this PR fixes some unrelated unit tests that fail when using swift-syntax-600.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -33,7 +33,7 @@ let package = Package(
   ],
 
   dependencies: [
-    .package(url: "https://github.com/apple/swift-syntax.git", from: "510.0.1"),
+    .package(url: "https://github.com/apple/swift-syntax.git", from: Version(stringLiteral: Context.environment["SWT_SWIFT_SYNTAX_VERSION"] ?? "510.0.1")),
   ],
 
   targets: [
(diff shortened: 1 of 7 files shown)


grynspan:
@swift-ci please test

briancroom (approved):
Very nice, thank you!

PR #364, opened 2024-04-19
Unlock expression expansions for negated expressions.

grynspan:
This PR enables expression expansion for expressions of the forms `!x`, `!f()`, `!y.g()`, etc. such that the value of the expressions they negate are available at runtime to inspect in a test failure. For example, consider this test:

```swift
func f() -> Int { 123 }
func g() -> Int { 123 }

@Test func myTest() {
  #expect(!(f() == g())) // ♦️ Expectation failed: !(f() == g())
}
```

(i.e., testing that the result of `f()` is not equal to the result of `g()`, but using `!(==)` instead of the more typical `!=`.)

This is a trivial/contrived example where we can tell at a glance that the values equal `123`, of course. Still, previously this test would (correctly) fail, but would not be able to provide the return values of `f()` and `g()`. With this change, the following test issue is recorded:

```swift
 #expect(!(f() == g())) // ♦️ Expectation failed: !((f() → 123) == (g() → 123) → true)
```

This increasing the amount of diagnostic information available.

This PR also makes adjustments to the `Expression` type to improve the information we log during test runs. In particular, when `--verbose` is passed to `swift test`, we will capture type information about more parts of a failed expectation/expression.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/Expression.swift
+++ b/Sources/Testing/SourceAttribution/Expression.swift
@@ -78,6 +78,18 @@ public struct Expression: Sendable {
     ///   - keyPath: The key path, relative to `value`, that was accessed, not
     ///     including a leading backslash or period.
     indirect case propertyAccess(value: Expression, keyPath: Expression)
+
+    /// The expression negates another expression.
+    ///
+    /// - Parameters:
+    ///   - expression: The expression that was negated.
+    ///   - isParenthetical: Whether or not `expression` was enclosed in
+    ///     parentheses (and the `!` operator was outside it.) This argument
+    ///     affects how this expression is represented as a string.
+    ///
+    /// Unlike other cases in this enumeration, this case affects the runtime
+    /// behavior of the `__check()` family of functions.
+    indirect case negation(_ expression: Expression, isParenthetical: Bool)
   }
 
   /// The kind of syntax node represented by this instance.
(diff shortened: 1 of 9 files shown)


briancroom (approved):
Very nice, this is a great enhancement.

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -237,58 +258,129 @@ public struct Expression: Sendable {
...
+  func expandedDebugDescription() -> String {
+    var context = _ExpandedDescriptionContext()
+    context.includeTypeNames = true
+    context.includeParenthesesIfNeeded = false
+    return _expandedDescription(in: context)
+  }
+
+  /// A structure describing the state tracked while calling
+  /// `_expandedDescription(in:)`.
+  private struct _ExpandedDescriptionContext {
```
stmontgomery, line 298: `Sendable`?

grynspan: No need to as it never gets sent anywhere.

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -237,58 +258,129 @@ public struct Expression: Sendable {
...
+      result.kind = .negation(
+        expression.capturingRuntimeValues(firstValue, repeat each additionalValues),
+        isParenthetical: isParenthetical
+      )
     }
 
     return result
   }
 
+
```
stmontgomery, line 271: nit: extra line here

stmontgomery (approved):
Nice, thanks!

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

PR #365, opened 2024-04-19
Improve type-checking time of `listTestsForSwiftPM`

xedin:
Type-checking of the chained return expression took over 10 secs.

### DIFF
--- a/Sources/Testing/Running/EntryPoint.swift
+++ b/Sources/Testing/Running/EntryPoint.swift
@@ -107,12 +107,14 @@ func listTestsForSwiftPM(_ tests: some Sequence<Test>) -> [String] {
   // Group tests by the name components of the tests' IDs. If the name
   // components of two tests' IDs are ambiguous, present their source locations
   // to disambiguate.
-  return Dictionary(
+  let initialGroups = Dictionary(
     grouping: tests.lazy.map(\.id),
     by: \.nameComponents
   ).values.lazy
     .map { ($0, isAmbiguous: $0.count > 1) }
-    .flatMap { testIDs, isAmbiguous in
+
+  // This operation is split to improve type-checking performance.
+  return initialGroups.flatMap { testIDs, isAmbiguous in
       testIDs.lazy
         .map { testID in
           if !isAmbiguous, testID.sourceLocation != nil {


stmontgomery:
@swift-ci please test

stmontgomery (approved):
Thank you!

grynspan:
Works for me. Thanks!

Sources/Testing/Running/EntryPoint.swift
```diff
@@ -107,12 +107,14 @@ func listTestsForSwiftPM(_ tests: some Sequence<Test>) -> [String] {
...
-  return Dictionary(
+  let initialGroups = Dictionary(
     grouping: tests.lazy.map(\.id),
     by: \.nameComponents
   ).values.lazy
     .map { ($0, isAmbiguous: $0.count > 1) }
-    .flatMap { testIDs, isAmbiguous in
+
+  // This operation is split to improve type-checking performance.
+  return initialGroups.flatMap { testIDs, isAmbiguous in
```
grynspan, line 117:
```suggestion
  return initialGroups.lazy
    .flatMap { testIDs, isAmbiguous in
```

Unless that would regress performance again. I'm trying to avoid creating intermediate array allocations.

xedin: Let me check

xedin: The time goes up from 0.6 sec. up to `1.4` secs. for that file with `lazy` added there.

grynspan: What happens if we remove the original `lazy`?

xedin: Only if I remove all `.lazy` it's comparably fast to the changes I have in the PR.

grynspan: Sorry, can you clarify? With your PR, there's still a `.lazy` there. I'm asking what it looks like if we remove that one too.

xedin: What I mean is unless we split this expression the way I did it would still be too slow unless all `.lazy` calls are removed.

grynspan: Okay, I'm suggesting changing this line above:

```swift
  ).values.lazy
```

To:

```swift
  ).values
```

And measuring the difference. Would that also make it any faster?

xedin: It doesn't make any difference. Adding `.lazy` in the change you suggested does it make slower.

grynspan: Okay! Ship it.

grynspan (approved):

PR #366, opened 2024-04-19
Lazily evaluate test arguments only after determining their test will run, and support throwing expressions

### LINKED ISSUE #166: Support throwing errors from within `@Test` attributes
### Description

The argument parameter of `Test(_:_:arguments:)` is currently very limited.

Adding a version of `Test(_:_:arguments: () async throws -> C)` would greatly improve the parameterisation possibilities (e.g. async loading of parameters from a file).

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

_No response_

stmontgomery:
This changes test declaration, discovery, planning, and running such that arguments to parameterized test functions are evaluated lazily, only after determining their test will run. It also adds support for throwing expressions in test arguments.

### Motivation:

When defining parameterized test functions, it can be useful to call throwing or asynchronous APIs — for example, to fetch the arguments from an external source. Today, the `@Test` macro supports `async` expressions but not `throws`. More problematic still: the arguments of _all_ tests get evaluated, even for tests which don't actually run (which may happen for various reasons). Evaluating the arguments of a test which won't run wastes time and resources, so we should try to avoid this, and support `throws` expressions too for better flexibility.

### Modifications:

- Modify the `@Test` macro and supporting library interfaces to surround test arguments in closures, so their evaluation can be lazy and deferred.
- Modify how test case arguments are stored on each `Test` instance accordingly.
- Modify the planning logic in `Runner.Plan` to defer evaluation of test cases until we determine whether each test will run.
- Update tests.

### Result:

Now, arguments to a parameterized test will only be evaluated if that test is actually going to run, and the expressions may include `try` and potentially throw an `Error`.

### Checklist:

Resolves #166
Resolves rdar://121531170

### DIFF
--- a/Sources/Testing/Test.swift
+++ b/Sources/Testing/Test.swift
@@ -221,7 +221,7 @@ public struct Test: Sendable {
     self.sourceLocation = sourceLocation
     self.containingTypeInfo = containingTypeInfo
     self.xcTestCompatibleSelector = xcTestCompatibleSelector
-    self.testCaseState = .unevaluated({ .init(try await testCases()) })
+    self.testCaseState = .unevaluated { .init(try await testCases()) }
     self.parameters = parameters
   }
 }
@@ -250,7 +250,7 @@ extension Test {
       case name
       case displayName
       case sourceLocation
-      case testCaseState
+      case testCases
       case parameters
       case comments
       case tags
@@ -302,11 +302,13 @@ extension Test {
       }
     }
 
-    /// The evaluation state of this test's cases, if any.
+    /// The set of test cases associated with this test, if any.
     ///
-    /// If this test represents a suite type, the value of this property is
-    /// `nil`.
-    public var testCaseState: TestCaseState?
+    /// If the ``Test`` this instance was snapshotted from represented a
+    /// parameterized test function but its test cases had not yet been
+    /// evaluated when the snapshot was taken, or the evaluation attempt failed,
+    /// the value of this property will be an empty array.
+    public var testCases: [Test.Case.Snapshot]?
 
     /// The test function parameters, if any.
     ///
@@ -345,13 +347,22 @@ extension Test {
       name = test.name
       displayName = test.displayName
       sourceLocation = test.sourceLocation
-      testCaseState = test.testCaseState.map(TestCaseState.init(snapshotting:))
       parameters = test.parameters
       comments = test.comments
       tags = test.tags
       associatedBugs = test.associatedBugs
       if #available(_clockAPI, *) {
         _timeLimit = test.timeLimit.map(TimeValue.init)
+      }
+
+      testCases = switch test.testCaseState {
+      case .unevaluated,
+           .evaluated(.failure):
+        []
+      case let .evaluated(.success(testCases)):
+        testCases.rawValue.map(Test.Case.Snapshot.init(snapshotting:))
+      case nil:
+        nil
       }
     }
 
@@ -370,69 +381,7 @@ extension Test {
     ///
     /// - ``Test/isSuite``
     public var isSuite: Bool {
-      testCaseState == nil
+      testCases == nil
     }
-  }
-}
-
-extension Test.Snapshot.TestCaseState: Codable {
-  /// A simplified version of ``TestCaseState`` suitable for encoding and
-  /// decoding.
-  private enum _EncodedForm: Sendable, Codable {
-    /// The test's cases have not yet been evaluated.
-    case unevaluated
-
-    /// A representation of `Swift.Result` suitable for encoding and decoding.
-    enum Result: Sendable, Codable {
-      /// A successful evaluation of a test's cases.
-      ///
-      /// - Parameters:
-      ///   - testCases: The returned set of test cases.
-      case success(_ testCases: [Test.Case.Snapshot])
-
-      /// A failed evaluation of a test's cases.
-      ///
-      /// - Parameters:
-      ///   - error: A snapshot of the error caught when evaluating the test's
-      ///     cases.
-      case failure(_ error: ErrorSnapshot)
-    }
-
-    /// The test's cases have been evaluated, and either returned a set of test
-    /// or failed by throwing an error.
-    ///
-    /// - Parameters:
-    ///   - result: The result of having evaluated the test's cases.
-    case evaluated(_ result: Result)
-  }
-
-  public init(from decoder: any Decoder) throws {
-    self = switch try _EncodedForm(from: decoder) {
-    case .unevaluated:
-      .unevaluated
-    case let .evaluated(result):
-      switch result {
-      case let .success(testCases):
-        .evaluated(.success(testCases))
-      case let .failure(error):
-        .evaluated(.failure(error))
-      }
-    }
-  }
-
-  public func encode(to encoder: any Encoder) throws {
-    let encodedForm: _EncodedForm = switch self {
-    case .unevaluated:
-      .unevaluated
-    case let .evaluated(result):
-      switch result {
-      case let .success(testCases):
-        .evaluated(.success(testCases))
-      case let .failure(error):
-        .evaluated(.failure(error))
-      }
-    }
-
-    try encodedForm.encode(to: encoder)
   }
 }
--- a/Sources/Testing/Test.swift
+++ b/Sources/Testing/Test.swift
@@ -76,7 +76,7 @@ public struct Test: Sendable {
   /// ([96960993](rdar://96960993)). It is also not possible to have a value of
   /// an underlying generic sequence type without specifying its generic
   /// parameters.
-  fileprivate enum TestCaseState: Sendable {
+  fileprivate enum TestCasesState: Sendable {
     /// The test's cases have not yet been evaluated.
     ///
     /// - Parameters:
@@ -95,7 +95,7 @@ public struct Test: Sendable {
   /// The evaluation state of this test's cases, if any.
   ///
   /// If this test represents a suite type, the value of this property is `nil`.
-  fileprivate var testCaseState: TestCaseState?
+  fileprivate var testCasesState: TestCasesState?
 
   /// The set of test cases associated with this test, if any.
   ///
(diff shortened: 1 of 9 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/Test.swift
```diff
@@ -154,7 +221,7 @@ public struct Test: Sendable {
     self.sourceLocation = sourceLocation
     self.containingTypeInfo = containingTypeInfo
     self.xcTestCompatibleSelector = xcTestCompatibleSelector
-    self._testCases = .init(rawValue: .init(testCases))
+    self.testCaseState = .unevaluated({ .init(try await testCases()) })
```
grynspan, line 224:
```suggestion
    self.testCaseState = .unevaluated { .init(try await testCases()) }
```

Sources/Testing/Test.swift
```diff
@@ -183,7 +250,7 @@ extension Test {
       case name
       case displayName
       case sourceLocation
-      case testCases
+      case testCaseState
```
grynspan, line 253: I think in order to encode a test snapshot, we need to have the full set of test cases evaluated already, in which case this can just be `testCases` again.

Regardless, the keys in the snapshot encoding should generally mirror the public (or SPI) property names, not their backing storage.

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -133,6 +133,19 @@ struct AttributeInfo {
...
 
+    // Any arguments of the test declaration macro which specify test arguments
+    // need to be wrapped a closure so they may be evaluated lazily by the
+    // testing library at runtime. If any such arguments are present, they will
+    // begin with a labeled argument named `arguments:` and include all
+    // subsequent unlabeled arguments.
+    if let argumentsIndex = otherArguments.firstIndex(where: { $0.label?.tokenKind == .identifier("arguments") }) {
+      for index in argumentsIndex ..< otherArguments.endIndex {
+        var argument = otherArguments[index]
+        argument.expression = .init(ClosureExprSyntax { argument.expression })
```
grynspan, line 144: This rewrites the arguments as detached nodes and will cause diagnostics we emit for them to fail to emit in the right location. Rewriting needs to be deferred to `functionArgumentList()`.

Tests/TestingTests/RunnerTests.swift
```diff
@@ -36,6 +36,25 @@ private let randomNumber = Int.random(in: 0 ..< .max)
...
+  fatalError("Should never crash, since this should never be called by any test")
+}
+
+/// A test whose arguments always cause a `fatalError()` crash.
+///
+/// This test should always remain `.hidden` and never be selected to run. Its
+/// purpose is to validate that the arguments of tests which are not run are
+/// never evaluated.
+@Test(.hidden, arguments: fatalArguments())
+func parameterizedWithFatalArguments(i: Int) {}
```
grynspan, line 56: Seems like an opportunity to write an exit test here!

Sources/Testing/Support/Graph.swift
```diff
@@ -676,6 +676,49 @@ extension Graph {
...
+  ///     The element representing that node is passed to this closure and its
+  ///     result is used as the corresponding value in the new graph.
+  ///
+  /// - Returns: A graph containing the transformed nodes of this graph at the
+  ///   same key paths.
+  ///
+  /// - Throws: Whatever is thrown by `transform`.
+  ///
+  /// This function iterates depth-first.
+  func mapValues<U>(_ transform: (Element) async throws -> U) async rethrows -> Graph<K, U> {
```
grynspan, line 694: It is intentional that we don't include the key path or key in the existing `mapValues()` definition (same as for `Dictionary`.) Can you elaborate why you need it here?

grynspan (changes requested):
See comments above. I don't know why GitHub makes me type this. :)

Sources/Testing/Support/Graph.swift
stmontgomery: I'm modifying all the mapValues APIs to accept key path in #372

Sources/TestingMacros/Support/AttributeDiscovery.swift
stmontgomery: Done!

stmontgomery:
@swift-ci please test

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -193,6 +180,22 @@ struct AttributeInfo {
         ArrayElementSyntax(expression: traitExpr)
       }
     }))
+
+    var otherArguments = self.otherArguments
+
```
grynspan, line 185: Stray blank line? Might be more concise to move the `var otherArguments` decl under the comment.

Sources/Testing/Test.swift
```diff
@@ -68,24 +68,91 @@ public struct Test: Sendable {
...
+  ///
   /// For parameterized tests, each test case is associated with a single
   /// combination of parameterized inputs. For non-parameterized tests, a single
   /// test case is synthesized. For test suite types (as opposed to test
   /// functions), the value of this property is `nil`.
-  @_spi(ForToolsIntegrationOnly)
-  public var testCases: (some Sequence<Test.Case> & Sendable)? {
-    _testCases?.rawValue
+  var testCases: (some Sequence<Test.Case> & Sendable)? {
+    switch testCaseState {
```
grynspan, line 111:
```swift
guard case let .evaluated(result) = testCaseState else {
  preconditionFailure(...)
}
guard case let .success(testCases) = result else {
  preconditionFailure(...)
}
return testCases.rawValue
```

Sources/Testing/Test.swift
```diff
@@ -207,12 +274,40 @@ extension Test {
     /// The source location of this test.
     public var sourceLocation: SourceLocation
 
+    /// An enumeration describing the evaluation state of a test's cases.
```
grynspan, line 277: Do we really need this in the snapshot?

Tests/TestingTests/MiscellaneousTests.swift
```diff
@@ -405,41 +409,49 @@ struct MiscellaneousTests {
...
-    let parameterizedTestFunction2TestCases = try #require(parameterizedTestFunction2.testCases)
-    #expect(parameterizedTestFunction2TestCases.underestimatedCount == 100 * 100)
-    let parameterizedTestFunction2Parameters = try #require(parameterizedTestFunction2.parameters)
-    #expect(parameterizedTestFunction2Parameters.count == 2)
-    let parameterizedTestFunction2FirstParameter = try #require(parameterizedTestFunction2Parameters.first)
-    #expect(parameterizedTestFunction2FirstParameter.firstName == "i")
-    let parameterizedTestFunction2SecondParameter = try #require(parameterizedTestFunction2Parameters.last)
-    #expect(parameterizedTestFunction2SecondParameter.firstName == "j")
-    #expect(parameterizedTestFunction2SecondParameter.secondName == "value")
+    do {
```
grynspan, line 412: Please double-check that we don't regress code coverage here. (I'm not saying we do, just being cautious.)

Tests/TestingTests/RunnerTests.swift
grynspan: I wasn't kidding—why not write one?

Tests/TestingTests/Test.Case.Argument.IDTests.swift
```diff
@@ -47,10 +49,11 @@ struct Test_Case_Argument_IDTests {
 
   @Test("One Identifiable parameter")
   func oneIdentifiableParameter() async throws {
-    let test = Test(
+    var test = Test(
       arguments: [MyIdentifiableArgument(id: "abc")],
       parameters: [Test.Parameter(index: 0, firstName: "value", type: MyIdentifiableArgument.self)]
     ) { _ in }
+    try await test.evaluateTestCases()
```
grynspan, line 56: Should we just have the fixture `Test.init` call this? It would make the overall change more concise.

grynspan (approved):

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,6 +180,22 @@ struct AttributeInfo {
...
+    // Any arguments of the test declaration macro which specify test arguments
+    // need to be wrapped a closure so they may be evaluated lazily by the
+    // testing library at runtime. If any such arguments are present, they will
+    // begin with a labeled argument named `arguments:` and include all
+    // subsequent unlabeled arguments.
+    if let argumentsIndex = otherArguments.firstIndex(where: { $0.label?.tokenKind == .identifier("arguments") }) {
+      for index in argumentsIndex ..< otherArguments.endIndex {
```
grynspan, line 192: I think you could do:

```swift
arguments[argumentsIndex...] = arguments[argumentsIndex...].map { argument in
  ...
}
```

Sources/Testing/Test.swift
stmontgomery: Meant to remove this - done!

Tests/TestingTests/Test.Case.Argument.IDTests.swift
stmontgomery: I have re-organized things a bit so that we no longer need to call this on individual fixture test instances

stmontgomery:
@swift-ci please test

Sources/Testing/Test.swift
```diff
@@ -68,24 +68,91 @@ public struct Test: Sendable {
...
   /// For parameterized tests, each test case is associated with a single
   /// combination of parameterized inputs. For non-parameterized tests, a single
   /// test case is synthesized. For test suite types (as opposed to test
   /// functions), the value of this property is `nil`.
-  @_spi(ForToolsIntegrationOnly)
-  public var testCases: (some Sequence<Test.Case> & Sendable)? {
-    _testCases?.rawValue
+  var testCases: (some Sequence<Test.Case> & Sendable)? {
+    guard let testCasesState else {
+      return nil as AnySequence<Test.Case>?
```
grynspan, line 112: Is the `as` clause necessary here?

grynspan (approved):

Sources/Testing/Test.swift
stmontgomery: Compiler can't infer the `some` result type without it, since this is the first `return` statement. (This technique was actually needed in the previous commit in this PR too, just in a different place)

Sources/TestingMacros/Support/AttributeDiscovery.swift
stmontgomery: Definitely open to this but I can't get this suggestion to compile, even if I drop `.lazy` or surround it with `Array(...)`

Sources/Testing/Test.swift
grynspan: Boooo okay.

Sources/TestingMacros/Support/AttributeDiscovery.swift
grynspan: It's not critical, just a bit closer to our house style of _use all the monads!_. :) No worries.

PR #367, opened 2024-04-19
Fix thread name length in exit tests' call to `pthread_setname_np()`.

grynspan:
On Linux, the maximum length of a thread's name is defined by `TASK_COMM_LEN` to be 16 bytes (including the trailing null character.) We currently manually create one thread on Linux in order to monitor for termination of exit tests' child processes. This PR changes the name of the monitor thread on Linux to one that is less than 16 characters long.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

briancroom (approved):

PR #381, opened 2024-04-29
Work around miscompilation of entry point function on Windows with Swift 5.10.

grynspan:
This PR works around two apparent miscompiles of the entry point function added with #360. These workarounds are only necessary on Windows when using the Swift 5.10 compiler (I've seen no indication of issues on other platforms or with the Swift 6 compiler.) We can revert this change once we no longer provide any support for the Swift 5.10 toolchain.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #383, opened 2024-04-29
Define and implement a strawman stable JSON schema for output.

### LINKED ISSUE #368: Define and implement a stable JSON schema for output.
### Description

Experimental event streaming support was added in #286. Issues are reported as individual events with richly structured JSON that describe the failure. This is great for tools that need to work with and transform raw issues, but many tools will just want to use the human readable failure string, i.e: `Expectation failed: 1 == 2`. 

Swift-testing has done a lot of work to build up a nice message, and it would be nice to avoid reimplementing this in other tools (and keeping it up to date).

Feature: Add a new `message` field to the `issueRecorded` event that provides the human readable failure string. Bonus points if there is a markdown version of the string, as many tools can render markdown directly.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

_No response_

grynspan:
This PR defines a JSON schema for events output from swift-testing via either `--experimental-event-stream-output` or `abiEntryPoint_v0()`. The JSON schema needs to be formally reviewed separately.

We are already writing an experimental JSON stream using "snapshot" types and we don't want to break folks experimenting with it before we have a final JSON schema, so for now that remains the default. To opt into the new schema, pass `--experimental-event-stream-version 0` (note that this argument is not forwarded from `swift test`. See https://github.com/apple/swift-package-manager/pull/7534.)

Resolves #368.

### Checklist:

### DIFF
--- a/Documentation/ABI/JSON.md
+++ b/Documentation/ABI/JSON.md
@@ -1,0 +1,222 @@
+# JSON schema
+
+<!--
+This source file is part of the Swift.org open source project
+
+Copyright (c) 2024 Apple Inc. and the Swift project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+This document outlines the JSON schemas used by the testing library for its ABI
+entry point and for the `--experimental-event-stream-output` command-line
+argument. For more information about the ABI entry point, see the documentation
+for [ABI.EntryPoint_v0](https://github.com/search?q=repo%3Aapple%2Fswift-testing%EntryPoint_v0&type=code).
+
+> [!WARNING]
+> This JSON schema is still being developed and is subject to any and all
+> changes including removal from the package.
+
+## Modified Backus-Naur form
+
+This schema is expressed using a modified Backus-Naur syntax. `{`, `}`, `:`, and
+`,` represent their corresponding JSON tokens. `\n` represents an ASCII newline
+character.
+
+The order of keys in JSON objects is not normative. Whitespace in this schema is
+not normative; it is present to help the reader understand the content of the
+various JSON objects in the schema. The event stream is output using the JSON
+Lines format and does not include newline characters (except **one** at the end
+of the `<output-record-line>` rule.)
+
+Trailing commas in JSON objects and arrays are only to be included where
+syntactically valid.
+
+### Common data types
+
+`<string>` and `<number>` are defined as in JSON. `<array:T>` represents an
+array (also defined as in JSON) whose elements all follow rule `<T>`.
+
+```
+<bool> ::= true | false ; as in JSON
+
+<source-location> ::= {
+  "fileID": <string>, ; the Swift file ID of the file
+  "line": <number>,
+  "column": <number>,
+}
+
+<version> ::= "version": 0 ; will be incremented as the format changes
+
+<boolean> ::= true | false ; boolean value as in JSON
+```
+
+<!--
+TODO: implement input/configuration
+
+### Configuration
+
+A single configuration is passed into the testing library prior to running any
+tests and, as the name suggests, configures the test run. The configuration is
+encoded as a single [JSON Lines](https://jsonlines.org) value.
+
+```
+<configuration-record> ::= {
+  <version>,
+  "kind": "configuration",
+  "payload": <configuration>
+}
+
+<configuration> ::= {
+  ["verbosity": <number>,] ; 0 is the default; higher means more verbose output
+                           ; while negative values mean quieter output.
+  ["filters": <array:test-filter>,] ; how to filter the tests in the test run
+  ["parallel": <bool>,] ; whether to enable parallel testing (on by default)
+  ; more TBD
+}
+
+<test-filter> ::= <test-filter-tag> | <test-filter-id>
+
+<test-filter-action> ::= "include" | "exclude"
+
+<test-filter-tag> ::= {
+  "action": <test-filter-action>,
+  "tags": <array:string>, ; the names of tags to include
+  "operator": <test-filter-tag-operator> ; how to combine the values in "tags"
+}
+
+<test-filter-tag-operator> ::= "any" | "all"
+
+<test-filter-id> ::= {
+  "action": <test-filter-action>,
+  "id": <test-id> ; the ID of the test to filter in/out
+}
+```
+-->
+
+### Streams
+
+A stream consists of a sequence of values encoded as [JSON Lines](https://jsonlines.org).
+A single instance of `<output-stream>` is defined per test process and can be
+accessed by passing `--experimental-event-stream-output` to the test executable
+created by `swift build --build-tests`.
+
+```
+<output-stream> ::= <output-record>\n | <output-record>\n <output-stream>
+```
+
+### Records
+
+Records represent the values produced on a stream. Each record is encoded on a
+single line and can be decoded independently of other lines. If a decoder
+encounters a record whose `"kind"` field is unrecognized, the decoder should
+ignore that line.
+
+```
+<output-record> ::= <metadata-record> | <test-record> | <event-record>
+
+<metadata-record> ::= {
+  <version>,
+  "kind": "metadata",
+  "payload": <metadata>
+}
+
+<test-record> ::= {
+  <version>,
+  "kind": "test",
+  "payload": <test>
+}
+
+<event-record> ::= {
+  <version>,
+  "kind": "event",
+  "payload": <event>
+}
+```
+
+### Metadata
+
+Metadata records are reserved for future use.
+
+```
+<metadata> ::= {
+  ; unspecified JSON object content
+}
+```
+
+### Tests
+
+Test records represent individual test functions and test suites. Test records
+are passed through the record stream **before** most events.
+
+<!--
+If a test record represents a parameterized test function whose inputs are
+enumerable and can be independently replayed, the test record will include an
+additional `"testCases"` field describing the individual test cases.
+--> 
+
+```
+<test> ::= {
+  "kind": <test-kind>,
+  "name": <string>, ; the unformatted function or non-qualified type name
+  ["displayName": <string>,] ; the user-supplied custom display name
+  "sourceLocation": <source-location>, ; where the test is defined
+  "id": <test-id>,
+}
+
+<test-kind> ::= "suite" | "function" | "parameterizedFunction"
+
+<test-id> ::= <string> ; an opaque string representing the test case
+```
+
+<!--
+  TODO: define a round-trippable format for a test case ID
+  ["testCases": <array:test-case>] ; if kind is "parameterizedFunction" and
+                                     ; the inputs are enumerable, all test case
+                                     ; IDs, otherwise not present
+
+<test-case> ::= {
+  "id": <string>, ; an opaque string representing the test case
+  "displayName": <string> ; a string representing the corresponding Swift value
+}
+```
+-->
+
+### Events
+
+Event records represent things that can happen during testing. They include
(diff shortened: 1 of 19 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -27,6 +27,11 @@ extension Event {
 
       /// The human-readable message.
       var stringValue: String
+
+      /// A concise version of ``stringValue``, if available.
+      ///
+      /// Not all messages include a concise string.
+      var conciseStringValue: String?
```
plemarquand, line 34: How is this intended to be used?

plemarquand (approved):

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: This is a shorter version of the message's string. The longer version is output to `stderr` by default. The shorter version is what gets passed back in the JSON.

Sources/Testing/Events/Event.swift
```diff
@@ -15,8 +15,11 @@ public struct Event: Sendable {
   public enum Kind: Sendable {
     /// A test run started.
     ///
+    /// - Parameters:
+    ///   - plan: The test plan of the run that started.
+    ///
     /// This event is the first event posted after ``Runner/run()`` is called.
-    case runStarted
+    indirect case runStarted(_ plan: Runner.Plan)
```
stmontgomery, line 22: I suspect this is added because it's used elsewhere, but isn't this quite a large payload potentially? I wonder about the impact of that

But also (ironically): if we do add this here, for parity should we also be adding it to the corresponding _run ended_ event?

Sources/Testing/EntryPoints/ABIv0/ABIv0.swift
```diff
@@ -0,0 +1,12 @@
...
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// A namespace for ABI version 0 symbols.
+enum ABIv0: Sendable {}
```
stmontgomery, line 12: Not for right now, but if/when we expose this outside the module I'd like to bikeshed alternative naming conventions

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedEvent.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// to the user.
+    var messages: [EncodedMessage]
+
+    /// The ID of the test associated with this event, if any.
+    var testID: EncodedTest.ID?
+
+    /// The ID of the test case associated with this event, if any.
+    ///
+    /// - Warning: Test cases are not yet part of the JSON schema.
+    var _testCase: EncodedTestCase?
```
stmontgomery, line 60: I know this is non-final indicated by the underscore, and that's good. I think we ought to move in the direction of this being an ID of a test case, rather than encoding the full test case itself inline.

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+            .pass
+          }
+        case .fail:
+            .fail
+        case .difference:
+            .difference
+        case .warning:
+            .warning
+        case .details:
+            .details
```
stmontgomery, line 53: nit: indentation in this init

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedMessage.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    /// The symbol associated with this message.
+    var symbol: Symbol
+
+    /// The human-readable, unformatted text associated with this message.
+    var text: String
+
+    /// The human-readable, Markdown-formatted text associated with this
+    /// message, if any.
+    var markdown: String?
```
stmontgomery, line 66: is this speculative right now? Actually needed?

Sources/Testing/Events/Event.swift
grynspan: It's pretty chunky, yes, but remember that the value is CoW, so not as large as you might think. I'm open to alternative designs.

I left it out of the `runEnded` case precisely because it was relatively large.

Sources/Testing/EntryPoints/ABIv0/ABIv0.swift
grynspan: This enum is definitely not ever meant to be _API_. It's a namespace for ABI-version-0 symbols only.

I'm open to alternatives.

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedEvent.swift
grynspan: We don't encode the entire test case structure here, no.

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedMessage.swift
grynspan: Will fix.

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedMessage.swift
grynspan: Speculative at the request of @plemarquand.

grynspan: I'll remove it for now. We can add it later if we actually decide to implement it.

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedEvent.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    /// The source location of the event, if applicable.
+    var sourceLocation: SourceLocation?
+
+    /// The instant at which the event occurred on the current system's
+    /// suspending clock.
+    var timestamp: Double
+
+    /// The instant at which the event occurred on the wall clock.
+    var timestampSince1970: Double
```
stmontgomery, line 48: These two timestamp properties feel like they could be grouped together into some kind of single type, maybe `EncodedInstant` ?

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedEvent.swift
```diff
@@ -0,0 +1,103 @@
...
+    /// Note that the set of encodable events is a subset of all events
+    /// generated at runtime by the testing library.
+    ///
+    /// For descriptions of individual cases, see ``Event/Kind``.
+    enum Kind: String, Sendable {
+      case runStarted
+      case testStarted
+      case testCaseStarted
+      case issueRecorded
+      case knownIssueRecorded
```
stmontgomery, line 30: This `knownIssueRecorded` case is a little odd to me because the underlying types represent an `Issue` in general, which can either be non-known or known. And then separately, there is an Issue.Kind representing "_no known issue was recorded_", which is conceptually different. Is this representation confusing those?

grynspan: This is a different abstraction than what we have in Swift, although we could approach it in other ways. I didn't build out a complete "issue" type in the schema because at the end of the day, handling an issue is a lot like handling any other event. `knownIssueRecorded` is distinct from `issueRecorded` because they're just kinds of event.

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedEvent.swift
grynspan: Maybe. I originally just had `timestamp` as a single numeric value, but added the wall clock value later. @plemarquand What do you think?

plemarquand: I think wrapping them up in an `instant` field would make sense. 

If timestamp is a https://developer.apple.com/documentation/swift/suspendingclock maybe we could link to those docs to make it more clear when clock suspension might happen.

Sources/Testing/EntryPoints/ABIv0/Encoded/ABIv0.EncodedEvent.swift
grynspan: Refactored.

dennisweissmann (approved):

Documentation/ABI/JSON.md
```diff
@@ -0,0 +1,227 @@
...
+```
+<test> ::= {
+  "kind": <test-kind>,
+  "name": <string>, ; the unformatted function or non-qualified type name
+  ["displayName": <string>,] ; the user-supplied custom display name
+  "sourceLocation": <source-location>, ; where the test is defined
+  "id": <test-id>,
+}
+
+<test-kind> ::= "suite" | "function" | "parameterizedFunction"
```
stmontgomery, line 172: I'm curious about rationale for parameterized tests being considered a distinct "kind"? Alternatively I could imagine them being represented as `test-kind ::= "function"`, with a distinct, optional Boolean property like `parameterized ::= true | false`

Documentation/ABI/JSON.md
```diff
@@ -0,0 +1,227 @@
...
+```
+<bool> ::= true | false ; as in JSON
+
+<source-location> ::= {
+  "fileID": <string>, ; the Swift file ID of the file
+  "line": <number>,
+  "column": <number>,
+}
+
+<timestamp> ::= {
```
stmontgomery, line 51: Is this `<timestamp>` used elsewhere in the spec?

Documentation/ABI/JSON.md
```diff
@@ -0,0 +1,227 @@
...
+<event-record> ::= {
+  <version>,
+  "kind": "event",
+  "payload": <event>
+}
+```
+
+### Metadata
+
+Metadata records are reserved for future use.
```
stmontgomery, line 144: This one feels quite speculative; if we need additional kinds of records, it seems like we can define them and extend this schema to describe them at that time, right? Rather than needing to explicitly reserve this now.

Documentation/ABI/JSON.md
grynspan: Should be `<instant>` now.

Documentation/ABI/JSON.md
grynspan: Will delete for now. I think we had the idea that we wanted to front-load some additional data, but we can amend this schema with that if/when we implement it.

Documentation/ABI/JSON.md
grynspan: No real rationale beyond "this is what I typed." If you feel strongly, I can refactor it.

grynspan: Refactored.

stmontgomery (approved):

grynspan:
@swift-ci please test

PR #384, opened 2024-04-29
Move `__swiftPMEntryPoint()` to its own file.

grynspan:
Tidying up the entry point code and moving the SwiftPM-specific entry points to their own file. Renames internal functions to not mention SwiftPM if they also support the general entry point mechanism.

### Checklist:


grynspan:
@swift-ci please test

PR #392, opened 2024-05-01
During test discovery, ignore Swift types in the Darwin/dyld shared cache.

grynspan:
This PR modifies the Darwin-specific implementation of runtime type enumeration such that types in Darwin's dyld shared cache are ignored. As of macOS Sonoma, there are approximately 300 images in the shared cache containing approximately 3,000 Swift types, none of which are reasonably expected to represent test declarations.

We don't need to enumerate these types. Skipping them gets us a speedup of approximately 27% (according to my very unscientific measurements.)

Images in the current process (including framework binaries) that come from the shared cache have the `MH_DYLIB_IN_CACHE` bit set in their Mach headers, so we can easily filter them out.

### Checklist:

### DIFF
--- a/Sources/TestingInternals/Discovery.cpp
+++ b/Sources/TestingInternals/Discovery.cpp
@@ -21,22 +21,9 @@
 #elif defined(__APPLE__)
 #include <dispatch/dispatch.h>
 #include <mach-o/dyld.h>
-#if __has_include(<mach-o/dyld_priv.h>)
-#include <mach-o/dyld_priv.h>
-#endif
 #include <mach-o/getsect.h>
 #include <objc/runtime.h>
 #include <os/lock.h>
-
-#if !__has_include(<mach-o/dyld_priv.h>)
-/// Get the range of addresses in the shared cache.
-///
-/// The Swift runtime uses this function on Apple platforms to optimize protocol
-/// conformance lookups. The testing library uses it to optimize test discovery.
-///
-/// - Note: This function is not public API on Darwin.
-SWT_IMPORT_FROM_STDLIB const void *_dyld_get_shared_cache_range(size_t *length);
-#endif
 #endif
 
 /// Enumerate over all Swift type metadata sections in the current process.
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

Sources/TestingInternals/Discovery.cpp
```diff
@@ -227,15 +234,32 @@ static SWTMachHeaderList getMachHeaders(void) {
...
+    size_t dyldSharedCacheLength = 0;
+    dyldSharedCacheBegin = reinterpret_cast<uintptr_t>(_dyld_get_shared_cache_range(&dyldSharedCacheLength));
+    dyldSharedCacheEnd = dyldSharedCacheBegin + dyldSharedCacheLength;
+
     objc_addLoadImageFunc([] (const mach_header *mh) {
       auto mhn = reinterpret_cast<SWTMachHeaderList::value_type>(mh);
 
+      // Ignore this Mach header if it is in the shared cache.
+      auto machHeaderAddress = reinterpret_cast<uintptr_t>(mhn);
+      if (machHeaderAddress >= dyldSharedCacheBegin && machHeaderAddress < dyldSharedCacheEnd) {
```
dennisweissmann, line 259: Should that be `machHeaderAddress =< dyldSharedCacheEnd`?

grynspan: I also asked myself that. But nope, it's just the upper bound, non-inclusive. Compare `startIndex` and `endIndex` in Swift ranges. (It won't matter either way since a valid Mach header can't be 1 byte long.)

grynspan:
@swift-ci please test

mikeash (approved):

grynspan:
@swift-ci please test

dennisweissmann (approved):

dennisweissmann:
Great, thanks Jonathan!

stmontgomery (approved):

PR #396, opened 2024-05-02
Make the typealias of the ABIEntryPiont function available everywhere.

dennisweissmann:
Code that needs to decide whether to use the ABI entry point or not might need to be able to work with the type even if it the function itself is not available.


dennisweissmann:
@swift-ci please test

stmontgomery (approved):

grynspan (changes requested):
I don't understand why this change would be necessary.

PR #403, opened 2024-05-07
On Darwin, lock access to the `environ` block.

grynspan:
This PR adds calls to Darwin's internal libc functions `environ_lock_np()` and `environ_unlock_np()` when directly accessing the `environ` block (via the CRT function `_NSGetEnviron()`.)

These calls are necessary because direct access to `environ` is not thread-safe by default. The standard library guards accesses to `getenv()` and `setenv()`, but `environ` is a global variable and neither it nor `_NSGetEnviron()` can be made thread-safe internally.

This change does not prevent data races when using the result of `getenv()`, but short of a new `copyenv_np()` or `getenv_r()` function, there's not much we can do about it. This change has no effect on Linux since it doesn't have these non-portable functions, but if equivalents are added in the future, we can add support too. Windows' `GetEnvironmentStringsW()` API returns a copy of the environment block, so it doesn't need a lock.

Compare Foundation's use [here](https://github.com/apple/swift-foundation/blob/main/Sources/_CShims/platform_shims.c).

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Environment.swift
+++ b/Sources/Testing/Support/Environment.swift
@@ -67,6 +67,28 @@ enum Environment {
   }
 #endif
 
+#if SWT_TARGET_OS_APPLE && !SWT_NO_ENVIRONMENT_VARIABLES && !SWT_NO_DYNAMIC_LINKING
+  /// A non-POSIX/non-portable function that locks for access to `environ`.
+  ///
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

grynspan:
@swift-ci please test

PR #404, opened 2024-05-07
Rename `TestingInternals` to `_TestingInternals`.

grynspan:
This module is not meant to be accessible outside of swift-testing, however it may be reachable by clients inadvertently. To minimize the likelihood of that, rename the module so it doesn't show up in documentation or Xcode autocomplete contexts.

### Checklist:


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #407, opened 2024-05-09
Ensure tests representing containing types of suites are synthesized if necessary

stmontgomery:
Ensure that when a suite is nested in one or more containing suite types, and those outer suites do not have an explicit `@Suite` attribute, we synthesize `Test` instances representing all containing suites.

### Checklist:

Resolves rdar://127770218

### DIFF
--- a/Sources/Testing/Parameterization/TypeInfo.swift
+++ b/Sources/Testing/Parameterization/TypeInfo.swift
@@ -45,11 +45,33 @@ public struct TypeInfo: Sendable {
     return nil
   }
 
-  init(fullyQualifiedName: String, unqualifiedName: String, mangledName: String?) {
+  /// Initialize an instance of this type with the specified names.
+  ///
+  /// - Parameters:
+  ///   - fullyQualifiedComponents: The fully-qualified name components of the
+  ///     type.
+  ///   - unqualified: The unqualified name of the type.
+  ///   - mangled: The mangled name of the type, if available.
+  init(fullyQualifiedNameComponents: [String], unqualifiedName: String, mangledName: String? = nil) {
     _kind = .nameOnly(
-      fullyQualifiedComponents: fullyQualifiedName.split(separator: ".").map(String.init),
+      fullyQualifiedComponents: fullyQualifiedNameComponents,
       unqualified: unqualifiedName,
-      mangled: mangledName
+      mangled: mangledName,
+    )
+  }
+
+  /// Initialize an instance of this type with the specified names.
+  ///
+  /// - Parameters:
+  ///   - fullyQualifiedName: The fully-qualified name of the type, with its
+  ///     components separated by a period character (`"."`).
+  ///   - unqualified: The unqualified name of the type.
+  ///   - mangled: The mangled name of the type, if available.
+  init(fullyQualifiedName: String, unqualifiedName: String, mangledName: String?) {
+    self.init(
+      fullyQualifiedNameComponents: fullyQualifiedName.split(separator: ".").map(String.init),
+      unqualifiedName: unqualifiedName,
+      mangledName: mangledName
     )
   }
 
@@ -243,44 +265,6 @@ func isClass(_ subclass: AnyClass, subclassOf superclass: AnyClass) -> Bool {
     isClass(subclassImmediateSuperclass, subclassOf: superclass)
   } else {
     false
-  }
-}
-
-// MARK: - Containing types
-
-extension TypeInfo {
-  /// An instance of this type representing the type immediately containing the
-  /// described type.
-  ///
-  /// For instance, given the following declaration in the `Example` module:
-  ///
-  /// ```swift
-  /// struct A {
-  ///   struct B {}
-  /// }
-  /// ```
-  ///
-  /// The value of this property for the type `A.B` would describe `A`, while
-  /// the value for `A` would be `nil` because it has no enclosing type.
-  var containingTypeInfo: Self? {
-    let fqnComponents = fullyQualifiedNameComponents
-    if fqnComponents.count > 2 { // the module is not a type
-      let fqn = fqnComponents.dropLast().joined(separator: ".")
-#if false // currently non-functional
-      if let type = _typeByName(fqn) {
-        return Self(describing: type)
-      }
-#endif
-      let name = fqnComponents[fqnComponents.count - 2]
-      return Self(fullyQualifiedName: fqn, unqualifiedName: name, mangledName: nil)
-    }
-    return nil
-  }
-
-  /// A sequence of instances of this type representing the types that
-  /// recursively contain it, starting with the immediate parent (if any.)
-  var allContainingTypeInfo: some Sequence<Self> {
-    sequence(first: self, next: \.containingTypeInfo).dropFirst()
   }
 }
 
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -192,7 +192,7 @@ extension Runner.Plan {
         // source location, so we use the source location of a close descendant
         // test. We do this instead of falling back to some "unknown"
         // placeholder in an attempt to preserve the correct sort ordering.
-        graph.value = Test(traits: [], sourceLocation: sourceLocation, containingTypeInfo: typeInfo)
+        graph.value = Test(traits: [], sourceLocation: sourceLocation, containingTypeInfo: typeInfo, isSynthesized: true)
       }
     }
 
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -154,6 +154,52 @@ extension Runner.Plan {
     }
   }
 
+  /// Recursively synthesize test instances representing suites for all missing
+  /// values in the specified test graph.
+  ///
+  /// - Parameters:
+  ///   - graph: The graph in which suites should be synthesized.
+  ///   - nameComponents: The name components of the suite to synthesize, based
+  ///     on the key path from the root node of the test graph to `graph`.
+  private static func _recursivelySynthesizeSuites(in graph: inout Graph<String, Test?>, nameComponents: [String] = []) {
+    // The recursive function. This is a local function to simplify the initial
+    // call which does not need to pass the `sourceLocation:` inout argument.
+    func synthesizeSuites(in graph: inout Graph<String, Test?>, nameComponents: [String] = [], sourceLocation: inout SourceLocation?) {
+      for (key, var childGraph) in graph.children {
+        synthesizeSuites(in: &childGraph, nameComponents: nameComponents + [key], sourceLocation: &sourceLocation)
+        graph.children[key] = childGraph
+      }
+
+      if let test = graph.value {
+        sourceLocation = test.sourceLocation
+      } else if let unqualifiedName = nameComponents.last, let sourceLocation {
+        // Don't synthesize suites representing modules.
+        if nameComponents.count <= 1 {
+          return
+        }
+
+        // Don't synthesize suites for nodes in the graph which are the
+        // immediate ancestor of a test function. That level of the hierarchy is
+        // used to disambiguate test functions which have equal names but
+        // different source locations.
+        if let firstChildTest = graph.children.values.first?.value, !firstChildTest.isSuite {
+          return
+        }
+
+        let typeInfo = TypeInfo(fullyQualifiedNameComponents: nameComponents, unqualifiedName: unqualifiedName)
+
+        // Note: When a suite is synthesized, it does not have an accurate
+        // source location, so we use the source location of a close descendant
+        // test. We do this instead of falling back to some "unknown"
+        // placeholder in an attempt to preserve the correct sort ordering.
+        graph.value = Test(traits: [], sourceLocation: sourceLocation, containingTypeInfo: typeInfo)
+      }
+    }
+
+    var sourceLocation: SourceLocation?
+    synthesizeSuites(in: &graph, sourceLocation: &sourceLocation)
+  }
+
   /// Construct a graph of runner plan steps for the specified tests.
   ///
   /// - Parameters:
@@ -197,6 +243,9 @@ extension Runner.Plan {
       // the only scenario where this will throw is when using regex filtering,
       // and that is already guarded earlier in the SwiftPM entry point.
     }
+
+    // Synthesize suites for nodes in the test graph for which they are missing.
+    _recursivelySynthesizeSuites(in: &testGraph)
 
     // Recursively apply all recursive suite traits to children.
     //
--- a/Sources/Testing/Test+Discovery.swift
+++ b/Sources/Testing/Test+Discovery.swift
@@ -28,24 +28,7 @@ extension Test {
   /// All available ``Test`` instances in the process, according to the runtime.
   ///
   /// The order of values in this sequence is unspecified.
-  static var all: some Sequence<Test> {
-    get async {
-      // Convert the raw sequence of tests to a dictionary keyed by ID.
-      var result = await testsByID(_all)
-
-      // Ensure test suite types that don't have the @Suite attribute are still
-      // represented in the result.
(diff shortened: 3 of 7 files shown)


Sources/Testing/Parameterization/TypeInfo.swift
```diff
@@ -276,11 +276,18 @@ extension TypeInfo {
     }
     return nil
   }
+}
 
-  /// A sequence of instances of this type representing the types that
-  /// recursively contain it, starting with the immediate parent (if any.)
-  var allContainingTypeInfo: some Sequence<Self> {
-    sequence(first: self, next: \.containingTypeInfo).dropFirst()
+extension Test {
```
grynspan, line 281: Just cleanup?

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test Linux

Tests/TestingTests/RunnerTests.swift
```diff
@@ -501,7 +501,7 @@ final class RunnerTests: XCTestCase {
 
   func testPoundIfTrueTestFunctionRuns() async throws {
     let testStarted = expectation(description: "Test started")
-    testStarted.expectedFulfillmentCount = 4
+    testStarted.expectedFulfillmentCount = 5
```
briancroom, line 504: These increments are because a synthesized suite for `RunnerTests` is now included in the runner plan?

Tests/TestingTests/PlanTests.swift
```diff
@@ -421,6 +421,15 @@ struct PlanTests {
     #expect(plan.stepGraph.subgraph(at: typeInfo.fullyQualifiedNameComponents + CollectionOfOne("reserved1(reserved2:)")) != nil)
   }
 
+  @Test("Containing suite types without @Suite are synthesized")
+  func containingSuiteSynthesis() async {
+    let plan = await Runner.Plan(selecting: OutermostSuite.self)
+    let testNames = plan.steps.map(\.test.name)
+    #expect(testNames.contains("OutermostSuite"))
```
briancroom, line 428: Is `PlanTests` part of this list also?

Sources/Testing/Test+Discovery.swift
```diff
@@ -80,32 +80,23 @@ extension Test {
...
+  private static func _synthesizeSuiteTypes(into tests: inout [ID: Self]) {
+    // We won't have an accurate source location for synthesized suites, so use
+    // a "runtime inferred" sentinel value instead.
+    //
+    // FIXME: Consider either changing `Test.sourceLocation` to Optional (an
+    // API-breaking change) or modifying Runner.Plan so that it doesn't require
+    // an actual Test instance to represent synthesized parent suites.
+    lazy var inferredSourceLocation = SourceLocation(fileID: "<runtime-inferred>", filePath: "<runtime-inferred>", line: 0, column: 0)
+
+    for test in tests.values {
```
briancroom, line 92: What do the performance characteristics of this look like? Do we possibly need more caching to keep the repeated lookups from becoming a bottleneck?

Tests/TestingTests/RunnerTests.swift
stmontgomery: Yes, exactly. Many of the fixture suites in this file are nested right within the `RunnerTests` XCTestCase subclass.

Tests/TestingTests/PlanTests.swift
stmontgomery: Yes. I nested the fixture suite within `PlanTests` right before posting the PR and forgot to come back and add it here for clarity. It isn't strictly necessary to validate that the fix works (checking for the presence of `OutermostSuite` is sufficient). But including `PlanTests` would be better, so I'll add that and also restructure the `#expect` so that it requires a strict order for the test names.

stmontgomery:
@swift-ci please test

Sources/Testing/Test+Discovery.swift
stmontgomery: I don't think the performance characteristics of this function are changed in a significant way. However, I have never really loved the approach that this method uses, which involves mapping the tests into a dictionary, iterating every test and all of its ancestors, constructing a test ID for each one, and testing whether it's already in the dictionary.

I decided to take a different approach with this PR, which requires broader changes but I think has more benefits overall. In my latest commit, we no longer construct this temporary dictionary, and instead synthesize suites during planning while traversing the test graph. While implementing and debugging this, I also noticed some opportunities to simplify and optimize the test graph by pruning nodes for tests which have been filtered out.

stmontgomery:
@swift-ci please test

Sources/Testing/Test+Discovery.swift
grynspan: The dictionary (or a set) is necessary to avoid duplicate test or suite records which can potentially appear multiple times in the raw data we're walking. Does your new code dedup the data?

stmontgomery: Yeah, they are de-duplicated by their placement in the test graph (which relies on a dictionary)

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -422,3 +485,60 @@ extension Configuration.TestFilter {
     self = combining(with: other, using: op)
   }
 }
+
+// MARK: - Filterable types
+
+extension Configuration.TestFilter.Kind {
```
grynspan, line 491: Are the changes in this file actually related to suite synthesis?

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -304,6 +344,20 @@ extension Runner.Plan {
...
+  ///
+  /// When a suite is synthesized, it does not have accurate source location, so
+  /// the value of this property is a "runtime inferred" source location which
+  /// can be used instead.
+  //
+  // FIXME: Consider either changing `Test.sourceLocation` to Optional (an
+  // API-breaking change) or modifying Runner.Plan so that it doesn't require
+  // an actual Test instance to represent synthesized parent suites, thereby
+  // avoiding the need for a source location for them.
+  fileprivate static let synthesized = SourceLocation(fileID: "<runtime-inferred>", filePath: "<runtime-inferred>", line: 0, column: 0)
```
grynspan, line 358: The malformed file ID will cause crashes.
Line 0 and column 0 aren't allowed and will also cause crashes. Use 1 for each.

"<runtime-inferred>" reads to me like the Swift runtime inferred them, not that they were synthesized at run time. Bikeshed a bit?

Sources/Testing/Test.swift
```diff
@@ -192,29 +192,18 @@ public struct Test: Sendable {
...
   }
 
-  /// Whether or not this instance was synthesized at runtime.
-  ///
-  /// During test planning, suites that are not explicitly marked with the
-  /// `@Suite` attribute are synthesized from available type information before
-  /// being added to the plan. For such suites, the value of this property is
-  /// `true`.
-  @_spi(ForToolsIntegrationOnly)
-  public var isSynthesized = false
```
grynspan, line 202: This property is useful for diagnosing issues with synthesized instances—why did you remove it?

grynspan (changes requested):
See comments above.

Sources/Testing/Running/Runner.Plan.swift
stmontgomery: > Line 0 and column 0 aren't allowed and will also cause crashes. Use 1 for each.

Hmm… why is this not crashing currently if that's the case? This property is being accessed.

> `"<runtime-inferred>"` reads to me like the Swift runtime inferred them, not that they were synthesized at run time. Bikeshed a bit?

The compiler sometimes generates code with locations that look like `<compiler-generated>:0`, and this is what inspired me to use the `<...>` notation. Any alternative wording suggestions? `"<inferred>"` ? 

I also want to draw attention to the "FIXME" comment above this, which suggests future ways we could avoid needing this property altogether.

Sources/Testing/Test.swift
stmontgomery: It doesn't appear to be used anywhere

Sources/Testing/Running/Runner.Plan.swift
stmontgomery: > Hmm… why is this not crashing currently if that's the case? This property is being accessed.

Ah, right. `didSet` isn't used for property assignments in initializers

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: I've expanded the scope of this PR to more than just the bug fix. I'm going to rewrite the PR description to summarize, although if you think it would help, I could tease this apart into separate PRs, and just use this one to focus on the original bug fix (related to test suite synthesis).

Sources/Testing/Running/Runner.Plan.swift
grynspan: > Ah, right. didSet isn't used for property assignments in initializers

Indeedy.

> Any alternative wording suggestions? "<inferred>" ?

Off the top of my head:

- `<synthesized>`
- `<compiler-generated>` (i.e. just lie for consistency with other bits of the stack)
- `<library-generated>`

But the string needs to be a valid file ID anyway (file path ideally should be valid too, but that's less urgent.) _This is why I was using the source location of the nearest real test_. It wasn't just arbitrary.

Sources/Testing/Test.swift
grynspan: It's not actively used but it's _useful_ when I'm debugging stuff.

Sources/Testing/Running/Configuration.TestFilter.swift
grynspan: Yes, I do think these unrelated changes belong in their own PR. Unless they really are related, in which case please elaborate.

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -304,6 +344,20 @@ extension Runner.Plan {
...
+  ///
+  /// When a suite is synthesized, it does not have accurate source location, so
+  /// the value of this property is a "runtime inferred" source location which
+  /// can be used instead.
+  //
+  // FIXME: Consider either changing `Test.sourceLocation` to Optional (an
+  // API-breaking change) or modifying Runner.Plan so that it doesn't require
+  // an actual Test instance to represent synthesized parent suites, thereby
+  // avoiding the need for a source location for them.
+  fileprivate static let synthesized = SourceLocation(fileID: "<unknown-module>/<unknown-file>", filePath: "<unknown-file>", line: 0, column: 0)
```
grynspan, line 358: Did you check to see if this breaks total ordering for tests now that they may appear to be located in a file with this name?

Also the module is not unknown for these types and should be correctly specified.

stmontgomery: > Did you check to see if this breaks total ordering for tests now that they may appear to be located in a file with this name?

Thanks for that context! I'm going to change the portion of this PR related to inferring the SourceLocation to ensure it continues using the location of some child, to ensure the sort ordering of tests remains unchanged. This should address this and your [other PR comment](https://github.com/swiftlang/swift-testing/pull/407#discussion_r1835109103) above.

One nice thing about my new approach in this PR (synthesizing while recursively traversing the graph during planning) is that it can ensure that any synthesized suite always uses a source location of a child test which is a minimum distance from it in the graph, meaning the inferred source location should be closer to accurate more often. I'll also add unit tests for it.

Sources/Testing/Running/Runner.Plan.swift
stmontgomery: > > Ah, right. didSet isn't used for property assignments in initializers
> 
> Indeedy.

I spun off a separate PR to improve this: https://github.com/swiftlang/swift-testing/pull/817

Sources/Testing/Running/Runner.Plan.swift
grynspan: I see. We should definitely fix the module names, but if the filename isn't an issue, then we can leave it stubbed out.

Sources/Testing/Test.swift
stmontgomery: What if we made `isSynthesized` into an `internal`-only, non-recursive `SuiteTrait` instead? Then we could eliminate this stored property, while keeping the ability to see whether a suite is synthesized while debugging.

And if we take that approach, I would like to conditionalize the trait's inclusion using `#if DEBUG`.

Tests/TestingTests/MiscellaneousTests.swift
```diff
@@ -552,17 +552,6 @@ struct MiscellaneousTests {
     #expect(id.keyPathRepresentation == [""])
   }
 
-  @Test("Test.all deduping")
```
grynspan, line 555: Why did you remove this test? Did it start failing? Can it be refactored?

stmontgomery: This PR removes `Test.testsByID()` and modifies `Test.all` to directly return all runtime-enumerated test instances, so I didn't see a remaining need for this test.

grynspan: We still want to test that content is deduped because the raw content we collect can have duplicates.

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: I've split this out into #820

Tests/TestingTests/MiscellaneousTests.swift
stmontgomery: Ok, I think that would be better to test at the end of the planning process, whereas this test is validating an intermediary stage (after runtime discovery but before planning). I'll refactor it.

grynspan: Deduping can be a natural consequence of us building the initial test graph, FWIW.

stmontgomery: Oh yeah, and I recall now that's actually why I didn't bother replacing or refactoring this test when I removed it. I think we're OK as-is

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -191,7 +191,7 @@ extension Configuration.TestFilter {
 extension Configuration.TestFilter {
   /// An enumeration which represents filtering logic to be applied to a test
   /// graph.
-  fileprivate enum Operation: Sendable {
```
grynspan, line 194: We don't _usually_ bother naming generic parameters more than `T`, `U`, etc.… should we start?

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -154,6 +154,52 @@ extension Runner.Plan {
...
   }
 
+  /// Recursively synthesize test instances representing suites for all missing
+  /// values in the specified test graph.
+  ///
+  /// - Parameters:
+  ///   - graph: The graph in which suites should be synthesized.
+  ///   - nameComponents: The name components of the suite to synthesize, based
+  ///     on the key path from the root node of the test graph to `graph`.
+  private static func _recursivelySynthesizeSuites(in graph: inout Graph<String, Test?>, nameComponents: [String] = []) {
```
grynspan, line 164: Test coverage of this function at 100%, yeah?

stmontgomery: Yes!

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: I can switch it to `T` for now, I just found this one instance a bit confusing out of context

stmontgomery:
@swift-ci please test

Sources/Testing/Test.swift
stmontgomery: For this PR, I have restored the `isSynthesized` property, and actually adopted it in my new unit tests so it's no longer unused

stmontgomery:
@swift-ci please test

PR #408, opened 2024-05-09
Behavior validation docs

iamleeg:
Introduces additional documentation for expectations and confirmations.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -16,17 +16,17 @@ Check for expected values, outcomes, and asynchronous events in tests.
 
 Use ``expect(_:_:sourceLocation:)`` and
 ``require(_:_:sourceLocation:)-5l63q`` macros to validate expected
-outcomes.  To validate that an error is thrown, or _not_ thrown, the
+outcomes. To validate that an error is thrown, or _not_ thrown, the
 testing library provides several overloads of the macros that you can
-use.  Use a ``Confirmation`` to confirm the occurrence of an
+use. Use a ``Confirmation`` to confirm the occurrence of an
 asynchronous event that you can't check directly using an expectation.
 
 ### Validate your code's result
 
 To validate that your code produces an expected value, use
-``expect(_:_:sourceLocation:)``.  ``expect(_:_:sourceLocation:)`` captures the
+``expect(_:_:sourceLocation:)``. ``expect(_:_:sourceLocation:)`` captures the
 expression you pass, and provides detailed information when the code doesn't
-satisfy the expectation:
+satisfy the expectation.
 
 ```swift
 class OrderCalculator {
@@ -44,7 +44,7 @@ class OrderCalculator {
 }
 ```
 
-Your test keeps running after ``expect(_:_:sourceLocation:)`` fails.  To stop
+Your test keeps running after ``expect(_:_:sourceLocation:)`` fails. To stop
 the test when the code doesn't satisfy a requirement, use
 ``require(_:_:sourceLocation:)-5l63q`` instead:
 
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -14,16 +14,19 @@ Check for expected values, outcomes, and events in tests.
 
 ## Overview
 
-Use ``expect(_:_:sourceLocation:)`` and ``require(_:_:sourceLocation:)-5l63q`` macros to validate expected outcomes.
-To validate that an error is thrown, or _not_ thrown, 
-the testing library provides several overloads of the macros that you can use.
-Use a ``Confirmation`` to confirm the occurrence of an asynchronous event that
-you can't check directly using an expectation.
+Use ``expect(_:_:sourceLocation:)`` and
+``require(_:_:sourceLocation:)-5l63q`` macros to validate expected
+outcomes.  To validate that an error is thrown, or _not_ thrown, the
+testing library provides several overloads of the macros that you can
+use.  Use a ``Confirmation`` to confirm the occurrence of an
+asynchronous event that you can't check directly using an expectation.
 
 ### Validate your code's result
 
-To validate that your code produces an expected value, use `#expect()`.
-`#expect()` captures the expression you pass, and provides detailed information when the code doesn't satisfy the expectation:
+To validate that your code produces an expected value, use
+`#expect()`.  `#expect()` captures the expression you pass, and
+provides detailed information when the code doesn't satisfy the
+expectation:
 
 ```swift
 func add(_ x: Int, to y: Int) -> Int {
@@ -38,8 +41,8 @@ func add(_ x: Int, to y: Int) -> Int {
 }
 ```
 
-Your test keeps running after `#expect()` fails.
-To stop the test when the code doesn't satisfy a requirement, use `#require()` instead:
+Your test keeps running after `#expect()` fails.  To stop the test
+when the code doesn't satisfy a requirement, use `#require()` instead:
 
 ```swift
 struct Square {
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -24,9 +24,9 @@ asynchronous event that you can't check directly using an expectation.
 ### Validate your code's result
 
 To validate that your code produces an expected value, use
-`#expect()`.  `#expect()` captures the expression you pass, and
-provides detailed information when the code doesn't satisfy the
-expectation:
+``expect(_:_:sourceLocation:)``.  ``expect(_:_:sourceLocation:)`` captures the
+expression you pass, and provides detailed information when the code doesn't
+satisfy the expectation:
 
 ```swift
 class OrderCalculator {
@@ -44,8 +44,9 @@ class OrderCalculator {
 }
 ```
 
-Your test keeps running after `#expect()` fails.  To stop the test
-when the code doesn't satisfy a requirement, use `#require()` instead:
+Your test keeps running after ``expect(_:_:sourceLocation:)`` fails.  To stop
+the test when the code doesn't satisfy a requirement, use
+``require(_:_:sourceLocation:)-5l63q`` instead:
 
 ```swift
 @Test func returningCustomerRemembersUsualOrder() throws {
@@ -55,7 +56,7 @@ when the code doesn't satisfy a requirement, use `#require()` instead:
 }
 ```
 
-`#require()` throws an instance of ``ExpectationFailedError`` when your code
+``require(_:_:sourceLocation:)-5l63q`` throws an instance of ``ExpectationFailedError`` when your code
 fails to satisfy the requirement.
 
 ## Topics
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -29,15 +29,18 @@ provides detailed information when the code doesn't satisfy the
 expectation:
 
 ```swift
-func add(_ x: Int, to y: Int) -> Int {
-    return x * y
+class OrderCalculator {
+    func total(of subtotals: [Int]) -> Int {
+        return subtotals.reduce(1) { partialResult, subtotal in
+            partialResult + subtotal
+        }
+    }
 }
 
-
-@Test func addingTwoNumbers() {
-    #expect(add(3, to: 3) == 6)
-    // Prints "Test addingTwoNumbers() recorded an issue at WritingTestsInSwiftTestingTests.swift:18:9:
-    //   Expectation failed: (add(3, to: 3) → 9) == 6"
+@Test func calculatingOrderTotal() {
+    let calculator = OrderCalculator()
+    #expect(calculator.total(of: [3, 3]) == 6)
+    // Prints "Expectation failed: (calculator.total(of: [3, 3]) → 7) == 6"
 }
 ```
 
@@ -45,23 +48,10 @@ Your test keeps running after `#expect()` fails.  To stop the test
 when the code doesn't satisfy a requirement, use `#require()` instead:
 
 ```swift
-struct Square {
-    let sideLength: Int
-
-    var perimeter: Int {
-        get { sideLength * 4 }
-    }
-
-    init?(length: Int) {
-        guard length > 0 else { return nil }
-        sideLength = length
-    }
-}
-
-@Test func zeroLengthSquare() throws {
-    let aSquare = Square(length: 0)
-    try #require(aSquare != nil)
-    #expect(aSquare?.perimeter == 0) // The test runner doesn't reach this line.
+@Test func returningCustomerRemembersUsualOrder() throws {
+    let customer = Customer(id: 123)
+    try #require(customer)
+    #expect(customer?.usualOrder?.countOfItems == 2) // The test runner doesn't reach this line if the customer is nil.
 }
 ```
 
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -86,6 +86,7 @@ struct Square {
 
 ### Confirming that asynchronous events occur
 
+- ``<doc:testing-asynchronous-code>
 - ``confirmation(_:expectedCount:fileID:filePath:line:column:_:)``
 - ``Confirmation``
 
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -10,15 +10,59 @@ See https://swift.org/LICENSE.txt for license information
 See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
 
-Check for expected values and outcomes in tests.
+Check for expected values, outcomes, and events in tests.
 
 ## Overview
 
-The testing library provides `#expect()` and `#require()` macros you use to 
-validate expected outcomes. To validate that an error is thrown, or _not_ thrown, 
+Use `#expect()` and `#require()` macros to validate expected outcomes.
+To validate that an error is thrown, or _not_ thrown, 
 the testing library provides several overloads of the macros that you can use.
 Use a ``Confirmation`` to confirm the occurrence of an asynchronous event that
 you can't check directly using an expectation.
+
+### Validate your code's result
+
+To validate that your code produces an expected value, use `#expect()`.
+`#expect()` captures the expression you pass, and provides detailed information when the code doesn't satisfy the expectation:
+
+```swift
+func add(_ x: Int, to y: Int) -> Int {
+    return x * y
+}
+
+
+@Test func addingTwoNumbers() {
+    #expect(add(3, to: 3) == 6)
+    // Prints "Test addingTwoNumbers() recorded an issue at WritingTestsInSwiftTestingTests.swift:18:9:
+    //   Expectation failed: (add(3, to: 3) → 9) == 6"
+}
+```
+
+Your test keeps running after `#expect()` fails.
+To stop the test when the code doesn't satisfy a requirement, use `#require()` instead:
+
+```swift
+struct Square {
+    let sideLength: Int
+
+    var perimeter: Int {
+        get { sideLength * 4 }
+    }
+
+    init?(length: Int) {
+        guard length > 0 else { return nil }
+        sideLength = length
+    }
+}
+
+@Test func zeroLengthSquare() throws {
+    let aSquare = Square(length: 0)
+    try #require(aSquare != nil)
+    #expect(aSquare?.perimeter == 0) // The test runner doesn't reach this line.
+}
+```
+
+`#require()` throws when your code fails to satisfy the requirement.
 
 ## Topics
 
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -18,8 +18,11 @@ Use ``expect(_:_:sourceLocation:)`` and
 ``require(_:_:sourceLocation:)-5l63q`` macros to validate expected
 outcomes. To validate that an error is thrown, or _not_ thrown, the
 testing library provides several overloads of the macros that you can
-use. Use a ``Confirmation`` to confirm the occurrence of an
+use. For more information, see <doc:testing-for-errors-in-swift-code>.
+
+Use a ``Confirmation`` to confirm the occurrence of an
 asynchronous event that you can't check directly using an expectation.
+For more information, see <doc:testing-asynchronous-code>.
 
 ### Validate your code's result
 
@@ -29,14 +32,6 @@ expression you pass, and provides detailed information when the code doesn't
 satisfy the expectation.
 
 ```swift
-class OrderCalculator {
-  func total(of subtotals: [Int]) -> Int {
-    return subtotals.reduce(1) { partialResult, subtotal in
-      partialResult + subtotal
-    }
-  }
-}
-
 @Test func calculatingOrderTotal() {
   let calculator = OrderCalculator()
   #expect(calculator.total(of: [3, 3]) == 6)
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -50,14 +50,14 @@ the test when the code doesn't satisfy a requirement, use
 
 ```swift
 @Test func returningCustomerRemembersUsualOrder() throws {
-    let customer = Customer(id: 123)
-    try #require(customer)
-    #expect(customer?.usualOrder?.countOfItems == 2) // The test runner doesn't reach this line if the customer is nil.
+    let customer = try #require(Customer(id: 123))
+    // The test runner doesn't reach this line if the customer is nil.
+    #expect(customer.usualOrder.countOfItems == 2)
 }
 ```
 
-``require(_:_:sourceLocation:)-5l63q`` throws an instance of ``ExpectationFailedError`` when your code
-fails to satisfy the requirement.
+``require(_:_:sourceLocation:)-5l63q`` throws an instance of
+``ExpectationFailedError`` when your code fails to satisfy the requirement.
 
 ## Topics
 
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -74,6 +74,7 @@ struct Square {
 
 ### Checking that errors are thrown
 
+- <doc:testing-for-errors-in-swift-code>
 - ``expect(throws:_:sourceLocation:performing:)-79piu``
 - ``expect(throws:_:sourceLocation:performing:)-1xr34``
 - ``expect(_:sourceLocation:performing:throws:)``
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -14,7 +14,7 @@ Check for expected values, outcomes, and events in tests.
 
 ## Overview
 
-Use `#expect()` and `#require()` macros to validate expected outcomes.
+Use ``expect(_:_:sourceLocation:)`` and ``require(_:_:sourceLocation:)-5l63q`` macros to validate expected outcomes.
 To validate that an error is thrown, or _not_ thrown, 
 the testing library provides several overloads of the macros that you can use.
 Use a ``Confirmation`` to confirm the occurrence of an asynchronous event that
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -30,17 +30,17 @@ satisfy the expectation.
 
 ```swift
 class OrderCalculator {
-    func total(of subtotals: [Int]) -> Int {
-        return subtotals.reduce(1) { partialResult, subtotal in
-            partialResult + subtotal
-        }
+  func total(of subtotals: [Int]) -> Int {
+    return subtotals.reduce(1) { partialResult, subtotal in
+      partialResult + subtotal
     }
+  }
 }
 
 @Test func calculatingOrderTotal() {
-    let calculator = OrderCalculator()
-    #expect(calculator.total(of: [3, 3]) == 6)
-    // Prints "Expectation failed: (calculator.total(of: [3, 3]) → 7) == 6"
+  let calculator = OrderCalculator()
+  #expect(calculator.total(of: [3, 3]) == 6)
+  // Prints "Expectation failed: (calculator.total(of: [3, 3]) → 7) == 6"
 }
 ```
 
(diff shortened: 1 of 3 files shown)


iamleeg:
@swift-ci please test

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,16 +10,60 @@ See https://swift.org/LICENSE.txt for license information
 See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
 
-Check for expected values and outcomes in tests.
+Check for expected values, outcomes, and events in tests.
```
grynspan, line 13: Events have special meaning to swift-testing and `#expect()` does not check for them.

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,16 +10,60 @@ See https://swift.org/LICENSE.txt for license information
...
 
 ## Overview
 
-The testing library provides `#expect()` and `#require()` macros you use to 
-validate expected outcomes. To validate that an error is thrown, or _not_ thrown, 
+Use `#expect()` and `#require()` macros to validate expected outcomes.
```
grynspan, line 17: Please use double-backticks and link to the `expect(_:_:)` macro rather than writing `#expect()`. Ibid for `#require()`, but note the overloads are different. (This is an error in the existing documentation.)

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,16 +10,60 @@ See https://swift.org/LICENSE.txt for license information
...
+To validate that an error is thrown, or _not_ thrown, 
 the testing library provides several overloads of the macros that you can use.
 Use a ``Confirmation`` to confirm the occurrence of an asynchronous event that
 you can't check directly using an expectation.
 
+### Validate your code's result
+
+To validate that your code produces an expected value, use `#expect()`.
+`#expect()` captures the expression you pass, and provides detailed information when the code doesn't satisfy the expectation:
```
grynspan, line 26: Wrap at 80 columns please.

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,16 +10,60 @@ See https://swift.org/LICENSE.txt for license information
...
+}
+
+@Test func zeroLengthSquare() throws {
+    let aSquare = Square(length: 0)
+    try #require(aSquare != nil)
+    #expect(aSquare?.perimeter == 0) // The test runner doesn't reach this line.
+}
+```
+
+`#require()` throws when your code fails to satisfy the requirement.
```
grynspan, line 65: Throws an instance of `ExpectationFailedError` specifically.

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,59 @@
...
+
+See https://swift.org/LICENSE.txt for license information
+See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+Validate whether your code causes expected events to happen.
+
+## Overview
+
+Swift Testing integrates with Swift concurrency, meaning that in many situations you can test asynchronous code using standard Swift features.
```
grynspan, line 17: The package is named `swift-testing`.

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,59 @@
...
+Mark your test function as `async` and, in the function body, `await` any asynchronous interactions:
+
+```swift
+@Test func asynchronousCalculationYieldsExpectedValue() async {
+    #expect(await asynchronousCalculation(with: 10) == 12)
```
grynspan, line 22: We don't currently recommend placing effectful expressions directly in `#expect()` or `#require()` because we can't expand them correctly on failure. Better would be:

```swift
let result = await asynchronousCalculation(with: 10)
#expect(result == 12)
```

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,59 @@
...
+The test passes if the number of occurrences during the test matches the expected count, and fails otherwise.
+
+### Confirm that an event doesn't happen
+
+To validate that a particular event doesn't occur during a test, create a `Confirmation` with an expected count of `0`:
+
+```swift
+@Test func asynchronousCalculatorEncountersNoErrors() async {
+    let calculator = AsynchronousCalculator()
+    await confirmation(expectedCount:0) { confirmation in
```
grynspan, line 54:
```suggestion
    await confirmation(expectedCount: 0) { confirmation in
```

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,100 @@
...
+## Overview
+
+The Swift language provides an idiomatic approach to [error handling](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/errorhandling), based on throwing errors where your code detects a failure for a caller to catch and react to.
+
+Write tests for your code that validate the conditions in which the code throws errors, and the conditions in which it returns without throwing an error.
+Use overrides of the `#expect()` and `#require()` macros that check for errors.
+
+### Validate that your code throws an expected error
+
+The Swift structure in this example represents a list that accepts any number of "tags" for items in the list.
```
grynspan, line 24: "Tags" are a term of art in swift-testing. To avoid confusion, use a different concept in this example.

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,100 @@
...
+@Test func canAddTagToObjectInPositionZero() throws {
+    var array = TaggedArray(list: [1,2,3])
+    #expect(throws: Never.self) {
+        try array.add(tag: "my tag", toObjectsIn: 0..<1)
+    }
+    #expect(try array.tags(forItemAt: 0) == ["my tag"])
+}
+```
+
+If the block throws an error, the testing library records an issue.
```
grynspan, line 96:
```suggestion
If the closure throws an error, the testing library records an issue.
```

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,100 @@
...
+```swift
+@Test func cannotAddTagToObjectBeforeStartOfList() {
+    var array = TaggedArray(list: [1,2,3])
+    #expect(throws: TaggedArray<Int>.TaggedArrayError.outOfRange) {
+        try array.add(tag: "my tag", toObjectsIn: -1..<0)
+    }
+}
+```
+
+If the block completes without throwing an error, the testing library records an issue.
```
grynspan, line 78:
```suggestion
If the closure completes without throwing an error, the testing library records an issue.
```

Sources/Testing/Testing.docc/Expectations.md
iamleeg: I agree, but `Confirmation` is in this API collection (see Topics) and `Confirmation` does check for events.

Sources/Testing/Testing.docc/Expectations.md
iamleeg: On first use only, or every time?

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,100 @@
...
+Other overloads of `#expect()` let you test that the code throws an error of a given type, or matching an arbitrary Boolean test.
+Similar overloads of `#require()` stop running your test if the code doesn't throw the expected error.
+
+### Validate that your code doesn't throw an error
+
+Validate that the code under test doesn't throw an error by comparing the error to `Never`:
```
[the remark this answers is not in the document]
iamleeg: What do you think of this, is it sufficient or would an example help?

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
iamleeg: _checks synonyms_

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -27,44 +27,44 @@ macros that check for errors.
 ### Validate that your code throws an expected error
 
 The Swift structure in this example represents a list that accepts any
-number of "tags" for items in the list.  The API contains a method for
-applying a tag to a range of items, and a method for retrieving the
-tags associated with the item at a given index.  Both of these methods
+number of attributes for items in the list.  The API contains a method for
```
grynspan, line 30: Attribute is also a Swift term of art.

Sources/Testing/Testing.docc/Expectations.md
grynspan: Every time: DocC does not render macros with leading `#` or `@` characters, so having them suddenly appear is likely to be confusing. If, later, DocC does add those characters, we'll get them everywhere "for free".

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,118 @@
...
+    #expect(try array.attributes(forItemAt: 0) == ["my attribute"])
+}
+```
+
+If the closure throws an error, the testing library records an issue.
+If you need the test to stop if the code throws an error, include the
+code inline in the test function instead of wrapping it in an
+`#expect(throws:)` block.
+
+> Note: `#require(throws:Never.self)` is deprecated, because calling
```
grynspan, line 117: We don't need to call this out here.

Also, is this style correct for DocC? Will this render as a note callout?

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,118 @@
...
+Validate that the code under test doesn't throw an error by comparing
+the error to `Never`:
+
+```swift
+@Test func canAddAttributeToObjectInPositionZero() throws {
+    var array = AttributedArray(list: [1,2,3])
+    #expect(throws: Never.self) {
+        try array.add(attribute: "my attribute", toObjectsIn: 0..<1)
+    }
+    #expect(try array.attributes(forItemAt: 0) == ["my attribute"])
```
grynspan, line 108:
```suggestion
    let attributes = try array.attributes(forItemAt: 0)
    #expect(attributes == ["my attribute"])
```

Sources/Testing/Testing.docc/Expectations.md
grynspan: How about:

> Check for expected values, outcomes, and asynchronous events in tests.

Since `confirmation()` should really only be used for testing out-of-band stuff.

Sources/Testing/Testing.docc/Expectations.md
grynspan: However, if your department already has guidelines for this sort of thing, I defer to you.

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
grynspan: Oh neat. We otherwise use:

```swift
- Note: lorem ipsum…
```

Sources/Testing/Testing.docc/testing-asynchronous-code.md
iamleeg: Fixed (and I've updated the sample to be more FoodTruck-esque)

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
iamleeg: Removed.

Sources/Testing/Testing.docc/Expectations.md
iamleeg: FWIW, what a developer _types_ is `#expect()` or `#require()`, so I think it's more helpful to show them that; nonetheless, I'll change this to match the local style.

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,15 +10,54 @@ See https://swift.org/LICENSE.txt for license information
...
+
+```swift
+@Test func returningCustomerRemembersUsualOrder() throws {
+    let customer = Customer(id: 123)
+    try #require(customer)
+    #expect(customer?.usualOrder?.countOfItems == 2) // The test runner doesn't reach this line if the customer is nil.
+}
+```
+
+``require(_:_:sourceLocation:)-5l63q`` throws an instance of ``ExpectationFailedError`` when your code
```
grynspan, line 59: Wrap at 80 columns please.

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,15 +10,54 @@ See https://swift.org/LICENSE.txt for license information
...
+
+Your test keeps running after ``expect(_:_:sourceLocation:)`` fails.  To stop
+the test when the code doesn't satisfy a requirement, use
+``require(_:_:sourceLocation:)-5l63q`` instead:
+
+```swift
+@Test func returningCustomerRemembersUsualOrder() throws {
+    let customer = Customer(id: 123)
+    try #require(customer)
+    #expect(customer?.usualOrder?.countOfItems == 2) // The test runner doesn't reach this line if the customer is nil.
```
grynspan, line 55: This comment may trigger a horizontal scrollbar. How about:

```suggestion
    // The test runner doesn't reach this line if the customer is nil.
    #expect(customer?.usualOrder?.countOfItems == 2)
```

Sources/Testing/Testing.docc/Expectations.md
grynspan: Mind making this change?

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,70 @@
...
+## Overview
+
+`swift-testing` integrates with Swift concurrency, meaning that in many
+situations you can test asynchronous code using standard Swift
+features.  Mark your test function as `async` and, in the function
+body, `await` any asynchronous interactions:
+
+```swift
+@Test func priceLookupYieldsExpectedValue() async {
+    #expect(await unitPrice(for: .mozarella) == 3)
```
grynspan, line 24: Because `#expect(await ...)` does not expand its arguments, please split this line up:

```suggestion
    let mozarellaPrice = await unitPrice(for: .mozarella)
    #expect(mozarellaPrice == 3)
```

The existing code will compile, but won't produce the best possible output on test failure.

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,15 +10,54 @@ See https://swift.org/LICENSE.txt for license information
...
+```
+
+Your test keeps running after ``expect(_:_:sourceLocation:)`` fails.  To stop
+the test when the code doesn't satisfy a requirement, use
+``require(_:_:sourceLocation:)-5l63q`` instead:
+
+```swift
+@Test func returningCustomerRemembersUsualOrder() throws {
+    let customer = Customer(id: 123)
+    try #require(customer)
```
grynspan, line 54: This line does not unwrap `customer` and will trigger an unused-result warning. Instead, combine the two lines:
```suggestion
    let customer = try #require(Customer(id: 123))
```

Sources/Testing/Testing.docc/Expectations.md
grynspan: Note if you apply my suggestion for the above line, `customer?` here becomes `customer`.

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,70 @@
...
+If your code under test doesn't use Swift concurrency, call
+``confirmation(_:expectedCount:fileID:filePath:line:column:_:)`` in
+your asynchronous test function to create a `Confirmation` for the
+expected event.  In the trailing block parameter, call the code under
+test.  Swift Testing passes a `Confirmation` as the parameter to the
+block, which you call as a function in the completion or event handler
+for the code under test when the event you're testing for occurs:
+
+```swift
+@Test func orderCalculatorSuccessfullyCalculatesSubtotalForNoPizzas() async {
```
grynspan, line 43: Test function names don't need to be this verbose. Consider using a display name if they get this long:

```suggestion
@Test("OrderCalculator successfully calculates subtotal for no pizzas")
func subtotalForNoPizzas() async {
```

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,70 @@
...
+    let calculator = OrderCalculator()
+    await confirmation() { confirmation in
+        calculator.successHandler = { _ in confirmation() }
+        calculator.subtotal(for: PizzaToppings(bases: []))
```
grynspan, line 47:
```suggestion
        _ = calculator.subtotal(for: PizzaToppings(bases: []))
```

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,131 @@
...
+### Validate that your code throws an expected error
+
+The Swift structure in this example represents a list that accepts any
+number of toppings for pizzas in the list.  The API contains a method for
+applying a topping to a range of pizzas, and a method for retrieving the
+toppings requested for the item at a given index.  Both of these methods
+throw errors if their parameters are outside the list's range.
+
+```swift
+enum PizzaBase {
```
grynspan, line 36: For conciseness, consider:

```swift
enum PizzaBase {
  case deepCrust, shallowCrust, calzone
}
```

(Whether or not that's good style in general is beyond the scope of my nitpicking.)

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,131 @@
...
+    let pizzas: [PizzaBase]
+    var toppings: [Int: [Topping]]
+
+    init(bases: [PizzaBase]) {
+        pizzas = bases
+        toppings = [Int: [Topping]]()
+    }
+
+    mutating func add(topping: Topping, toPizzasIn range: Range<Int>) throws {
+        guard Int(range.startIndex) >= 0 && Int(range.endIndex) < pizzas.count else {
```
grynspan, line 65: This is awkward phrasing. What are you trying to do here? We can probably simplify it.

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,131 @@
...
+            if var toppingList = toppings[index] {
+                toppingList.append(topping)
+                toppings[index] = toppingList
+            } else {
+                toppings[index] = [topping]
+            }
+        }
+    }
+
+    func toppings(forPizzaAt index: Int) throws -> [Topping] {
```
grynspan, line 78: There's a _lot_ of code here that isn't actually going to be used by developers. Are we sure we need to actually build out this type for the sake of explaining the topic? Is there some way to factor some of it out of the documentation?

Sources/Testing/Testing.docc/Expectations.md
grynspan: We'd much prefer the DocC compiler would add the `#` for us in this case. There's an issue or radar or some such floating around out there about it.

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,131 @@
...
+
+If the closure completes without throwing an error, the testing library
+records an issue.  Other overloads of ``expect(_:_:sourceLocation:)`` let you test that
+the code throws an error of a given type, or matching an arbitrary
+Boolean test.  Similar overloads of ``require(_:_:sourceLocation:)-5l63q`` stop running your
+test if the code doesn't throw the expected error.
+
+### Validate that your code doesn't throw an error
+
+Validate that the code under test doesn't throw an error by comparing
```
grynspan, line 114: We should emphasize here that a test function that throws will fail if an error is thrown, so _usually_ you don't need to use `#expect(throws: Never.self)`. You'd really only use it if you want to record an issue for a thrown error but _not_ stop execution of the test function. It's pretty nuanced, to the point that I'm not sure we should even call it out in the documentation.

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,70 @@
...
+}
+```
+
+In more complex situations, where the code you test doesn't use Swift
+concurrency, you use ``Confirmation`` to discover whether an expected
+event happens.
+
+### Confirm that an event happens
+
+If your code under test doesn't use Swift concurrency, call
```
grynspan, line 34: This paragraph is misleading. `confirmation()` does not make synchronous code asynchronous. The code under test must still complete by the time `confirmation {}` returns, and can be asynchronous. For example, this is valid:

```swift
confirmation { didThing in
  let x = X()
  x.eventHandler = { didThing() }
  await x.f()
}
```

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,131 @@
...
+## Overview
+
+The Swift language provides an idiomatic approach to [error
+handling](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/errorhandling),
+based on throwing errors where your code detects a failure for a
+caller to catch and react to.
+
+Write tests for your code that validate the conditions in which the
+code throws errors, and the conditions in which it returns without
+throwing an error.  Use overrides of the `#expect()` and
```
grynspan, line 24: `#expect()` here probably should be a callout to `expect(_:_:sourceLocation:)`.

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,131 @@
...
+``require(_:_:sourceLocation:)-5l63q`` macros that check for errors.
+
+### Validate that your code throws an expected error
+
+The Swift structure in this example represents a list that accepts any
+number of toppings for pizzas in the list.  The API contains a method for
```
grynspan, line 30:
```suggestion
number of toppings for pizzas in the list.  The interface contains a method for
```

grynspan: Or `type` or `structure`. It's not an API; it's not `public`.

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,131 @@
...
+enum Topping {
+    case tomato
+    case cheese
+    case caper
+    case anchovy
+    case prosciutto
+    case pineapple
+}
+
+struct PizzaToppings {
```
grynspan, line 51: You have a `Toppings` type and a `PizzaToppings` type. But this type also describes the base. Perhaps this should just be `Pizza` and the other types should be nested `Pizza.Topping` and `Pizza.Base`?

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
grynspan: There's some history here that we don't need to document, but for your own edification/amusement: XCTest in Objective-C has `XCTAssertNoThrow()` that asserts a given expression doesn't throw an _Objective-C or C++ exception_. This is different from a Swift error because if an exception is thrown through a test method's frame, it tends to wreak some havoc on the way out before being caught by XCTest, so you want to proactively catch them if you know they might occur. This was inherited as `XCTAssertNoThrow()` in Swift which checks if a function throws a Swift error rather than an exception.

grynspan (changes requested):
See comments above.

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
iamleeg: I'm trying to test that the range is within the domain of indices in the `pizzas` array, but given that you go on below to say that I don't need to show the code under test, I'll avoid fixing this (which isn't a complete check anyway).

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
iamleeg: No, I'll remove it.

chuckdude (approved):
All good here. Thanks.

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+In more complex situations you use ``Confirmation`` to discover whether an
+expected event happens.
+
+### Confirm that an event happens
+
+Call ``confirmation(_:expectedCount:fileID:filePath:line:column:_:)``
+in your asynchronous test function to create a `Confirmation` for the
+expected event. In the trailing block parameter, call the code under
```
grynspan, line 36:
```suggestion
expected event. In the trailing closure parameter, call the code under
```

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,70 @@
...
+body, `await` any asynchronous interactions:
+
+```swift
+@Test func priceLookupYieldsExpectedValue() async {
+  let mozarellaPrice = await unitPrice(for: .mozarella)
+  #expect(mozarellaPrice == 3)
+}
+```
+
+In more complex situations you use ``Confirmation`` to discover whether an
```
grynspan, line 29:
```suggestion
In more complex situations you can use ``Confirmation`` to discover whether an
```

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,68 @@
...
+See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+Ensure that your code handles errors in the way you expect.
+
+## Overview
+
+Write tests for your code that validate the conditions in which the
+code throws errors, and the conditions in which it returns without
+throwing an error. Use overrides of the ``expect(_:_:sourceLocation:)`` and
```
grynspan, line 19:
```suggestion
throwing an error. Use overloads of the ``expect(_:_:sourceLocation:)`` and
```

Or "variants", etc. "Overrides" refers to a feature of inheritance which doesn't apply to macros.

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,68 @@
...
+  #expect(throws: PizzaToppings.Error.outOfRange) {
+    try order.add(topping: .mozarella, toPizzasIn: -1..<0)
+  }
+}
+```
+
+If the closure completes without throwing an error, the testing library
+records an issue. Other overloads of ``expect(_:_:sourceLocation:)`` let you
+test that the code throws an error of a given type, or matching an arbitrary
+Boolean test. Similar overloads of ``require(_:_:sourceLocation:)-5l63q`` stop
```
grynspan, line 44:
```suggestion
boolean test. Similar overloads of ``require(_:_:sourceLocation:)-5l63q`` stop
```

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,68 @@
...
+  var order = PizzaToppings(bases: [.calzone, .deepCrust])
+  #expect(throws: PizzaToppings.Error.outOfRange) {
+    try order.add(topping: .mozarella, toPizzasIn: -1..<0)
+  }
+}
+```
+
+If the closure completes without throwing an error, the testing library
+records an issue. Other overloads of ``expect(_:_:sourceLocation:)`` let you
+test that the code throws an error of a given type, or matching an arbitrary
```
grynspan, line 43:
```suggestion
test that the code throws an error of a given type, or matches an arbitrary
```

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -0,0 +1,68 @@
...
+  var order = PizzaToppings(bases: [.thinCrust, .thinCrust])
+  #expect(throws: Never.self) {
+    try order.add(topping: .caper, toPizzasIn: 0..<1)
+  }
+  let toppings = try order.toppings(forPizzaAt: 0)
+  #expect(toppings == [.caper])
+}
+```
+
+If the closure throws an error, the testing library records an issue.
```
grynspan, line 65:
```suggestion
If the closure throws _any_ error, the testing library records an issue.
```

Sources/Testing/Testing.docc/testing-asynchronous-code.md
```diff
@@ -0,0 +1,70 @@
...
+
+See https://swift.org/LICENSE.txt for license information
+See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+Validate whether your code causes expected events to happen.
+
+## Overview
+
+`swift-testing` integrates with Swift concurrency, meaning that in many
```
grynspan, line 17:
```suggestion
The testing library integrates with Swift concurrency, meaning that in many
```

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,15 +10,54 @@ See https://swift.org/LICENSE.txt for license information
...
+
+### Validate your code's result
+
+To validate that your code produces an expected value, use
+``expect(_:_:sourceLocation:)``. ``expect(_:_:sourceLocation:)`` captures the
+expression you pass, and provides detailed information when the code doesn't
+satisfy the expectation.
+
+```swift
+class OrderCalculator {
```
grynspan, line 32: We don't need to define this class in order to demonstrate the testing library (as with other files.)

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -10,15 +10,54 @@ See https://swift.org/LICENSE.txt for license information
...
 
-The testing library provides `#expect()` and `#require()` macros you use to 
-validate expected outcomes. To validate that an error is thrown, or _not_ thrown, 
-the testing library provides several overloads of the macros that you can use.
-Use a ``Confirmation`` to confirm the occurrence of an asynchronous event that
-you can't check directly using an expectation.
+Use ``expect(_:_:sourceLocation:)`` and
+``require(_:_:sourceLocation:)-5l63q`` macros to validate expected
+outcomes. To validate that an error is thrown, or _not_ thrown, the
+testing library provides several overloads of the macros that you can
```
grynspan, line 20: These overloads aren't discussed in this file. Did you want to link out to them?

grynspan (changes requested):

Sources/Testing/Testing.docc/Expectations.md
iamleeg: Yes, I'll also link to the confirmation article as that isn't discussed here either.

grynspan (approved):

PR #410, opened 2024-05-09
Avoid Regex API availability check when no filter or skip flags have been passed

stmontgomery:
Fix several test failures in `SwiftPMTests` when running on older Apple OS versions before `Regex` was added, which is due to unnecessarily checking whether that API is available. One example:

```text
✘ Test "No --filter or --skip argument" recorded an issue at SwiftPMTests.swift:40:4: Caught error: The `--filter' option is not supported on this OS version.
```

In tests which do not involve `--filter` or `--skip` arguments, this check is unnecessary and can be avoided altogether.

### Checklist:

### DIFF
--- a/Sources/Testing/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/EntryPoints/EntryPoint.swift
@@ -158,10 +158,10 @@ public struct __CommandLineArguments_v0: Sendable {
   public var experimentalEventStreamOutput: String?
 
   /// The value(s) of the `--filter` argument.
-  public var filter: [String] = []
+  public var filter: [String]?
 
   /// The value(s) of the `--skip` argument.
-  public var skip: [String] = []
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

grynspan (approved):

Sources/Testing/EntryPoints/EntryPoint.swift
```diff
@@ -158,10 +158,10 @@ public struct __CommandLineArguments_v0: Sendable {
   public var experimentalEventStreamOutput: String?
 
   /// The value(s) of the `--filter` argument.
-  public var filter: [String]?
+  public var filter: [String] = []
```
grynspan, line 161: These need to be optional so that they can be omitted from the encoded JSON.

stmontgomery:
@swift-ci please test

PR #411, opened 2024-05-09
Support forcing color output

plemarquand:
Add support for the `FORCE_COLOR` environment variable. It is possible to run in an environment that does not look like either a TTY or a named pipe and still want ANSI escape codes emitted. The `NO_COLOR` environment variable takes precedence.

### Checklist:

### DIFF
--- a/Sources/Testing/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/EntryPoints/EntryPoint.swift
@@ -439,7 +439,9 @@ extension Event.ConsoleOutputRecorder.Options {
     var result = Self()
 
     result.useANSIEscapeCodes = _fileHandleSupportsANSIEscapeCodes(fileHandle)
-    if result.useANSIEscapeCodes {
+
+    // Respect the FORCE_COLOR environment variable. SEE: https://www.force-color.org
+    if result.useANSIEscapeCodes || Environment.variable(named: "FORCE_COLOR")?.isEmpty == false {
       if let noColor = Environment.variable(named: "NO_COLOR"), !noColor.isEmpty {
         // Respect the NO_COLOR environment variable. SEE: https://www.no-color.org
         result.ansiColorBitDepth = 1


grynspan:
Please update the markup for `XCTestScaffold.runAllTests(hostedBy:_:)` too?

Sources/Testing/EntryPoints/EntryPoint.swift
```diff
@@ -439,7 +439,9 @@ extension Event.ConsoleOutputRecorder.Options {
     var result = Self()
 
     result.useANSIEscapeCodes = _fileHandleSupportsANSIEscapeCodes(fileHandle)
-    if result.useANSIEscapeCodes {
+
+    // Respect the FORCE_COLOR environment variable. SEE: https://www.force-color.org
+    if result.useANSIEscapeCodes || Environment.variable(named: "FORCE_COLOR")?.isEmpty == false {
```
grynspan, line 444: What is the intended interaction between `FORCE_COLOR` and `NO_COLOR` if both are specified?

grynspan (changes requested):
I appreciate the intent here, but `NO_COLOR` is supported specifically for the benefit of SwiftPM which uses pipes rather than TTYs to pass data along to/from `swift test`. `FORCE_COLOR` is not widely supported and the interactions with `NO_COLOR` don't seem to be defined.

Rather than adding another environment variable check, this seems like something we could pass as a command-line argument instead, e.g. `--color` or `--enable-color` or some such.

Sources/Testing/EntryPoints/EntryPoint.swift
plemarquand: The NO_COLOR environment variable takes precedence.

plemarquand:
My thoughts are supporting only `NO_COLOR` is only half of the implementation of controlling terminal colours. Forcing it _off_ when you _do_ have a TTY is just as valid as forcing it _on_ when you _don't_. 

Adding another CLI flag would mean plumbing it through SwiftPM to the swift-testing binary. An environment variable lets anyone opt in all the way down the chain.

There is some [debate](https://bixense.com/clicolors/) around whether to use `FORCE_COLOR` or `CLICOLOR_FORCE`. No preference there.

grynspan:
Neither `CLICOLOR_FORCE ` nor `FORCE_COLOR` is widely used. If you'd like to pursue a solution here, let's use a command-line argument. Thanks!

PR #412, opened 2024-05-14
Dedicated `.bug()` functions for URLs and IDs

grynspan:
One of the features of swift-testing is a test traits system that allows associating metadata with a test suite or test function. One trait in particular, `.bug()`, has the potential for integration with development tools but needs some refinement before integration would be practical.

Read the full proposal [here]().

### Checklist:


Uncommon:
Full proposal link goes right back to this page.

grynspan:
Oops.

PR #414, opened 2024-05-14
Avoid specialization of `unsafeBitCast(_:to:)` when getting an error's backtrace.

grynspan:
It appears the Swift compiler has recently started specializing `unsafeBitCast(_:to:)` more aggressively than before, resulting in a call to that function in `Backtrace.init(forFirstThrowOf:)` crashing because the size of the input error (expected to be an existential box of type `NSError` or `SwiftError`, i.e. actually an object) doesn't match the size of `AnyObject`.

This PR explicitly asks the compiler not to specialize `unsafeBitCast(_:to:)` quite so precisely, preserving the intended semantics.

(This method is already known to contain non-zero quantities of "wat.")

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/Backtrace.swift
+++ b/Sources/Testing/SourceAttribution/Backtrace.swift
@@ -220,7 +220,7 @@ extension Backtrace {
   ///   existential containers with different addresses.
   @inline(never)
   init?(forFirstThrowOf error: any Error) {
-    let errorID = ObjectIdentifier(unsafeBitCast(error, to: AnyObject.self))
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

grynspan:
@swift-ci please test

grynspan:
Continuing to fail on Linux due to the issue fixed in https://github.com/apple/swift-testing/pull/413. Continuing to fail on Windows due to rdar://128092675. Merging as-is.

PR #422, opened 2024-05-17
Always set the XCTest bundle path on Darwin when running exit tests.

grynspan:
Xcode is a complicated beast and there are code paths that lead to us running a test hosted by XCTest that isn't discoverable by the existing exit test implementation. This PR makes that implementation more robust in these scenarios and fixes spurious exit test failures that were resulting.

### Checklist:

### DIFF
--- a/Sources/_TestingInternals/include/Includes.h
+++ b/Sources/_TestingInternals/include/Includes.h
@@ -95,6 +95,10 @@
 #include <wasi/libc-environ.h>
 #endif
 
+#if __has_include(<libgen.h>)
+#include <libgen.h>
+#endif
+
 #if defined(_WIN32)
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci please test

Sources/_TestingInternals/include/Includes.h
```diff
@@ -95,6 +95,10 @@
 #include <wasi/libc-environ.h>
 #endif
 
+#if __has_include(<libgen.h>)
+#include <libgen.h>
```
grynspan, line 99: For `basename_r()`.

briancroom (approved):

grynspan:
Windows failure is a known issue in the compiler.

PR #424, opened 2024-05-21
Explicitly add visionOS availability to tests that check availability.

briancroom:
Since https://github.com/apple/swift/pull/72834, the Swift compiler supports visionOS availability annotations. For completeness, add this platform to existing tests which declare Apple platform availability.

PR #426, opened 2024-05-21
Revert "Explicitly add visionOS availability to tests that check availability. (#424)"

grynspan:
This reverts commit 406564e919a329aab52f803b4c69d712e597a834.

The revert is needed because Swift 5.10 does not recognize visionOS and swift-testing does not currently have the ability to conditionalize availability with `#if compiler(>=5.11)`.

We can reapply the change in the future if/when we require Swift 6.

### Checklist:


grynspan:
@swift-ci please test

PR #431, opened 2024-05-23
Add an umbrella import to the `_TestingInternals` module map file.

grynspan:
This PR removes the individual header declarations from the `_TestingInternals` module map file and replaces them with a single umbrella import. This fixes an existing workflow used by some of our teammates.

~~This PR moves the module map file to align with an existing workflow used by some of our teammates.~~

### Checklist:


grynspan:
@swift-ci please test

Sources/_TestingInternals/module.modulemap
stmontgomery, line 1: If you move this I imagine the CMake build may need to be adjusted

grynspan: I asked Jeff/Saleem to verify I didn't break their changes.

grynspan:
> Are you sure that, when a module map is placed in this location, SwiftPM actually honors it and does not revert back to its default behavior of synthesizing a module map representing all the headers in `include`?

It doesn't actually matter. SwiftPM is fine to synthesize a module here. The module map is only used by CMake and the Xcode project generated by it.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
Windows failure is a known compiler issue.

jeffdav:
I'm on pto but will verify tomorrow!

compnerd:
Tested 004994fef30548d6370dd141dfbb5154cb5bc59c with the latest build from https://github.com/thebrowsercompany/swift-build as the current swift.org toolchains are too old to support the package at this point (due to the use of the stdlib macros).

grynspan:
Curious. I wonder what's confusing CMake there given that there aren't actually any source changes involved.

compnerd:
It's unclear if this change triggered this, just something that I noticed :)

PR #433, opened 2024-05-23
Remove references to Radar from `.bug()` documentation.

grynspan:
Radar is an Apple-only internal tool, so remove references to it from our documentation.

### Checklist:


grynspan:
Documentation change only, skipping CI.

PR #434, opened 2024-05-24
Make Citrine compile in environments that don't have a UTC clock

dennisweissmann:
Add a compile time guard for an access of `Test.Clock.Instant.wall`.

Resolves: rdar://128629302


dennisweissmann:
@swift-ci please test

PR #436, opened 2024-05-24
Fix a misuse of `.bug()` in documentation.

grynspan:
Fixes a typo in an example using `.bug()`.

### Checklist:


grynspan:
Self-reviewing. Documentation only, skipping CI.

PR #448, opened 2024-05-30
Ensure `@Test(arguments:)` macro expansion trims trivia from argument expressions

stmontgomery:
This fixes a build error which can arise if the `@Test(arguments:)` attribute applied to a parameterized test function includes trailing trivia such as a `//`-style line comment on the end of the line. For example:

```swift
@Test(
  arguments: [...] // some comment
)
func example(...) { ... }
```

This regressed with my changes in #366 which began surrounding the argument expressions in a closure expression, in order to facilitate lazy evaluation. The fix ensures that the wrapped expression is trimmed, in addition to the outer closure (which was, and still is, trimmed).

### Motivation:

Comments in this position may not be used often, but can be useful to provide context about the role or meaning of a particular collection of arguments passed to a parameterized test. So we'd like to continue supporting this.

### Modifications:

- Trim trivia from the argument expression included in the `@Test` macro expansion code.
- Add test reproducing and validating this scenario.

### Result:

Parameterized test functions with trailing line comments on the `arguments:` line now produce valid macro expansion code and build successfully.

### Checklist:

Resolves rdar://128609944

### DIFF
--- a/Sources/TestingMacros/Support/AttributeDiscovery.swift
+++ b/Sources/TestingMacros/Support/AttributeDiscovery.swift
@@ -190,7 +190,7 @@ struct AttributeInfo {
     if let argumentsIndex = otherArguments.firstIndex(where: { $0.label?.tokenKind == .identifier("arguments") }) {
       for index in argumentsIndex ..< otherArguments.endIndex {
         var argument = otherArguments[index]
-        argument.expression = .init(ClosureExprSyntax { argument.expression })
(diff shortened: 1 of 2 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test Linux

briancroom (approved):

stmontgomery:
Windows CI failure is known and tracked elsewhere, merging…

PR #450, opened 2024-06-03
Include fix-its in the diagnostic on a parameterized test when `@Test` is missing `arguments:`

stmontgomery:
This adds fix-its to the diagnostic emitted by the `@Test` macro when it has been applied to a function which includes parameters but does not specify `arguments: ...`.

### Motivation:

When transforming a non-parameterized `@Test` function into a parameterized one, it's common to begin by adding one or more parameters to the function's signature. Immediately after doing that, the `@Test` macro emits an error diagnostic if `arguments:` is not passed to `@Test`. Since this is a common workflow and it may not always be obvious to new users where to add `arguments:`, it would be beneficial for this diagnostic to include a fix-it.

### Modifications:

- Add a couple of styles of fix-it to this diagnostic, with variations for single- and multiple-parameter functions.
- Enhance a utility on `EditorPlaceholderExprSyntax` to prefer typed placeholders (see code comment).

### Checklist:

Resolves rdar://128396593

### DIFF
--- a/Sources/TestingMacros/Support/Additions/EditorPlaceholderExprSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/EditorPlaceholderExprSyntaxAdditions.swift
@@ -11,12 +11,28 @@
 import SwiftSyntax
 
 extension EditorPlaceholderExprSyntax {
-  /// Initialize an instance of this type with the given placeholder string.
+  /// Initialize an instance of this type with the given display name string and
+  /// optional type.
   ///
   /// - Parameters:
-  ///   - placeholder: The placeholder string, not including surrounding angle
+  ///   - displayName: The display name string, not including surrounding angle
   ///     brackets or pound characters.
-  init(_ placeholder: String) {
-    self.init(placeholder: .identifier("<# \(placeholder) #" + ">"))
+  ///   - type: The type which this placeholder have, if any. When non-`nil`,
+  ///     the expression will use typed placeholder syntax.
+  init(_ displayName: String, type: String? = nil) {
+    let placeholderString = if let type {
+      // This uses typed placeholder syntax, which allows the compiler to
+      // type-check the expression successfully. The resulting code still does
+      // not compile due to the placeholder, but it makes the diagnostic more
+      // clear. See
+      // https://developer.apple.com/documentation/swift-playgrounds/specifying-editable-regions-in-a-playground-page#Mark-Editable-Areas-with-Placeholder-Tokens
+      "T##\(displayName)##\(type)"
+    } else {
+      displayName
+    }
+
+    // Manually concatenate the string to avoid it being interpreted as a
+    // placeholder when editing this file.
+    self.init(placeholder: .identifier("<#\(placeholderString)#" + ">"))
   }
 }
--- a/Sources/TestingMacros/Support/Additions/EditorPlaceholderExprSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/EditorPlaceholderExprSyntaxAdditions.swift
@@ -20,7 +20,26 @@ extension EditorPlaceholderExprSyntax {
   ///   - type: The type which this placeholder have, if any. When non-`nil`,
   ///     the expression will use typed placeholder syntax.
   init(_ placeholder: String, type: String? = nil) {
-    self.init(placeholder: .identifier(_editorPlaceholder(placeholder, type: type)))
+    let placeholderContent = if let type {
+      // These use typed placeholder syntax, which allows the compiler to
+      // type-check the expression successfully. The resulting code still does
+      // not compile due to the placeholder, but it makes the diagnostic more
+      // clear. See
+      // https://developer.apple.com/documentation/swift-playgrounds/specifying-editable-regions-in-a-playground-page#Mark-Editable-Areas-with-Placeholder-Tokens
+      if placeholder == type {
+        // When the placeholder string is exactly the same as the type string,
+        // use the shorter typed placeholder format.
+        "T##\(placeholder)"
+      } else {
+        "T##\(placeholder)##\(type)"
+      }
+    } else {
+      placeholder
+    }
+
+    // Manually concatenate the string to avoid it being interpreted as a
+    // placeholder when editing this file.
+    self.init(placeholder: .identifier("<#\(placeholderContent)#" + ">"))
   }
 
   /// Initialize an instance of this type with the given type, using that as the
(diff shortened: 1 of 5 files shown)


stmontgomery:
@swift-ci please test

Sources/TestingMacros/Support/Additions/EditorPlaceholderExprSyntaxAdditions.swift
```diff
@@ -15,12 +15,28 @@ public import SwiftSyntax
 #endif
 
 extension EditorPlaceholderExprSyntax {
-  /// Initialize an instance of this type with the given placeholder string.
+  /// Initialize an instance of this type with the given display name string and
```
grynspan, line 18: Neat!

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -438,32 +438,106 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+
+      baseArguments = existingLabeledArguments
+    } else {
+      baseArguments = .init()
+    }
+
+    var fixIts: [FixIt] = []
+    func addFixIt(_ message: String, appendingArguments arguments: some Collection<LabeledExprSyntax>) {
+      var newAttribute = attribute
+      newAttribute.attributeName = newAttribute.attributeName.trimmed
```
grynspan, line 496: We might want to preserve trailing trivia (specifically comments) after `rightParen`.

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -438,32 +438,106 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+    // type.
+    do {
+      let argumentsCollectionType = if parameters.count == 1, let parameter = parameters.first {
+        "[\(parameter.baseTypeName)]"
+      } else {
+        "[(\(parameters.map(\.baseTypeName).joined(separator: ", ")))]"
+      }
+
+      addFixIt(
+        parameters.count == 1 ? "Add 'arguments:'" : "Add 'arguments:' with one collection",
```
grynspan, line 518: What's the distinction here? Can we just have the same label?

Sources/TestingMacros/Support/Additions/EditorPlaceholderExprSyntaxAdditions.swift
```diff
@@ -15,12 +15,28 @@ public import SwiftSyntax
...
+  /// optional type.
   ///
   /// - Parameters:
-  ///   - placeholder: The placeholder string, not including surrounding angle
+  ///   - displayName: The display name string, not including surrounding angle
```
grynspan, line 22: This isn't a display name. It's the placeholder text.

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -438,32 +438,106 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+        let label = index == parameters.startIndex ? "arguments" : nil
+        let argumentsCollectionType = "[\(parameters[index].baseTypeName)]"
+        return LabeledExprSyntax(
+          label: label.map { .identifier($0) },
+          colon: label == nil ? nil : .colonToken(trailingTrivia: .space),
+          expression: EditorPlaceholderExprSyntax(argumentsCollectionType, type: argumentsCollectionType),
+          trailingComma: parameters.index(after: index) < parameters.endIndex ? .commaToken(trailingTrivia: .space) : nil
+        )
+      }
+      addFixIt("Add 'arguments:' with all combinations of \(parameters.count) collections", appendingArguments: additionalArguments)
```
grynspan, line 537: I'm confused as to the different use cases for different Fix-Its here.

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -438,32 +438,106 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
     }
+
+    // Fix-It to add 'arguments:' with all combinations of <N> collections,
+    // where <N> is the count of the function's parameters. Only offered for
+    // functions with 2 or more parameters.
+    if parameters.count >= 2 {
+      let additionalArguments = parameters.indices.map { index in
+        let label = index == parameters.startIndex ? "arguments" : nil
+        let argumentsCollectionType = "[\(parameters[index].baseTypeName)]"
+        return LabeledExprSyntax(
```
grynspan, line 530: Consider using our built-in `Argument` type instead, which when cast to `LabeledExprSyntax` ought to always do the right thing.

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -438,32 +438,106 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+    do {
+      let argumentsCollectionType = if parameters.count == 1, let parameter = parameters.first {
+        "[\(parameter.baseTypeName)]"
+      } else {
+        "[(\(parameters.map(\.baseTypeName).joined(separator: ", ")))]"
+      }
+
+      addFixIt(
+        parameters.count == 1 ? "Add 'arguments:'" : "Add 'arguments:' with one collection",
+        appendingArguments: [LabeledExprSyntax(label: "arguments", expression: EditorPlaceholderExprSyntax(argumentsCollectionType, type: argumentsCollectionType))]
```
grynspan, line 519: I'm not certain this would be the right placeholder text. It's, AFAIK, typically some short human-readable string. In this case, maybe "arguments" or "collection". I wouldn't expect to see the name of the type (`"[T]"`) here, and an arbitrary string in the position runs the risk of breaking Fix-It display if it happens to contain special characters.

grynspan (changes requested):
Please add a new parameterized test that covers all the code paths for the new Fix-It code.

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -438,32 +438,106 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
         message: "Attribute \(_macroName(attribute)) cannot specify arguments when used with '\(functionDecl.completeName)' because it does not take any",
         severity: .error
       )
-    case 1:
+    } else {
       return Self(
         syntax: Syntax(functionDecl),
-        message: "Attribute \(_macroName(attribute)) must specify an argument when used with '\(functionDecl.completeName)'",
-        severity: .error
+        message: "Attribute \(_macroName(attribute)) must specify arguments when used with '\(functionDecl.completeName)'",
```
grynspan, line 456: We should add "function" to both messages so it reads like "Attribute 'Test' must specify arguments when used with **function** 'foo(x:)'". I think that's stylistically closer to what the compiler emits.

grynspan:
@swift-ci please test Windows

Sources/TestingMacros/Support/DiagnosticMessage.swift
stmontgomery: I see quite a few examples in the Swift compiler repo which use this pattern, where the placeholder string itself provides a hint about the _type_ of the expression which the user should replace it with. In our case here, I think it's especially helpful when there are two collection arguments because it clarifies which one is which.

Sources/TestingMacros/Support/DiagnosticMessage.swift
stmontgomery: Updated the logic to preserve trailing trivia, and the new unit tests verify this

stmontgomery:
> Please add a new parameterized test that covers all the code paths for the new Fix-It code.

I've now added a comprehensive unit test to validate the fix-its on this diagnostic under various scenarios

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/TestingMacros/Support/Additions/EditorPlaceholderExprSyntaxAdditions.swift
```diff
@@ -11,12 +11,70 @@
...
+
+/// Format a string to be included in an editor placeholder expression using the
+/// specified type, using that type as the placeholder text.
+///
+/// - Parameters:
+///   - type: The type to use both as the placeholder text and as the
+///     expression's type.
+///
+/// - Returns: A formatted editor placeholder string.
+func editorPlaceholder(forType type: String) -> String {
```
grynspan, line 78: Does this actually need to be fully broken out?

Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
```diff
@@ -158,6 +154,91 @@ struct TestDeclarationMacroTests {
...
 
+  @Test("Error diagnostics which include fix-its emitted on API misuse", arguments: [
+    // 'Test' attribute must specify arguments to parameterized test functions.
+    "@Test func f(i: Int) {}":
+      (
+        message: "Attribute 'Test' must specify arguments when used with function 'f(i:)'",
+        fixIts: [
+          ExpectedFixIt(
+            message: "Add 'arguments:' with one collection",
+            changes: [.replace(oldSourceCode: "@Test ", newSourceCode: "@Test(arguments: \(editorPlaceholder(forType: "[Int]"))) ")]
```
grynspan, line 165: You should be able to pass an instance of `EditorPlaceholderExprSyntax` here and get it stringified correctly, no?

Sources/TestingMacros/Support/DiagnosticMessage.swift
grynspan: I still would suggest using `Argument` here.

Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
stmontgomery: Nice simplification, thanks

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/TestingMacros/Support/DiagnosticMessage.swift
stmontgomery: I've attempted to do this but found that as it stands currently, the `Argument` helper type still requires me to do the same book-keeping for trailing trivia like `commaToken` on the `LabeledExprSyntax` nodes. I'm not quite sure what the benefit of this type is in this circumstance.

I'm open to it, but would ask for a bit more specific guidance on how you recommend applying it (once you're available again). In the mean time I'm planning to proceed with merging this since we've gone through a couple rounds and I think the PR is in good shape overall at this point

PR #455, opened 2024-06-05
Reduce the number of calls to `Environment.variable(named:)`.

grynspan:
This PR stores the value of the `"SWT_PRETTY_PRINT_JSON"` environment variable that we sometimes use for debugging JSON output. We were previously reading that environment variable hundreds of thousands of times just in our own tests. Oops.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/JSON.swift
+++ b/Sources/Testing/Support/JSON.swift
@@ -13,6 +13,12 @@ private import Foundation
 #endif
 
 enum JSON {
+  /// Whether or not pretty-printed JSON is enabled for this process.
+  ///
+  /// This is a debugging tool that can be used by developers working on the
+  /// testing library to improve the readability of JSON output.
+  private static let _prettyPrintingEnabled = Environment.flag(named: "SWT_PRETTY_PRINT_JSON") == true
+
   /// Encode a value as JSON.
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

Sources/Testing/Support/JSON.swift
```diff
@@ -13,6 +13,12 @@ private import Foundation
 #endif
 
 enum JSON {
+  /// Whether or not pretty-printed JSON is enabled for this process.
+  ///
+  /// This is a debugging tool that can be used by developers working on the
+  /// testing library to improve the readability of JSON output.
+  private static let _prettyPrintingEnabled = Environment.flag(named: "SWT_PRETTY_PRINT_JSON") == true
```
stmontgomery, line 20: Nit
```suggestion
  private static let _isPrettyPrintingEnabled = Environment.flag(named: "SWT_PRETTY_PRINT_JSON") == true
```

stmontgomery (approved):

Sources/Testing/Support/JSON.swift
grynspan: Fair, but not worth running CI again.

PR #470, opened 2024-06-12
Switch doc generation to use 6.0

finestructure:
NB: we always build docs with the latest _release_ version of Swift by default. So once Swift 6 is released this change will be redundant and moreover would pin doc gen to 6.0 even once 6.1 released and should be reverted by then.

Tested to be working ok with our doc gen mechanism:

CleanShot 2024-06-12 at 14 39 35@2x

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -3,5 +3,8 @@ metadata:
   authors: Apple Inc.
 builder:
   configs:
-    - documentation_targets: [Testing]
+    - swift_version: 5.10
+      scheme: Testing
+    - swift_version: 6.0
+      documentation_targets: [Testing]
       scheme: Testing


.spi.yml
```diff
@@ -3,5 +3,8 @@ metadata:
   authors: Apple Inc.
 builder:
   configs:
-    - documentation_targets: [Testing]
+    - swift_version: 5.10
```
grynspan, line 6: We're planning to drop 5.10 support shortly (see #467.) Would we remove these two lines then?

grynspan:
@swift-ci please test

.spi.yml
finestructure: Yes, just change it to

```
builder:
  configs:
    - swift_version: 6.0
      documentation_targets: [Testing]
      scheme: Testing
```

then! (And drop the `swift_version: 6.0` line once Swift 6.0 is released.)

stmontgomery (approved):

.spi.yml
grynspan: Filed #471 to track.

PR #476, opened 2024-06-12
Add an experimental console output prefix option for command-line output.

### LINKED ISSUE #472: Configurable log prefix
### Description

We running test with lots of log output that are hard to run locally end-to-end, so we often have to rely on CI to get the results.

Getting these results is difficult because of how noisy the logs are. It'd be very helpful if we could set a custom prefix that is prepended to all testing output lines. That way we can filter/search the log chaos on a know string in the CI viewer and zero in on just the test failure output.

Would be great if this was a global config or even a per-test trait.

### Expected behavior

-

### Actual behavior

-

### Steps to reproduce

_No response_

_No response_

_No response_

grynspan:
This PR adds the ability to specify a prefix that should be applied to all output from the command line when using `swift test`. For example, if you specify the prefix "`[WHEE]`", then you'll get output like:

> [WHEE] ◇ Test run started.
> [WHEE] ↳ Testing Library Version: ff2775952f346dd94757adffa7f7ba57bcc4b2d1 (modified)
> [WHEE] ◇ Suite "My suite" started.
(etc.)

Since the `--experimental-console-output-prefix` argument is not recognized by Swift Package Manager, you can also specify the prefix using an environment variable, `"SWT_EXPERIMENTAL_CONSOLE_OUTPUT_PREFIX"`.

Resolves #472.
Resolves rdar://129677014.

### Checklist:

### DIFF
--- a/Sources/Testing/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/EntryPoints/EntryPoint.swift
@@ -66,6 +66,9 @@ func entryPoint(passing args: __CommandLineArguments_v0?, eventHandler: Event.Ha
       var options = Event.ConsoleOutputRecorder.Options()
       options = .for(.stderr)
       options.verbosity = args.verbosity
+      if let prefix = args.experimentalConsoleOutputPrefix {
+        options.prefix = prefix
+      }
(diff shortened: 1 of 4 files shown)


grynspan:
@swift-ci please test

Issue #474, opened 2024-06-12
Allow direct creation of Issues


younata:
### Description

Xcode's XCTest has the XCTIssue API for reporting issues, but the closest thing to creating and reporting an XCTIssue is `Issue.record`. Which has limitations and prevents me from providing semantics for why an expectation failed.

This limits the ability of [Nimble](https://github.com/Quick/nimble) to integrate with Swift Testing.

grynspan:
Discussed in a lab.

We've so far resisted doing this in the general case because we want to steer developers toward specific API like `#expect()`. @younata's use case is integration with tools however, which requires a different set of interfaces from what we provide for test authors.

We currently offer a number of interfaces guarded by `@_spi(ForToolsIntegrationOnly)`. These interfaces are available when building Swift Testing from a package. However, SPI is not available when using the copy of Swift Testing included in Xcode 16.

Would it be useful if we added an `Issue.init` member guarded by `@_spi` in this way? Would that be sufficient for your needs?

younata:
Hi @grynspan! Thanks for the quick response!

My concern with using `@_spi` (which I am unfamiliar with) is if this will affect Nimble's Cocoapods & Carthage integrations. For the Swift Packages integration, `@_spi` will work fine. But, naively, I'm planning to try to get Nimble's Cocoapods & Carthage integrations to use the built-in Swift Testing. Last I checked (and google isn't helping me here).
This is because there isn't a way for me to tell Cocoapods that I have a dependency on something that's only available as a swift package. I know for a fact that if I were to embed Swift Testing as a Package for the Carthage integration, then the tests will refuse to build unless you also add the Swift Testing package as a dependency to your app (which I know because [Nimble's Carthage integration is currently broken for the same reason](https://github.com/Quick/Quick/issues/1275)).

Blocking the integration on tools-only `@_spi` looks promising, but I have reservations about if it will work with the other package manager Nimble purports to support.

stmontgomery:
I think we do need to expose the ability to programmatically initialize and record an `Issue` and have that eventually become regular API. I can take this. I can't currently commit to any particular timeframe for delivering this functionality, however.

If anyone would like to get started drafting an API proposal (using our [template](https://github.com/apple/swift-testing/blob/main/Documentation/Proposals/0000-proposal-template.md)) for this, that would certainly be welcome!

stmontgomery:
Tracked internally as rdar://114106842

younata:
[...]

PR #481, opened 2024-06-13
[SWT-NNNN] Make the Issues API public & provide a public `Issue.record` instance method.

younata:
See #474 for the related bug.

Read the full proposal [here](https://github.com/younata/swift-testing/blob/public-issues-pitch/Documentation/Proposals/NNNN-public-issues-api.md)!

### Motivation

To better facilitate integrating Swift Testing with third party assertion tools, Swift Testing should allow public access to creating and recording Issues directly.

### Modifications:

At this stage, the PR is just the pitch doc, and I'm more than happy to contribute the small change as we discuss it. I figured it would be better to agree on the pitch first?

### Result:

Once merged, this will open up third-party assertion tools to better integrate with Swift Testing.

### Checklist:

### DIFF
--- a/Documentation/Proposals/NNNN-public-issues-api.md
+++ b/Documentation/Proposals/NNNN-public-issues-api.md
@@ -71,7 +71,7 @@ arguments, and simply calls the `Issue.record(configuration:)` instance
 method with nil:
 
 ```swift
-extension Issue: {
+extension Issue {
   /// Record this issue by wrapping it in an ``Event`` and passing it to the
   /// current event handler.
   ///
--- a/Documentation/Proposals/NNNN-public-issues-api.md
+++ b/Documentation/Proposals/NNNN-public-issues-api.md
@@ -1,6 +1,6 @@
-# Feature name
+# Public Issues API
 
-* Proposal: [SWT-0002](0002-public-issues-api.md)
+* Proposal: [SWT-NNNN](NNNN-public-issues-api.md)
 * Authors: [Rachel Brindle](https://github.com/younata)
 * Status: **Awaiting implementation**
 * Bug: [apple/swift-testing#474](https://github.com/apple/swift-testing/issues/474)
@@ -14,18 +14,24 @@ better integration with third-party assertion tools.
 
 ## Motivation
 
+There are several third-party assertion libraries that developers are used to
+using, and Swift Testing should make it as easy as possible for the developers
+and maintainers of these third-party assertion libraries to integrate with
+Swift Testing.
+
 Swift Testing currently offers the `Issue.record` family of static methods,
 which provides an acceptable interface for one-off/custom assertions without
 using the `#expect` or `#require` macros. Unfortunately, `Issue.record` is not
 robust enough for dedicated assertion libraries like Nimble. For example,
 there is no current way to provide an `Issue.Kind` to `Issue.record`.
 
+
 ## Proposed solution
 
 I propose making the initializer for `Issue` public, and provide a public
 `Issue.record()` instance method to report that Issue. This provides a clean,
-easy to maintain way to create and record custom Issues that assertion libraries
-are already used to using because it is analogous to the XCTIssue API in XCTest.
+easy to maintain way to create and record custom Issues that assertion tools
+are already used to because it is analogous to the XCTIssue API in XCTest.
 
 ## Detailed design
 
@@ -60,8 +66,9 @@ public struct Issue: Sendable {
 }
 ```
 
-Additionally, create a public `record()` method on Issue, which takes no arguments,
-and simply calls the `Issue.record(configuration:)` instance method with nil:
+Additionally, create a public `record()` method on Issue, which takes no
+arguments, and simply calls the `Issue.record(configuration:)` instance
+method with nil:
 
 ```swift
 extension Issue: {
@@ -83,41 +90,56 @@ private. Only new code and making existing code public.
 
 ## Integration with supporting tools
 
-Third-party assertion libraries would be able to directly create an `Issue`
+Third-party assertion tools would be able to directly create an `Issue`
 and then report it using the `Issue.report()` instance method. `Issue.report()`
-would then work similarly to how it does now. This flow is analogous to reporting
-an issue in XCTest using the XCTIssue API.
+would then work similarly to how it does now. This flow is analogous to
+reporting an issue in XCTest using the XCTIssue API.
 
 ## Future directions
 
-A future direction is to remove all of the `@_spi(ForToolsIntegrationOnly)`
-annotations. This would allowing third party tools to integrate with Swift
-Testing regardless of the distribution method they use.
+Also necessary for supporting third-party assertion tools is providing a stable
+and reliable API for accessing the current test. This somewhat exists with the
+current `Test.current` static property, but that returns the
+semantically-incorrect value of nil if you access it from a detached task. This
+will be defined in a future proposal.
 
+A future direction is to remove all of the `@_spi(ForToolsIntegrationOnly)`
+annotations. This would allow third party tools to integrate with Swift
+Testing regardless of the distribution method they use.
 This was left out to keep this proposal small & focused, but the same rationale
-behind making the `Issue` initializer public and available everywhere applies to
-removing the `@_spi(ForToolsIntegrationOnly)` annotation from everything.
+behind making the `Issue` initializer public and available everywhere applies
+to removing the `@_spi(ForToolsIntegrationOnly)` annotation from everything.
 
 ## Alternatives considered
 
 One potential approach is to extend or provide overloads to `Issue.report` that
 allows developers to specify every property on an `Issue`. However, this is
 undesirable from a maintenance point of view: every time a property is added to
-or removed from `Issue`, we would similarly have to update `Issue.report`. While
-trivial, having that extra source of truth is mildly annoying.
+or removed from `Issue`, we would similarly have to update `Issue.report`.
+While trivial, having that extra source of truth is mildly annoying.
 
-Another approach discussed in [apple/swift-testing#474](https://github.com/apple/swift-testing/issues/474)
+Another approach discussed in
+[apple/swift-testing#474](https://github.com/apple/swift-testing/issues/474)
 is making `Issue` public, but tagged with `@_spi(ForToolsIntegrationOnly)`.
-Which is also not desirable because that limits third party integrations to only
-supporting Swift Testing when managed through Swift Package Manager. Which means
-that third party integrations would not be able to support Swift Testing when
-managed using Cocoapods, Carthage, or even just by depending on the raw code.
-This increases the support burden on those third party tools because they then
-have to explain why they only support Swift Testing when managed through
-Swift Package Manager.
+Which is also not desirable because that limits third party integrations to
+only supporting Swift Testing when managed through Swift Package Manager.
+Which means that third party integrations would not be able to support Swift
+Testing when managed using Cocoapods, Carthage, or even just by depending on
+the raw code. This increases the support burden on those third party tools
+because they then have to explain why they only support Swift Testing when
+managed through Swift Package Manager.
+
+I also considered combining this with a proposal to provide a more reliable
+API for accessing the current test, as mentioned in the Future Directions
+section. I decided against this because while that problem is motivation by the
+same motivation behind this proposal, it is also a significantly harder
+problem. Which is deserving of a full document describing the problem and the
+solution. Additionally, the concern behind this proposal is to provide the same
+access to issue reporting that `#expect` and `#require` enjoy, while making
+accessing the current test more reliable and robust is a separate concern.
 
 ## Acknowledgments
 
 I'd like to thank [Jonathon Grynspan](https://github.com/grynspan) and
-[Stuart Montgomery](https://github.com/stmontgomery) for fielding my Issue report
-and encouraging me to contribute more to this community.
+[Stuart Montgomery](https://github.com/stmontgomery) for fielding my Issu
+report and encouraging me to contribute more to this community.
--- a/Documentation/Proposals/NNNN-public-issues-api.md
+++ b/Documentation/Proposals/NNNN-public-issues-api.md
@@ -2,9 +2,9 @@
 
 * Proposal: [SWT-NNNN](NNNN-public-issues-api.md)
 * Authors: [Rachel Brindle](https://github.com/younata)
-* Status: **Awaiting implementation**
+* Status: **Awaiting review**
 * Bug: [apple/swift-testing#474](https://github.com/apple/swift-testing/issues/474)
-* Implementation: n/a
+* Implementation: [apple/swift-testing#481](https://github.com/apple/swift-testing/pull/481)
 * Review: n/a
 
 ## Introduction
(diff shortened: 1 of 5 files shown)


grynspan:
One minor thing that worries me is that this allows creating issues of type `.system` which is meant to be reserved by the testing library. We can relax that definition, but I'm not sure we _want_ to?

As well, should we consider deprecating the static `record()` functions? They exist to allow recording arbitrary issues without exposing a full initializer, so they're redundant with this change.

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,145 @@
...
+```swift
+extension Issue: {
+  /// Record this issue by wrapping it in an ``Event`` and passing it to the
+  /// current event handler.
+  ///
+  /// - Returns: The issue that was recorded (`self` or a modified copy of it.)
+  @discardableResult
+  public func record() -> Self {
+    record(configuration: nil)
+  }
```
stmontgomery, line 82: For the purposes of a proposal, you can omit the body of the method

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,145 @@
...
+
+## Source compatibility
+
+This is strictly an additive change. No deletion of code, nothing is being made
+private. Only new code and making existing code public.
+
+## Integration with supporting tools
+
+Third-party assertion tools would be able to directly create an `Issue`
+and then report it using the `Issue.report()` instance method. `Issue.report()`
```
stmontgomery, line 94: These say `report()` but the actual proposed method is named `record()`, can you update those to be consistent throughout the document?

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,145 @@
...
+current `Test.current` static property, but that returns the
+semantically-incorrect value of nil if you access it from a detached task. This
+will be defined in a future proposal.
+
+A future direction is to remove all of the `@_spi(ForToolsIntegrationOnly)`
+annotations. This would allow third party tools to integrate with Swift
+Testing regardless of the distribution method they use.
+This was left out to keep this proposal small & focused, but the same rationale
+behind making the `Issue` initializer public and available everywhere applies
+to removing the `@_spi(ForToolsIntegrationOnly)` annotation from everything.
```
stmontgomery, line 111: I don't think it's actually a goal per se to eventually remove _every_ usage of `@_spi(ForToolsIntegrationOnly)`. There may be things that users of this library may never need, but tools always need. APIs marked `@_spi(Experimental)`, however, are more likely to be considered for promotion to public API. There's more context and background about this in `SPI.md`(https://github.com/apple/swift-testing/blob/main/Documentation/SPI.md).

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,145 @@
...
+same motivation behind this proposal, it is also a significantly harder
+problem. Which is deserving of a full document describing the problem and the
+solution. Additionally, the concern behind this proposal is to provide the same
+access to issue reporting that `#expect` and `#require` enjoy, while making
+accessing the current test more reliable and robust is a separate concern.
+
+## Acknowledgments
+
+I'd like to thank [Jonathon Grynspan](https://github.com/grynspan) and
+[Stuart Montgomery](https://github.com/stmontgomery) for fielding my Issu
```
stmontgomery, line 144: typo
```suggestion
[Stuart Montgomery](https://github.com/stmontgomery) for fielding my Issue
```

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,145 @@
...
+  ///     arguments.
+  public init(
+    kind: Kind,
+    comments: [Comment],
+    sourceContext: SourceContext = .init()
+  ) {
+    self.kind = kind
+    self.comments = comments
+    self.sourceContext = sourceContext
+  }
```
stmontgomery, line 63: Here too, you can omit the body of the initializer — only the signature is needed for review purposes

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,145 @@
...
+  ///   - kind: The kind of issue this value represents.
+  ///   - comments: An array of comments describing the issue. This array may be
+  ///     empty.
+  ///   - sourceContext: A ``SourceContext`` indicating where and how this issue
+  ///     occurred. This defaults to a default source context returned by
+  ///     calling ``SourceContext/init(backtrace:sourceLocation:)`` with zero
+  ///     arguments.
+  public init(
+    kind: Kind,
+    comments: [Comment],
```
stmontgomery, line 57: I think we can provide a default `= []` empty array value for the `comments:` parameter

stmontgomery:
One thing which is likely to pose a challenge with _using_ this API as proposed is that other related types don't yet expose a `public` initializer, either. Perhaps most important is the `Expectation` type. It's used as an associated value of the `Issue.Kind.expectationFailed(_ expectation: Expectation)` enum case.

In order to correctly represent an expectation failure, a client would need to first programmatically construct an `Expectation`, which implies that more APIs may need to be exposed than just the ones mentioned here. One approach could be to start conservatively and identify only a small subset of the `Expectation` API which is likely needed to express basic expectation failures, and not attempt to include as much detail as the built-in `#expect` does.

I'd recommend looking through the `Issue.Kind` enum and evaluating how each one might be used, or whether it _should_ be used at all (as @grynspan mentioned, `.system` doesn't seem relevant).

younata:
> One thing which is likely to pose a challenge with using this API as proposed is that other related types don't yet expose a public initializer, either. Perhaps most important is the Expectation type. It's used as an associated value of the Issue.Kind.expectationFailed(_ expectation: Expectation) enum case.

Thanks for the review!

I think a decent next-steps for me would be to go and implement this in my branch, and check out what needs to be modified to actually do this.

younata:
Update: I added an implementation, which I used with Nimble to actually create an integration.

You can check out the [basic-swift-testing-support branch on Nimble](https://github.com/Quick/Nimble/tree/basic-swift-testing-support) if you'd like to try it out. Both the xcodeproj and the Package.swift use this branch of swift-testing. Run the [`SwiftTestingSupportSuite`](https://github.com/Quick/Nimble/blob/basic-swift-testing-support/Tests/NimbleTests/SwiftTestingSupportTest.swift) to verify that the error is properly recorded in Swift Testing.

Sources/Testing/Expectations/Expectation.swift
```diff
@@ -18,7 +18,6 @@ public struct Expectation: Sendable {
   ///
   /// If this expectation passed, the value of this property is `nil` because no
   /// error mismatch occurred.
-  @_spi(ForToolsIntegrationOnly)
```
grynspan, line 21: This property should probably stay SPI for now as it's a bit hacky and probably not the final design for this data.

Sources/Testing/Expectations/Expectation.swift
```diff
@@ -27,7 +26,6 @@ public struct Expectation: Sendable {
   /// If this expectation passed, the value of this property is `nil` because
   /// the difference is only computed when necessary to assist with diagnosing
   /// test failures.
-  @_spi(ForToolsIntegrationOnly)
```
grynspan, line 30: This property should probably stay SPI for now as it's a bit hacky and probably not the final design for this data.

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -54,6 +54,15 @@ extension Issue {
     return issue.record(configuration: configuration)
   }
 
+  /// Record this issue by wrapping it in an ``Event`` and passing it to the
```
grynspan, line 57: I don't think we really discuss events or event handlers in our API documentation because they're not otherwise part of our API surface. How about simply:

```swift
/// Record this issue.
```

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -54,6 +54,15 @@ extension Issue {
...
+  /// current event handler.
+  ///
+  /// - Returns: The issue that was recorded (`self` or a modified copy of it.)
```
grynspan, line 60: We should make sure the API documentation explains _what_ might be modified (the SPI documentation is, and this is a technical term here, "lazy".)

Sources/Testing/Issues/Issue.swift
```diff
@@ -65,6 +65,7 @@ public struct Issue: Sendable {
 
     /// An issue due to a failure in the underlying system, not due to a failure
     /// within the tests being run.
+    @_spi(ForToolsIntegrationOnly)
```
grynspan, line 68: We should leave `.system` as SPI for now since we don't generally want external components to generate system issues. These represent problems with Swift Testing itself. (We could maybe add an additional case e.g. `case integratedToolFailed(description: String)`, what do you think @stmontgomery ?)

Sources/Testing/Issues/Issue.swift
```diff
@@ -93,9 +94,9 @@ public struct Issue: Sendable {
   ///     occurred. This defaults to a default source context returned by
   ///     calling ``SourceContext/init(backtrace:sourceLocation:)`` with zero
   ///     arguments.
-  init(
+  public init(
```
grynspan, line 97: This is fine but will produce a source context for the issue that points into Swift Testing rather than the call site, which is an oversight of the existing code to be fair. How about we expose a different initializer with the signature:

```swift
init(
  _ kind: Kind,
  comments: [Comment] = [],
)
```

(Whether or not `kind` deserves a label is something we can bikeshed.)

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,178 @@
...
+  ///     be empty.
+  ///   - sourceContext: A ``SourceContext`` indicating where and how this
+  ///     issue occurred. This defaults to a default source context returned by
+  ///     calling ``SourceContext/init(backtrace:sourceLocation:)`` with zero
+  ///     arguments.
+  public init(
+    kind: Kind,
+    comments: [Comment] = [],
+    sourceContext: SourceContext = .init()
+  ) {}
```
grynspan, line 64: Proposals don't need to include the bodies of methods/initializers/etc. You can leave the squiggly brackets out here.

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,178 @@
...
+This change requires us to eliminate the default value of `nil` for the
+`Issue.record(configuration:)` instance method:
+
+```swift
+extension Issue {
+  func record(configuration: Configuration?)  -> Self {}
+}
+```
+
+Furthermore, to support passing in any `Issue.Kind` to `Issue`, `Expectation`
```
grynspan, line 94: It is intentional that `Expectation` does not have a public initializer: it is always supposed to be the product of `#expect` or `#require`. If you record an issue that does not come from one of those macros, it should not be producing an instance of this type.

Documentation/Proposals/NNNN-public-issues-api.md
```diff
@@ -0,0 +1,178 @@
...
+section. I decided against this because while that problem is motivation by the
+same motivation behind this proposal, it is also a significantly harder
+problem. Which is deserving of a full document describing the problem and the
+solution. Additionally, the concern behind this proposal is to provide the same
+access to issue reporting that `#expect` and `#require` enjoy, while making
+accessing the current test more reliable and robust is a separate concern.
+
+## Acknowledgments
+
+I'd like to thank [Jonathon Grynspan](https://github.com/grynspan) and
```
grynspan, line 176: Jonath**a**n. :)

Sources/Testing/Issues/Issue.swift
younata: Just so I understand, you're saying this line is correct? Or is there a more-correct `@_spi` annotation than `ForToolsIntegrationOnly`? [The SPI doc](https://github.com/apple/swift-testing/blob/main/Documentation/SPI.md) only lists `Experimental` and `ForToolsIntegrationOnly`.

grynspan: Yes, adding the SPI marker here is probably the right thing to do, but I'd like @stmontgomery to chime in too.

stmontgomery: I think it's OK to leave `.system` _without_ an `@_spi` attribute. Although it's true that users _shouldn't_ generally record system issues themselves, I do think it can be useful for them to _match against_ issues of that kind using other APIs like `withKnownIssue`. So personally, I don't think we need to add it as part of this proposal/PR.

Sources/Testing/Expectations/Expectation.swift
```diff
@@ -41,6 +39,38 @@ public struct Expectation: Sendable {
...
+    isRequired: Bool,
+    sourceLocation: SourceLocation
+  ) {
+    self.evaluatedExpression = Expression("")
+    self.mismatchedErrorDescription = mismatchedErrorDescription
+    self.differenceDescription = differenceDescription
+    self.isPassing = isPassing
+    self.isRequired = isRequired
+    self.sourceLocation = sourceLocation
+  }
```
stmontgomery, line 57: I'd like to explore ideas for ways to structure this that avoid needing to publicly expose an `init` on `Expectation`. I know I brought that idea up in an earlier comment on this PR, but after reviewing this type again, and taking @grynspan 's comments about the two properties above into consideration, I'm wary to open this type up for general API use right now until it's a bit better designed.

As one alternative, I wonder if we could expose a `static func` on `Issue.Kind` which takes a generic message, as well as one or more values of the evaluated expression, and internally construct an `Expectation` with the relevant `Expression` and `Expression.Value` values representing all of the passed-in values? Something like

```swift
extension Issue.Kind {
  public static func expectationFailed<each T>(
    _ description: String,
    values: repeat each T,
    isRequired: Bool,
    sourceLocation: SourceLocation
  ) -> Self
}
```

This would require some internal change to `__Expression` but that seems doable, and it would allow capturing rich details of the left- and right-hand side values of binary expressions.

Documentation/Proposals/NNNN-public-issues-api.md
stmontgomery: See my [alternative suggestion](https://github.com/apple/swift-testing/pull/481#discussion_r1658106595)

grynspan:
If we were to proceed with #490, would we still need this API change?

younata:
> If we were to proceed with https://github.com/apple/swift-testing/issues/490, would we still need this API change?

No, this PR can be closed if we proceed with #490, which solves this with less code required of third-party tools.

grynspan:
> No, this PR can be closed if we proceed with #490, which solves this with less code required of third-party tools.

With that in mind, I'd like to suggest we close out this PR as well as #474, and focus on that PR. (If you agree that it's an approach that works for you.) We'll want to do an API PR and forum post for that one too, if you're interested in writing it?

younata:
> With that in mind, I'd like to suggest we close out this PR as well as https://github.com/apple/swift-testing/issues/474, and focus on that PR. (If you agree that it's an approach that works for you.) We'll want to do an API PR and forum post for that one too, if you're interested in writing it?

Ok, sure. I'll take that on.

PR #479, opened 2024-06-13
[SWT-0002] A stable JSON-based ABI for tools integration

grynspan:
One of the core components of Swift Testing is its ability to interoperate with Xcode 16, VS Code, and other tools. Swift Testing has been fully open-sourced across all platforms supported by Swift, and can be added as a package dependency (or—eventually—linked from the Swift toolchain.)

Because Swift Testing may be used in various forms, and because integration with various tools is critical to its success, we need it to have a stable interface that can be used regardless of how it's been added to a package.

Read the full proposal [here](https://github.com/apple/swift-testing/blob/main/Documentation/Proposals/0002-json-abi.md).

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #482, opened 2024-06-13
Don't emit JSON blobs for test cases if a test function isn't parameterized.

grynspan:
Internally, Swift Testing models non-parameterized test functions as parameterized over a single input, `()`. This is a very useful abstraction internally, but when it comes to the JSON output for parameterized tests, it can be confusing or surprising to see information about a test case that was synthesized by the library.

This PR suppresses events and metadata for the single test case of a non-parameterized test function.

### Checklist:


grynspan:
@swift-ci please test

briancroom:
Have you thought about what kind/layering of test coverage we should have for this code?

grynspan:
> Have you thought about what kind/layering of test coverage we should have for this code?

We already technically have tests that cover this code, it's just never been a failure for _extra_ data to be present. I intend to add some additional test coverage once the JSON format is finalized per #479.

briancroom (approved):

PR #488, opened 2024-06-14
Fix miscompile with inferred base expression of member access expression.

### LINKED ISSUE #487: #require cannot use implicit typing like XCTUnwrap can.
### Description

This is surely highly related to #486. I tried this way first, as a direct translation, and then I tried that way, as a backup. Neither worked.

### Expected behavior

`#require` replicates all of the type inference capability that `XCTUnwrap` allows.

For example, this.

```swift
import XCTest

class TestCase: XCTestCase {
  func test() throws {
    _ = try XCTUnwrap(.ƒ(0))
  }
}

extension Optional {
  static func ƒ(_ optional: Self) -> Self { optional }
}
```

### Actual behavior

> Cannot find 'ƒ' in scope

### Steps to reproduce

```swift
@Test func test() throws {
  try #require(.ƒ(0))
}
```

Whatever's in the first Xcode 16 beta.

swift-driver version: 1.109.2 Apple Swift version 6.0 (swiftlang-6.0.0.3.300 clang-1600.0.20.10)
Target: arm64-apple-macosx14.0
Darwin Jessys-2021-MBP.local 23.5.0 Darwin Kernel Version 23.5.0: Wed May 1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64

grynspan:
This PR fixes a bug where an inferred base expression in a member function call (`.foo()` instead of `T.foo()`) would be expanded incorrectly in `#expect()` and `#require()` as if it were a free function call.

For example:

```swift
extension Bool {
  static func truth() -> Bool {
    true
  }
}

@Test func f() {
  #expect(.truth())
}
```

Resolves #487.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/ConditionArgumentParsing.swift
+++ b/Sources/TestingMacros/Support/ConditionArgumentParsing.swift
@@ -393,12 +393,20 @@ private func _parseCondition(from expr: FunctionCallExprSyntax, for macro: some 
     // closure body since it is unused.
     let parameterList = forwardedArguments.isEmpty ? "_ in" : ""
     conditionArguments.append(Argument(expression: "()"))
+
+    // If memberAccessExpr is not nil here, that means it had a nil base
+    // expression (i.e. the base is inferred.)
+    var dot: TokenSyntax?
+    if memberAccessExpr != nil {
+      dot = .periodToken()
+    }
+
     conditionArguments.append(
       Argument(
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

PR #494, opened 2024-06-18
Store a collection of patterns to match against in Configuration.TestFilter instead of chaining multiple pattern-matcher filters

stmontgomery:
This fixes a crash which can occur if there are many filter or skip arguments passed to the testing library's entry point.

In this situation, before this fix, one `Configuration.TestFilter` was created for each filter or skip argument and they were chained together to form a tree using `combining(with:using:)`. Then, they were applied to a test graph recursively. If this tree was deep due to having many filter or skip arguments chained together, the recursion could exceed the stack and cause a crash.

The fix is to change `Configuration.TestFilter.Kind.pattern` so that it can store multiple patterns to match against instead of just one. This makes it more aligned with the other cases in its enum, which also store collections, and prevents the crash since it typically results in much shallower filter trees.

### Result:

Crash is resolved, validated by a new unit test.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Configuration.TestFilter.swift
+++ b/Sources/Testing/Running/Configuration.TestFilter.swift
@@ -54,9 +54,9 @@ extension Configuration {
       /// The test filter contains a pattern to predicate test IDs against.
       ///
       /// - Parameters:
-      ///   - pattern: The pattern to predicate test IDs against.
+      ///   - patterns: The patterns to predicate test IDs against.
       ///   - membership: How to interpret the result when predicating tests.
-      case pattern(_ pattern: String, membership: Membership)
+      case patterns(_ patterns: [String], membership: Membership)
 
       /// The test filter is a combination of other test filter kinds.
       ///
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

dennisweissmann (approved):

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -126,21 +126,23 @@ extension Configuration.TestFilter {
...
     // representing it, but do not preserve it. This type only represents
     // the pattern in the abstract, and is not responsible for actually
     // applying it to a test graph — that happens later during planning.
     //
     // Performing this validation here currently makes such errors easier to
     // surface when using the SwiftPM entry point. But longer-term, we should
     // make the planning phase throwing and propagate errors from there instead.
-    _ = try Regex(pattern)
+    for pattern in patterns {
+      _ = try Regex(pattern)
```
grynspan, line 142: I think at this point we can store the regexes directly.

Tests/TestingTests/ABIEntryPointTests.swift
```diff
@@ -20,6 +20,35 @@ private import _TestingInternals
 @Suite("ABI entry point tests")
 struct ABIEntryPointTests {
   @Test func v0() async throws {
+    var arguments = __CommandLineArguments_v0()
```
grynspan, line 23: Were the changes in this file necessary? Is there an ABI break? Will this conflict with my draft PR?

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: I can try but I think we still won't be able to due to partial API availability

Tests/TestingTests/ABIEntryPointTests.swift
stmontgomery: None of the changes here were in response to API or ABI changes, I was only wanting to refactor things to make it easier to write a second test of the entry point without needing to duplicate a bunch of set up code

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: Yeah, confirmed it still cannot be stored in the `Kind` enum case due to higher API availability

Tests/TestingTests/ABIEntryPointTests.swift
grynspan: The irony being I'm pitching a different signature for the entry point so this code ends up dead. 🥲

Sources/Testing/Running/Configuration.TestFilter.swift
grynspan: Booooo okay.

PR #497, opened 2024-06-20
Update CMake settings to match recent package changes

stmontgomery:
A few fixes to the CMake and package build settings:

- Enable Swift 6 language mode, to match the package which now uses this language mode due to the change which bumped `swift-tools-version` to 6.0.
- Add SourceLocation macro-related source files, which were missing.
- Add fake `visionOS 99.0` OS version to our `_distantFuture` availability definition.
- Clean up experimental/upcoming features which are no longer needed.

### Checklist:


stmontgomery:
@swift-ci please test

stmontgomery:
@compnerd @jeffdav FYI about these adjustments to the CMake rules.

cmake/modules/shared/CompilerSettings.cmake
```diff
@@ -8,14 +8,14 @@
 
 # Settings intended to be applied to every Swift target in this package.
 # Analogous to project-level build settings in an Xcode project.
+add_compile_options(
+  "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-swift-version 6>")
```
compnerd, line 12: I think that this is better spelt as the following at the top level CMakeLists.txt:

```cmake
set(Swift_LANGUAGE_VERSION 6)
```

compnerd (approved):

cmake/modules/shared/CompilerSettings.cmake
stmontgomery: Looks like it's spelled **CMAKE**_Swift_LANGUAGE_VERSION, but that does work!

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

briancroom (approved):

PR #498, opened 2024-06-21
Avoid linker symbol de-duplication in 'failureBreakpoint()' marker function

stmontgomery:
This fixes an issue where the setting a debugger breakpoint on the special `failureBreakpoint()` marker function (which is used to break on test failures) may cause the debugger to resolve to the incorrect symbol in release builds due to linker de-duplication. See code comment for additional details.

### Result:

Confirmed this resolved the issue in a manually built release build of the testing library.

### Checklist:

Resolves rdar://129872111

### DIFF
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -259,5 +259,13 @@ extension Issue {
 @inline(never) @_optimize(none)
 @usableFromInline
 func failureBreakpoint() {
-  // Empty.
+  // This function's body cannot be completely empty or else linker symbol
+  // de-duplication will cause its symbol to be consolidated with that of
+  // another, arbitrarily chosen empty function in this module. This linker
+  // behavior can be disabled by passing the `-no_deduplicate` flag described in
+  // ld(1), but that would disable it module-wide and sacrifice optimization
+  // opportunities elsewhere, so instead this function performs some trivial
+  // work to ensure its body is unique.
+  func noOp() {}
+  noOp()
 }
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

adrian-prantl (approved):
Note that if two libraries were do this _exact_ same trick, this could still get deduplicated. A more fool-proof approach might be to have a real side-effect here, such as call to os_log() with a unique message or something like that.

dennisweissmann (approved):

stmontgomery:
> > os_log()
> 
> Or something more portable, like a print to stderr...

I'm open to the idea of something with actual side-effects, but don't want to do a console or log print statement because we already print to the console for a test failure elsewhere after this, and we don't want to print each failure twice since it would pollute the output. But if there's something more resilient that _isn't_ a write to stdio, I can definitely improve this. Any ideas?

stmontgomery:
@swift-ci please test

grynspan:
Could you just increment a `nonisolated(unsafe) static var`? Would that be sufficient?

stmontgomery:
We don't need the "no-op" function to have _observable_ results, such as assigning to a variable, but we do need the function call itself to be unique enough to ensure it never gets consolidated with another function which calls it. My revised approach passes the function's own _name_ as a String to a no-op function, which I think is likely very unique and unlikely to trigger de-duplication, but I agree @grynspan, we could make that _even less_ likely by passing a global var `inout`, so I'll make that change. The no-op function doesn't need to assign a value to that variable though, and I'd prefer it didn't so we're never tempted to use this value for a real purpose in the future (such as a failure counter). This failure breakpoint symbol is meant to have low overhead and we're intentionally not synchronizing access to the global variable.

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

PR #507, opened 2024-06-27
Skip evaluating parameterized test arguments for disabled tests

briancroom:
Similar to how parameterized test arguments are left unevaluated for tests that aren't included in a `Plan`, also leave them unevaluated for tests in a `Plan` that won't run, e.g. because of having the `.disabled()` trait attached.

### Motivation:

Since the changes in #366, parameterized test arguments are evaluated lazily, only for tests which are included in the `Runner.Plan` that is being prepared for execution. A code comment included in that change further suggested that arguments are only evaluated for tests which will actually run as part of the plan, as opposed to being skipped. However the actual code didn't reflect that behavior, by mistake. That means wasted work is being performed when setting up a `Runner.Plan`, evaluating the arguments of tests which won't end up actually running.

### Modifications:

Only perform the step of evaluating test arguments/cases for tests which are marked with a `.run` action in the `Runner.Plan`.

### Result:

Less throwaway work is performed when setting up a `Runner.Plan` for greater efficiency.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -253,12 +253,14 @@ extension Runner.Plan {
       // can be evaluated lazily only once it is determined that the test will
       // run, to avoid unnecessary work. But now is the appropriate time to
       // evaluate them.
-      do {
-        try await test.evaluateTestCases()
-      } catch {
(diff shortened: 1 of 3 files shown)


briancroom:
@swift-ci please test

stmontgomery (approved):
Thanks! Good catch

PR #509, opened 2024-06-27
Remove an obsolete workaround for Result.get()

stmontgomery:
This removes a workaround for `Result.get()` which was briefly necessary (added in #215) due to a compiler issue but has since been resolved.

I took the opportunity to also make these two helper methods `@inlinable` since they are included in many macro expansions and in certain situations that optimization opportunity could be helpful.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/ResultAdditions.swift
+++ b/Sources/Testing/Support/Additions/ResultAdditions.swift
@@ -19,7 +19,7 @@ extension Result {
   ///
   /// - Warning: This function is used to implement the `#expect()` and
   ///   `#require()` macros. Do not call it directly.
-  @inlinable public func __required() throws(Failure) -> Success {
+  @inlinable public func __required() throws -> Success {
     try get()
   }
 }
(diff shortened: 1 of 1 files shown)


dennisweissmann (approved):

Sources/Testing/Support/Additions/ResultAdditions.swift
```diff
@@ -13,20 +13,13 @@ extension Result {
...
-  public func __required() throws -> Success {
-    /// `get()` is current broken in the Swift standard library, so switch
-    /// manually to work around the problem. ([122797397](rdar://122797397))
-    switch self {
-    case let .success(result):
-      return result
-    case let .failure(error):
-      throw error
-    }
+  @inlinable public func __required() throws(Failure) -> Success {
```
grynspan, line 22: This should continue to be an untyped `throws` or it'll lose source location information and potentially be ABI-breaking with Xcode 16 Beta 1.

grynspan (changes requested):
See comment.

grynspan:
@swift-ci please test

Sources/Testing/Support/Additions/ResultAdditions.swift
stmontgomery: Thanks. I'm not sure the ABI breakage with Xcode 16 Beta matters (since there is no general expectation of ABI compatibility in testing frameworks across Xcode releases), but you're right, the loss of thrown source location tracking is a good reason to avoid typed throws for now! I hope we can overcome that limitation eventually.

stmontgomery:
@swift-ci please test

grynspan (approved):

PR #511, opened 2024-06-28
Emit ABIv0 JSON objects for tests when listing them.

### LINKED ISSUE #506: Provide a JSON API to *just* list tests (via the callback, not stdout)
### Description

If the JSON ABI entry point is used with `listTests: true` in its configuration, then the framework prints the list of all tests to stdout and then continues on executing tests ([implementation](https://github.com/apple/swift-testing/blob/2cdfccf6e891ef181be3091fc81670c3b6dd641c/Sources/Testing/EntryPoints/EntryPoint.swift#L35-L42)).

However, a client of the JSON entry point may want to query which tests are available without running anything yet. I think this can be hacked up today by passing in a filter regex that could not possibly match any test ID; then, the framework would terminate after running nothing. (Disclaimer: I haven't actually tried this.) However, doing so would still require the calling process to redirect stdout temporarily and then parse the output line-by-line.

Would it be possible for the JSON entry point to provide a mode where `listTests: true` (or something like it with a different name, if we need to preserve the current behavior) would report the list of tests back to the caller via the callback as a JSON payload instead, returning immediately without running anything?

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

main

_No response_

grynspan:
This PR changes the behavior of `swift test list` and its various synonyms to allow reporting the list of tests via the ABI-stable JSON mechanism described in #479.

As it is not currently possible to directly call `swift test list --event-stream-output-path ... --event-stream-version 0`, it's a bit hard to test this code. However, it is possible to opt into this mode using `--configuration-path` and passing a path to a JSON file that includes `"listTests": true` (as noted by @allevato.) Support for these arguments with `swift test list` is tracked by https://github.com/swiftlang/swift-package-manager/pull/7768.

Resolves #506.
Resolves rdar://130627856.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/v0/ABIv0.Record+Streaming.swift
+++ b/Sources/Testing/ABI/v0/ABIv0.Record+Streaming.swift
@@ -31,18 +31,15 @@ extension ABIv0.Record {
   ) -> Event.Handler {
     let humanReadableOutputRecorder = Event.HumanReadableOutputRecorder()
     return { event, context in
-      if case let .runStarted(plan) = event.kind {
-        // Gather all tests in the run and forward records for them.
-        for test in plan.steps.lazy.map(\.test) {
-          try? JSON.withEncoding(of: Self(encoding: test)) { testJSON in
-            eventHandler(testJSON)
-          }
+      if case .testDiscovered = event.kind, let test = context.test {
+        try? JSON.withEncoding(of: Self(encoding: test)) { testJSON in
+          eventHandler(testJSON)
         }
-      }
-
-      let messages = humanReadableOutputRecorder.record(event, in: context)
-      if let eventRecord = Self(encoding: event, in: context, messages: messages) {
-        try? JSON.withEncoding(of: eventRecord, eventHandler)
+      } else {
+        let messages = humanReadableOutputRecorder.record(event, in: context)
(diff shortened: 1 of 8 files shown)


grynspan:
@swift-ci please test

Sources/Testing/ABI/v0/ABIv0.Record+Streaming.swift
```diff
@@ -31,18 +31,15 @@ extension ABIv0.Record {
   ) -> Event.Handler {
     let humanReadableOutputRecorder = Event.HumanReadableOutputRecorder()
     return { event, context in
-      if case let .runStarted(plan) = event.kind {
```
briancroom, line 34: This diff produces in no change to the contents of the event stream no non-listing scenarios, right?

briancroom (approved):

Sources/Testing/ABI/v0/ABIv0.Record+Streaming.swift
grynspan: Correct. When running `swift test`, the JSON output is identical. The actual enumeration happens slightly earlier in the execution flow, but from the POV of the caller, it's the same because the logic inside Swift Testing is responsible for emitting the test objects before the `runStarted` event object in both cases.

PR #512, opened 2024-06-28
Add experimental overload of `confirmation()` that takes a range.

grynspan:
Wouldn't it be useful to be able to specify a range of expected counts for a confirmation rather than a specific number? For instance, if you expect an event to occur exactly zero or one times:

```swift
await confirmation(expectedCount: 0 ... 1) { thingHappened in
  // ...
}
```

Or maybe you expect that an event will occur _at least_ 10 times, but maybe many more?

```swift
await confirmation(expectedCount: 10...) { thingHappened in
  // ...
}
```

Well... good news, everyone! Here's an experimental overload of `confirmation()` that allows you to do exactly that!

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Issues/Confirmation.swift
```diff
@@ -96,13 +96,92 @@ public func confirmation<R>(
...
+///
+/// When the closure returns, the testing library checks if the confirmation's
+/// preconditions have been met, and records an issue if they have not.
+///
+/// If an exact count is expected, use
+/// ``confirmation(_:expectedCount:sourceLocation:_:)-7kfko`` instead.
+@_spi(Experimental)
+public func confirmation<R>(
+  _ comment: Comment? = nil,
+  expectedCount: some RangeExpression<Int>,
```
Kyle-Ye, line 161: Should we considering using `expectedCountRange: some RangeExpression<Int>` here?

grynspan: I think the name of the argument would be something we'd discuss if/when we promote this function to API, but to me right now it reads like "expected count is in the range _n_".

Sources/Testing/Issues/Confirmation.swift
```diff
@@ -96,13 +96,92 @@ public func confirmation<R>(
...
     let actualCount = confirmation.count.rawValue
-    if actualCount != expectedCount {
+    if !expectedCount.contains(actualCount) {
+      var comment = comment
+      let issueKind: Issue.Kind
+      if let expectedCount = expectedCount as? ClosedRange<Int>,
+         expectedCount.lowerBound == expectedCount.upperBound {
+        issueKind = .confirmationMiscounted(actual: actualCount, expected: expectedCount.lowerBound)
+      } else {
+        // TODO: define an issue kind for out-of-range confirmation failures
```
Kyle-Ye, line 175: Tried to help add it in #515.

grynspan: Much appreciated—as discussed off of GitHub, we'll hold off on a resolution here until we have a bit more team buy-in and know how best to stage in a change (as I'd like to actually _replace `.confirmationMiscounted` here.)

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Linux

Kyle-Ye (approved):

grynspan:
@swift-ci please test

Sources/Testing/Issues/Confirmation.swift
```diff
@@ -190,3 +178,52 @@ public func confirmation<R>(
...
+  ///
+  /// - Parameters:
+  ///   - actualCount: The actual count for the failed confirmation.
+  ///
+  /// - Returns: An instance of ``Issue/Kind-swift.enum`` that describes `self`.
+  fileprivate func issueKind(forActualCount actualCount: Int) -> Issue.Kind {
+    switch self {
+    case let expectedCount as ClosedRange<Int> where expectedCount.lowerBound == expectedCount.upperBound:
+      return .confirmationMiscounted(actual: actualCount, expected: expectedCount.lowerBound)
+    case let expectedCount as Range<Int> where expectedCount.lowerBound == expectedCount.upperBound - 1:
```
Kyle-Ye, line 208: Nit:

Considering 5..<5 is a valid range expression.

If someone is passing `Int.min ..< Int.min`, this will cause a "Swift runtime failure: arithmetic overflow".

We can just use `&-` here to avoid the corner case crash. (But I guess probably no one would pass `Int.min ..< Int.min` as expectedCount here LOL)

```suggestion
    case let expectedCount as Range<Int> where expectedCount.lowerBound == expectedCount.upperBound &- 1:
```

Sources/Testing/Issues/Confirmation.swift
```diff
@@ -190,3 +178,52 @@ public func confirmation<R>(
...
+  /// ``confirmation(_:expectedCount:sourceLocation:_:)-41gmd``.
+  ///
+  /// This protocol represents any expression that describes a range of
+  /// confirmation counts. For example, the expression `1 ..< 10` automatically
+  /// conforms to it.
+  ///
+  /// You do not generally need to add conformances to this type yourself. It is
+  /// used by the testing library to abstract away the different range types
+  /// provided by the Swift standard library.
+  public protocol ExpectedCount: Sendable, RangeExpression<Int> {}
```
Kyle-Ye, line 194: Nice abstraction 👍

Sources/Testing/Issues/Issue.swift
```diff
@@ -300,6 +318,8 @@ extension Issue.Kind {
           .expectationFailed(Expectation.Snapshot(snapshotting: expectation))
       case let .confirmationMiscounted(actual: actual, expected: expected):
           .confirmationMiscounted(actual: actual, expected: expected)
+      case let .confirmationOutOfRange(actual: actual, expected: _):
+          .confirmationMiscounted(actual: actual, expected: 0)
```
Kyle-Ye, line 322: Why not also adding a corresponding new case for Snapshot here?

Is this on purpose or this is a todo item?

grynspan: We are not planning to continue developing the snapshot types. See #479.

Sources/Testing/Issues/Confirmation.swift
grynspan: `n ..< n` is not a valid range expression for integer types as the upper bound (effectively `n - 1`) is less than the lower bound.

Kyle-Ye:
```
Swift/arm64e-apple-macos.swiftinterface:22083: Fatal error: Range requires lowerBound <= upperBound
```

Swift Standard library will accept the form eg. lowerBound ..< upperBound.

Do you mean we have a more strict rule for Int range and the standard library has not enforce such rule?

grynspan: It may _accept_ construction of such an expression, but it will result in an invalid range. As well, a constraint that a condition be met fewer than zero times cannot be satisfied in reality, so passing such a range constitutes a logic error in your test.

Kyle-Ye: Yes. I agree such input is an invalid input. And the behavior shall be undefined.

We can use `&-` to make it not crash and we can also choose do nothing and make it crash.

grynspan: What I'm saying is that it is _correct_ that such code crashes, because an upper bound of `.min` is never going to be valid in the first place. 🙂

If we wanted to be pedantic, we could add a protocol requirement to `ExpectedCount` of the form `func validate() throws` that, for a half-open range like `n ..< m`, tests that `n < m && m >= 0`, but that's overdesigning the feature for the sake of a corner case that, again, will never function correctly in the first place.

dennisweissmann (approved):

SeanROlszewski (approved):
Looks good to me. I think the tests for this would need to be extended if we decided to no longer consider this experimental, but for now this is fine.

grynspan:
> Looks good to me. I think the tests for this would need to be extended if we decided to no longer consider this experimental, but for now this is fine.

DM me any specific tests you have in mind and I can add them when time permits.

Sources/Testing/Issues/Confirmation.swift
stmontgomery: I also think we should consolidate the new `.confirmationMiscounted` enum case here — can we do that, or file an issue to track it?

PR #520, opened 2024-07-02
@Test applied to function with newline before its name causes error in expanded code

stmontgomery:
This fixes an issue where certain trivia, such as a newline (`\n`) placed in between the `func` keyword and a function's name on a function with the `@Test` attribute causes an error in the macro expansion code. For example:

```swift
@Test func
nameOnSecondLine() { ... }
```

### Modifications:

Trim trivia from function name in macro expansion code.

### Result:

These examples now build without errors.

### Checklist:

Resolves rdar://130797660

### DIFF
--- a/Sources/TestingMacros/TestDeclarationMacro.swift
+++ b/Sources/TestingMacros/TestDeclarationMacro.swift
@@ -299,13 +299,13 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
       thunkBody = ""
     } else if let typeName {
       if functionDecl.isStaticOrClass {
-        thunkBody = "_ = \(forwardCall("\(typeName).\(functionDecl.name)\(forwardedParamsExpr)"))"
+        thunkBody = "_ = \(forwardCall("\(typeName).\(functionDecl.name.trimmed)\(forwardedParamsExpr)"))"
       } else {
         let instanceName = context.makeUniqueName(thunking: functionDecl)
         let varOrLet = functionDecl.isMutating ? "var" : "let"
         thunkBody = """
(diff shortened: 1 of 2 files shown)


stmontgomery:
@swift-ci please test

grynspan (approved):

suzannaratcliff (approved):

JetForMe:
Wild. This literally wasn't working when I wrote #561 a day or two ago, but now it does. It seems both CLI Swift and Xcode will implicitly include a version of Swift Testing. Adding the explicit dependency to `Package.swift` on branch `main` picks up this commit and tests behave as expected. Sorry for the noise!

JetForMe:
@stmontgomery May I ask: It seems like this is a shortcoming of swift-syntax, right? In that it produces `functionDecl.name` with surrounding whitespace?

bnbarham:
> @stmontgomery May I ask: It seems like this is a shortcoming of swift-syntax, right? In that it produces functionDecl.name with surrounding whitespace? Should I open an issue there?

This is intentional on the swift-syntax side - the tree produced by swift-syntax is intended to provide source fidelity, ie. the tree is always a faithful recreation of the source code it was given as input. `name` here is actually a `TokenSyntax` which includes trivia around it (both whitespace and eg. comments). It was being used directly to produce a call, which thus also included its trivia. Using `.trimmed` is the correct change.

JetForMe:
Probably a discussion to have over there, but how do they decide that the surrounding whitespace belongs to the function name and not some other token? Seems like it would be better to have tokens like `<func>``<whitespace>``<func name>`, etc. Because I'd argue that intuitively, the whitespace is definitely *not* part of the function name.

grynspan:
Let's continue this discussion in the forums.

PR #523, opened 2024-07-04
Avoid calling `malloc()` while holding a lock that may block dyld or libobjc.

grynspan:
This PR pre-sizes the vector result of `getMachHeaders()` on Darwin (used during test discovery) so that it is very unlikely that it will need to be reallocated while holding the function's lock. If `malloc()` is called while this lock is held, there is a theoretical possibility of deadlock with the Objective-C runtime or dyld itself if `malloc()` ends up calling into either.

### Checklist:


grynspan:
@mikeash, I wouldn't mind your eyes on this (smol) diff.

grynspan:
@swift-ci please test

stmontgomery (approved):

mikeash (commented):
Sorry, was out, but this looks good to me.

grynspan:
> Sorry, was out, but this looks good to me.

No worries, thanks!

PR #525, opened 2024-07-07
Remove `"(unknown context at $xxxxxxxx)"` from fully-qualified type names.

grynspan:
This PR ensures that the aforementioned gunk doesn't show up in our string representations of suites and type IDs. These substrings indicate that a type is `private`, nested in a function, or otherwise might have a non-unique Swift name. We don't care: Swift Testing will treat two private types with the same (otherwise qualified) name in the same module as the same type and will ultimately disambiguate tests in them by source location if needed.

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #526, opened 2024-07-08
Ensure that a `CustomExecutionTrait`'s teardown logic occurs _after_ nested tests run.

grynspan:
This PR reorders operations in `Runner` such that a test with a custom execution trait will execute operations in this order:

1. Custom execution trait "before" logic;
1. Test logic;
1. Child test logic; and finally
1. Custom execution trait "after" logic.

There is currently a bug here where child test logic runs _after_ the custom execution trait runs its "after" logic, which is not intentional.

Resolves rdar://125115497.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -209,11 +209,33 @@ extension Runner {
             if let testCases = step.test.testCases {
               try await _runTestCases(testCases, within: step)
             }
+
+            // Run the children of this test (i.e. the tests in this suite.)
+            try await _runChildren(of: stepGraph, depth: depth, lastAncestorStep: lastAncestorStep)
           }
         }
       }
+    } else {
+      // There is no test at this node in the graph, so just skip down to the
+      // child nodes.
+      try await _runChildren(of: stepGraph, depth: depth, lastAncestorStep: lastAncestorStep)
     }
+  }
 
+  /// Recursively run the tests that are children of a given plan step.
+  ///
+  /// - Parameters:
+  ///   - stepGraph: The subgraph whose root value, a step, is to be run.
+  ///   - depth: How deep into the step graph this call is. The first call has a
+  ///     depth of `0`.
+  ///   - lastAncestorStep: The last-known ancestral step, if any, of the step
+  ///     at the root of `stepGraph`. The options in this step (if its action is
+  ///     of case ``Runner/Plan/Action/run(options:)``) inform the execution of
+  ///     `stepGraph`.
+  ///
+  /// - Throws: Whatever is thrown from the test body. Thrown errors are
+  ///   normally reported as test failures.
+  private func _runChildren(of stepGraph: Graph<String, Plan.Step?>, depth: Int, lastAncestorStep: Plan.Step?) async throws {
     // Figure out the last-good step, either the one at the root of `stepGraph`
     // or, if it is nil, the one passed into this function. We need to track
     // this value in case we run into sparse sections of the graph so we don't
--- a/Tests/TestingTests/Traits/CustomExecutionTraitTests.swift
+++ b/Tests/TestingTests/Traits/CustomExecutionTraitTests.swift
@@ -10,26 +10,6 @@
 
 @testable @_spi(Experimental) @_spi(ForToolsIntegrationOnly) import Testing
 
-private struct CustomTrait: CustomExecutionTrait, TestTrait {
-  var before: Confirmation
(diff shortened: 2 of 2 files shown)


grynspan:
@swift-ci please test

dennisweissmann (approved):
Thank you Jonathan! That looks great!

Tests/TestingTests/Traits/CustomExecutionTraitTests.swift
```diff
@@ -76,4 +56,53 @@ struct CustomExecutionTraitTests {
...
+  func execute(_ function: @escaping @Sendable () async throws -> Void, for test: Testing.Test, testCase: Testing.Test.Case?) async throws {
+    #expect(Self.state.increment() == 1)
+
+    try await function()
+    #expect(Self.state.increment() == 3)
+  }
+}
+
+@Suite(.hidden, DoSomethingBeforeAndAfterTrait())
+struct TestsWithCustomTrait { // Trait should only excecute once for each test since it is a suite trait, if we want to execute trait logic for each test set isRecursive to true
```
dennisweissmann, line 104: > // Trait should only excecute once for each test since it is a suite trait, if we want to execute trait logic for each test set isRecursive to true

Do you think it's worth adding a test for that as well?

dennisweissmann, line 104:
```suggestion
struct TestsWithCustomTrait { // Trait should only execute once for each test since it is a suite trait, if we want to execute trait logic for each test set isRecursive to true
```

grynspan: That's not new code, I just moved it down to a new "fixtures" section. Not going to refactor it as part of this PR, but feel free to open a new PR with changes if you think they're worthwhile. :)

grynspan:
@swift-ci please test

Tests/TestingTests/Traits/CustomExecutionTraitTests.swift
```diff
@@ -76,4 +56,53 @@ struct CustomExecutionTraitTests {
       }.run(configuration: configuration)
     }
   }
+
+  @Test("Teardown occurs after child tests run")
+  func teardownOccursAtEnd() async throws {
+    var configuration = Configuration()
+    configuration.eventHandler = { event, _ in
+      print(event)
```
stmontgomery, line 64: Did you mean to leave in this `print`?

Sources/Testing/Running/Runner.swift
```diff
@@ -209,11 +209,33 @@ extension Runner {
             if let testCases = step.test.testCases {
               try await _runTestCases(testCases, within: step)
             }
+
+            // Run the children of this test (i.e. the tests in this suite.)
+            try await _runChildren(of: stepGraph, depth: depth, lastAncestorStep: lastAncestorStep)
           }
         }
       }
+    } else {
```
stmontgomery, line 218: This `else` is for this line above, right?
```
if let step = stepGraph.value, case .run = step.action {
```
Isn't it possible that `value` is non-`nil` but `step.action` is not `.run`? In which case, we should not continue to run the children, right?

Tests/TestingTests/Traits/CustomExecutionTraitTests.swift
grynspan: lol nope

Sources/Testing/Running/Runner.swift
grynspan: No, I think we still want to traverse them in order to indicate we've skipped them, for instance (remember that the action is propagated downward during planning.)

This isn't a regression at any rate, so if we want to rethink that, let's do so in a separate PR dedicated to it?

stmontgomery: Ah right — the function name "**run**Children(...)" threw me off. I recall now it doesn't necessarily _run_ them, but more like "evaluate" or "process" them, depending on their action. Sounds good

stmontgomery (approved):
LGTM thanks, pending that stray `print` removal

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Tests/TestingTests/Traits/CustomExecutionTraitTests.swift
grynspan: Actually… not sure where the comment came from. Brain worms?

PR #528, opened 2024-07-10
Avoid fork-bombing ourselves when hosted by `swift-test`.

grynspan:
This PR amends exit tests' child process spawning logic to ensure that they don't trigger a fork bomb or fail outright when the test process is hosted by Swift Package Manager (as in [this branch](https://github.com/swiftlang/swift-package-manager/tree/jgrynspan/one-test-product-two-libraries)).

Exit tests are an experimental feature, but one that we use internally in our own test suite, so we need to make sure our tests continue to function if/when that SwiftPM branch is merged.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -232,12 +232,16 @@ extension ExitTest {
     // can precompute them.
     let childProcessExecutablePath = Result { try CommandLine.executablePath }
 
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

stmontgomery (approved):

grynspan:
See also: https://github.com/swiftlang/swift-package-manager/pull/7766

PR #529, opened 2024-07-10
Use the '-define-availability' frontend flag directly in CMake rules

stmontgomery:
This replaces usages of the `-enable-experimental-feature AvailabilityMacro=X` Swift frontend flag in swift-testing's CMake rules with `-define-availability X`.

### Motivation:

The former flags are used in `Package.swift` as a workaround because SwiftPM does not allow using unsafe flags in package dependencies. But this workaround is not necessary when building with CMake—we can use the `-define-availability` flag directly without restriction.

Using this workaround when building via CMake with Library Evolution enabled causes every specified availability macro flag to be printed in the textual `.swiftinterface` file for the module. This information is not needed by clients of the `Testing` module, so the primary motivation of this change is to remove those unnecessary flags from that interface file.

### Checklist:

### DIFF
--- a/cmake/modules/shared/AvailabilityDefinitions.cmake
+++ b/cmake/modules/shared/AvailabilityDefinitions.cmake
@@ -7,14 +7,10 @@
 # See http://swift.org/CONTRIBUTORS.txt for Swift project authors
 
 # Settings which define commonly-used OS availability macros.
-#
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

briancroom (approved):

grynspan (approved):

PR #530, opened 2024-07-10
Change the default JSON schema version to v0.

### LINKED ISSUE #527: `__CommandLineArguments_v0` should default to JSON schema v0 rather than snapshot JSON
Per this paragraph in #479:

> If --event-stream-output is specified but --event-stream-version is not, the format currently used is based on direct JSON encodings of the internal Swift structures used by Swift Testing. This format is necessary to support Xcode 16 Beta 1. In the future, the default value of this argument will be assumed to be 0 instead (i.e. the JSON schema will match what we are proposing here.)

We should switch the default behaviour. To continue supporting Xcode 16 betas, we can continue to emit the older snapshot format if the requested version is **-1** instead (which is not a supported version and would only be used until Xcode can adopt the v0 JSON schema.)

grynspan:
This PR changes the default JSON schema version for event stream output to version 0 (the version documented in #479.) Previously, if a caller did not specify a version number for the event stream, it would default to encoding "snapshots" of various values from the testing library.

Xcode 16 Beta 1 uses this legacy format, so to continue supporting newer betas of Xcode 16 transitionally, this PR temporarily enables a "-1" JSON schema version that continues to emit snapshot encodings. This JSON schema version will be removed (along with the snapshots code in general) in a future update once Xcode has migrated.

Resolves #527.

### Checklist:

### DIFF
--- a/Sources/Testing/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/EntryPoints/EntryPoint.swift
@@ -220,12 +220,12 @@ public struct __CommandLineArguments_v0: Sendable {
   /// The version of the event stream schema to use when writing events to
   /// ``experimentalEventStreamOutput``.
   ///
-  /// If the value of this property is `nil`, events are encoded verbatim (using
-  /// ``Event/Snapshot``.) Otherwise, the corresponding stable schema is used
-  /// (e.g. ``ABIv0/Record`` for `0`.)
+  /// The corresponding stable schema is used to encode events to the event
+  /// stream (for example, ``ABIv0/Record`` is used if the value of this
+  /// property is `0`.)
   ///
-  /// - Warning: The behavior of this property will change when the ABI version
-  ///   0 JSON schema is finalized.
+  /// If the value of this property is `nil`, the testing library assumes a
+  /// value of `0` instead.
   public var experimentalEventStreamVersion: Int?
 
   /// The value(s) of the `--filter` argument.
@@ -482,9 +482,11 @@ public func configurationForEntryPoint(from args: __CommandLineArguments_v0) thr
 /// - Throws: If `version` is not a supported ABI version.
 func eventHandlerForStreamingEvents(version: Int?, forwardingTo eventHandler: @escaping @Sendable (UnsafeRawBufferPointer) -> Void) throws -> Event.Handler {
   switch version {
-  case nil:
+  case -1:
+    // Legacy support for Xcode 16 betas. Support for this undocumented version
+    // will be removed in a future update. Do not use it.
     eventHandlerForStreamingEventSnapshots(to: eventHandler)
-  case 0:
+  case nil, 0:
     ABIv0.Record.eventHandler(forwardingTo: eventHandler)
   case let .some(unsupportedVersion):
     throw _EntryPointError.invalidArgument("--experimental-event-stream-version", value: "\(unsupportedVersion)")
--- a/Sources/Testing/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/EntryPoints/EntryPoint.swift
@@ -224,8 +224,8 @@ public struct __CommandLineArguments_v0: Sendable {
   /// stream (for example, ``ABIv0/Record`` is used if the value of this
   /// property is `0`.)
   ///
-  /// If the value of this property is `nil`, the testing library assumes a
-  /// value of `0` instead.
+  /// If the value of this property is `nil`, the testing library assumes that
+  /// the newest available schema should be used.
   public var experimentalEventStreamVersion: Int?
 
   /// The value(s) of the `--filter` argument.
--- a/Sources/Testing/Test.swift
+++ b/Sources/Testing/Test.swift
@@ -128,6 +128,23 @@ public struct Test: Sendable {
     }
   }
 
+  /// Equivalent to ``testCases``, but without requiring that the test cases be
+  /// evaluated first.
+  ///
+  /// Most callers should not use this property and should prefer ``testCases``
+  /// since it will help catch logic errors in the testing library. Use this
+  /// property if you are interested in the test's test cases, but the test has
+  /// not been evaluated by an instance of ``Runner/Plan`` (e.g. if you are
+  /// implementing `swift test list`.)
+  var uncheckedTestCases: (some Sequence<Test.Case>)? {
+    testCasesState.flatMap { testCasesState in
+      if case let .evaluated(testCases) = testCasesState {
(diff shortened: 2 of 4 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/EntryPoints/EntryPoint.swift
```diff
@@ -220,12 +220,12 @@ public struct __CommandLineArguments_v0: Sendable {
...
-  /// ``Event/Snapshot``.) Otherwise, the corresponding stable schema is used
-  /// (e.g. ``ABIv0/Record`` for `0`.)
+  /// The corresponding stable schema is used to encode events to the event
+  /// stream (for example, ``ABIv0/Record`` is used if the value of this
+  /// property is `0`.)
   ///
-  /// - Warning: The behavior of this property will change when the ABI version
-  ///   0 JSON schema is finalized.
+  /// If the value of this property is `nil`, the testing library assumes a
+  /// value of `0` instead.
```
stmontgomery, line 228: I'd say it should assume whatever is the _newest_ (i.e. latest) version at the time it is called, if nil. Right now that is 0, but will presumably increment as subsequent versions are added

grynspan:
@swift-ci please test Linux

Sources/Testing/Test.swift
```diff
@@ -128,6 +128,23 @@ public struct Test: Sendable {
...
 
+  /// Equivalent to ``testCases``, but without requiring that the test cases be
+  /// evaluated first.
+  ///
+  /// Most callers should not use this property and should prefer ``testCases``
+  /// since it will help catch logic errors in the testing library. Use this
+  /// property if you are interested in the test's test cases, but the test has
+  /// not been evaluated by an instance of ``Runner/Plan`` (e.g. if you are
+  /// implementing `swift test list`.)
+  var uncheckedTestCases: (some Sequence<Test.Case>)? {
```
stmontgomery, line 139: What's the connection to this PR of this change?

grynspan: Crashes without it :)

stmontgomery: I do think it would be useful to evaluate test cases when listing tests by default (perhaps with an opt-out). That can be considered out of scope for this PR, although I also would't object if you instead changed this to evaluate them as a way to fix this crash.

Sources/Testing/EntryPoints/EntryPoint.swift
stmontgomery: Yeah, it's a classic problem with any versioning scheme: what should the default be if no version is explicitly specified? If you default to oldest, newcomers may not be aware there is a newer version they can be using and that newer version may struggle to gain traction among the user base. But if you default to latest, a newer version may break older users who weren't explicitly passing any version and assuming an older one.

Two thoughts about this:

1. When it comes to an **entry point _function_**, such as `swt_abiv0_getEntryPoint`, the version is already specified explicitly in the symbol name, so there will never be a scenario where it's not explicit. Good! That avoids that class of problems (…right?).

    Although that raises a different question: what if someone called that function but passed a different version in the `__CommandLineArguments_v0` struct property?
2. When it comes to **command-line users of SwiftPM**, they might not pass the new `--event-stream-version` flag at all, so that's the scenario where we need to decide what the default should be.

It's common for Swift packages to predicate breaking behavioral changes on the `// swift-tools-version: <vers>` line in the package's manifest file. To solve problem 2 above, what if we planned to do this as well? SwiftPM could be modified such that, if no version flag is explicitly passed, it passes the version which was known to be latest at the time when the package's tools version was initially released. Then, it would always pass the version flag to the runner process, so the runner always receives an explicit version.

grynspan: > Although that raises a different question: what if someone called that function but passed a different version in the __CommandLineArguments_v0 struct property?

This is going to be moot eventually because `__CommandLineArguments_v0` will be replaced by an "ABI stable" JSON format, same as how the snapshot types are being replaced. We just haven't gotten to that point yet. Once that's done, there will probably be a mandatory  `"version": 0` field in the JSON blob, and we'll teach our JSON decoding code to try a few different formats before throwing up.

grynspan: > To solve problem 2 above, what if we planned to do this as well? SwiftPM could be modified such that, if no version flag is explicitly passed, it passes the version which was known to be latest at the time when the package's tools version was initially released. Then, it would always pass the version flag to the runner process, so the runner always receives an explicit version.

This way lies danger for tools though. If I write a tool that doesn't pass `--event-stream-version` today, and I build somebody else's package that says `// swift-tools-version: 7.0`, I am not expecting to get the version 2 (or whatever) of the JSON format. My code was written against what was current when I wrote it. The problem here is that the package being built and tested may have no relationship to the tool that's reading the JSON output.

It's better to default to the lowest common denominator because it's stable in the long term. We can/should advise tools authors to be explicit about their requirements, but having it spontaneously change formats is asking for trouble down the line.

Sources/Testing/Test.swift
grynspan: Let's consider that a "future direction" for now as the evaluated test cases aren't formally part of the JSON spec yet. (This part of the change doesn't affect snapshots.)

Sources/Testing/EntryPoints/EntryPoint.swift
stmontgomery: Another alternative would be to mandate that if you pass the flag requesting the JSON output, you must pass the flag specifying a version explicitly too.

But that being said, I honestly think it may be best for us to just document that we default to _newest_, and strongly recommend that tools consuming this info via the SwiftPM CLI explicitly pass the version flag (but not mandate it). This decision can be aligned with the toolchain version once swift-testing is part of toolchains, so that means it would only potentially break if/when a newer toolchain is installed, and there's already some precedent for downstream tools needing to be adjusted to accommodate new toolchains.

Sources/Testing/Test.swift
stmontgomery: Sounds good

stmontgomery (approved):

Sources/Testing/EntryPoints/EntryPoint.swift
grynspan: Discussed further and we'll have `nil` equal "whatever the newest version is" and advise tools authors to always be explicit.

grynspan:
@swift-ci please test

PR #533, opened 2024-07-12
Do not write to `stderr` when testing `swift test list`.

grynspan:
When testing the JSON ABI _vis-à-vis_ `swift test list`, we're writing to `stderr`. For the purposes of our tests, this is just noise, so block output to `stderr` from the `list` subcommand if verbosity is set to `.min` (as it is in the tests.)

### Checklist:


grynspan:
@swift-ci please test

grynspan:
Self-reviewing.

PR #536, opened 2024-07-13
Emit a specific exit code when no tests match inputs to `swift test`.

grynspan:
This PR modifies the entry point function used by Swift Package Manager so that when no tests match the inputs (or, in the case of `swift test list`, when there are no Swift Testing tests in the test target), Swift Package Manager can detect it and respond appropriately.

Because exit codes exist in a flat namespace, and because on POSIX-y systems only the low 8 bits of an exit code are reliable, our options here are limited. I have selected the `EX_UNAVAILABLE` code from sysexits.h to represent this state. On Windows, all 32 bits of the exit code are propagated, but there is no `EX_UNAVAILABLE` value, so I've used the Win32 `ERROR_NOT_FOUND` error code instead. My assumption is that neither of these codes is likely to _actually_ be passed to `exit()` by test code (because why would they be?)

A separate PR is required in Swift Package Manager to correctly handle this exit code and emit the `noMatchingTests` diagnostic that's currently emitted only for XCTest.

Resolves rdar://131704539.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/EntryPoints/ABIEntryPoint.swift
+++ b/Sources/Testing/ABI/EntryPoints/ABIEntryPoint.swift
@@ -118,6 +118,11 @@ private func entryPoint(
 
   let eventHandler = try eventHandlerForStreamingEvents(version: args?.eventStreamVersion, forwardingTo: recordHandler)
   let exitCode = await entryPoint(passing: args, eventHandler: eventHandler)
+
+  // To maintain compatibility with Xcode 16 Beta 1, suppress custom exit codes.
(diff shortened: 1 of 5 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #539, opened 2024-07-16
Lower `Backtrace` and `SourceContext` to SPI.

grynspan:
`Backtrace` is our implementation of a stack trace type. Xcode 16 uses it to provide richer information about test failures than can be provided by _just_ an instance of `SourceLocation`.

The standard library has its own type, `_Backtracing.Backtrace`, but due to our deployment requirements on Apple platforms, we can't use it yet. It would make sense for us to adopt the `Backtrace` type in the standard library as soon as possible. To avoid staging issues in the future, and because our type has not undergone any sort of formal review, I'm lowering it to tools-only SPI.

I'm also lowering `SourceContext` because it consists of an instance of `Backtrace` and an instance of `SourceLocation`—but if `Backtrace` is SPI, then the public interface of `SourceContext` is simply a wrapper around `SourceLocation`. This does not eliminate the internal need for the type, but it does make it redundant from an API perspective. If and when we regain `Backtrace` as a public interface, we can look at raising `SourceContext` too.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedIssue.swift
+++ b/Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedIssue.swift
@@ -22,9 +22,13 @@ extension ABIv0 {
     /// The location in source where this issue occurred, if available.
     var sourceLocation: SourceLocation?
 
+    /// The backtrace where this issue occurred, if available.
+    var _backtrace: [Backtrace.Address]?
+
     init(encoding issue: borrowing Issue) {
       isKnown = issue.isKnown
       sourceLocation = issue.sourceLocation
+      _backtrace = issue.sourceContext.backtrace.map(\.addresses)
     }
   }
 }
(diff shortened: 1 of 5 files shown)


grynspan:
@swift-ci please test

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedIssue.swift
```diff
@@ -22,9 +22,13 @@ extension ABIv0 {
     /// The location in source where this issue occurred, if available.
     var sourceLocation: SourceLocation?
 
+    /// The backtrace where this issue occurred, if available.
+    var _backtrace: [Backtrace.Address]?
```
grynspan, line 26: This is not formally part of the JSON schema (hence the underscore) but I realized it was missing and should be present here.

grynspan:
@swift-ci please test

briancroom (approved):

PR #540, opened 2024-07-16
Remove `Snapshot` types from non-Darwin builds.

grynspan:
Now that we've landed and tagged #479 in 0.11.0, we no longer need to keep building the snapshot types for non-Darwin platforms because we have a stable, permanent replacement for them.

We need to keep them around on Darwin to support Xcode 16 betas.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
Linux failure is due to too many jobs running at a time.

grynspan:
@swift-ci please test Linux

briancroom (approved):

PR #541, opened 2024-07-16
Work around a bug in the older Swift compiler used to build swift-syntax in the toolchain, which affected certain result builder APIs

stmontgomery:
This works around a bug which was identified while working towards adding swift-testing to Swift toolchains. See the code comment for a more detailed explanation.

### Modifications:

Use a different variant of this swift-syntax API in our macro plugin to avoid the bug.

### Result:

Building in the toolchain avoids the failure this was causing.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
@@ -142,11 +142,12 @@ extension FunctionParameterSyntax {
     // Enclose the base type declaration reference in a 1-element tuple, e.g.
     // `(<baseType>)`. It will be used in a member access expression below, and
     // some types (such as function types) require this.
-    let metatypeMemberAccessBase = TupleExprSyntax {
-      LabeledExprListSyntax {
-        LabeledExprSyntax(expression: baseTypeDeclReferenceExpr)
-      }
-    }
+    //
+    // Intentionally avoid using the result builder variant of these APIs due to
+    // a bug which affected a range of Swift compilers (including the one
+    // currently used to build swift-syntax in the toolchain) and caused one of
+    // these APIs to have an incorrect representation in the module interface.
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

grynspan:
Booooourns

grynspan (approved):

Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
```diff
@@ -142,11 +142,12 @@ extension FunctionParameterSyntax {
     // Enclose the base type declaration reference in a 1-element tuple, e.g.
     // `(<baseType>)`. It will be used in a member access expression below, and
     // some types (such as function types) require this.
-    let metatypeMemberAccessBase = TupleExprSyntax {
-      LabeledExprListSyntax {
-        LabeledExprSyntax(expression: baseTypeDeclReferenceExpr)
-      }
-    }
+    //
+    // Intentionally avoid using the result builder variant of these APIs due to
```
grynspan, line 146: Is there a swift-syntax or swift PR/issue you can point to here? For future reference.

stmontgomery: Nobody involved in the investigation was sure what exactly the problem was, but in any event, the bug itself has long been fixed. We believe it was fixed in Swift 5.10. The problem is Swift's CI, in this case at least, is building some early dependencies using Swift 5.9 or earlier (intentionally), and that compiler had the bug.

briancroom (approved):

Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
grynspan: Alrighty. And the bug _only_ affected _this_ use of a result builder? How strange!

PR #548, opened 2024-07-17
Lower some symbols to SPI.

grynspan:
This PR lowers some API symbols to SPI, mostly as experimental tools-only. These symbols are, we believe, important but deserve more thought and design before we promote them to API.

As well, this PR merges the `comment` and `comments` properties of `ConditionTrait` as redundant.

Resolves rdar://131937376.

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -215,6 +215,7 @@ extension Issue {
     public var sourceContext: SourceContext
 
     /// Whether or not this issue is known to occur.
+    @_spi(ForToolsIntegrationOnly)
     public var isKnown = false
 
     /// Initialize an issue instance with the specified details.
--- a/Sources/Testing/Parameterization/Test.Case.ID.swift
+++ b/Sources/Testing/Parameterization/Test.Case.ID.swift
@@ -8,7 +8,6 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-@_spi(ForToolsIntegrationOnly)
 extension Test.Case: Identifiable {
   /// The ID of a test case.
   ///
@@ -21,8 +20,10 @@ extension Test.Case: Identifiable {
     ///
     /// The value of this property is `nil` if _any_ of the associated test
     /// case's arguments has a `nil` ID.
+    @_spi(ForToolsIntegrationOnly)
     public var argumentIDs: [Argument.ID]?
 
+    @_spi(ForToolsIntegrationOnly)
     public init(argumentIDs: [Argument.ID]?) {
       self.argumentIDs = argumentIDs
     }
--- a/Sources/Testing/Traits/ConditionTrait.swift
+++ b/Sources/Testing/Traits/ConditionTrait.swift
@@ -19,9 +19,6 @@
 /// - ``Trait/disabled(if:_:sourceLocation:)``
 /// - ``Trait/disabled(_:sourceLocation:_:)``
 public struct ConditionTrait: TestTrait, SuiteTrait {
-  /// An optional, user-specified comment describing this trait.
-  public var comment: Comment?
-
   /// An enumeration describing the kinds of conditions that can be represented
   /// by an instance of this type.
   enum Kind: Sendable {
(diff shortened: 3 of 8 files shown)


grynspan:
@swift-ci please test

briancroom (approved):

Sources/Testing/Traits/ConditionTrait.swift
```diff
@@ -77,6 +75,8 @@ public struct ConditionTrait: TestTrait, SuiteTrait {
     }
   }
 
+  public var comments: [Comment]
```
stmontgomery, line 78: DocC comment

grynspan: Not needed, it's a protocol requirement

stmontgomery (approved):

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

Sources/Testing/Parameterization/Test.Case.ID.swift
```diff
@@ -8,7 +8,6 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-@_spi(ForToolsIntegrationOnly)
```
stmontgomery, line 11: I'm worried about this change -- making the `Identifiable` conformance public and also making its `ID` type public.  `Test.Case.ID` is not yet fully designed and, in some scenarios will not actually _uniquely_ identify a test case. When its `arguments` property is `nil` (which is the case if any argument is non-`Codable` or equivalent), it will be considered equal to all other instances representing non-`Codable` arguments. This could be confusing to users or lead to bugs.

This is a known limitation of this `ID` type currently which we need to come back and fix, but the current tool adopters already use it properly with that consideration in mind. But if we make this change it will publish this `ID` type as API and I don't think it is mature or robust enough yet.

The existing solution was to wrap this entire extension in `@_spi`. Can you elaborate on why that is problematic? A runtime `as? Identifiable` cast may succeed, but since the concrete `ID` type is not public, I did not think this would be a problem.

grynspan: Protocol conformances are always as visible as the type/protocol in question. See here: https://developer.apple.com/documentation/testing/test/case#relationships

If we don't want the conformance to be public, we'll have to remove it outright.

grynspan: Behold:

```swift
func idForTestCase(_ testCase: Test.Case) -> some Hashable {
  func thunk(_ x: some Identifiable) -> some Hashable {
    x.id
  }
  return thunk(testCase)
}
```

Contrived, but the point is the IDs are publicly accessible already.

grynspan: Actually, due to opaque types, this is valid even when everything's SPI and you haven't imported it:

```swift
let x = Test.Case.current!.id
let y = Test.Case.current!.id
#expect(x == y)
```

So yeah, we need to pick a side of the fence to be on here.

stmontgomery: I think it would certainly be preferable to remove the `Identifiable` conformance entirely, until the `ID` type is ready for broader use.

Also, the thing I was worried about was that this `ID` type might be used in `Equatable` comparisons. Your example showed a `Hashable` thunk, and luckily I think hash-ability is much less worrisome for this type. That said, we could also remove the `ID` type's conformance to `Equatable`, unless it's needed for some other reason I've forgotten. 

Either way, it sounds like we should completely remove `Test.Case`'s conformance to `Identifiable` for now.

stmontgomery: Yeah, let's just remove the Identifiable conformance

grynspan: @briancroom Thoughts?

Sources/Testing/Issues/Issue.swift
```diff
@@ -215,6 +215,7 @@ extension Issue {
     public var sourceContext: SourceContext
 
     /// Whether or not this issue is known to occur.
+    @_spi(ForToolsIntegrationOnly)
```
briancroom, line 218: This change was intended to be applied to the property in `Issue`, not `Issue.Snapshot`, wasn't it?

grynspan: 🤦

PR #549, opened 2024-07-18
Add skipped test tracking to JUnit output.

grynspan:
This PR adds skipped test reporting to our JUnit XML output. For example, given this test:

```swift
@Test(.disabled("Because I said so"))
func f() {}
```

The XML output would be, approximately:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
  <testsuite name="TestResults" errors="0" tests="1" failures="0" skipped="1" time="12345.0">
    <testcase classname="MyTests" name="f()" time="12344.0">
      <skipped>Because I said so</skipped>
    </testcase>
  </testsuite>
</testsuites>
```

See also https://github.com/swiftlang/swift-package-manager/issues/7383 which asks for this for XCTest.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Recorder/Event.JUnitXMLRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.JUnitXMLRecorder.swift
@@ -11,6 +11,9 @@
 extension Event {
   /// A type which handles ``Event`` instances and outputs representations of
   /// them as JUnit-compatible XML.
+  ///
+  /// The maintainers of JUnit do not publish a formal XML schema. A _de facto_
+  /// schema is described in the [JUnit repository](https://github.com/junit-team/junit5/blob/main/junit-platform-reporting/src/main/java/org/junit/platform/reporting/legacy/xml/XmlReportWriter.java).
   @_spi(ForToolsIntegrationOnly)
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

briancroom:
Ah nice, thanks!

briancroom (approved):

Sources/Testing/Events/Recorder/Event.JUnitXMLRecorder.swift
```diff
@@ -183,14 +209,19 @@ extension Event.JUnitXMLRecorder {
   ///
   /// - Returns: `character`, or a string containing its escaped form.
   private static func _escapeForXML(_ character: Character) -> String {
-    if character == #"""# {
+    switch character {
```
jakepetroules, line 212: `<` and `>` aren't _required_ to be escaped in XML attributes, but `&` definitely is. Could you add that?

grynspan: Gonna leave `>` out of pedantry, but yeah I'll add `&`.

PR #550, opened 2024-07-18
Allow non-`final` classes to act as suites.

grynspan:
Now that we require Swift 6, we can see the names of classes during `@Test` macro expansion. This was the primary blocker preventing us from allowing non-`final` classes, because we had to substitute `Self` for the class name and this was invalid if the suite type could be covariant. We can drop the constraint now.

Note that test functions are _not_ heritable.

### Checklist:


grynspan:
@swift-ci please test

briancroom:
Nice, thank you!

briancroom (approved):

PR #551, opened 2024-07-18
Update the messaging around the use of deinit for teardown

medreisbach:
resolves rdar://132007317

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/MigratingFromXCTest.md
+++ b/Sources/Testing/Testing.docc/MigratingFromXCTest.md
@@ -97,7 +97,7 @@ For more information about suites and how to declare and customize them, see
 In XCTest, you can schedule code to run before a test using the
 [`setUp()`](https://developer.apple.com/documentation/xctest/xctest/3856481-setup)
 family of functions. When writing tests using the testing library, you can use
-`init()` for setup. Your suite's initializer can be `async`, `throwing`, or
+`init()` for setup. Your suite's initializer can be `async`, `throws`, or
 actor-isolated if necessary.
 
 @Row {


Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -127,9 +127,13 @@ family of functions. When writing tests using the testing library, implement
...
-The use of `async` and `throws` is optional. If teardown is needed, declare your
-test suite as a `final` class or as an actor rather than as a structure and
-implement `deinit`:
+In XCTest, you can also schedule work to run after a test using the [`tearDown()`](https://developer.apple.com/documentation/xctest/xctest/3856482-teardown)
+family of functions. When writing tests using the testing library, adopt structured
+  [concurrency](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/concurrency/) 
+and take advantage of Swift to handle cleanup for you where possible rather 
+than relying on a dedicated teardown method. If teardown is necessary, declare
+your test suite as a `final` class or actor and implement `deinit`. Your cleanup
+operations should be quick, synchronous, and non-throwing.
```
stmontgomery, line 136: Not sure we need to say that cleanup operations should be "quick"; it's perfectly valid to write a long-running cleanup task if you want. Of course, overall test duration will still be subject to any time limit trait which has been applied, if any.

Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -94,11 +94,11 @@ For more information about suites and how to declare and customize them, see
...
 ### Convert setup and teardown functions
 
-In XCTest, code can be scheduled to run before and after a test using the
+In XCTest, you can schedule code to run before a test using the
 [`setUp()`](https://developer.apple.com/documentation/xctest/xctest/3856481-setup)
-and [`tearDown()`](https://developer.apple.com/documentation/xctest/xctest/3856482-teardown)
-family of functions. When writing tests using the testing library, implement
-`init()` and/or `deinit` instead:
+family of functions. When writing tests using the testing library, you can use
+`init()` for setup. Your suite's initializer can be async, throwing, or
```
grynspan, line 100: Let's explicitly use the `async` and `throws` keywords like we do for test functions here: https://github.com/apple/swift-testing/blob/main/Sources/Testing/Testing.docc/DefiningTests.md

grynspan (changes requested):
Minor quibbles as noted.

Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -127,9 +127,13 @@ family of functions. When writing tests using the testing library, implement
...
 
-The use of `async` and `throws` is optional. If teardown is needed, declare your
-test suite as a `final` class or as an actor rather than as a structure and
-implement `deinit`:
+In XCTest, you can also schedule work to run after a test using the [`tearDown()`](https://developer.apple.com/documentation/xctest/xctest/3856482-teardown)
+family of functions. When writing tests using the testing library, adopt structured
+  [concurrency](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/concurrency/) 
+and take advantage of Swift to handle cleanup for you where possible rather 
+than relying on a dedicated teardown method. If teardown is necessary, declare
+your test suite as a `final` class or actor and implement `deinit`. Your cleanup
```
grynspan, line 135: I just merged #550 so when you go to resolve conflicts, note the word `final` has been dropped.

Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -127,9 +127,13 @@ family of functions. When writing tests using the testing library, implement
   }
 }
 
-The use of `async` and `throws` is optional. If teardown is needed, declare your
-test suite as a `final` class or as an actor rather than as a structure and
-implement `deinit`:
+In XCTest, you can also schedule work to run after a test using the [`tearDown()`](https://developer.apple.com/documentation/xctest/xctest/3856482-teardown)
+family of functions. When writing tests using the testing library, adopt structured
+  [concurrency](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/concurrency/) 
+and take advantage of Swift to handle cleanup for you where possible rather
```
stmontgomery, line 133: Can you clarify what feature(s) of Swift concurrency, or which APIs, you mean here which assist with cleanup? I can think of `defer`, but I would consider that a basic language feature rather than a _concurrency_ feature.

stmontgomery (commented):
Thanks for improving these docs!

Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -97,7 +97,7 @@ For more information about suites and how to declare and customize them, see
 In XCTest, you can schedule code to run before a test using the
 [`setUp()`](https://developer.apple.com/documentation/xctest/xctest/3856481-setup)
 family of functions. When writing tests using the testing library, you can use
-`init()` for setup. Your suite's initializer can be async, throwing, or
+`init()` for setup. Your suite's initializer can be `async`, `throwing`, or
```
stmontgomery, line 100:
```suggestion
`init()` for setup. Your suite's initializer can be `async`, `throws`, or
```

Sources/Testing/Testing.docc/MigratingFromXCTest.md
FranzBusch: I think what we should aim to push people towards is using `with` style methods e.g. `try await withNetworkConnection { connection in ... }`. This is the pattern that recommend to people to have guaranteed lifetimes with structured concurrency.

grynspan: That's something they have to write themselves, and explaining how to write one would get us into the weeds a bit.

iamleeg: There is precedent for describing how testing helps you design nicer APIs and vice versa, and it's something Apple has already published in the Xcode documentation: https://developer.apple.com/documentation/xcode/updating-your-existing-codebase-to-accommodate-unit-tests

So maybe we could take a similar approach here, where we show a `with…` function as an example of designing API+test together, but (as a separate task) create another doc that shows the thinking behind that design?

grynspan: Would this be an appropriate place to use a [snippet](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0356-swift-snippets.md)?

0xTim: I think an example here would be extremely helpful for users trying to migrate

Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -127,9 +127,13 @@ family of functions. When writing tests using the testing library, implement
...
 
-The use of `async` and `throws` is optional. If teardown is needed, declare your
-test suite as a `final` class or as an actor rather than as a structure and
-implement `deinit`:
+In XCTest, you can also schedule work to run after a test using the [`tearDown()`](https://developer.apple.com/documentation/xctest/xctest/3856482-teardown)
+family of functions. When writing tests using the testing library, adopt structured
+  [concurrency](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/concurrency/) 
+and take advantage of Swift to handle cleanup for you where possible rather 
+than relying on a dedicated teardown method. If teardown is necessary, declare
+your test suite as a class or actor, and implement `deinit` with operations
```
weissi, line 135: If they're synchronous and non-throwing, `defer { ... }` is more appropriate because it actually deterministically happens at a precise, user-visible place.

PR #553, opened 2024-07-19
Fix incorrect implementation-only import for CMake build

stmontgomery:
This resolves a warning seen when building the project with CMake, since that style currently still requires `@_implementationOnly import` instead of `private import` or other access levels.

```
ABIEntryPoint.swift:12:16: warning: '_TestingInternals' inconsistently imported as implementation-only
```

### Checklist:


stmontgomery:
@swift-ci please test

briancroom (approved):

PR #557, opened 2024-07-19
Simplify `failureBreakpoint()` and add a unit test.

grynspan:
This PR simplifies the "unique" action performed by `failureBreakpoint()` to avoid de-duplication and implements a unit test to confirm that action is actually occurring and not being optimized away (at least under the current compiler settings at the time of test.)

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -236,6 +236,9 @@ extension Issue {
 
 // MARK: - Debugging failures
 
+/// A unique value used by ``failureBreakpoint()``.
+@usableFromInline nonisolated(unsafe) var failureBreakpointValue = 0
+
 /// A function called by the testing library when a failure occurs.
 ///
 /// Whenever a test failure (specifically, a non-known ``Issue``) is recorded,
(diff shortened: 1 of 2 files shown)


grynspan:
@stmontgomery What do you think? This eliminates the need to capture a value `inout` or call a second-order function. Also now we can unit test it (not a particularly interesting test, but _code coverage intensifies_.)

grynspan:
@swift-ci please test

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -265,11 +268,7 @@ func failureBreakpoint() {
...
-  // function call, passing it a sufficiently unique value to avoid
-  // de-duplication.
-  struct NoOp {
-    nonisolated(unsafe) static var ignored: Int = 0
-    static func perform(_: inout Int) {}
-  }
-  NoOp.perform(&NoOp.ignored)
+  // operation on a usable-from-inline value, which the compiler must assume
+  // cannot be optimized away.
+  failureBreakpointValue = 1
```
stmontgomery, line 273: If we were to enable the Thread Sanitizer and run our test suite, would it flag this as a data race? (Despite always assigning the same value)

grynspan: Nope, but if it did I would expect the original implementation would have the same issue.

stmontgomery: I'd be surprised if both of them did not report a TSan race. But you're saying you tried and they did not? 🤔

Either way, this PR seems fine. I just really want to avoid transforming this into a global incrementing counter in the future, since then it would require a lock and would increase the overhead of every test failure.

stmontgomery (approved):

PR #559, opened 2024-07-19
Remove `Test.Case`'s conformance to `Identifiable`.

grynspan:
With further consideration after #548, we've decided to remove conformance to `Identifiable` from `Test.Case` rather than promoting it to API. The identity of a test case is fuzzy because it depends on the arguments to a test, not all of which may be uniquely identifiable. We need more time to consider the design of this part of the interface before making it a permanent part of our API.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

stmontgomery (approved):

PR #560, opened 2024-07-19
Disable a couple of tests on older OSes that are reliant on `Regex`.

grynspan:
The tests in question use `Regex`, so they won't run on older Darwin versions.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #567, opened 2024-07-23
Explicitly import Foundation in IssueTests.swift.

grynspan:
If XCTest is unavailable, IssueTests.swift fails to compile due to a Foundation dependency (that was transitively satisfied by importing XCTest which reexports Foundation). Add an explicit import of Foundation in this file to resolve the issue.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
Self-reviewing.

PR #568, opened 2024-07-23
Ensure `stdout` is line-buffered when testing starts.

grynspan:
Swift Package Manager reroutes test output through pipes which are block-buffered by default, but convention in C is that `stdout` is line-buffered. This results (when using `swift test`) in output to `stdout` (e.g. from `print()`) not being displayed until after most test events have been displayed.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

compnerd (approved):

briancroom (approved):

PR #569, opened 2024-07-23
Ensure serially-scheduled tests run in source order.

### LINKED ISSUE #566: Tests should run in source-location order in serialized contexts
### Description

When running tests within a suite with the `.serialized` trait, or when parallelization is disabled for the entire run, the testing library should use the location (line number) within their source file as a primary sorting key. This expectation was also recently stated [here](https://forums.swift.org/t/pitch-make-serialized-trait-public-api/73147/4), in the discussion about the `.serialized` trait. However currently, tests are being executed in alphabetic order.

### Expected behavior

Tests in serialized suites should be ordered by source location.

### Actual behavior

Tests in serialized suites are ordered alphabetically.

### Steps to reproduce

1. Run the following test suite using `swift test`:

```
import Testing

@Suite(.serialized)
struct Tests {
    @Test func c() {}
    @Test func b() {}
    @Test func a() {}
}
```

2. Observe the alphabetic ordering of the tests in the console output:

```
􀟈  Test run started.
􀄵  Testing Library Version: 91 (arm64e-apple-macos13.0)
􀟈  Suite Tests started.
􀟈  Test a() started.
􁁛  Test a() passed after 0.001 seconds.
􀟈  Test b() started.
􁁛  Test b() passed after 0.001 seconds.
􀟈  Test c() started.
􁁛  Test c() passed after 0.001 seconds.
􁁛  Suite Tests passed after 0.001 seconds.
􁁛  Test run with 3 tests passed after 0.001 seconds.
```

e3e7e940fff9ce6bc0294b68a8cbbdbcd7033d8d

Xcode 16 Beta

grynspan:
This PR ensures that tests, when scheduled serially, run in declared source order (or as close to it as suite layout allows.)

When tests are run in parallel, we don't bother to sort them because they will run non-deterministically anyway. Although we don't explicitly shuffle them, they are stored in a `Dictionary` and in Swift, hashcodes are salted with a random value, so their iteration order will change on every test run. (In other words, they're already effectively shuffled.)

Resolves #566.
Resolves rdar://132285155.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -222,6 +222,22 @@ extension Runner {
     }
   }
 
+  /// Get the source location corresponding to the root of a plan step graph.
+  ///
+  /// - Parameters:
+  ///   - stepGraph: The plan step graph whose root node is of interest.
+  ///
+  /// - Returns: The source location of the root node of `stepGraph`, or of the
+  ///   first descendant node thereof (sorted by source location.)
+  private func _sourceLocation(of stepGraph: Graph<String, Plan.Step?>) -> SourceLocation? {
+    if let result = stepGraph.value?.test.sourceLocation {
+      return result
+    }
+    return stepGraph.children.lazy
+      .compactMap { _sourceLocation(of: $0.value) }
+      .min()
+  }
+
   /// Recursively run the tests that are children of a given plan step.
   ///
   /// - Parameters:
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci please test

SeanROlszewski (approved):

grynspan:
There's a natural inconsistency that may occur here:

```swift
extension Inner {
  @Test func f() {}
}

@Suite struct Outer {
  @Test g() {}
  struct Inner {
    @Test func h() {}
  }
}
```

In this example, `Inner` is assigned the source location of `f()` and sorts _before_ `g()` because `f()` is the lowest-ordered child of `Inner`. If I add `@Suite` to `Inner`, then suddenly we have its real source location, and `Inner` will sort _after_ `g()`. There's really nothing I can do about that unless we always ignore source location information for suites (which may be even _more_ confusing.) 🤷‍♂️

PR #570, opened 2024-07-23
`Graph.count` is computed incorrectly.

grynspan:
`Graph.count` is computed with the `underestimatedCount` of its children, not the `count` of them, which results in a shallow sum. Oops.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #572, opened 2024-07-24
Report a skipped event for parameterized tests with no cases

briancroom:
When a parameterized test's argument list is empty, a "skip" event is a better reflection of the fact that the test function doesn't actually get executed.

With this change, a `.testSkipped` event will get posted for such tests, instead of a `.testStarted` + `.testEnded` event pair. 

This also means that the following tests will be reported very similarly (with just a different comment on the `SkipInfo`):

```
@Test(arguments: Array<Int>()) func t1(x: Int) {}
@Test(.disabled(), arguments: [1]) func t2(x: Int) {}
```

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -264,7 +264,7 @@ extension Runner.Plan {
       }
 
       // If the test is parameterized but has no cases, mark it as skipped.
-      if case .run = action, test.isParameterized, test.testCases?.lazy.first(where: { _ in true }) == nil {
+      if case .run = action, test.isParameterized, test.testCases?.first(where: { _ in true }) == nil {
         action = .skip(SkipInfo(comment: "No test cases found."))
       }
 
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -264,7 +264,7 @@ extension Runner.Plan {
       }
 
       // If the test is parameterized but has no cases, mark it as skipped.
-      if case .run = action, test.isParameterized, test.testCases?.first(where: { _ in true }) == nil {
+      if case .run = action, let testCases = test.testCases, testCases.first(where: { _ in true }) == nil {
         action = .skip(SkipInfo(comment: "No test cases found."))
       }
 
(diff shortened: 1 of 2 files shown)


dennisweissmann (approved):

Tests/TestingTests/RunnerTests.swift
```diff
@@ -201,6 +201,24 @@ final class RunnerTests: XCTestCase {
     await fulfillment(of: [testSkipped], timeout: 0.0)
   }
 
+  func testParameterizedTestWithNoCasesIsSkipped() async throws {
+    let testSkipped = expectation(description: "Test was skipped")
+    var configuration = Configuration()
+    configuration.eventHandler = { event, _ in
+      if case .testSkipped(_) = event.kind {
```
grynspan, line 208:
```suggestion
      if case .testSkipped = event.kind {
```

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -263,6 +263,11 @@ extension Runner.Plan {
         }
       }
 
+      // If the test is parameterized but has no cases, mark it as skipped.
+      if case .run = action, test.isParameterized, test.testCases?.lazy.first(where: { _ in true }) == nil {
```
grynspan, line 267:
```suggestion
      if case .run = action, test.isParameterized, test.testCases?.first(where: { _ in true }) == nil {
```

grynspan (changes requested):
Nitpicks only.

briancroom:
@swift-ci please test

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -263,6 +263,11 @@ extension Runner.Plan {
         }
       }
 
+      // If the test is parameterized but has no cases, mark it as skipped.
+      if case .run = action, test.isParameterized, test.testCases?.first(where: { _ in true }) == nil {
```
grynspan, line 267: We could also express this as:

```swift
      if case .run = action, let testCases = test.testCases, testCases.first(where: { _ in true }) == nil {
```

Which might be slightly more efficient.

briancroom:
@swift-ci please test

grynspan (approved):

grynspan:
Merging on Brian's behalf.

PR #575, opened 2024-07-24
Remove "unconditional" messaging from Issue kind `.unconditional`.

### LINKED ISSUE #574: Remove "Unconditionally" from `Issue.record` messaging
### Description

While a more official API for extending `Issue.record` with more issue kinds is being fleshed out, it might also be worth removing the "Unconditionally" from the standard output. `Issue.record` is still an API that is convenient to reach for to write test helpers, but when a test helper _should_ fail given a condition, the "unconditionally" part of the messaging can be confusing. "Failed" is probably enough to communicate when an issue is recorded.

### Expected behavior

`Issue.record` ideally emits a basic "Failed" message, like XCTest does.

### Actual behavior

`Issue.record` emits "Unconditionally failed" instead.

### Steps to reproduce

Invoke `Issue.record`.

Xcode 16 beta 4

N/A

grynspan:
This PR changes the default message for an issue of kind `.unconditional` from `"Unconditionally failed"` to `"Issue recorded"`.

Although the failure is unconditional at the point it is recorded, the code that recorded the issue may not be unconditionally executing, so we shouldn't describe it as unconditional (we just don't know!)

For example, given the following code:

```swift
if case .empty = potatoSack.contents {
  Issue.record("No potatoes left.")
}
```

We would previously log:

> Unconditionally failed: No potatoes left.

Now we'll log:

> Issue recorded: No potatoes left.

Resolves #574.

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -171,7 +171,10 @@ extension Issue.Kind: CustomStringConvertible {
   public var description: String {
     switch self {
     case .unconditional:
-      "Unconditionally failed"
+      // Although the failure is unconditional at the point it is recorded, the
+      // code that recorded the issue may not be unconditionally executing, so
+      // we shouldn't describe it as unconditional (we just don't know!)
+      "Issue recorded"
     case let .expectationFailed(expectation):
       if let mismatchedErrorDescription = expectation.mismatchedErrorDescription {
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci please test

suzannaratcliff (approved):
Looks good!

stephencelis:
Looks good, thank you!

PR #576, opened 2024-07-25
Move the `stdout` line-buffering call.

grynspan:
Move the `stdout` line-buffering call so it affects both overloads of `__swiftPMEntryPoint()`.

This has no effect on SwiftPM because it currently uses the overload that returns `Never`. Two overloads are provided to allow SwiftPM some flexibility, but we're not actively taking advantage of that in Swift 6.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #578, opened 2024-07-25
Run a subtask of the runner task group for each individual test even when running serially.

grynspan:
Right now, when tests are run serially, they're simply run in a for-loop. This has the side effect of potentially building up a lot of autorelease pool cruft (on platforms with such a concept) if those tests never introduce any suspension points.

As well, this produces a subtle difference in task cancellation propagation because calling `UnsafeCurrentTask.cancel()` in a serialized test will cancel the entire task group, not just the test's own subtask.

Resolves rdar://130034310.

### Checklist:


grynspan:
@swift-ci please test

briancroom (approved):

PR #579, opened 2024-07-26
Use `access()` instead of `stat()` to determine if the SF Pro font is installed.

grynspan:
This PR uses the slightly more efficient `access()` syscall to determine if SF Pro is installed on macOS.

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #584, opened 2024-07-31
Complete WebAssembly/WASI Support

kateinoigakukun:
### Motivation:

Trying to support WebAssembly/WASI before shipping swift-testing as a part of toolchain

### Modifications:

Thanks to prior WASI compatibility efforts by @grynspan, most of the parts work as is now. Here are a few items I changed:

- Remove `SWT_NO_FILE_IO` definition since wasi-libc provides all IO functionalities we need
- Stop including `signal.h`
- Fix metadata accessor function pointer representation on WASI to use absolute pointer instead of relative pointer
- Define primitive lock type as `Void` for WASI
  It's not the optimal representation but I couldn't find a better way 😓 

### Result:

Tested with `DEVELOPMENT-SNAPSHOT-2024-07-15-a` toolchain

```console
$ swift sdk install https://github.com/swiftwasm/swift/releases/download/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-07-16-a/swift-wasm-DEVELOPMENT-SNAPSHOT-2024-07-16-a-wasm32-unknown-wasi.artifactbundle.zip

$ swift package init --name Hello --enable-experimental-swift-testing
# Replace swit-testing dependency entry in Package.swift with:
#   .package(url: "https://github.com/kateinoigakukun/swift-testing.git", branch: "yt/wasi-support"),

$ swift build --swift-sdk DEVELOPMENT-SNAPSHOT-2024-07-16-a-wasm32-unknown-wasi --build-tests --enable-experimental-swift-testing
$ wasmkit-cli run --dir . ./.build/debug/HelloPackageTests.wasm
◇ Test run started.
↳ Testing Library Version: 3702b3a53779ccca0f8205b7a76a6f5038197120
◇ Test example() started.
✔ Test example() passed after 0.041244319 seconds.
✔ Test run with 1 test passed after 0.081816901 seconds.
```

### Checklist:


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

Sources/_TestingInternals/include/Includes.h
```diff
@@ -27,7 +27,7 @@
 /// - Note: Avoid including headers that aren't actually used.
 
 #include <errno.h>
-#if __has_include(<signal.h>)
+#if __has_include(<signal.h>) && !defined(__wasi__)
```
grynspan, line 30: What goes wrong if signal.h is included on WASI? Is it harmful?

grynspan:
Thanks for taking a look at all this! We really appreciate it. :)

Sources/_TestingInternals/include/Includes.h
grynspan: I see you commented about this in the commit description. Please document the issue in a comment here so we can refer back to it in the future. Thanks!

Package.swift
```diff
@@ -130,7 +130,6 @@ extension Array where Element == PackageDescription.SwiftSetting {
 
       .define("SWT_TARGET_OS_APPLE", .when(platforms: [.macOS, .iOS, .macCatalyst, .watchOS, .tvOS, .visionOS])),
 
-      .define("SWT_NO_FILE_IO", .when(platforms: [.wasi])),
```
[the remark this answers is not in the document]
grynspan: At an educated guess, most `os(Linux)` branches in FileHandle.swift should also be used by WASI. :)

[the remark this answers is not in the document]
kateinoigakukun: Oh, good catch. I missed them!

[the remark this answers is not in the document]
grynspan: I'd search the package for any uses of `SWT_NO_FILE_IO` and sanity-check 'em too.

Sources/Testing/Support/Locked.swift
```diff
@@ -42,6 +42,7 @@ struct Locked<T>: RawRepresentable, Sendable where T: Sendable {
   private typealias _Lock = SRWLOCK
 #elseif os(WASI)
   // No locks on WASI without multithreaded runtime.
+  private typealias _Lock = Void
```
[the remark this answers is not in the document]
grynspan: Note I have adoption of `Mutex` on a branch here: https://github.com/apple/swift-testing/tree/jgrynspan/adopt-mutex

It defines the "no supported lock type" case to `Never`, but I'd prefer to defer to `Mutex` and let it decide to be a no-op if possible.

I'm not planning to merge that branch in the near future, but it's relevant.

Sources/_TestingInternals/Discovery.cpp
```diff
@@ -86,6 +86,23 @@ struct SWTRelativePointer {
...
 };
 
+/// A type representing a 32-bit absolute function pointer, usually used on platforms
+/// where relative function pointers are not supported.
+///
+/// This type is derived from `AbsoluteFunctionPointer` in the Swift repository.
+template <typename T, int32_t maskValue = 0>
+class SWTAbsoluteFunctionPointer {
+  T *Pointer;
+  static_assert(sizeof(T *) == sizeof(int32_t),
```
grynspan, line 96: This `static_assert()` can just be on one line, I think. Code doesn't need to wrap at 80 columns, just comments.

Sources/_TestingInternals/Discovery.cpp
```diff
@@ -86,6 +86,23 @@ struct SWTRelativePointer {
   }
 };
 
+/// A type representing a 32-bit absolute function pointer, usually used on platforms
+/// where relative function pointers are not supported.
+///
+/// This type is derived from `AbsoluteFunctionPointer` in the Swift repository.
+template <typename T, int32_t maskValue = 0>
```
grynspan, line 93: Is the mask value used? If not, does this type need to be generic over it? (I assume yes, for the sake of calling code?)

Sources/_TestingInternals/Discovery.cpp
```diff
@@ -98,6 +115,15 @@ struct SWTRelativePointerIntPair: public SWTRelativePointer<T, maskValue> {
   }
 };
 
+
+template <typename T, int32_t maskValue = 0>
+#if defined(__wasm32__)
+using SWTCompactFunctionPointer = SWTAbsoluteFunctionPointer<T, maskValue>;
+#else
+using SWTCompactFunctionPointer = SWTRelativePointer<T, maskValue>;
```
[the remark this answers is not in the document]
kateinoigakukun: Yes, we still use `SWTRelativePointer` to reference a data space on wasm32. We need absolute pointer just for references to code space.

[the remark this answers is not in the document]
grynspan: Okay! Makes sense.

Sources/_TestingInternals/Discovery.cpp
kateinoigakukun: At least for now, there is no metadata that embeds payload within a function pointer, so we actually don't use the mask value

kateinoigakukun:
@swift-ci test

Sources/_TestingInternals/include/Includes.h
kateinoigakukun: Added a comment there :)

grynspan:
One other thing! Can you modify the project readme to include this line in the supported OS table? 😁

```md
| **Wasm** | | | Experimental |
```

Sources/Testing/Support/Locked.swift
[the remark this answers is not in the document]
grynspan: Perfect!

Sources/Testing/Support/FileHandle.swift
```diff
@@ -371,7 +373,7 @@ extension FileHandle {
 extension FileHandle {
   /// Is this file handle a TTY or PTY?
   var isTTY: Bool {
-#if SWT_TARGET_OS_APPLE || os(Linux)
+#if SWT_TARGET_OS_APPLE || os(Linux) || os(WASI)
```
grynspan, line 376: Can we assume WASI doesn't support TTYs, or is it possible for this to come back `true`?

Sources/Testing/Traits/Tags/Tag.Color+Loading.swift
```diff
@@ -49,22 +49,28 @@ private var _appDataDirectoryPath: String? {
 /// On Apple platforms and on Linux, this path is equivalent to
 /// `"~/.swift-testing"`. On Windows, it is equivalent to
 /// `"%HOMEPATH%\AppData\Local\.swift-testing"`.
-var swiftTestingDirectoryPath: String {
+/// Returns `nil` if the platform does not support the concept of a user
```
grynspan, line 52:
```suggestion
/// The value of this property is `nil` if the platform does not support the concept of a home directory, or if the home directory could not be determined.
```

(Wrap to 80 columns.)

Sources/Testing/Traits/Tags/Tag.Color+Loading.swift
```diff
@@ -49,22 +49,28 @@ private var _appDataDirectoryPath: String? {
...
+/// directory.
+var swiftTestingDirectoryPath: String? {
   // The (default) name of the .swift-testing directory.
   let swiftTestingDirectoryName = ".swift-testing"
 
 #if os(macOS) || (os(iOS) && targetEnvironment(macCatalyst)) || os(Linux)
   if let homeDirectoryPath = _homeDirectoryPath {
     return appendPathComponent(swiftTestingDirectoryName, to: homeDirectoryPath)
   }
+  return ""
```
grynspan, line 62:
```suggestion
  return nil
```

Sources/Testing/Traits/Tags/Tag.Color+Loading.swift
```diff
@@ -49,22 +49,28 @@ private var _appDataDirectoryPath: String? {
...
 #elseif os(Windows)
   if let appDataDirectoryPath = _appDataDirectoryPath {
     return appendPathComponent(swiftTestingDirectoryName, to: appDataDirectoryPath)
   }
+  return ""
```
grynspan, line 67:
```suggestion
  return nil
```

Sources/Testing/Support/FileHandle.swift
kateinoigakukun: wasi-libc provides `isatty` implementation (not a stub!), so it can be `true` https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/sources/isatty.c

Sources/Testing/Traits/Tags/Tag.Color+Loading.swift
```diff
@@ -83,7 +89,11 @@ var swiftTestingDirectoryPath: String {
 /// assumed to contain a JSON object (a dictionary) where the keys are tags'
 /// string values and the values represent tag colors. For a list of the
 /// supported formats for tag colors in this dictionary, see <doc:AddingTags>.
-func loadTagColors(fromFileInDirectoryAtPath swiftTestingDirectoryPath: String = swiftTestingDirectoryPath) throws -> [Tag: Tag.Color] {
+func loadTagColors(fromFileInDirectoryAtPath swiftTestingDirectoryPath: String? = swiftTestingDirectoryPath) throws -> [Tag: Tag.Color] {
+  guard let swiftTestingDirectoryPath else {
+    // If the platform does not support user-specific configuration, skip custom tag colors.
```
grynspan, line 94:
```suggestion
    // If the platform does not support user-specific configuration, skip custom
    // tag colors.
```

Sources/Testing/Support/FileHandle.swift
MaxDesiatov: You can read from a WASI-compatible module's stdout, so it's a matter of the consumer of that stdout whether they support terminal emulation or not. There are definitely in-browser terminal emulation libraries, and a Wasm runtime itself could be running as CLI in a terminal emulator. IMO seems premature to exclude WASI from that.

Sources/Testing/Support/FileHandle.swift
```diff
@@ -397,7 +399,7 @@ extension FileHandle {
 
   /// Is this file handle a pipe or FIFO?
   var isPipe: Bool {
-#if SWT_TARGET_OS_APPLE || os(Linux)
+#if SWT_TARGET_OS_APPLE || os(Linux) || os(WASI)
```
grynspan, line 402: As with TTYs, is it actually possible for a file to be a pipe?

kateinoigakukun: WASI Preview 1 doesn't have fifo nor pipe but Preview 2 does! https://github.com/WebAssembly/wasi-filesystem/blob/fe0ebbcb8bab1ee1f860812c25a6cc83b725e47b/wit/types.wit#L52

grynspan:
@swift-ci please test

Sources/Testing/Support/FileHandle.swift
grynspan: Okay, works for me!

Sources/Testing/Support/FileHandle.swift
grynspan: @MaxDesiatov I want to avoid accidentally opting us into a state where the output is garbled. If we're not certain a feature works properly, it's better to disable it now and come back to it later than to preemptively enable it.

Pipes are problematic in this way because you can run a series of pipes that ultimately ends at a terminal emulator that supports ANSI escape codes, but actually determining whether outputtting ANSI escape codes is one of the Hard Problems™ in Computer Science™.

But, since @kateinoigakukun pointed out that `isatty()` is fully implemented, we can just defer to it. (Adding a flag to enable writing escape codes to an arbitrary file would be out of the scope of this PR, but we can chat about it in the forums.)

grynspan (approved):
:shipit:

grynspan:
Ready to merge when you are.

kateinoigakukun:
⛵

PR #588, opened 2024-08-01
Simplify `swiftTestingDirectoryPath`.

grynspan:
Since all platform-specific paths in this function end in `return nil`, we can simplify a bit.

Also a couple of drive-by newline changes.

### Checklist:


grynspan:
@swift-ci please test

PR #421, opened 2024-05-17
Add support for `swift test --quiet`.

grynspan:
This PR adds support for `swift test --quiet` and generalizes various "is verbose" checks into "verbosity level" checks instead.

Several of the properties of `__CommandLineArguments_v0` have been made optional because they are not required to be specified (and should never have been non-optional.)

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #592, opened 2024-08-01
Fix a crash when an issue is recorded on a detached task.

grynspan:
Given the following code:

```swift
@Test func f() async {
  await Task.detached {
    #expect(1 == 2)
  }.value
}
```

An issue is recorded on a detached task, but Swift Testing is unable to determine at runtime which test it corresponds to. This should simply result in the issue being attributed to an arbitrary test, or to the aether, but #421 caused a regression that triggers an optional unwrap crash instead.

This PR fixes the crash. :)

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #598, opened 2024-08-02
Implement an overload of `confirmation()` that takes an unbounded range.

grynspan:
Unbounded ranges are not meaningful when used with `confirmation()` because _any_ confirmation count matches. As well, the `...` operator produces an instance of `UnboundedRange` which is a non-nominal type and cannot conform to `RangeExpression`. It may be non-obvious to a developer why `...` doesn't work as the `expectedCount` argument to that function when other range operators work as expected.

This PR implements a stub overload of `confirmation()` that takes an unbounded range. The stub overload is marked unavailable and cannot be called.

Example usage:

```swift
await confirmation("Stuff happens", expectedCount: ...) { stuff in
  // ...
}
```

Generated diagnostic:

> 🛑 'confirmation(\_:expectedCount:sourceLocation:\_:)' is unavailable: Unbounded range '...' has no effect when used with a confirmation.

As a reminder, using a range expression with `confirmation()` is an experimental feature and has not been API-reviewed.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Environment.swift
+++ b/Sources/Testing/Support/Environment.swift
@@ -188,7 +188,7 @@ enum Environment {
               return nil
             case let errorCode:
               let error = Win32Error(rawValue: errorCode)
-              fatalError("unexpected error when getting environment variable '\(name)': \(error) (\(errorCode))")
+              fatalError("Unexpected error when getting environment variable '\(name)': \(error) (\(errorCode))")
             }
           } else if count > buffer.count {
             // Try again with the larger count.
--- a/Sources/TestingMacros/Support/AvailabilityGuards.swift
+++ b/Sources/TestingMacros/Support/AvailabilityGuards.swift
@@ -118,7 +118,7 @@ private func _createAvailabilityTraitExpr(
     return ".__unavailable(message: \(message), sourceLocation: \(sourceLocationExpr))"
 
   default:
-    fatalError("Unsupported keyword \(whenKeyword) passed to \(#function)")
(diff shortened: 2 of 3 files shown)


grynspan:
@swift-ci please test

Sources/Testing/Support/Environment.swift
```diff
@@ -188,7 +188,7 @@ enum Environment {
               return nil
             case let errorCode:
               let error = Win32Error(rawValue: errorCode)
-              fatalError("unexpected error when getting environment variable '\(name)': \(error) (\(errorCode))")
+              fatalError("Unexpected error when getting environment variable '\(name)': \(error) (\(errorCode))")
```
grynspan, line 191: Drive-by formatting tweak.

Sources/TestingMacros/Support/AvailabilityGuards.swift
```diff
@@ -118,7 +118,7 @@ private func _createAvailabilityTraitExpr(
     return ".__unavailable(message: \(message), sourceLocation: \(sourceLocationExpr))"
 
   default:
-    fatalError("Unsupported keyword \(whenKeyword) passed to \(#function)")
+    fatalError("Unsupported keyword \(whenKeyword) passed to \(#function). Please file a bug report at https://github.com/swiftlang/swift-testing/issues/new")
```
grynspan, line 121: Drive-by amendment to fatal error message.

grynspan:
@swift-ci please test

grynspan:
Self-reviewing (tsk.)

PR #603, opened 2024-08-07
Clarify that 8-bit exit codes aren't a problem on macOS and Windows.

grynspan:
The documentation for the experimental exit tests feature currently says that on POSIX-like systems, only the low 8 bits of a process' exit code are preserved. This would be true if we used `wait()`, `wait4()`, etc. and `WEXITSTATUS()`, but we use `waitid()` instead which is [supposed to](https://pubs.opengroup.org/onlinepubs/9699919799/functions/exit.html) preserve the full exit code. It does so on Darwin, but not on Linux; Windows doesn't use `waitid()` but does report the full exit code.

Now, we're not currently building for any other POSIX-like systems that support processes (WASI/Wasm doesn't count here), so I've left in some weasel words and added a canary unit test. It will let us know if/when we add a platform that where `waitid()` doesn't preserve all the bits of the exit code, and we can amend the documentation in that case.

Exit tests remain an experimental feature only.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
Looks like Linux, or at least our CI's version, does not correctly implement `waitid()` per the POSIX rules. I suspected as much.

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
Superseded by #615.

PR #610, opened 2024-08-12
Reduce overhead of `.expectationChecked` event handling in `#expect()`.

grynspan:
This PR refactors the implementation of `#expect()` and `#require()` a bit such that they don't incur more than minimal overhead posting `.expectationChecked` events if nobody is listening for them (which, currently, nobody is.)

We considered removing `.expectationChecked` outright, but XCTest has historically had a number of requests for a way to observe calls to `XCTAssert()` etc. even when they pass, so we opted not to remove the event kind at this time.

This PR also introduces a cache for fully-qualified type names so that we don't need to call into the runtime to get them as often.

Overall speedup is approximately **90% or 11x**. Test time for a tight loop of 1,000,000 `#expect()` calls goes from 5.898893 seconds down to 0.515558291 seconds (as measured on my work computer.)

Resolves rdar://133517028.

### Checklist:


grynspan:
@swift-ci please test

suzannaratcliff (approved):

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

PR #612, opened 2024-08-13
Fix a crash if an empty JSON configuration file is specified.

grynspan:
If you pass an empty file to `swift test --configuration-path`, we crash on an unexpected `nil` value. Fix it fix it fix it!

### Checklist:


grynspan:
@swift-ci please test

dennisweissmann (approved):

PR #614, opened 2024-08-15
Allow throwing an error from an exit test's body.

grynspan:
This PR amends the signatures of the exit test macros (`#expect(exitsWith:) {}` and `try #require(exitsWith:) {}`) to allow bodies to throw errors. If they do, they are treated as uncaught errors and the child process terminates abnormally (in the same way it does if an error is thrown from the main function of a Swift program.)

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #613, opened 2024-08-15
Implement an equality operator for `ExitCondition`.

grynspan:
This PR implements `==` and `===` for `ExitCondition`, part of the experimental exit tests feature. These operators are necessary in order to allow for exit tests to support more complex matching by trailing closure (e.g. to support inspecting `stdout`.) Because `.failure` is a fuzzy case, `==` fuzzy-matches while `===` exactly matches. `Hashable` conformance is unavailable.

Example usage:

```swift
let lhs: ExitCondition = .failure
let rhs: ExitCondition = .signal(SIGTERM)
print(lhs == rhs) // prints "true"
print(lhs === rhs) // prints "false"
```

Exit tests remain an experimental feature.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #615, opened 2024-08-15
Fixes to exit tests.

grynspan:
This PR supersedes #603, #613, and #614. Exit tests remain an experimental feature.

 ## Clarify that 8-bit exit codes aren't a problem on macOS and Windows. (#603)

The documentation for the experimental exit tests feature currently says that on POSIX-like systems, only the low 8 bits of a process' exit code are preserved. This would be true if we used `wait()`, `wait4()`, etc. and `WEXITSTATUS()`, but we use `waitid()` instead which is [supposed to](https://pubs.opengroup.org/onlinepubs/9699919799/functions/exit.html) preserve the full exit code. It does so on Darwin, but not on Linux; Windows doesn't use `waitid()` but does report the full exit code.

Now, we're not currently building for any other POSIX-like systems that support processes (WASI/Wasm doesn't count here), so I've left in some weasel words and added a canary unit test. It will let us know if/when we add a platform that where `waitid()` doesn't preserve all the bits of the exit code, and we can amend the documentation in that case.

 ## Implement an equality operator for ExitCondition. (#613)

This PR implements `==` and `===` for `ExitCondition`, part of the experimental exit tests feature. These operators are necessary in order to allow for exit tests to support more complex matching by trailing closure (e.g. to support inspecting `stdout`.) Because `.failure` is a fuzzy case, `==` fuzzy-matches while `===` exactly matches. `Hashable` conformance is unavailable.

Example usage:

```swift
let lhs: ExitCondition = .failure
let rhs: ExitCondition = .signal(SIGTERM)
print(lhs == rhs) // prints "true"
print(lhs === rhs) // prints "false"
```

 ## Allow throwing an error from an exit test's body. (#614)

This PR amends the signatures of the exit test macros (`#expect(exitsWith:) {}` and `try #require(exitsWith:) {}`) to allow bodies to throw errors. If they do, they are treated as uncaught errors and the child process terminates abnormally (in the same way it does if an error is thrown from the main function of a Swift program.)

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitCondition.swift
+++ b/Sources/Testing/ExitTests/ExitCondition.swift
@@ -104,6 +104,20 @@ extension ExitCondition: Equatable {
     }
   }
 
+  /// Check whether or not two values of this type are identical.
+  ///
+  /// - Parameters:
+  ///   - lhs: One value to compare.
+  ///   - rhs: Another value to compare.
+  ///
+  /// - Returns: Whether or not `lhs` and `rhs` are identical.
+  ///
+  /// This operator differs from [`==(lhs:rhs:)`](https://developer.apple.com/documentation/swift/equatable/==(_:_:)-3axv1)
+  /// in that ``failure`` will only compare equal to itself using this operator,
+  /// but will compare equal to any value except ``success`` when using
+  /// [`==(lhs:rhs:)`](https://developer.apple.com/documentation/swift/equatable/==(_:_:)-3axv1).
+  ///
+  /// For any values `a` and `b`, `a === b` implies that `a !== b` is false.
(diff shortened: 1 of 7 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test Windows

grynspan:
@swift-ci please test

suzannaratcliff (approved):

dennisweissmann (approved):

grynspan:
@swift-ci please test

grynspan:
@swift-ci please test

PR #616, opened 2024-08-16
[CMake/Macros] Set correct RPATH

rintaro:
Previously it expected the host (i.e. the compiler) already loaded all dependencies in the process. But since that's not guaranteed, let's set the correct RPATH.
Also, stop importing string-processing and backtracing.


rintaro:
@swift-ci Please test

grynspan (approved):

PR #617, opened 2024-08-17
[CMake] Temporarily remove Foundation dependency

rintaro:
As a workaround for testing swift-corelibs-foundation. SwiftPM is currently always link test targets with 'Testing', that causes 2 Foundation in the process when testing Foundation itself.


grynspan:
We absolutely need that code to compile.

PR #619, opened 2024-08-19
Allow move-only types as suites.

grynspan:
This PR enables using move-only types as suites. For example:

```swift
@Suite struct NumberOfBeesTests: ~Copyable {
  @Test consuming func countBees() {
    var count = 0
    for species in allSpecies {
      if species is Bee {
        count += species.populationCount
      }
    }

    #expect(count > 0)
  }
}
```

Move-only types have a number of constraints in Swift, and those constraints aren't lifted in a test target, but generally speaking a move-only type should be able to do all the things any other type can do _as a test suite_.

### Checklist:


grynspan:
@swift-ci please test

suzannaratcliff (approved):

SeanROlszewski (approved):

PR #620, opened 2024-08-20
Warn when passing a non-optional value to `try #require(T?)`.

grynspan:
This PR adds an overload of `try #require()` that warns the developer if they pass a non-optional, non-`Bool` value For example, this code:

```swift
let x = 0
let y = try #require(x)
```

Will produce the diagnostic:

> ⚠️ '#require(\_:\_:)' is redundant because 'x' never equals 'nil'

### Checklist:

### DIFF
--- a/Sources/TestingMacros/ConditionMacro.swift
+++ b/Sources/TestingMacros/ConditionMacro.swift
@@ -293,6 +293,26 @@ public struct AmbiguousRequireMacro: RefinedConditionMacro {
   }
 }
 
+/// A type describing the expansion of the `#require()` macro when it is passed
+/// a non-optional, non-`Bool` value.
+///
+/// This type is otherwise exactly equivalent to ``RequireMacro``.
+public struct NonOptionalRequireMacro: RefinedConditionMacro {
+  public typealias Base = RequireMacro
+
+  public static func expansion(
+    of macro: some FreestandingMacroExpansionSyntax,
+    in context: some MacroExpansionContext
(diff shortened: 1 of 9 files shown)


Sources/TestingMacros/ConditionMacro.swift
```diff
@@ -293,6 +293,26 @@ public struct AmbiguousRequireMacro: RefinedConditionMacro {
...
+///
+/// This type is otherwise exactly equivalent to ``RequireMacro``.
+public struct NonOptionalRequireMacro: RefinedConditionMacro {
+  public typealias Base = RequireMacro
+
+  public static func expansion(
+    of macro: some FreestandingMacroExpansionSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> ExprSyntax {
+    if let argument = macro.arguments.first {
```
suzannaratcliff, line 307: Where does this check if it is a Bool?

suzannaratcliff: I see, you are trying to not show a warning for a bool value, which this code does not do

suzannaratcliff (approved):

Sources/TestingMacros/ConditionMacro.swift
grynspan: `Bool` and `Bool?` won't hit this particular macro variant, they'll hit the normal one (or the "ambiguous `Bool?`" one.)

PR #624, opened 2024-08-20
Add `isolation` argument to functions taking non-sendable async closures.

### LINKED ISSUE #622: confirmation() does not work from @MainActor tests
### Description

The following test function will not compile in the Swift 6 language mode:

```swift
@MainActor @Test func example() async {
	await confirmation { _ in } // Error: Sending main actor-isolated value of type '(Confirmation) async -> ()' with later accesses to nonisolated context risks causing data races
}
```

### Expected behavior

I think this should just work. `confirmation()` should take (and respect) an additional isolation argument `isolation: isolated (any Actor)? = #isolation`.

Proposed new signature (also discussed [here](https://forums.swift.org/t/where-to-order-isolation-and-sourcelocation-arguments/74070)):
```swift
func confirmation<R>(
    _ comment: Testing.Comment? = nil,
    expectedCount: Int = 1,
    isolation: isolated (any Actor)? = #isolation,
    sourceLocation: Testing.SourceLocation = #_sourceLocation,
    _ body: (Testing.Confirmation) async throws -> R
) async rethrows -> R
```

### Actual behavior

_No response_

### Steps to reproduce

_No response_

Xcode 16 beta 5

swift-driver version: 1.113 Apple Swift version 6.0 (swiftlang-6.0.0.7.6 clang-1600.0.24.1)
Darwin Mac.localdomain 24.0.0 Darwin Kernel Version 24.0.0: Wed Aug  7 03:08:56 PDT 2024; root:xnu-11215.1.9~22/RELEASE_ARM64_T6020 arm64

grynspan:
This PR adds an `isolation` parameter to several API functions that take `async` closures that are not required to be sendable. As well, it adds `sending` to those closures' return types where appropriate.

This change is necessary in Swift 6 because a non-sendable async closure could be called in the wrong isolation domain otherwise. In particular, if the caller is `@MainActor`-isolated, the closure will not be called on the main actor and will therefore always hop, and a concurrency error occurs:

```swift
@MainActor func f() async {
  await confirmation {
    // this inner closure is not actor-isolated, but can't be sent across
    // isolation domains.
  }
}
```

`withKnownIssue()` and `confirmation()` are affected, as are several async overloads of the internal-but-public `__check()` function family.

This change is necessary for correctness, but is potentially source-breaking if calling code directly references the modified functions by full name.

Resolves #622.

### Checklist:

### DIFF
--- a/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
+++ b/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
@@ -835,10 +835,11 @@ public func __checkClosureCall<E>(
 ///   `#require()` macros. Do not call it directly.
 public func __checkClosureCall<E>(
   throws errorType: E.Type,
-  performing body: () async throws -> some Any,
+  performing body: () async throws -> sending some Any,
   expression: __Expression,
   comments: @autoclosure () -> [Comment],
   isRequired: Bool,
+  isolation: isolated (any Actor)? = #isolation,
   sourceLocation: SourceLocation
 ) async -> Result<Void, any Error> where E: Error {
   if errorType == Never.self {
@@ -848,6 +849,7 @@ public func __checkClosureCall<E>(
       expression: expression,
       comments: comments(),
       isRequired: isRequired,
+      isolation: isolation,
       sourceLocation: sourceLocation
     )
   } else {
@@ -858,6 +860,7 @@ public func __checkClosureCall<E>(
       expression: expression,
       comments: comments(),
       isRequired: isRequired,
+      isolation: isolation,
       sourceLocation: sourceLocation
     )
   }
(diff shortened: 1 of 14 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

macguru:
I'm really not an expert on this, but I just stumbled across `@_inheritActorContext` in the [Closure isolation control](https://github.com/sophiapoirier/swift-evolution/blob/closure-isolation/proposals/nnnn-closure-isolation-control.md#closure-isolation-control) proposal, and am wondering what the tradeoff between that and this pull request's solution is.

As far as I understand, the following change should provide the same behavior to clients, with the benefit of not adding an additional argument to the signature:

```swift
func confirmation<R>(
    _ comment: Testing.Comment? = nil,
    expectedCount: Int = 1,
    sourceLocation: Testing.SourceLocation = #_sourceLocation,
    @_inheritActorContext _ body: (Testing.Confirmation) async throws -> R
) async rethrows -> R
```

Then again, `@_inheritActorContext` seems to be experimental and there is maybe more flexibility in having an `(any Actor)?` inside the function bodies…

grynspan:
`@_inheritActorContext` is not the "modern" way to propagate actor isolation into a closure, no. :)

grynspan:
@swift-ci test

briancroom (approved):

PR #625, opened 2024-08-21
Generalize optional chaining detection used by `#require()`.

### LINKED ISSUE #623: #require broken when accessing optional property
### Description

The following code does not compile currently. It produces an incorrect warning **and** an error inside the `#expect` macro compilation:

```swift
protocol MyType {
    func function() -> Something
}

protocol Something {
    var optional: Any? { get }
}

@MainActor @Test func mainExample() throws {
    let object: MyType?
    _ = try #require(object?.function().optional) // Warning: No calls to throwing functions occur within 'try' expression
    // Error: Value of optional type '(any Something)?' must be unwrapped to a value of type 'any Something'
}
```

As far as my testing goes, this needs a chaining of three things inside the `#require`:
- an optional base value
- a function returning a non-optional value (which is transitively optional, because of the base)
- an optional property defined on the function's return value

I tried to be general here, but a common scenario is some object that has a function like `makeArray() -> [Int]` which is then accessed using `first` (e.g because we know there's just one element in this particular situation).

### Expected behavior

The code should compile. The workaround is to split this into two expectations at any point in the chain. For example like so:
```swift
@MainActor @Test func mainExample() throws {
    let object: MyType?
    let some = try #require(object?.function())
    _ = try #require(some.optional)
}
```

### Actual behavior

_No response_

### Steps to reproduce

_No response_

Xcode 16 beta 5

swift-driver version: 1.113 Apple Swift version 6.0 (swiftlang-6.0.0.7.6 clang-1600.0.24.1)
Darwin Mac.localdomain 24.0.0 Darwin Kernel Version 24.0.0: Wed Aug  7 03:08:56 PDT 2024; root:xnu-11215.1.9~22/RELEASE_ARM64_T6020 arm64

grynspan:
This PR generalizes the optional chaining detection used in the implementation of unwrapping `#require()` such that it works with more than just member access expressions.

Resolves #623.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/ConditionArgumentParsing.swift
+++ b/Sources/TestingMacros/Support/ConditionArgumentParsing.swift
@@ -242,6 +242,18 @@ private func _parseCondition(from expr: ClosureExprSyntax, for macro: some Frees
   return Condition(expression: expr)
 }
 
+/// A class that walks a syntax tree looking for optional chaining expressions
+/// such as `a?.b.c`.
+private final class _OptionalChainFinder: SyntaxVisitor {
+  /// Whether or not any optional chaining was found.
+  var optionalChainFound = false
+
+  override func visit(_ node: OptionalChainingExprSyntax) -> SyntaxVisitorContinueKind {
+    optionalChainFound = true
+    return .skipChildren
+  }
+}
+
 /// Extract the underlying expression from an optional-chained expression as
 /// well as the number of question marks required to reach it.
 ///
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci please test

grynspan:
@swift-ci test Linux

Sources/TestingMacros/Support/ConditionArgumentParsing.swift
```diff
@@ -279,15 +291,9 @@ private func _exprFromOptionalChainedExpr(_ expr: some ExprSyntaxProtocol) -> (E
...
-    func isOptionalChained(_ expr: some ExprSyntaxProtocol) -> Bool {
-      if expr.is(OptionalChainingExprSyntax.self) {
-        return true
-      } else if let memberAccessBaseExpr = expr.as(MemberAccessExprSyntax.self)?.base {
-        return isOptionalChained(memberAccessBaseExpr)
-      }
-      return false
-    }
-    if isOptionalChained(originalExpr) {
+    let optionalChainFinder = _OptionalChainFinder(viewMode: .sourceAccurate)
```
briancroom, line 294: Nit: did you consider defining a `static` helper on the new class that would allow the usage here to be a one-liner?

briancroom (approved):

Sources/TestingMacros/Support/ConditionArgumentParsing.swift
grynspan: That would total more lines of code than just calling it as-is. :)

grynspan:
@swift-ci test

PR #626, opened 2024-08-21
Disallow `@Test` on member functions of `XCTestCase` subclasses.

grynspan:
This PR adds a diagnostic if we detect that `@Test` has been used on a member function of an `XCTestCase` subclass. This was meant to be disallowed, but we neglected to add a check after adopting `lexicalContext`.

### Checklist:


grynspan:
@swift-ci test

suzannaratcliff (approved):
This makes sense! Do we want to do something like this for `@Suite` as well?

grynspan:
`@Suite` already does this, note the comment I modified in SuiteDeclarationMacro.swift.

PR #633, opened 2024-08-21
Fix incorrectly-specified platform list in CMake file.

grynspan:
Our CMake file that sets compiler conditionals has the wrong syntax when setting `SWT_NO_EXIT_TESTS` conditionally. Fix that. (With thanks to @compnerd.)

### Checklist:


compnerd (approved):

grynspan:
@swift-ci test

PR #648, opened 2024-08-28
Plumb through the testing library version into a function instead of a C++ macro.

grynspan:
This PR moves the consumption of the testing library version (via C++ macro `SWT_TESTING_LIBRARY_VERSION`) to a C++ function. It seems that defining it in CMake for C++ targets doesn't trigger for the Swift side when it imports symbols from C headers.

Will verify with a toolchain build.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Versions.swift
+++ b/Sources/Testing/Support/Versions.swift
@@ -125,7 +125,7 @@ let simulatorVersion: String = {
 ///
 /// This value is not part of the public interface of the testing library.
 var testingLibraryVersion: String {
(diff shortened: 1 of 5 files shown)


grynspan:
@swift-ci test

grynspan:
Verified this works in a toolchain build:

> ◇ Test run started.
> ↳ Testing Library Version: 6.1-dev (26451155133ecb2)
> ✔ Test run with 0 tests passed after 0.001 seconds.

grynspan:
Self-reviewing. @briancroom is aware of the change.

PR #651, opened 2024-08-29
CMake: Install `_TestingInternals` in static library builds

kateinoigakukun:
When building the Testing library as a static library (BUILD_SHARED_LIBS=FALSE), `_TestingInternals` is not included in the `libTesting.a` archive by default. (when building as a shared library, `libTesting.so` includes `_TestingInternals`). This causes the linker to complain about missing symbols, so we need to manually merge the objects from `_TestingInternals` into `libTesting.a`.

This change is required to ship swift-testing as a part of Swift SDK for WebAssembly

### Checklist:

### DIFF
--- a/Sources/Testing/CMakeLists.txt
+++ b/Sources/Testing/CMakeLists.txt
@@ -94,11 +94,16 @@ add_library(Testing
   Traits/Trait.swift)
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
+if(NOT BUILD_SHARED_LIBS)
+  # When building a static library, tell clients to autolink the internal
+  # library.
+  target_compile_options(Testing PRIVATE
+    "SHELL:-Xfrontend -public-autolink-library -Xfrontend _TestingInternals")
+endif()
 add_dependencies(Testing
   TestingMacros)
 target_compile_options(Testing PRIVATE
   -enable-library-evolution
   -emit-module-interface -emit-module-interface-path $<TARGET_PROPERTY:Testing,Swift_MODULE_DIRECTORY>/Testing.swiftinterface)
 
-include(SwiftModuleInstallation)
 _swift_testing_install_target(Testing)
--- a/cmake/modules/SwiftModuleInstallation.cmake
+++ b/cmake/modules/SwiftModuleInstallation.cmake
@@ -18,27 +18,43 @@ function(get_swift_host_os result_var_name)
   set(${result_var_name} ${SWIFT_SYSTEM_NAME} PARENT_SCOPE)
 endfunction()
 
-function(_swift_testing_install_target module)
+# Returns the path to the Swift Testing library installation directory
+#
+# Usage:
+#   get_swift_testing_install_lib_dir(type result_var_name)
+#
+# Arguments:
+#   type: The type of the library (STATIC_LIBRARY, SHARED_LIBRARY, or EXECUTABLE).
+#         Typically, the value of the TYPE target property.
+#   result_var_name: The name of the variable to set
+function(get_swift_testing_install_lib_dir type result_var_name)
   get_swift_host_os(swift_os)
-  get_target_property(type ${module} TYPE)
-
   if(type STREQUAL STATIC_LIBRARY)
     set(swift swift_static)
   else()
     set(swift swift)
   endif()
 
+  if(APPLE)
+    set(${result_var_name} "lib/${swift}/${swift_os}/testing" PARENT_SCOPE)
+  else()
+    set(${result_var_name} "lib/${swift}/${swift_os}" PARENT_SCOPE)
+  endif()
+endfunction()
+
+function(_swift_testing_install_target module)
   target_compile_options(Testing PRIVATE "-no-toolchain-stdlib-rpath")
 
-  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
-    set(lib_destination_dir "lib/${swift}/${swift_os}/testing")
+  if(APPLE)
     set_property(TARGET ${module} PROPERTY
       INSTALL_RPATH "@loader_path/..")
   else()
-    set(lib_destination_dir "lib/${swift}/${swift_os}")
     set_property(TARGET ${module} PROPERTY
       INSTALL_RPATH "$ORIGIN")
   endif()
+
+  get_target_property(type ${module} TYPE)
+  get_swift_testing_install_lib_dir(${type} lib_destination_dir)
 
   install(TARGETS ${module}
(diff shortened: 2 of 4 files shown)


grynspan:
@etcwilde @rintaro Please review.

Sources/Testing/CMakeLists.txt
```diff
@@ -94,6 +94,11 @@ add_library(Testing
   Traits/Trait.swift)
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
+if(NOT BUILD_SHARED_LIBS)
+  # Merge private dependencies into single static objects archive
+  set_property(TARGET Testing PROPERTY STATIC_LIBRARY_OPTIONS
+    $<TARGET_OBJECTS:_TestingInternals>)
```
rintaro, line 100: Is there a reasons not to use `target_link_options(Testing $<TARGET_OBJECTS:_TestingInternals>)`?

kateinoigakukun: > Note This command cannot be used to add options for static library targets, since they do not use a linker. To add archiver or MSVC librarian flags, see the [STATIC_LIBRARY_OPTIONS](https://cmake.org/cmake/help/latest/prop_tgt/STATIC_LIBRARY_OPTIONS.html#prop_tgt:STATIC_LIBRARY_OPTIONS) target property.

https://cmake.org/cmake/help/latest/command/target_link_options.html

Yes, it's intentional. Options specified to `target_link_options` are passed only when linking but not archiving, so we need to use `STATIC_LIBRARY_OPTIONS` here.

etcwilde (commented):
I mean, this is kind of how static archives work, there should be a `lib_TestingInternals.a` that `libTesting.a` depends on.
You'll have to remove the dependency on `_TestingInternals.a` or you'll have duplicate symbols.
I think rather than going through the link options, you can actually list the objects as "sources" to the target, so
```cmake
add_library(BlahBlah ...)

# ...

target_sources(BlahBlah PRIVATE $<TARGET_OBJECTS:_TestingInternals>)
```
which I think looks a bit cleaner than mucking with the properties directly.

Are you testing with older versions of CMake as well? I seem to recall the old build model having no idea how to deal with object files getting linked into a Swift target so it would just omit them.

@compnerd, do you have opinions here?

grynspan:
We definitely would prefer it not to be possible to directly link to _TestingInternals as it's not, as the name suggests, public. It's definitely not ABI- or API-stable and just houses our C++ gunk.

etcwilde:
From the shared library perspective, that's fine, they have things like "visibility". But that's not really one of the facilities that static archives provide in general. Whether it's all in one archive or two, the symbols are still there for the poking. If people do poke at it, that's on them though.

TBH, we should probably update the minimum CMake version and take the sources that are currently going into `lib_TestingInternals.a` and shove them directly into `libTesting.a`. Then there is no _TestingInternals library to link against. The difference here is that we

grynspan:
… is that we…?

etcwilde:
That's likely me talking to future me. :)

compnerd:
I think that I would prefer that we expose the internal library. The way that you build the library matters as that is encoded into the module and the static nature of the library (e.g. if `_TestingInternals` is static and `Testing` is dynamic, the modules would be isolated but then merged into a single library and `_TestingInternals` would be fully internalized).

kateinoigakukun:
I like @compnerd's option as its simplicity and it does not require the new split build model in CMake.
Updated the change to do that.

cmake/modules/SwiftModuleInstallation.cmake
```diff
@@ -18,28 +18,44 @@ function(get_swift_host_os result_var_name)
...
+# Returns the path to the Swift library installation directory
+#
+# Usage:
+#   get_swift_install_lib_dir(type result_var_name)
+#
+# Arguments:
+#   type: The type of the library (STATIC_LIBRARY, SHARED_LIBRARY, or EXECUTABLE).
+#         Typically, the value of the TYPE target property.
+#   result_var_name: The name of the variable to set
+function(get_swift_install_lib_dir type result_var_name)
```
etcwilde, line 30: I kind of wonder if we should factor this out into its own module, sort of like `GNUInstallDirs`, and have it set a bunch of cached variables so that it's easier for distributors to override? Would be fine for a different PR though.

etcwilde (approved):

etcwilde:
CC @compnerd

cmake/modules/SwiftModuleInstallation.cmake
compnerd: I'm not sure if we should bother. I'd really rather just sink this down into CMake itself and support `install(TARGETS ...)` properly for Swift targets.

cmake/modules/SwiftModuleInstallation.cmake
```diff
@@ -18,28 +18,44 @@ function(get_swift_host_os result_var_name)
...
   get_swift_host_os(swift_os)
-  get_target_property(type ${module} TYPE)
-
   if(type STREQUAL STATIC_LIBRARY)
     set(swift swift_static)
   else()
     set(swift swift)
   endif()
 
+  if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```
compnerd, line 38:
```suggestion
  if(APPLE)
```

cmake/modules/SwiftModuleInstallation.cmake
```diff
@@ -18,28 +18,44 @@ function(get_swift_host_os result_var_name)
...
+    set(${result_var_name} "lib/${swift}/${swift_os}/testing" PARENT_SCOPE)
+  else()
+    set(${result_var_name} "lib/${swift}/${swift_os}" PARENT_SCOPE)
+  endif()
+endfunction()
+
+function(_swift_testing_install_target module)
   target_compile_options(Testing PRIVATE "-no-toolchain-stdlib-rpath")
 
   if(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
```
compnerd, line 48:
```suggestion
  if(APPLE)
```

compnerd (approved):

cmake/modules/SwiftModuleInstallation.cmake
kateinoigakukun: Either way, the library installation directory of swift-testing is a bit different than the conventional one (not `lib/swift/<os>` but `lib/swift/<os>/testing` on darwin), so we need a project specific logic that cannot be shared across packages.

kateinoigakukun:
@swift-ci test

PR #652, opened 2024-08-29
Expose the target triple as a separate string from the testing library version.

grynspan:
This PR adds a script to our CMake build to capture the target triple being used to build Swift Testing so that we can include it in our diagnostic output.

Apple's fork of Swift Testing (as included with Xcode 16) also includes this information, and it is very useful to have when reviewing bug reports as it can tell us if a problem is specific to a given architecture or OS variant.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@kateinoigakukun FYI some WASI-specific tweaks here.

Sources/_TestingInternals/Versions.cpp
```diff
@@ -10,11 +10,32 @@
 
 #include "Versions.h"
 
+#if defined(_SWT_TESTING_LIBRARY_VERSION) && !defined(SWT_TESTING_LIBRARY_VERSION)
+#warning _SWT_TESTING_LIBRARY_VERSION is deprecated
+#warning Define SWT_TESTING_LIBRARY_VERSION and optionally SWT_TARGET_TRIPLE instead
```
stmontgomery, line 15: Do we even need to bother deprecating it, since it was never public?

grynspan: This is a note to ourselves so that when Apple's fork of the project upstreams this change, we'll remember to adjust the relevant code.

grynspan: (I'll come back and rip this bit out when the dust has settled.)

stmontgomery (approved):

kateinoigakukun (approved):

PR #654, opened 2024-08-30
Special-case `FileHandle(forWritingAtPath: "CONOUT$")` on Windows.

grynspan:
This PR special-cases calls to `FileHandle.init(forWritingAtPath:)` when passed `"CONOUT$"` (the reserved "console output" file) on Windows such that it does not open a new file but just returns a copy of `.stdout` instead.

This change is necessary because, on Windows, opening and locking `"CONOUT$"` does not lock `stdout` (they're considered entirely different files) and output written to `"CONOUT$"` is wrapped at the console's column limit even though it's being written as binary data. The VSCode Swift plugin authors would like to be able to specify writing the JSON event stream to `"CONOUT$"`, but the observed behaviour is stopping them from doing so.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/FileHandle.swift
+++ b/Sources/Testing/Support/FileHandle.swift
@@ -88,6 +88,22 @@ struct FileHandle: ~Copyable, Sendable {
   ///
   /// - Throws: Any error preventing the stream from being opened.
   init(forWritingAtPath path: String) throws {
+#if os(Windows)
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci test

dschaefer2:
LOL, got to love Windows. Nice catch!

plemarquand (approved):

grynspan:
@swift-ci test

PR #661, opened 2024-09-04
Remove `#expect(throws: Never.self)` and `#require(throws: Never.self)` as distinct overloads.

grynspan:
`#expect(throws: Never.self)` and `#require(throws: Never.self)` do not need to be specified as separate overloads. This PR removes them as distinct symbols:

- The documentation for `#expect(throws: Never.self)` is merged into the documentation for `#expect(throws:)`.
- The diagnostic emitted when using `#require(throws: Never.self)` is lowered from a deprecation warning to a macro-generated custom warning.

This change has no other compile-time or runtime effects: code that passes `Never.self` to either macro will continue to compile, and macros cannot be referenced so there is no risk of source-level breakage like there might be from a function signature change.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

PR #666, opened 2024-09-06
Don't use `Synchronization.Atomic` for `deliverExpectationCheckedEvents`.

grynspan:
This PR replaces the new use of `Atomic<Int>` with a `Locked<Int>` in the implementation of `deliverExpectationCheckedEvents`. Why? Because we're running into some environments where the Synchronization module isn't available (e.g. older host macOSes) and this is simpler. The performance profile is comparable: on my system, running the `repeatedlyExpect()` test takes 0.55s instead of 0.49s to call `#expect()` 1,000,000 times, so it's still a significant win over the implementation we had earlier.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

PR #670, opened 2024-09-06
Exit tests shouldn't emit crash logs.

grynspan:
When a process crashes, the platforms we support all write a crash log (or core, or error report) to disk. This is not necessary for exit test child processes (since we intentionally crash them.)

This PR suppresses generation of those logs from within an exit test:

- On Darwin, we disable the exception port for crash report files;
- On Linux, we set the maximum core dump size to zero bytes; and
- On Windows, we disable Windows Error Reporting.

The above effects are only applied to the exit test's child process, not to the parent test process nor the system as a whole.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -37,6 +37,9 @@ public struct ExitTest: Sendable {
     // test. In the future, we might want to investigate actually setting up a
     // listener port in the parent process and tracking interesting exceptions
     // as separate exit conditions.
+    //
+    // BUG: The system may still opt to write crash logs to /Library/Logs
+    // instead of the user's home folder. rdar://47982238
     _ = task_set_exception_ports(
       swt_mach_task_self(),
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci test

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -37,6 +37,32 @@ public struct ExitTest: Sendable {
   /// terminate the process in a way that causes the corresponding expectation
   /// to fail.
   public func callAsFunction() async -> Never {
+#if SWT_TARGET_OS_APPLE && !SWT_NO_MACH_PORTS
```
briancroom, line 40: Maybe extract all this out into a helper function?

grynspan: Sure!

grynspan:
@swift-ci test

compnerd (approved):

stmontgomery (approved):

grynspan:
@swift-ci test

PR #672, opened 2024-09-08
Fix a typo in a call of install path getter function name

kateinoigakukun:
This is a follow-up fix to f078f7af1e5 that I made at the last minute before merging...

### DIFF
--- a/Sources/_TestingInternals/include/Includes.h
+++ b/Sources/_TestingInternals/include/Includes.h
@@ -58,7 +58,7 @@
 #include <sys/fcntl.h>
 #endif
 
-#if __has_include(<sys/resource.h>)
+#if __has_include(<sys/resource.h>) && !defined(__wasi__)
 #include <sys/resource.h>
 #endif
 
--- a/Sources/_TestingInternals/CMakeLists.txt
+++ b/Sources/_TestingInternals/CMakeLists.txt
@@ -22,7 +22,7 @@ if(NOT BUILD_SHARED_LIBS)
   # When building a static library, install the internal library archive
   # alongside the main library. In shared library builds, the internal library
   # is linked into the main library and does not need to be installed separately.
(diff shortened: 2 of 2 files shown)


Sources/_TestingInternals/include/Includes.h
```diff
@@ -58,7 +58,7 @@
 #include <sys/fcntl.h>
 #endif
 
-#if __has_include(<sys/resource.h>)
+#if __has_include(<sys/resource.h>) && !defined(__wasi__)
```
grynspan, line 61: What's the issue here? (I assume this fix is needed to unbreak Wasm.)

grynspan:
@swift-ci test

Sources/_TestingInternals/include/Includes.h
grynspan: Oh, I see it's in the commit message. n/m

grynspan (approved):

Sources/_TestingInternals/include/Includes.h
kateinoigakukun: Including `sys/resource.h` from wasi-libc requires a feature macro to be defined (`-D_WASI_EMULATED_PROCESS_CLOCKS`) and emits compilation errors otherwise. However, the header is not used in WASI paths, so it is safe to exclude it from the list of includes.

See `sys/times.h` that is transitively included by `sys/resource.h` https://github.com/WebAssembly/wasi-libc/blob/wasi-sdk-22/libc-top-half/musl/include/sys/times.h

MaxDesiatov (approved):

stmontgomery (approved):

grynspan:
@kateinoigakukun Note if you need this fix in Swift 6.0.1, please clone the PR and nominate it using the procedure outlined [here](https://forums.swift.org/t/swift-6-0-release-process/70220).

PR #673, opened 2024-09-09
Ask Foundation to capture `NSError`/`CFError` backtraces for us.

grynspan:
On Apple platforms, when an error occurs in an Objective-C method or C function, convention is to return the error as an instance of `NSError`/`CFError` via an out-parameter. When that Objective-C method or C function is called by a Swift function, the Swift function detects the error, then effectively rethrows it, at which point our `swift_willThrow` hook is triggered. This can obscure the real origin of the error which may be many stack frames down from the point Swift takes over.

This PR asks Foundation, via a relatively new internal function, to capture backtraces for instances of `NSError` and `CFError` at the point they are created. Then, when Swift Testing attempts to look up the backtrace for an error it has caught, if Foundation has generated one, then Swift Testing can substitute it in place of the one it generated in the `swift_willThrow` hook.

Resolves rdar://114386243.

### Checklist:

### DIFF
--- a/Tests/TestingTests/BacktraceTests.swift
+++ b/Tests/TestingTests/BacktraceTests.swift
@@ -108,7 +108,7 @@ struct BacktraceTests {
   }
 
   @Test("Thrown NSError has a different backtrace than we generated", .enabled(if: Backtrace.isFoundationCaptureEnabled))
-  func foundationGeneratedNSError() throws {
+  func foundationGeneratedNSError() {
     do {
       try throwNSError()
     } catch {
@@ -129,7 +129,6 @@ struct BacktraceTests {
     }
   }
 #endif
-
 
   @Test("Backtrace.current() is populated")
   func currentBacktrace() {
--- a/Tests/TestingTests/BacktraceTests.swift
+++ b/Tests/TestingTests/BacktraceTests.swift
@@ -95,7 +95,41 @@ struct BacktraceTests {
       await runner.run()
     }
   }
+
+  @inline(never)
+  func throwNSError() throws {
+    let error = NSError(domain: "Oh no!", code: 123, userInfo: [:])
+    throw error
+  }
+
+  @inline(never)
+  func throwBacktracedRefCountedError() throws {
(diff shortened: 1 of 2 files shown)


Tests/TestingTests/BacktraceTests.swift
```diff
@@ -95,8 +95,42 @@ struct BacktraceTests {
...
+    throw error
+  }
+
+  @inline(never)
+  func throwBacktracedRefCountedError() throws {
+    throw BacktracedRefCountedError()
+  }
+
+  @Test("Thrown NSError has a different backtrace than we generated", .enabled(if: Backtrace.isFoundationCaptureEnabled))
+  func foundationGeneratedNSError() throws {
```
stmontgomery, line 111: Guess it doesn't need `throws`?
```suggestion
  func foundationGeneratedNSError() {
```

Tests/TestingTests/BacktraceTests.swift
```diff
@@ -95,8 +95,42 @@ struct BacktraceTests {
...
+      try throwBacktracedRefCountedError()
+    } catch {
+      let backtrace1 = Backtrace(forFirstThrowOf: error, checkFoundation: true)
+      let backtrace2 = Backtrace(forFirstThrowOf: error, checkFoundation: false)
+      #expect(backtrace1 == backtrace2)
+    }
+  }
 #endif
 
+
```
stmontgomery, line 133: stray newline

stmontgomery (approved):
Nice!

Tests/TestingTests/BacktraceTests.swift
grynspan: Oops!

Tests/TestingTests/BacktraceTests.swift
grynspan: Yeah fair

grynspan:
@swift-ci test

grynspan:
@swift-ci test

PR #676, opened 2024-09-10
Optionally symbolicate backtraces.

grynspan:
This PR adds the ability to symbolicate backtraces on Darwin and Windows. A few different modes are provided: mangled, demangled, and "precise demangled" (which includes symbol addresses and instruction pointer offsets.) Tools such as the Swift VS Code plugin will be able to adopt this new feature along with VS Code's new call stacks API (https://github.com/microsoft/vscode/pull/222126).

Note that on Linux, it is not currently possible to symbolicate backtraces meaningfully. The standard library's `Backtrace` type has the ability to do this for us, but we'll need some tweaks to its interface before we can adopt it.

Note also that Apple's internal Core Symbolication framework is not used; we may be able to add support for it in a future PR (or Apple may opt to use it in their internal fork of Swift Testing.)

There is no way to emit backtraces to the command line right now. I considered having `--very-verbose` imply backtraces, but it's something I'm going to reserve for a future PR after discussion with the community.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

plemarquand (approved):
Very nice!

compnerd (approved):

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedBacktrace.swift
```diff
@@ -0,0 +1,42 @@
...
+extension ABIv0 {
+  /// A type implementing the JSON encoding of ``Backtrace`` for the ABI entry
+  /// point and event stream output.
+  ///
+  /// This type is not part of the public interface of the testing library. It
+  /// assists in converting values to JSON; clients that consume this JSON are
+  /// expected to write their own decoders.
+  struct EncodedBacktrace: Sendable {
+    /// The frames in the backtrace.
+    var symbolicatedAddresses: [Backtrace.SymbolicatedAddress]
```
stmontgomery, line 20: I notice this property, which gets included in the `Codable` representation, it not itself a type from the `ABIv0` namespace; it's the actual underlying type from the testing library. Do we feel it's appropriate to share types which end up getting encoded like that? I just worry it makes it slightly harder to ensure the encoded types remain stable as the upstream type evolves.

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedBacktrace.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+
+extension ABIv0 {
+  /// A type implementing the JSON encoding of ``Backtrace`` for the ABI entry
+  /// point and event stream output.
+  ///
+  /// This type is not part of the public interface of the testing library. It
+  /// assists in converting values to JSON; clients that consume this JSON are
+  /// expected to write their own decoders.
+  struct EncodedBacktrace: Sendable {
```
stmontgomery, line 18: Can/should we document this new encoding type in the ABI documentation schema?

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedBacktrace.swift
```diff
@@ -0,0 +1,42 @@
...
+    /// The frames in the backtrace.
+    var symbolicatedAddresses: [Backtrace.SymbolicatedAddress]
+
+    init(encoding backtrace: borrowing Backtrace, in eventContext: borrowing Event.Context) {
```
stmontgomery, line 22: Nit: this nontrivial initializer might warrant DocC

Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -222,3 +222,12 @@ extension Test.Case {
     return try await Runner.RuntimeState.$current.withValue(runtimeState, operation: body)
   }
 }
+
+// MARK: - Event.Context convenience initializer
+
+extension Event.Context {
+  init() {
```
stmontgomery, line 229: Documentation?

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedBacktrace.swift
grynspan: It's a pure data type with a reasonable JSON representation already. We do the same with some other structures like `SourceLocation`. The `ABIv0` types aren't API or SPI surface and if, for whatever reason, this data type stops producing the JSON we want, we can add a bespoke encoded-form type for it in the future.

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedBacktrace.swift
grynspan: This remains experimental and underscored. We aren't ready to promote it to API.

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedBacktrace.swift
grynspan: These types all have `init(encoding:in:)` initializers already. This isn't new.

Sources/Testing/Running/Runner.RuntimeState.swift
grynspan: See #677—this `init()` won't stick around.

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedBacktrace.swift
stmontgomery: Yeah. We don't have to address this in this PR given that precedent. But I think it may be worthwhile for _all_ of these encoded types to have counterparts in the `ABIv0` namespace to help protect against breakages due to changes in their upstream types. Either we, or outside contributors, may change an existing property and not realize it will break the encoded type, leading to tools instability.

This is just a general suggestion about protecting the encoded representation of all these types. I can file a GH issue to follow up there.

stmontgomery (approved):

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedBacktrace.swift
grynspan: Let's hold off for now—we're likely going to have a v1 ABI to accomodate new features in Swift 6.1, at which point we will want to freeze the v0 ABI stuff.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

PR #677, opened 2024-09-11
Add the current configuration to `Event.Context`.

grynspan:
This PR adds a `var configuration: Configuration` property to `Event.Context` and adds a `verbosity` property to `Configuration`. We solve two problems:

1. An event handler cannot see the current configuration (without querying `Configuration.current`) which limits its ability to customize its behaviour; and
2. The verbosity of human-readable output is tracked by individual event handlers when it is intended to be part of a test run's configuration.

Since the configuration contains the event handler to which we pass the event context, the `eventHandler` property of the context is cleared before the event handler is called (thus breaking any reference cycles.)

Callers of the internal `Event.post()` function are now required to either pass _both_ the test and test case for the event, or to pass neither (in which case we now gather both from the runtime state in a single call rather than hitting the task-local twice.) This is effected with a tuple as an argument to `Event.post()` which is admittedly a bit wonky but does force our call sites to explicitly pass values for both the test and test case, where previously we could accidentally pass a specific test and then go look up the task-local current test case (which should have been `nil` in all the cases where we were doing this anyway.)

A couple of interfaces on the event recorder types have changed in this PR in source-breaking ways. The old interfaces remain present but deprecated; they will be removed later in the Swift 6.1 development cycle after tools authors have had time to migrate.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Event.swift
+++ b/Sources/Testing/Events/Event.swift
@@ -186,8 +186,9 @@ public struct Event: Sendable {
   ///
   /// - Parameters:
   ///   - kind: The kind of event that occurred.
-  ///   - test: The test for which the event occurred, if any.
-  ///   - testCase: The test case for which the event occurred, if any.
+  ///   - testAndTestCase: The test and test case for which the event occurred,
+  ///     if any. The default value of this argument is ``Test/current`` and
+  ///     ``Test/Case/current``.
   ///   - instant: The instant at which the event occurred. The default value
   ///     of this argument is `.now`.
   ///   - configuration: The configuration whose event handler should handle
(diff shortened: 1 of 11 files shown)


grynspan:
@swift-ci test

Sources/Testing/Running/Configuration+EventHandling.swift
```diff
@@ -20,6 +20,9 @@ extension Configuration {
   /// `eventHandler` but this method may also be used as a customization point
   /// to change how the event is passed to the event handler.
   func handleEvent(_ event: borrowing Event, in context: borrowing Event.Context) {
-    eventHandler(event, context)
+    var contextCopy = copy context
```
stmontgomery, line 23: How heavyweight do we think this copy operation could be? For example, in scenarios where there's a selected or skipped test filter applied with many entries?

grynspan: Not very. Copy-on-write means that we incur minimal actual memory traffic here.

Sources/Testing/Events/Event.swift
```diff
@@ -186,24 +186,28 @@ public struct Event: Sendable {
...
     // Create both the event and its associated context here at same point, to
-    // ensure their task local-derived values are the same.
+    // ensure their task local-derived values are the same. Note we set the
+    // configuration property of Event.Context to nil initially because we'll
+    // reset it to the actual configuration that handles the event when we call
+    // handleEvent() later, so there's no need to make a copy of it yet.
+    let (test, testCase) = testAndTestCase
     let event = Event(kind, testID: test?.id, testCaseID: testCase?.id, instant: instant)
-    let context = Event.Context(test: test, testCase: testCase)
+    let context = Event.Context(test: test, testCase: testCase, configuration: nil)
```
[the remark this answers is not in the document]
grynspan: This is intentional. Read the comment immediately above. :)

stmontgomery (approved):

Sources/Testing/Running/Configuration+EventHandling.swift
grynspan: For those at home reading this PR or these comments: the reason `Event.Context` is `borrowing` is to emphasize that its contents—instances of `Test`, `Test.Case`, and now `Configuration`, aren't _free_ and that copies should be avoided where possible. That doesn't preclude the possibility of _any_ copies, just forces us to think about them when they are needed.

PR #687, opened 2024-09-12
[6.0.1] Fix WASI build with CMake

kateinoigakukun:
Explanation: When building the Testing library as a static library (BUILD_SHARED_LIBS=FALSE), `_TestingInternals` is not included in the `libTesting.a` archive by default. (when building as a shared library, `libTesting.so` includes `_TestingInternals` objects). This causes the linker to complain about missing symbols, so we need to ship `lib_TestingInternals.a` and autolink it too.
 
Original PR: https://github.com/swiftlang/swift-testing/pull/651 https://github.com/swiftlang/swift-testing/pull/672
Risk: Low; NFC for non WASI platforms
Testing: Tested in ci.swift.org
Reviewer: @grynspan

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,4 +38,5 @@ if(NOT SWIFT_SYSTEM_NAME)
   endif()
 endif()
 
+include(SwiftModuleInstallation)
 add_subdirectory(Sources)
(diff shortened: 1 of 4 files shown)


shahmishal:
I updated the targeted branch to release/6.0.2 from release/6.0.1.

briancroom (approved):

PR #688, opened 2024-09-12
Require explicit backtrace and source location when creating and recording issues

stmontgomery:
This refactors some of the logic for constructing `SourceContext` and `Issue` instances to require explicit arguments in more places.

### Motivation:

We can provide clearer and more relevant details about recorded issues by ensuring that their backtrace and/or source location are accurate. In general, whenever collecting a backtrace, it's better to elide unhelpful frames from the testing library itself so that frames from user code are as close to the top of the call stack as possible. Similarly, whenever an explicit source location is known, it's best to propagate that in any recorded issues which originate at that location.

Currently, the `SourceContext` type has default parameter values for both its `backtrace` and `sourceLocation` properties. Having default values for these parameters means that callers elsewhere in the codebase may accidentally not specify a value, thereby not propagating useful information. Or, in the case of backtraces, they may rely on the default value which results in additional, irrelevant frames in the call stack (representing thunks inserted by the compiler for the default value). The initializer for `Issue` also includes a default value for one of its parameters.

### Modifications:

- Change `SourceContext.init(...)` to remove the default values for both of its parameters: `backtrace:` and `sourceLocation:`.
- Change `Issue.init(...)` to no longer have default values for any of its parameters.
- Remove the non-public, static `Issue.record(...)` functions, and change callers to rely on the instance `record(configuration:)` method. This consolidates the usage pattern across the `Testing` module for creating an `Issue` and then recording it.
- Change `SkipInfo.init(...)` to remove the default value for its `sourceContext:` parameter.
- Add a convenience `init(for:sourceLocation:)` which accepts an `Error` and correctly creates a `SourceContext` for it.
- Add deprecated overloads of all SPIs which were modified to remove default parameter values.
- Add convenience overloads to `TestingAdditions.swift` which _preserve_ default parameter values, to avoid unnecessary boilerplate in test-only code where it's safe to rely on such default values.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -178,7 +178,9 @@ func callExitTest(
     // common issues, however they would constitute a failure of the test
     // infrastructure rather than the test itself and perhaps should not cause
     // the test to terminate early.
-    Issue.record(.errorCaught(error), comments: comments(), backtrace: .current(), sourceLocation: sourceLocation, configuration: configuration)
+    let issue = Issue(kind: .errorCaught(error), comments: comments(), sourceContext: .init(sourceLocation: sourceLocation))
+    issue.record(configuration: configuration)
+
     return __checkValue(
       false,
       expression: expression,
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -178,7 +178,7 @@ func callExitTest(
     // common issues, however they would constitute a failure of the test
     // infrastructure rather than the test itself and perhaps should not cause
     // the test to terminate early.
-    let issue = Issue(kind: .errorCaught(error), comments: comments(), sourceContext: .init(sourceLocation: sourceLocation))
+    let issue = Issue(kind: .errorCaught(error), comments: comments(), sourceContext: .init(backtrace: .current(), sourceLocation: sourceLocation))
     issue.record(configuration: configuration)
 
     return __checkValue(
--- a/Sources/Testing/Issues/Confirmation.swift
+++ b/Sources/Testing/Issues/Confirmation.swift
@@ -176,7 +176,7 @@ public func confirmation<R>(
       let issue = Issue(
         kind: expectedCount.issueKind(forActualCount: actualCount),
         comments: Array(comment),
-        sourceContext: .init(sourceLocation: sourceLocation)
+        sourceContext: .init(backtrace: .current(), sourceLocation: sourceLocation)
       )
       issue.record()
     }
--- a/Sources/Testing/Issues/Confirmation.swift
+++ b/Sources/Testing/Issues/Confirmation.swift
@@ -173,12 +173,12 @@ public func confirmation<R>(
   defer {
     let actualCount = confirmation.count.rawValue
     if !expectedCount.contains(actualCount) {
-      Issue.record(
-        expectedCount.issueKind(forActualCount: actualCount),
+      let issue = Issue(
+        kind: expectedCount.issueKind(forActualCount: actualCount),
         comments: Array(comment),
-        backtrace: .current(),
-        sourceLocation: sourceLocation
+        sourceContext: .init(sourceLocation: sourceLocation)
       )
+      issue.record()
     }
   }
   return try await body(confirmation)
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -105,15 +105,15 @@ public struct Issue: Sendable {
   ///
   /// - Parameters:
   ///   - kind: The kind of issue this value represents.
-  ///   - comments: An array of comments describing the issue. This array may be
-  ///     empty.
+  ///   - comments: An array of comments describing the issue. The default value
+  ///     is an empty array.
   ///   - sourceContext: A ``SourceContext`` indicating where and how this issue
   ///     occurred. This defaults to a default source context returned by
   ///     calling ``SourceContext/init(backtrace:sourceLocation:)`` with zero
   ///     arguments.
   init(
     kind: Kind,
-    comments: [Comment],
+    comments: [Comment] = [],
     sourceContext: SourceContext = .init()
   ) {
     self.kind = kind
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -105,20 +105,41 @@ public struct Issue: Sendable {
   ///
   /// - Parameters:
   ///   - kind: The kind of issue this value represents.
-  ///   - comments: An array of comments describing the issue. The default value
-  ///     is an empty array.
+  ///   - comments: An array of comments describing the issue. This array may be
+  ///     empty.
   ///   - sourceContext: A ``SourceContext`` indicating where and how this issue
   ///     occurred. This defaults to a default source context returned by
   ///     calling ``SourceContext/init(backtrace:sourceLocation:)`` with zero
   ///     arguments.
   init(
     kind: Kind,
-    comments: [Comment] = [],
-    sourceContext: SourceContext = .init()
+    comments: [Comment],
+    sourceContext: SourceContext
   ) {
     self.kind = kind
     self.comments = comments
     self.sourceContext = sourceContext
+  }
+
+  /// Initialize an issue instance representing a caught error.
+  ///
+  /// - Parameters:
+  ///   - error: The error which was caught which this issue is describing.
+  ///   - sourceLocation: The source location of the issue. This value is used
+  ///     to construct an instance of ``SourceContext``.
+  ///
+  /// The ``sourceContext`` property will have a ``SourceContext/backtrace``
+  /// property whose value is the backtrace for the first throw of `error`.
+  init(
+    forCaughtError error: any Error,
+    sourceLocation: SourceLocation? = nil
+  ) {
+    let sourceContext = SourceContext(backtrace: Backtrace(forFirstThrowOf: error), sourceLocation: sourceLocation)
+    self.init(
+      kind: .errorCaught(error),
+      comments: [],
+      sourceContext: sourceContext
+    )
   }
 
   /// The error which was associated with this issue, if any.
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -108,9 +108,7 @@ public struct Issue: Sendable {
   ///   - comments: An array of comments describing the issue. This array may be
   ///     empty.
   ///   - sourceContext: A ``SourceContext`` indicating where and how this issue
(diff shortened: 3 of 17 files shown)


grynspan:
This is internal code, so it doesn't make a huge difference, but… the intent behind not having a default value is to force us callers to always spell it out, lest we accidentally forget to forward comments from our own callers.

Is this fixing a known issue or just housekeeping?

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/Issues/Confirmation.swift
```diff
@@ -173,12 +173,12 @@ public func confirmation<R>(
   defer {
     let actualCount = confirmation.count.rawValue
     if !expectedCount.contains(actualCount) {
-      Issue.record(
-        expectedCount.issueKind(forActualCount: actualCount),
+      let issue = Issue(
```
grynspan, line 176: This captures the wrong backtrace.

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -178,7 +178,9 @@ func callExitTest(
     // common issues, however they would constitute a failure of the test
     // infrastructure rather than the test itself and perhaps should not cause
     // the test to terminate early.
-    Issue.record(.errorCaught(error), comments: comments(), backtrace: .current(), sourceLocation: sourceLocation, configuration: configuration)
+    let issue = Issue(kind: .errorCaught(error), comments: comments(), sourceContext: .init(sourceLocation: sourceLocation))
```
grynspan, line 181: This captures the wrong backtrace.

Sources/Testing/Running/Runner.swift
```diff
@@ -340,13 +340,12 @@ extension Runner {
             try await testCase.body()
           }
         } timeoutHandler: { timeLimit in
-          Issue.record(
-            .timeLimitExceeded(timeLimitComponents: timeLimit),
+          let issue = Issue(
```
grynspan, line 343: This captures the wrong backtrace.

Sources/Testing/Issues/Issue.swift
```diff
@@ -114,13 +114,34 @@ public struct Issue: Sendable {
...
+  ///
+  /// - Parameters:
+  ///   - error: The error which was caught which this issue is describing.
+  ///   - sourceLocation: The source location of the issue. This value is used
+  ///     to construct an instance of ``SourceContext``.
+  ///
+  /// The ``sourceContext`` property will have a ``SourceContext/backtrace``
+  /// property whose value is the backtrace for the first throw of `error`.
+  init(
+    forCaughtError error: any Error,
```
grynspan, line 134:
```suggestion
    for error: any Error,
```

Sources/Testing/Issues/Issue.swift
```diff
@@ -114,13 +114,34 @@ public struct Issue: Sendable {
...
+    sourceLocation: SourceLocation? = nil
```
grynspan, line 135: This initializer makes it possible to accidentally produce an `Issue` with no source context at all, because `sourceLocation` defaults to `nil` and the backtrace is not guaranteed to be available.

grynspan (changes requested):
This change would introduce regressions in our backtrace quality and could cause us to record issues with no source context at all.

Sources/Testing/Issues/Issue.swift
stmontgomery: That's already true prior to this PR—for example, in the way `.errorCaught` issues are created in `Runner.Plan`.

grynspan: Right, but in the cases where it happens now, we had to explicitly write it out, right?

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -41,7 +41,7 @@ extension Runner {
       ///
       /// - Parameters:
       ///   - skipInfo: A ``SkipInfo`` representing the details of this skip.
-      indirect case skip(_ skipInfo: SkipInfo = .init())
+      indirect case skip(_ skipInfo: SkipInfo)
```
grynspan, line 44: This is a source-breaking tools SPI change. Can we keep the old signature but deprecate it? (We don't need ABI compatibility, just source, so a `public tatic func skip() -> Self` may be sufficient.)

Sources/Testing/Running/SkipInfo.swift
```diff
@@ -34,10 +34,11 @@ public struct SkipInfo: Sendable {
   ///     Defaults to `nil`.
   ///   - sourceContext: A source context indicating where this skip occurred.
   ///     Defaults to a source context returned by calling
-  ///     ``SourceContext/init(backtrace:sourceLocation:)`` with zero arguments.
+  ///     ``SourceContext/init(backtrace:sourceLocation:)`` passing only the
+  ///     current backtrace.
   public init(
     comment: Comment? = nil,
-    sourceContext: SourceContext = .init()
+    sourceContext: SourceContext = .init(backtrace: .current())
```
grynspan, line 41: This will still capture the wrong backtrace.

Sources/Testing/SourceAttribution/SourceContext.swift
```diff
@@ -26,11 +26,10 @@ public struct SourceContext: Sendable {
...
   /// - Parameters:
-  ///   - backtrace: The backtrace associated with the new instance. Defaults to
-  ///     the current backtrace (obtained via
-  ///     ``Backtrace/current(maximumAddressCount:)``).
-  ///   - sourceLocation: The source location associated with the new instance.
-  public init(backtrace: Backtrace? = .current(), sourceLocation: SourceLocation? = nil) {
+  ///   - backtrace: The backtrace associated with the new instance.
+  ///   - sourceLocation: The source location associated with the new instance,
+  ///     if available.
+  public init(backtrace: Backtrace?, sourceLocation: SourceLocation? = nil) {
```
grynspan, line 32: Can we get rid of the default value for `sourceLocation`? We should probably make ourselves spell it out explicitly.

grynspan: Also, removing the default value for `backtrace` is source-breaking.

Sources/Testing/Running/SkipInfo.swift
stmontgomery: Discussed offline; this one _already_ contained undesired frames in the backtrace. This change should reduce the number of junk frames, improving things slightly. But I'm actually going to change this further to remove the default `SourceContext` from `SkipInfo.init` entirely.

grynspan: We'll need to keep a deprecated source-compatible initializer around since this is tools SPI.

stmontgomery:
@swift-ci please test

Sources/Testing/Running/SkipInfo.swift
```diff
@@ -34,10 +34,11 @@ public struct SkipInfo: Sendable {
   ///     Defaults to `nil`.
   ///   - sourceContext: A source context indicating where this skip occurred.
   ///     Defaults to a source context returned by calling
-  ///     ``SourceContext/init(backtrace:sourceLocation:)`` with zero arguments.
+  ///     ``SourceContext/init(backtrace:sourceLocation:)`` passing only the
```
grynspan, line 37: Comment is out of date.

Sources/Testing/Issues/Issue.swift
stmontgomery: The common case for caught errors is to attempt to get the error's backtrace, if one is available, else `nil`. And typically caught errors don't include an explicitly-supplied `SourceLocation`. So I think this new convenience initializer strikes the right balance, and the PR overall helps make many things more explicit

Sources/Testing/SourceAttribution/SourceContext.swift
```diff
@@ -41,3 +40,17 @@ extension SourceContext: Equatable, Hashable {}
 // MARK: - Codable
 
 extension SourceContext: Codable {}
+
+// MARK: - Deprecated
+
+extension SourceContext {
+  @available(*, deprecated, message: "Use init(backtrace:sourceLocation:) and pass both arguments explicitly instead.")
```
grynspan, line 47: One more needed with no args, right?

stmontgomery: The second deprecated overload below, `init(sourceLocation:)`, still includes a default value, so that covers the zero-argument usage scenario. Only two overloads needed!

stmontgomery:
@swift-ci please test

Tests/TestingTests/TestSupport/TestingAdditions.swift
```diff
@@ -366,3 +366,27 @@ extension Trait where Self == TimeLimitTrait {
...
+    self.init(kind: kind, sourceContext: .init())
+  }
+
+  init(kind: Kind, sourceContext: SourceContext) {
+    self.init(kind: kind, comments: [], sourceContext: sourceContext)
+  }
+}
+
+extension SourceContext {
+  init() {
```
grynspan, line 381: If you want to be really concise, I suppose this initializer could just have two default args, yeah?

Tests/TestingTests/TestSupport/TestingAdditions.swift
```diff
@@ -366,3 +366,27 @@ extension Trait where Self == TimeLimitTrait {
     return Self(timeLimit: timeLimit)
   }
 }
+
+extension Issue {
+  init(kind: Kind) {
```
grynspan, line 371: Also for conciseness, this could be `kind: Kind, sourceContext: SourceContext = .init()`.

I wouldn't want to reintroduce that into production code, but this is just for tests. :)

grynspan (approved):

Tests/TestingTests/TestSupport/TestingAdditions.swift
stmontgomery: I tried that originally, but the compiler notes that it will cause infinite recursion:

```swift
  init(backtrace: Backtrace? = .current(), sourceLocation: SourceLocation? = nil) {
    self.init(backtrace: backtrace, sourceLocation: sourceLocation) // warning: function call causes an infinite recursion
  }
```

I'm not aware of a way to request that an overload in another module such as this call the _original_ initializer, to resolve that warning.

Tests/TestingTests/TestSupport/TestingAdditions.swift
stmontgomery: Yes, this one _can_ be simplified as you suggest without ambiguity. Thanks

stmontgomery:
@swift-ci please test

Tests/TestingTests/TestSupport/TestingAdditions.swift
grynspan: Dang it, alright.

PR #689, opened 2024-09-13
Remove the `Confirmation.ExpectedCount` marker protocol.

grynspan:
This PR replaces uses of `Confirmation.ExpectedCount` in our API surface with `RangeExpression<Int> & Sendable`. A composed protocol type `P<A> & Q` can only be expressed as `some`, not `any` as the latter is not yet implemented in the compiler (rdar://96960993), so we were using a separate protocol in place of that combination.

The downside of doing so is that it makes the legibility of the interface worse. Xcode and other IDEs will offer you `confirmation(expectedCount: Confirmation.ExpectedCount)` but don't tell you that you need to write some range expression such as `0...2`. You have to dig into the protocol definition/documentation to figure it out.

So this change changes the signature of the experimental `confirmation()` overload to take a `some RangeExpression<Int> & Sendable` and plumbs that through everywhere it's valid. Once we get to generating an `Issue`, we need an existential box (`any`) at which point we will just drop the `<Int>` bit until the compiler feature is added.

We have a parameterized unit test that takes an array of these values, so to keep it building and passing, I moved a copy of the `ExpectedCount` protocol to the fixtures section of the test file containing that test; it's present only in our test target and not in our API surface.

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Confirmation.swift
+++ b/Sources/Testing/Issues/Confirmation.swift
@@ -164,7 +164,7 @@ public func confirmation<R>(
 @_spi(Experimental)
 public func confirmation<R>(
   _ comment: Comment? = nil,
-  expectedCount: some Confirmation.ExpectedCount,
+  expectedCount: some RangeExpression<Int> & Sendable,
   isolation: isolated (any Actor)? = #isolation,
   sourceLocation: SourceLocation = #_sourceLocation,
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

stmontgomery (approved):

PR #693, opened 2024-09-16
[CMake] Explicitly link Testing to Foundation

rintaro:
Explicitly link to `Foundation` in CMake builds

If `Foundation_DIR` is passed, `find_package` finds the `Foundation` target and `target_link_libraries` adds required `-I` and `-l`/`-L`. If not, `find_package` fails and `target_link_libraries` just assumes it's a system library name and just adds `-lFoundation`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,9 @@ project(SwiftTesting
   LANGUAGES CXX Swift)
 
 if(NOT APPLE)
-  find_package(dispatch CONFIG)
+  if(NOT CMAKE_SYSTEM_NAME STREQUAL WASI)
+    find_package(dispatch CONFIG)
+  endif()
   find_package(Foundation CONFIG)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,11 @@ endif()
 project(SwiftTesting
   LANGUAGES CXX Swift)
 
+if(NOT APPLE)
+  find_package(dispatch CONFIG)
+  find_package(Foundation CONFIG)
+endif()
+
 include(GNUInstallDirs)
 
 list(APPEND CMAKE_MODULE_PATH
--- a/Sources/Testing/CMakeLists.txt
+++ b/Sources/Testing/CMakeLists.txt
@@ -97,8 +97,11 @@ add_library(Testing
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
 if(NOT APPLE)
+  if(NOT CMAKE_SYSTEM_NAME STREQUAL WASI)
+    target_link_libraries(Testing PUBLIC
+      dispatch)
+  endif()
   target_link_libraries(Testing PUBLIC
-    dispatch
     Foundation)
 endif()
 if(NOT BUILD_SHARED_LIBS)
--- a/Sources/Testing/CMakeLists.txt
+++ b/Sources/Testing/CMakeLists.txt
@@ -96,6 +96,11 @@ add_library(Testing
   Traits/Trait.swift)
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
+if(NOT APPLE)
+  target_link_libraries(Testing PUBLIC
+    dispatch
+    Foundation)
+endif()
 if(NOT BUILD_SHARED_LIBS)
   # When building a static library, tell clients to autolink the internal
   # library.


Sources/Testing/CMakeLists.txt
```diff
@@ -96,6 +96,8 @@ add_library(Testing
   Traits/Trait.swift)
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
+target_link_libraries(Testing PUBLIC
```
grynspan, line 99: Will this fail if for some reason we _can't_ find Foundation?

rintaro: I think it should fail with linker errors.

CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ endif()
 project(SwiftTesting
   LANGUAGES CXX Swift)
 
+if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
```
etcwilde, line 18:
```suggestion
if(NOT APPLE)
```

Sources/Testing/CMakeLists.txt
```diff
@@ -96,6 +96,11 @@ add_library(Testing
   Traits/Trait.swift)
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
```
etcwilde, line 99:
```suggestion
if(NOT APPLE)
```

CMakeLists.txt
```diff
@@ -15,6 +15,11 @@ endif()
...
+  find_package(dispatch)
```
etcwilde, line 19: If we go this route instead of using the build-directory configurations, we'll need to include a `Finddispatch.cmake` and `FindFoundation.cmake` in our modules directory and move these down below the `list(APPEND CMAKE_MODUKE_PATH ...` on line 25.

rintaro: I think `find_package()` tries Module mode first then falls backs to Config mode IIUC.
> When the [basic signature](https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature) is used, the command searches in Module mode first. If the package is not found, the search falls back to Config mode.

etcwilde: Sure, but I think we still would want the `Find<PackageName>.cmake` modules or I don't really see much point in not just using the `CONFIG` form.

rintaro: Let's just use `CONFIG` then.

Sources/Testing/CMakeLists.txt
```diff
@@ -96,6 +96,11 @@ add_library(Testing
   Traits/Trait.swift)
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
+if(NOT APPLE)
+  target_link_libraries(Testing PUBLIC
+    dispatch
```
kateinoigakukun, line 101: Can we optionalize this `dispatch` link as WASI doesn't support libdispatch? You can ensure the change doesn't break WASI build by `@swift-ci test WebAssembly` in cross repo testing.

rintaro: What would be the condition? `if(NOT WASI)`?

kateinoigakukun: I think it would be `if(NOT CMAKE_SYSTEM_NAME STREQUAL WASI)`.

rintaro: 👍 Since `WASI` is equivalent to `CMAKE_SYSTEM_NAME STREQUAL WASI`. I went with `WASI`

kateinoigakukun: `WASI` is available only in the bleeding edge CMakes (https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9659), so can we stick on `CMAKE_SYSTEM_NAME STREQUAL WASI` for now?

CMakeLists.txt
```diff
@@ -15,6 +15,13 @@ endif()
 project(SwiftTesting
   LANGUAGES CXX Swift)
 
+if(NOT APPLE)
+  if(NOT CMAKE_SYSTEM_NAME STREQUAL WASI)
+    find_package(dispatch CONFIG)
+  endif()
```
compnerd, line 21: Any reason to not just do this unconditionally? We could do this, it shouldn't prevent the configure or build.

Sources/Testing/CMakeLists.txt
```diff
@@ -96,6 +96,14 @@ add_library(Testing
   Traits/Trait.swift)
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
+if(NOT APPLE)
+  if(NOT CMAKE_SYSTEM_NAME STREQUAL WASI)
+    target_link_libraries(Testing PUBLIC
+      dispatch)
+  endif()
```
compnerd, line 103: @etcwilde what do you think of the generator expression way of writing this instead?

CMakeLists.txt
rintaro: Just to make it clear `WASI` won't use dispatch. Although we can do this unconditionally, I thought there's no reason we'd call this for WASI.

Sources/Testing/CMakeLists.txt
etcwilde: eh, I think for this I'm fine keeping the explicit `if`.
Info on whether we should be linking dispatch into swift-testing seems like it should be coming from the Foundation configuration though, and then we wouldn't need it here at all. It would be part of the exposed interface from Foundation.

CMakeLists.txt
compnerd: Right, I get that. My reasoning is that we should eventually get dispatch working on WASI as well - we want the platforms to be at parity whenever possible.

Sources/Testing/CMakeLists.txt
```diff
@@ -96,6 +96,14 @@ add_library(Testing
   Traits/Trait.swift)
 target_link_libraries(Testing PRIVATE
   _TestingInternals)
+if(NOT APPLE)
```
stmontgomery, line 99: Curious why we need the `if(NOT APPLE)` guard for these `target_link_libraries`? I get why it's needed for `find_package`, but don't we want to link Foundation even on Darwin?

stmontgomery (approved):
Had one question but LGTM regardless

Sources/Testing/CMakeLists.txt
rintaro: In Darwin, Foundation is a framework, so `-lFoundation` doesn't work.
Yes it links Foundation, but that's done via auto-linking. I.e. `import Foundation` automatically sets required linker flags.

PR #697, opened 2024-09-17
Record issues generated within exit tests.

grynspan:
This PR introduces a "back channel" file handle to exit tests, allowing us to record issues that occur within exit test bodies. For example:

```swift
await #expect(exitsWith: .failure) {
  let context = try #require(possiblyMissingContext)
  ...
}
```

In this example, if the call to `try #require()` finds `nil`, it will record an issue, but that issue today will be lost because there's no mechanism to forward the issue back to the parent process hosting the exit test. This PR fixes that!

Issues are converted to JSON using the same schema we use for event handling, then written over a pipe back to the parent process where they are decoded. This decoding is lossy, so there will be further refinement needed here to try to preserve more information about the recorded issues. That said, "it's got good bones" right?

On Darwin, Linux, and FreeBSD, the pipe's write end is allowed to survive into the child process (i.e. no `FD_CLOEXEC`). On Windows, the equivalent is to tell `CreateProcessW()` to explicitly inherit a `HANDLE`. The identity of this file descriptor or handle is passed to the child process via environment variable. The child process then parses the file descriptor or handle out of the environment and converts it back to a `FileHandle` that is then connected to an instance of `Configuration` with an event handler set, and off we go.

Because we can now report these issues back to the parent process, I've removed the compile-time diagnostic in the `#expect(exitsWith:)` macro implementation that we emit when we see a nested `#expect()` or `#require()` call.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -66,6 +66,32 @@ public struct ExitTest: Sendable, ~Copyable {
 #endif
   }
 
+  /// Find a back channel file handle set up by the parent process.
+  ///
+  /// - Returns: A file handle open for writing to which events should be
+  ///   written, or `nil` if the file handle could not be resolved.
+  private static func _findBackChannel() -> FileHandle? {
+    guard let backChannelEnvironmentVariable = Environment.variable(named: "SWT_EXPERIMENTAL_BACKCHANNEL_FD") else {
+      return nil
+    }
+
+    var fd: CInt?
+#if SWT_TARGET_OS_APPLE || os(Linux) || os(FreeBSD)
+    fd = CInt(backChannelEnvironmentVariable).map(dup)
+#elseif os(Windows)
+    if let handle = UInt(backChannelEnvironmentVariable).flatMap(HANDLE.init(bitPattern:)) {
+      fd = _open_osfhandle(Int(bitPattern: handle), _O_WRONLY | _O_BINARY)
+    }
+#else
+#warning("Platform-specific implementation missing: back-channel pipe unavailable")
+#endif
+    guard let fd, fd >= 0 else {
+      return nil
+    }
+
+    return try? FileHandle(unsafePOSIXFileDescriptor: fd, mode: "wb")
+  }
+
   /// Call the exit test in the current process.
   ///
   /// This function invokes the closure originally passed to
@@ -76,8 +102,27 @@ public struct ExitTest: Sendable, ~Copyable {
   public consuming func callAsFunction() async -> Never {
     Self._disableCrashReporting()
 
+    // Set up the configuration for this process.
+    var configuration = Configuration()
+    if let backChannel = Self._findBackChannel() {
+      // Encode events as JSON and write them to the back channel file handle.
+      var eventHandler = ABIv0.Record.eventHandler(encodeAsJSONLines: true) { json in
+        try? backChannel.write(json)
+      }
+
+      // Only forward issue-recorded events. (If we start handling other kinds
+      // of event in the future, we can forward them too.)
+      eventHandler = { [eventHandler] event, eventContext in
+        if case .issueRecorded = event.kind {
+          eventHandler(event, eventContext)
+        }
+      }
+
+      configuration.eventHandler = eventHandler
+    }
+
     do {
-      try await body()
+      try await Configuration.withCurrent(configuration, perform: body)
     } catch {
       _errorInMain(error)
     }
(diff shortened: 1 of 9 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
swift test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test Linux

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -342,11 +387,109 @@ extension ExitTest {
...
+    let record = try JSON.decode(ABIv0.Record.self, from: recordJSON)
+
+    if case let .event(event) = record.kind, let issue = event.issue {
+      // Translate the issue back into a "real" issue and record it
+      // in the parent process. This translation is, of course, lossy
+      // due to the process boundary, but we make a best effort.
+      let comments: [Comment] = event.messages.compactMap { message in
+        message.symbol == .details ? Comment(rawValue: message.text) : nil
+      }
+      let issue = Issue(kind: .unconditional, comments: comments, sourceContext: .init(backtrace: nil, sourceLocation: issue.sourceLocation))
```
grynspan, line 491: We can flesh out this implementation going forward to handle more nuanced issues. The primary focus of this PR is just plumbing the data stream through.

Sources/Testing/ExitTests/SpawnProcess.swift
```diff
@@ -88,15 +127,130 @@ func spawnAndWait(
...
+      0,
+      swt_PROC_THREAD_ATTRIBUTE_HANDLE_LIST(),
+      inheritedHandlesBuffer.baseAddress!,
+      SIZE_T(MemoryLayout<HANDLE>.stride * inheritedHandlesBuffer.count),
+      nil,
+      nil
+    )
+
+    let commandLine = _escapeCommandLine(CollectionOfOne(executablePath) + arguments)
+    let environ = environment.map { "\($0.key)=\($0.value)"}.joined(separator: "\0") + "\0\0"
```
compnerd, line 160:
```suggestion
    let environ = environment.map { "\($0.key)=\($0.value)" }.joined(separator: "\0") + "\0\0"
```

compnerd (approved):

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -343,11 +404,115 @@ extension ExitTest {
...
+  ///   - backChannel: The file handle to read from. Reading continues until an
+  ///     error is encountered or the end of the file is reached.
+  private static func _processRecordsFromBackChannel(_ backChannel: borrowing FileHandle) {
+    let bytes: [UInt8]
+    do {
+      bytes = try backChannel.readToEnd()
+    } catch {
+      // NOTE: an error caught here indicates an I/O problem.
+      // TODO: should we record these issues as systemic instead?
+      Issue.record(error)
```
stmontgomery, line 473: I think we ought to refine `.system` a little bit and perhaps add an optional `error: any Error` associated value to it, so that a situation like this could be considered a system error but still propagate the underlying Error. Not for this PR tho

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -343,11 +404,115 @@ extension ExitTest {
...
+    }
+  }
+
+  /// Read lines from the given back channel file handle and process them as
+  /// event records.
+  ///
+  /// - Parameters:
+  ///   - backChannel: The file handle to read from. Reading continues until an
+  ///     error is encountered or the end of the file is reached.
+  private static func _processRecordsFromBackChannel(_ backChannel: borrowing FileHandle) {
```
stmontgomery, line 466:
```suggestion
  private static func _processRecords(from backChannel: borrowing FileHandle) {
```

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -240,16 +267,50 @@ extension ExitTest {
...
+      return result
+    }
+
+    // Set up the configuration for this process.
+    var configuration = Configuration()
+
+    // Encode events as JSON and write them to the back channel file handle.
+    // Only forward issue-recorded events. (If we start handling other kinds of
+    // events in the future, we can forward them too.)
+    let eventHandler = ABIv0.Record.eventHandler(encodeAsJSONLines: true) { json in
```
stmontgomery, line 301: I notice at least two places in this PR are hardcoding `ABIv0`: the sender and received sides of the pipe. I assume that, if/when we ever bump this ABI version, we'd want the parent and child process of any exit test to always use the latest version.

Given that, I wonder if we ought to have an (`internal`) typealias representing "latest ABI version", and reference that from these places rather than hardcode `ABIv0` ?

grynspan: Maybe. At least for this use case, we control both ends and can use anything we want for the communications channel. I don't think that's necessary for this PR but once we introduce ABIv1, it may be a good idea.

grynspan:
@swift-ci test

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -343,11 +404,115 @@ extension ExitTest {
...
+
+    if case let .event(event) = record.kind, let issue = event.issue {
+      // Translate the issue back into a "real" issue and record it
+      // in the parent process. This translation is, of course, lossy
+      // due to the process boundary, but we make a best effort.
+      let comments: [Comment] = event.messages.compactMap { message in
+        message.symbol == .details ? Comment(rawValue: message.text) : nil
+      }
+      let sourceContext = SourceContext(
+        backtrace: nil, // `issue._backtrace` will have the wrong address space.
```
stmontgomery, line 509: Can we file an issue to follow up here and eventually propagate and preserve _some_ of the symbolicated address of the backtrace? I know the specific numeric addresses will not be useful, but if the backtrace was symbolicated, it would be great to propagate those symbol names. But I assume that will require a bit more effort and isn't in scope here

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -343,11 +404,115 @@ extension ExitTest {
...
+        sourceLocation: issue.sourceLocation
       )
+      // TODO: improve fidelity of issue kind reporting (especially those without associated values)
+      var issueCopy = Issue(kind: .unconditional, comments: comments, sourceContext: sourceContext)
```
stmontgomery, line 513: You already mentioned it in the "TODO" comment thankfully, but just to echo: I think preserving the `kind` will be important!

stmontgomery (approved):
Nice!

Sources/Testing/ExitTests/ExitTest.swift
grynspan: It's something we can do once we have a proper symbolication mechanism rather than the POC that's in the repo now.

Sources/Testing/ExitTests/ExitTest.swift
grynspan: Yes! This is very much a case of putting the cart before the horse though.

PR #699, opened 2024-09-17
Disable pipes code on WASI.

grynspan:
WASI does not support pipes.

Resolves #698.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #700, opened 2024-09-17
CMake: Use `CMAKE_SYSTEM_NAME STREQUAL "WASI"` instead of `WASI`

kateinoigakukun:
The `WASI` variable is only available in the bleeding-edge CMake and not released versions yet
(https://gitlab.kitware.com/cmake/cmake/-/merge_requests/9659). So, we should use the old way for now.

I forgot to check CMake build in https://github.com/swiftlang/swift-testing/pull/699 🤦


stmontgomery (approved):

stmontgomery:
@swift-ci please test

grynspan (approved):

PR #703, opened 2024-09-17
Support: unwrap storage for parameter to `pipe`

compnerd:
Android declares pipe as:

```
int pipe(int __fds[_Nonnull 2]);
```

Ensure that we unwrap the storage before we pass along the parameter to the syscall. This repairs the build for Android.


compnerd:
@swift-ci please test

grynspan (approved):

PR #711, opened 2024-09-18
Add a document about porting to new platforms.

grynspan:
This PR adds a new Porting.md file including information for developers who want to port Swift Testing to new platforms. The document uses Classic Mac OS as a strawman/example platform. That was fun. Shoutout to @belkadan.

It also adds a `SWT_NO_TIMESPEC` conditional because I realized that there are platforms that don't have C11's `struct timespec`. (Because Classic Mac OS should obviously have C types standardized 2011.)

Formatted version [here](https://github.com/swiftlang/swift-testing/blob/jgrynspan/porting-doc/Documentation/Porting.md).

### Checklist:

### DIFF
--- a/Documentation/Porting.md
+++ b/Documentation/Porting.md
@@ -1,0 +1,239 @@
+# Porting to new platforms
+
+<!--
+This source file is part of the Swift.org open source project
+
+Copyright (c) 2024 Apple Inc. and the Swift project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+When the Swift toolchain is ported to a new platform, it is necessary to port
+Swift Testing as well. This document contains various information, trivia, and
+deep wisdoms about porting Swift Testing.
+
+> [!NOTE]
+> This document uses Classic Mac OS ("Classic") as an example target platform.
+> In this hypothetical scenario, we assume that the Swift compiler identifies
+> Classic with `os(Classic)` and that the C++ compiler identifies it with
+> `defined(macintosh)`. Other platforms would be identified differently.
+
+## Getting started
+
+Before you start the porting process, make sure you are very familiar with Swift
+and C++ as well as the C standard library and platform SDK for your target
+platform.
+
+Your first task when porting Swift Testing is ensuring that it builds.
+
+We've made an effort to ensure that as much of our code as possible is
+platform-agnostic. When building the toolchain for a new platform, you will
+hopefully find that Swift Testing builds out-of-the-box with few, if any,
+errors.
+
+> [!NOTE]
+> Swift Testing relies on the Swift [standard library](https://github.com/swiftlang/swift),
+> Swift macros (including the [swift-syntax](https://github.com/swiftlang/swift-syntax) package),
+> and [Foundation](https://github.com/apple/swift-foundation). These components
+> must build and (minimally) function before you will be able to successfully
+> build Swift Testing regardless of which platform you are porting to.
+
+### Swift or C++?
+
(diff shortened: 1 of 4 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

Documentation/Porting.md
```diff
@@ -0,0 +1,249 @@
...
+needs to support running tests for all Swift targets except, for the moment, the
+Swift standard library itself. Adding a dependency on another Swift component
+means that that component may be unable to link to Swift Testing. If you find
+yourself needing to link to a Swift component, please reach out to us in the
+Swift forums for advice.
+
+> [!WARNING]
+> Swift Testing has some dependencies on Foundation, specifically to support our
+> JSON event stream. Do not add new uses of Foundation without talking to us
+> first.
```
stmontgomery, line 228: When it comes to Foundation, I would recommend specifically noting that although we do _use_ Foundation in our implementation, we are careful not to _publicly import_ its module, since that would exacerbate existing challenges with that dependency.

Documentation/Porting.md
```diff
@@ -0,0 +1,249 @@
...
+ #else
+ #warning Platform-specific implementation missing: Runtime test discovery unavailable
+ template <typename SectionEnumerator>
+ static void enumerateTypeMetadataSections(const SectionEnumerator& body) {}
+ #endif
+```
+
+## C++ stub implementations
+
+Some symbols defined in C and C++ headers, especially "complex" macros, cannot
```
stmontgomery, line 159: I've sometimes seen these referred to as "non-trivial" macros

grynspan: Six of one, half-dozen of the other?

grynspan:
@swift-ci test

Documentation/Porting.md
```diff
@@ -0,0 +1,251 @@
...
+
+We've made an effort to ensure that as much of our code as possible is
+platform-agnostic. When building the toolchain for a new platform, you will
+hopefully find that Swift Testing builds out-of-the-box with few, if any,
+errors.
+
+> [!NOTE]
+> Swift Testing relies on the Swift [standard library](https://github.com/swiftlang/swift),
+> Swift macros (including the [swift-syntax](https://github.com/swiftlang/swift-syntax) package),
+> and [Foundation](https://github.com/apple/swift-foundation). These components
```
briancroom, line 39: All our Foundation uses are guarded by `#if canImport(Foundation)`. Should Foundation rather be described as an optional dependency here?

Documentation/Porting.md
```diff
@@ -0,0 +1,251 @@
...
+> must build and (minimally) function before you will be able to successfully
+> build Swift Testing regardless of which platform you are porting to.
+
+### Swift or C++?
```
briancroom, line 43: Perhaps reorder this section later in the document? Reading through it as a step-by-step, this paragraph feels a little bit out of place. What changes would I even be making, where the choice of programming language is applicable?

Documentation/Porting.md
grynspan: Eh, I don't think so. I would expect most people attempting a port are doing so to a system where Foundation will be properly available. (Embedded Swift is a different animal entirely.) The fact that we compile without it is a quirk of our build process, not really something we want to support.

Documentation/Porting.md
grynspan: If I'm reading along, I want to know "you should be using Swift" before I get into the technical details, I think? It would read oddly to be two sections in and see "oh by the way don't use C++ unless you absolutely have to." I've already started writing code by then!

I think.

I dunno. Where do you want to move it to?

grynspan: Going to merge this as-is, but we can revise.

PR #718, opened 2024-09-20
Distinguish when an error is part of a recorded issue in an exit test.

grynspan:
Hot on the heels of #697…

If an exit test records an issue of kind `.errorCaught()`, we currently treat it the same as any other issue and translate it to a flat `.unconditional` issue in the parent process:

```swift
await #expect(exitsWith: .failure) {
  Issue.record(NoSuchWeaselError()) // unconditional issue recorded
}
```

This PR distinguishes the `.errorCaught` issue kind and records an `.errorCaught` issue in the parent process. Fidelity of errors across a process boundary is weak, but we encode a minimal representation of the error that should be good enough for most uses. If you need `withKnownIssue()`, or `#expect(throws:)`, you can call them _inside_ the exit test.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedError.swift
+++ b/Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedError.swift
@@ -1,0 +1,67 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+extension ABIv0 {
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci test

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedError.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    init(encoding error: some Error, in eventContext: borrowing Event.Context) {
+      description = String(describingForTest: error)
+      domain = error._domain
+      code = error._code
+    }
+  }
+}
+
+// MARK: - Error, CustomNSError
```
stmontgomery, line 40: Did you mean to also implement `CustomNSError`?

grynspan: I looked into it, but it's in Foundation. I'll remove that.

grynspan:
@swift-ci test

PR #724, opened 2024-09-21
Avoid truncating the path to the test executable on Windows.

grynspan:
On Windows, we get the path to the (current) test executable by calling `GetModuleFileNameW()`. This function has odd behaviour when the input buffer is too short. Rather than returning `false` or `-1` or some such, as you might expect, it returns successfully but truncates the path (and null-terminates while doing so.)

The function _does_ set the last error in this case to `ERROR_INSUFFICIENT_BUFFER`, indicating we need a larger buffer. (The error behaviour on Windows XP is different, but we don't support Windows XP. If you decide to add support for Windows XP to Swift Testing, keep this in mind.)

So this PR checks for `ERROR_INSUFFICIENT_BUFFER` and loops with a larger buffer, similar to what we do on Darwin.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/SpawnProcess.swift
+++ b/Sources/Testing/ExitTests/SpawnProcess.swift
@@ -46,9 +46,9 @@ func spawnExecutable(
 ) throws -> ProcessID {
   // Darwin and Linux differ in their optionality for the posix_spawn types we
   // use, so use this typealias to paper over the differences.
-#if SWT_TARGET_OS_APPLE
+#if SWT_TARGET_OS_APPLE || os(FreeBSD)
   typealias P<T> = T?
-#elseif os(Linux) || os(FreeBSD)
+#elseif os(Linux)
   typealias P<T> = T
 #endif
 
--- a/Sources/Testing/Support/Additions/CommandLineAdditions.swift
+++ b/Sources/Testing/Support/Additions/CommandLineAdditions.swift
@@ -16,7 +16,11 @@ extension CommandLine {
     get throws {
 #if os(macOS)
       var result: String?
+#if DEBUG
+      var bufferCount = UInt32(1) // force looping
+#else
       var bufferCount = UInt32(PATH_MAX)
+#endif
       while result == nil {
         withUnsafeTemporaryAllocation(of: CChar.self, capacity: Int(bufferCount)) { buffer in
           // _NSGetExecutablePath returns 0 on success and -1 if bufferCount is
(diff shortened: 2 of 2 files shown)


grynspan:
@swift-ci test

Sources/Testing/ExitTests/SpawnProcess.swift
```diff
@@ -46,9 +46,9 @@ func spawnExecutable(
 ) throws -> ProcessID {
   // Darwin and Linux differ in their optionality for the posix_spawn types we
   // use, so use this typealias to paper over the differences.
-#if SWT_TARGET_OS_APPLE
```
grynspan, line 49: Drive-by fix: FreeBSD defines these POSIX types as pointers to opaque structs, so they get imported similarly to Darwin (where they're just raw pointers) rather than Linux (where they're actually structures.)

Sources/Testing/Support/Additions/CommandLineAdditions.swift
```diff
@@ -16,7 +16,7 @@ extension CommandLine {
     get throws {
 #if os(macOS)
       var result: String?
-      var bufferCount = UInt32(1024)
```
grynspan, line 19: Drive-by fix to use a named constant.

Sources/Testing/Support/Additions/CommandLineAdditions.swift
```diff
@@ -40,7 +40,7 @@ extension CommandLine {
       }
 #elseif os(FreeBSD)
       var mib = [CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, -1]
-      try mib.withUnsafeMutableBufferPointer { mib in
```
grynspan, line 43: Drive-by fix.

grynspan:
@swift-ci test

Sources/Testing/Support/Additions/CommandLineAdditions.swift
```diff
@@ -16,16 +16,15 @@ extension CommandLine {
     get throws {
 #if os(macOS)
       var result: String?
-      var bufferCount = UInt32(1024)
+      var bufferCount = UInt32(PATH_MAX)
       while result == nil {
-        result = withUnsafeTemporaryAllocation(of: CChar.self, capacity: Int(bufferCount)) { buffer in
```
grynspan, line 21: Let's consistently _set_ `result` in the buffer's scope rather than returning it. This saves us a `return nil` line and will be easier to translate if/when we get non-escaping buffers.

grynspan:
@swift-ci test

compnerd (approved):
There's a good number of ancillary drive by fixes. Generally, it's nicer to split those out into a separate change.

grynspan:
Eh, all but one are in the same function, just different platforms. I can live with it. 🙃

PR #726, opened 2024-09-23
Remove extraneous @escaping and @Sendable attributes from CustomExecutionTrait.execute(...) SPI

stmontgomery:
This refines the `CustomExecutionTrait` SPI protocol by removing two unnecessary attributes from the signature of its `execute(_:for:testCase:)` method requirement:

1. The `@escaping` attribute on the `function` parameter was needed originally to avoid a warning, but `function` should _not_ actually be considered escaping and part of the contract of this interface is that the implementor must call it before returning. The compiler no longer complains about this, so we can enforce this requirement by removing the unnecessary `@escaping` attribute now.
2. The outer `@Sendable` was necessary when this SPI was first written to avoid a Swift concurrency warning, since a reference to the function is taken elsewhere in the codebase, but this is no longer needed in recent Swift compilers.

### Checklist:

### DIFF
--- a/Sources/Testing/Traits/Trait.swift
+++ b/Sources/Testing/Traits/Trait.swift
@@ -110,5 +110,5 @@ public protocol CustomExecutionTrait: Trait {
   ///
   /// - Note: If a test function or test suite is skipped, this function does
   ///   not get invoked by the runner.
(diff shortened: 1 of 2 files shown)


stmontgomery:
@swift-ci please test

grynspan:
This is technically ABI-breaking, but it's also experimental SPI so it's fine.

grynspan (approved):

PR #728, opened 2024-09-24
Walk PE files' sections instead of using `swift_enumerateAllMetadataSections()`.

grynspan:
On Windows, the platform provides sufficient API for us to walk images at runtime looking for those that contain Swift metadata (similar to what we do on Darwin.) Therefore, we don't need to use the `swift_enumerateAllMetadataSections()` function from the Swift runtime on Windows.

This change also plumbs through an `imageAddress` argument to the discovery callback function. We're not currently using it, but it should be useful in the future for diagnostics (e.g. indicating that some data from a given image is malformed, or for passing to downstream testing tools that can make use of it.)

This change _also_ also plumbs the `stop` argument _down_ into the platform implementations so that we don't keep iterating images after the caller has told us to stop. In practice this has not been a correctness problem, but it does impact performance of type discovery.

### Checklist:

### DIFF
--- a/Sources/_TestingInternals/Discovery.cpp
+++ b/Sources/_TestingInternals/Discovery.cpp
@@ -10,15 +10,16 @@
 
 #include "Discovery.h"
 
+#include <algorithm>
+#include <array>
 #include <atomic>
 #include <cstring>
 #include <iterator>
 #include <type_traits>
 #include <vector>
+#include <optional>
 
-#if defined(SWT_NO_DYNAMIC_LINKING)
-#include <algorithm>
-#elif defined(__APPLE__)
+#if defined(__APPLE__) && !defined(SWT_NO_DYNAMIC_LINKING)
 #include <dispatch/dispatch.h>
 #include <mach-o/dyld.h>
 #include <mach-o/getsect.h>
@@ -199,7 +200,8 @@ extern "C" const char sectionEnd __asm("section$end$__TEXT$__swift5_types");
 template <typename SectionEnumerator>
 static void enumerateTypeMetadataSections(const SectionEnumerator& body) {
   auto size = std::distance(&sectionBegin, &sectionEnd);
-  body(&sectionBegin, size);
+  bool stop = false;
+  body(nullptr, &sectionBegin, size, &stop);
 }
 
 #elif defined(__APPLE__)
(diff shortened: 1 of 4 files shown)


grynspan:
@swift-ci test

Sources/_TestingInternals/Discovery.cpp
```diff
@@ -301,13 +303,96 @@ static void enumerateTypeMetadataSections(const SectionEnumerator& body) {
...
+  auto section = IMAGE_FIRST_SECTION(ntHeader);
+  for (size_t i = 0; i < sectionCount; i++, section += 1) {
+    if (section->VirtualAddress == 0) {
+      continue;
+    }
+
+    auto start = reinterpret_cast<const void *>(reinterpret_cast<uintptr_t>(dosHeader) + section->VirtualAddress);
+    size_t size = std::min(section->Misc.VirtualSize, section->SizeOfRawData);
+    if (start && size > 0) {
+      // FIXME: Handle longer names ("/%u") from string table
```
compnerd, line 359: This is not a FIXME. The extended meant is part of the COFF spec not the PE spec. That is, this will never occur.

Sources/_TestingInternals/Discovery.cpp
```diff
@@ -301,13 +303,96 @@ static void enumerateTypeMetadataSections(const SectionEnumerator& body) {
...
+/// - Parameters:
+///   - module: The module to inspect.
+///   - sectionName: The name of the section to look for. Long section names are
+///     not supported.
+///
+/// - Returns: A pointer to the start of the given section along with its size
+///   in bytes, or `std::nullopt` if the section could not be found. If the
+///   section was emitted by the Swift toolchain, be aware it will have leading
+///   and trailing bytes (`sizeof(uintptr_t)` each.)
+static std::optional<std::pair<const void *, size_t>> findSection(HMODULE module, const char *sectionName) {
```
compnerd, line 329: This is technically the segment as per macho. What macho considers sections are lost.

Sources/_TestingInternals/Discovery.cpp
grynspan: We _could_ add support for looking up sections (segments) with longer names. Microsoft documents that they're supported. However, Swift doesn't use any, so it's not important to our use case here.

Sources/_TestingInternals/Discovery.cpp
grynspan: The API and Microsoft's documentation refers to them as [sections](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers) either way.

Sources/_TestingInternals/Discovery.cpp
grynspan: See [here](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#section-table-section-headers).

grynspan:
@swift-ci test

stmontgomery (approved):

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/_TestingInternals/Discovery.cpp
compnerd: Yes, ELF and PE/COFF do not have segment/section, they only have sections.

stmontgomery (approved):

Sources/_TestingInternals/Discovery.cpp
```diff
@@ -366,12 +482,11 @@ static void enumerateTypeMetadataSections(const SectionEnumerator& body) {}
 #pragma mark -
 
 void swt_enumerateTypesWithNamesContaining(const char *nameSubstring, void *context, SWTTypeEnumerator body) {
-  enumerateTypeMetadataSections([=] (const void *section, size_t size) {
+  enumerateTypeMetadataSections([=] (const void *imageAddress, const void *section, size_t size, bool *stop) {
```
compnerd, line 485: Can we use `bool &` instead of `bool *` if this is meant to be non-null?

grynspan: This is just cribbed from what we do in Foundation for block-based enumeration.

Aside: it is not codified in any style guides anywhere, but I'm of the opinion that an `inout` or `out` argument should be a pointer so that at the call site you have to write `&` and within the callee you have to write `*` and it's clear to the caller that mutation may occur, whereas a mutable reference argument looks just like any other argument at the call site. We apply the same logic to `inout` in Swift, and we really wish this code were Swift. 🙃

Also we pass this parameter into Swift where there are no references, so it would be a bit confusing for it to be a reference here and a pointer there.

compnerd (approved):

PR #730, opened 2024-09-24
Don't use `swift_enumerateAllMetadataSections()` on WASI.

grynspan:
On WASI, there is only a single statically-linked binary, so we don't need to use `swift_enumerateAllMetadataSections()` and can just directly access the bounds of the metadata section(s) we care about.

### Checklist:

### DIFF
--- a/Sources/_TestingInternals/Discovery.cpp
+++ b/Sources/_TestingInternals/Discovery.cpp
@@ -422,8 +422,8 @@ static void enumerateTypeMetadataSections(const SectionEnumerator& body) {
 #elif defined(__wasi__)
 #pragma mark - WASI implementation (statically linked)
 
-extern "C" const char __start_swift5_type_metadata __attribute__((__visibility__("hidden"), __aligned__(1)));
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci test

Sources/_TestingInternals/Discovery.cpp
```diff
@@ -213,24 +213,8 @@ struct SWTTypeMetadataRecord {
   }
 };
 
-#if defined(SWT_NO_DYNAMIC_LINKING)
```
grynspan, line 216: Moved this bit down to clarify it's "Apple with static linking" rather than _all_ static-linked cases. The code isn't changed.

stmontgomery (approved):

grynspan:
Tested with wasmtime!

grynspan:
@swift-ci test

grynspan:
@swift-ci test

kateinoigakukun (approved):

PR #732, opened 2024-09-25
Reset default signal handlers on child processes before spawning them.

grynspan:
On platforms that use `posix_spawn()`, we should take care to reset the signal handlers of child processes to their defaults, right? Or should this be valid?

```swift
signal(SIGABRT, SIG_IGN)
await #expect(exitsWith: .success) {
  raise(SIGABRT) // does nothing because we set it to SIG_IGN in the parent
  exit(0)
}
```

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

Issue #36, opened 2023-09-28
API or mechanism to perform an action once before/after running all of the tests in a suite type


stmontgomery:
### Description

We should have an API or mechanism to perform an action once inside a suite type before or after running _all_ of that suite's tests.

Conceptually, this would be similar to XCTest's class-level `+setUp` and `+tearDown` APIs, or what some other testing libraries call "BeforeAll" and "AfterAll". However, for a more modern Swift solution, it may be better to use a closure- or scoped access- style API, since that interoperates better with Swift concurrency and allows using `@TaskLocal`.

rdar://103616215

grynspan:
I think the right basic shape for this is some protocol, i.e.:

```swift
protocol CustomWhateverable {
static func beforeAll(in suiteType: Self.Type) async throws
static func afterAll(in suiteType: Self.Type) async throws
}
```

Or something like that. `Runner` can then detect that a suite type conforms to this protocol and, if so, trivially call those functions at the appropriate times.

yonaskolb:
Are there any updates on this one?

grynspan:
Tracked internally as rdar://138099506.

PR #733, opened 2024-09-25
[SWT-NNNN] Introduce API allowing traits to customize test execution

stmontgomery:
This includes an API proposal and code changes to introduce new API for custom traits to customize test execution.

View the [API proposal](https://github.com/stmontgomery/swift-testing/blob/publicize-CustomExecutionTrait/Documentation/Proposals/NNNN-custom-test-execution-traits.md) for more details.

### Motivation:

One of the primary motivations for the trait system in Swift Testing, as [described in the vision document](https://github.com/swiftlang/swift-evolution/blob/main/visions/swift-testing.md#trait-extensibility), is to provide a way to customize the behavior of tests which have things in common. If all the tests in a given suite type need the same custom behavior, `init` and/or `deinit` (if applicable) can be used today. But if only _some_ of the tests in a suite need custom behavior, or tests across different levels of the suite hierarchy need it, traits would be a good place to encapsulate common logic since they can be applied granularly per-test or per-suite. This aspect of the vision for traits hasn't been realized yet, though: the `Trait` protocol does not offer a way for a trait to customize the execution of the tests or suites it's applied to.

Customizing a test's behavior typically means running code either before or after it runs, or both. Consolidating common set-up and tear-down logic allows each test function to be more succinct with less repetitive boilerplate so it can focus on what makes it unique.

### Checklist:

### DIFF
--- a/Documentation/Proposals/NNNN-custom-test-execution-traits.md
+++ b/Documentation/Proposals/NNNN-custom-test-execution-traits.md
@@ -4,7 +4,7 @@
 * Authors: [Stuart Montgomery](https://github.com/stmontgomery)
 * Status: **Awaiting review**
 * Implementation: [swiftlang/swift-testing#733](https://github.com/swiftlang/swift-testing/pull/733), [swiftlang/swift-testing#86](https://github.com/swiftlang/swift-testing/pull/86)
-* Review: ([pitch](https://forums.swift.org/...))
+* Review: ([pitch](https://forums.swift.org/t/pitch-custom-test-execution-traits/75055))
 
 ### Revision history
 
@@ -403,6 +403,10 @@ extension Trait where Self == MockAPICredentialsTrait {
 ## Source compatibility
 
 The proposed APIs are purely additive.
+
+This proposal will replace the existing `CustomExecutionTrait` SPI, and after
+further refactoring we anticipate it will obsolete the need for the
+`SPIAwareTrait` SPI as well.
 
 ## Integration with supporting tools
 
--- a/Documentation/Proposals/NNNN-custom-test-execution-traits.md
+++ b/Documentation/Proposals/NNNN-custom-test-execution-traits.md
@@ -422,6 +422,8 @@ supporting tools.
 
 ## Future directions
 
+### Access to suite type instances
+
 Some test authors have expressed interest in allowing custom traits to access
 the instance of a suite type for `@Test` instance methods, so the trait could
 inspect or mutate the instance. Currently, only instance-level members of a
@@ -429,6 +431,15 @@ suite type (including `init`, `deinit`, and the test function itself) can access
 `self`, so this would grant traits applied to an instance test method access to
 the instance as well. This is certainly interesting, but poses several technical
 challenges that puts it out of scope of this proposal.
+
+### Convenience trait for setting task locals
+
+Some reviewers of this proposal pointed out that the hypothetical usage example
+shown earlier involving setting a task local value while a test is executing
+will likely become a common use of these APIs. To streamline that pattern, it
+would be very natural to add a built-in trait type which facilitates this. I
+have prototyped this idea and plan to add it once this new trait functionality
+lands.
 
 ## Alternatives considered
 
--- a/Documentation/Proposals/NNNN-custom-test-execution-traits.md
+++ b/Documentation/Proposals/NNNN-custom-test-execution-traits.md
@@ -119,7 +119,7 @@ extension APICredentials {
 ```
 
 Binding task local values requires using the scoped access
-[`TaskLocal.withValue()`](https://developer.apple.com/documentation/swift/tasklocal/withvalue(_:operation:isolation:file:line:)
+[`TaskLocal.withValue()`](https://developer.apple.com/documentation/swift/tasklocal/withvalue(_:operation:isolation:file:line:))
 API though, and that would not be possible if `Trait` exposed separate methods
 like `setUp()` and `tearDown()`.
 
@@ -347,10 +347,10 @@ problem described above.
 
 ### Extend the `Trait` protocol
 
-The original SPI implementation of this feature included a protocol named
-`CustomExecutionTrait` which extended `Trait` and had roughly the same method
-requirement as the `CustomTestExecuting` protocol proposed above. This design
-worked, provided scoped access, and avoided the lengthy backtrace problem.
+The original, experimental implementation of this feature included a protocol
+named`CustomExecutionTrait` which extended `Trait` and had roughly the same
+method requirement as the `CustomTestExecuting` protocol proposed above. This
+design worked, provided scoped access, and avoided the lengthy backtrace problem.
 
 After evaluating the design and usage of this SPI though, it seemed unfortunate
 to structure it as a sub-protocol of `Trait` because it means that the full
--- a/Documentation/Proposals/NNNN-custom-test-execution-traits.md
+++ b/Documentation/Proposals/NNNN-custom-test-execution-traits.md
@@ -212,6 +212,14 @@ Below are the proposed interfaces:
 /// A protocol that allows customizing the execution of a test function (and
 /// each of its cases) or a test suite by performing custom code before or after
 /// it runs.
+///
+/// Types conforming to this protocol may be used in conjunction with a
+/// ``Trait``-conforming type by implementing the
+/// ``Trait/customTestExecutor-1dwpt`` property, allowing custom traits to
+/// customize the execution of tests. Consolidating common set-up and tear-down
+/// logic for tests which have similar needs allows each test function to be
+/// more succinct with less repetitive boilerplate so it can focus on what makes
+/// it unique.
 public protocol CustomTestExecuting: Sendable {
   /// Execute a function for the specified test and/or test case.
   ///
@@ -222,20 +230,29 @@ public protocol CustomTestExecuting: Sendable {
   ///     test function (including all cases if it is parameterized.)
   ///   - test: The test under which `function` is being performed.
   ///   - testCase: The test case, if any, under which `function` is being
-  ///     performed. This is `nil` when invoked on a suite.
+  ///     performed. When invoked on a suite, the value of this argument is
+  ///     `nil`.
   ///
   /// - Throws: Whatever is thrown by `function`, or an error preventing
-  ///   execution from running correctly.
+  ///   execution from running correctly. An error thrown from this method is
+  ///   recorded as an issue associated with `test`. If an error is thrown
+  ///   before `function` is called, the corresponding test will not run.
   ///
-  /// This function is called for each ``Trait`` on a test suite or test
-  /// function which has a non-`nil` value for ``Trait/customTestExecutor-1dwpt``.
-  /// It allows additional work to be performed before or after the test runs.
+  /// When the testing library is preparing to run a test, it finds all traits
+  /// applied to that test (including those inherited from containing suites)
+  /// and asks each for the value of its
+  /// ``Trait/customTestExecutor-1dwpt`` property. It then calls this method on
+  /// all non-`nil` instances, giving each an opportunity to perform
+  /// arbitrary work before or after invoking `function`.
   ///
-  /// This function is invoked once for the test its associated trait is applied
+  /// This method should either invoke `function` once before returning or throw
+  /// an error if it is unable to perform its custom logic successfully.
+  ///
+  /// This method is invoked once for the test its associated trait is applied
   /// to, and then once for each test case in that test, if applicable. If a
-  /// test is skipped, this function is not invoked for that test or its cases.
+  /// test is skipped, this method is not invoked for that test or its cases.
   ///
-  /// Issues recorded by this function are associated with `test`.
+  /// Issues recorded by this method are associated with `test`.
   func execute(_ function: @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws
 }
 
@@ -244,7 +261,10 @@ public protocol Trait: Sendable {
 
   /// The type of the custom test executor for this trait.
   ///
-  /// The default type is `Never`.
+  /// The default type is `Never`, which cannot be instantiated. This means the
+  /// value of the ``customTestExecutor-1dwpt`` property for all traits with the
+  /// default custom executor type is `nil`, meaning such traits will not
+  /// perform any custom execution for the tests they're applied to.
   associatedtype CustomTestExecutor: CustomTestExecuting = Never
 
   /// The custom test executor for this trait, if any.
@@ -254,20 +274,14 @@ public protocol Trait: Sendable {
   /// test execution. This is the most straightforward way to implement a trait
   /// which customizes the execution of tests.
   ///
-  /// However, if the value of this property is an instance of another type
+  /// If the value of this property is an instance of a different type
   /// conforming to ``CustomTestExecuting``, that instance will be used to
-  /// perform custom test execution instead.  Otherwise, the default value of
-  /// this property is `nil` (with the default type `Never?`), meaning that
-  /// custom test execution will not be performed for tests this trait is
-  /// applied to.
+  /// perform custom test execution instead.
+  ///
+  /// The default value of this property is `nil` (with the default type
+  /// `Never?`), meaning that instances of this type will not perform any custom
+  /// test execution for tests they are applied to.
   var customTestExecutor: CustomTestExecutor? { get }
-}
-
-extension Trait {
-  // ...
-
-  // The default implementation, which returns `nil`.
-  public var customTestExecutor: CustomTestExecutor? { get }
 }
 
 extension Trait where CustomTestExecutor == Self {
@@ -275,9 +289,12 @@ extension Trait where CustomTestExecutor == Self {
   public var customTestExecutor: CustomTestExecutor? { get }
 }
 
-extension Never: CustomTestExecuting {
-  public func execute(_ function: @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws
+extension Trait where CustomTestExecutor == Never {
+  // Returns `nil`.
+  public var customTestExecutor: CustomTestExecutor? { get }
 }
+
+extension Never: CustomTestExecuting {}
 ```
 
 Here is a complete example of the usage scenario described earlier, showcasing
--- a/Documentation/Proposals/NNNN-custom-test-execution-traits.md
+++ b/Documentation/Proposals/NNNN-custom-test-execution-traits.md
@@ -284,9 +284,9 @@ public protocol Trait: Sendable {
   var customTestExecutor: CustomTestExecutor? { get }
 }
 
-extension Trait where CustomTestExecutor == Self {
+extension Trait where Self: CustomTestExecuting {
   // Returns `self`.
-  public var customTestExecutor: CustomTestExecutor? { get }
+  public var customTestExecutor: Self? { get }
 }
 
 extension Trait where CustomTestExecutor == Never {
--- a/Documentation/Proposals/NNNN-custom-test-execution-traits.md
+++ b/Documentation/Proposals/NNNN-custom-test-execution-traits.md
@@ -11,6 +11,9 @@
 * **v1**: Initial pitch.
 * **v2**: Dropped 'Custom' prefix from the proposed API names (although kept the
   word in certain documentation passages where it clarified behavior).
+* **v3**: Changed the `Trait` requirement from a property to a method which
+  accepts the test and/or test case, and modify its default implementations such
+  that custom behavior is either performed per-suite or per-test case by default.
 
 ## Introduction
 
@@ -192,6 +195,42 @@ to mitigate these downsides it's important that there be some way to distinguish
 traits which customize test behavior. That way, the testing library can limit
 these scoped access calls to only the traits which require it.
 
+### Avoiding unnecessary (re-)execution
+
+Traits can be applied to either test functions or suites, and traits applied to
+suites can optionally support inheritance by implementing the `isRecursive`
+property of the `SuiteTrait` protocol. When a trait is directly applied to a
+test function, if the trait customizes the behavior of tests it's applied to, it
+should be given the opportunity to perform its custom behavior once for every
+invocation of that test function. In particular, if the test function is
+parameterized and runs multiple times, then the trait applied to it should
+perform its custom behavior once for every invocation. This should not be
+surprising to users, since it's consistent with the behavior of `init` and
+`deinit` for an instance `@Test` method.
+
+It may be useful for certain kinds of traits to perform custom logic once for
+_all_ the invocations of a parameterized test. Although this should be possible,
+we believe it shouldn't be the default since it could lead to work being
+repeated multiple times needlessly, or unintentional state sharing across tests,
+unless the trait is implemented carefully to avoid those problems.
+
+When a trait conforms to `SuiteTrait` and is applied to a suite, the question of
+when its custom behavior (if any) should be performed is less obvious. Some
+suite traits support inheritance and are recursively applied to all the test
+functions they contain (including transitively, via sub-suites). Other suite
+traits don't support inheritance, and only affect the specific suite they're
+applied to. (It's also worth noting that a sub-suite _can_ have the same
+non-recursive suite trait one of its ancestors has, as long as it's applied
+explicitly.)
+
+As a general rule of thumb, we believe most traits will either want to perform
+custom logic once for _all_ children or once for _each_ child, not both.
+Therefore, when it comes to suite traits, the default behavior should depend on
+whether it supports inheritance: a recursive suite trait should by default
+perform custom logic before each test, and a non-recursive one per-suite. But
+the APIs should be flexible enough to support both, for advanced traits which
+need it.
+
 ## Detailed design
 
 I propose the following new APIs:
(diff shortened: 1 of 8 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test Windows

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,376 @@
...
+concurrency-safe and still allow tests accessing this state to run in parallel:
+
+```swift
+extension APICredentials {
+  @TaskLocal static var current: Self?
+}
+```
+
+Binding task local values requires using the scoped access
+[`TaskLocal.withValue()`](https://developer.apple.com/documentation/swift/tasklocal/withvalue(_:operation:isolation:file:line:)
```
grynspan, line 122:
```suggestion
[`TaskLocal.withValue()`](https://developer.apple.com/documentation/swift/tasklocal/withvalue(_:operation:isolation:file:line:))
```

Markdown angry, missing paren.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,376 @@
...
+In practice, most traits probably will _not_ need to customize test behavior, so
+to mitigate these downsides it's important that there be some way to distinguish
+traits which customize test behavior. That way, the testing library can limit
+these scoped access calls to only the traits which require it.
+
+## Detailed design
+
+I propose the following new APIs:
+
+- A new protocol `CustomTestExecuting` with a single required `execute(...)`
```
grynspan, line 193: Gerunds for protocol names are common in Objective-C, but in Swift we usually prefer an adjective form, e.g. (bikeshedding here!) `CustomTestExecutable` or `CustomExecutable` or `CustomTestSetuppableAndTeardownable`.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,376 @@
...
+  method. This will be called to run a test, and allows the conforming type to
+  perform custom logic before or after.
+- A new property `customTestExecutor` on the `Trait` protocol whose type is an
```
grynspan, line 196: It occurs to me that types that conform to this protocol aren't actually responsible for executing the test, they're just adding before/after cladding to it. We might want to mull over the name a bit more. Thesaurus?

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,376 @@
...
+```swift
+@Test(.mockAPICredentials)
+func example() {
+  // ...validate API usage, referencing `APICredentials.current`...
+}
+
+struct MockAPICredentialsTrait: TestTrait, CustomTestExecuting {
+  func execute(_ function: @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws {
+    let mockCredentials = APICredentials(apiKey: "...")
+    try await APICredentials.$current.withValue(mockCredentials) {
```
grynspan, line 295: We think task locals are going to be a particularly common use case, right? Perhaps we should build out a trait type that sets a task local:

```swift
@Test(.withValue(123, for: $foo)) func f() { ... }
```

Or even if we don't make it a trait, can we make it easier to write a trait that sets a task local with minimal boilerplate?

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,376 @@
...
+
+The proposed `execute(...)` method could be added as a requirement of the
+`Trait` protocol instead of being part of a separate `CustomTestExecuting`
+protocol, and it could have a default implementation which directly invokes the
+passed-in closure. But this approach would suffer from the lengthy backtrace
+problem described above.
+
+### Extend the `Trait` protocol
+
+The original SPI implementation of this feature included a protocol named
```
grynspan, line 350:
```suggestion
The original experimental implementation of this feature included a protocol named
```

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: Agree that would be a very useful, built-in trait to have. Let's consider that separately, since it would ultimately leverage the API proposed here

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: Definitely open to these alternate suggestions, thanks. But I'll note: as I was considering names myself, I found this in the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/#protocols-describing-capability-should-use-suffixes):

> - Protocols that describe a capability should be named using the suffixes `able`, `ible`, or `ing` (e.g. `Equatable`, `ProgressReporting`).

Based on that, I think an `ing` suffix would be acceptable.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: Well they do work "surrounding" the test, but _also_ execute it.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
grynspan: Oh it's not _wrong_, just uncommon. So it's worth thinking about something more idiomatic if we can come up with something.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
grynspan: Doesn't need to be part of this proposal, but I think it does _affect_ this proposal. Might be worth coming up with an example of when you'd use this stuff other than to set a task local.

Sources/Testing/Testing.docc/Traits/Trait.md
```diff
@@ -39,8 +39,15 @@ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
...
 - ``Trait/comments``
 
 ### Preparing internal state
 
 - ``Trait/prepare(for:)-3s3zo``
+
+### Customizing the execution of tests
+
+- ``CustomTestExecuting``
+- ``Trait/customTestExecutor-1dwpt``
```
grynspan, line 52: Is this showing up as a separate symbol in DocC? Is the DocC bundle building cleanly with no warnings here?

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
   ///
   /// By default, the value of this property is an empty array.
   var comments: [Comment] { get }
+
+  /// The type of the custom test executor for this trait.
+  ///
+  /// The default type is `Never`.
```
grynspan, line 47: Please elaborate _why_ the default type is `Never` and what it means.

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
...
+  associatedtype CustomTestExecutor: CustomTestExecuting = Never
+
+  /// The custom test executor for this trait, if any.
+  ///
+  /// If this trait's type conforms to ``CustomTestExecuting``, the default
+  /// value of this property is `self` and this trait will be used to customize
+  /// test execution. This is the most straightforward way to implement a trait
+  /// which customizes the execution of tests.
+  ///
+  /// However, if the value of this property is an instance of another type
```
grynspan, line 57:
```suggestion
  /// If the value of this property is an instance of a different type
  /// that conforms to [...]
```

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: One thing I slightly dislike about the `able` suffixes is that it makes it seem as if the work is optional. "This type is _able_ to execute the test… but will it? Who knows!?". But if you define a trait this way, it **will** execute tests it's applied to, and to me the `ing` conveys that a little more clearly. "This type is responsible for _executing_ the test."

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
...
+  /// conforming to ``CustomTestExecuting``, that instance will be used to
+  /// perform custom test execution instead.  Otherwise, the default value of
```
grynspan, line 59:
```suggestion
  /// perform custom test execution instead.
  ///
  /// The default value of [...]
```

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
...
+  /// this property is `nil` (with the default type `Never?`), meaning that
+  /// custom test execution will not be performed for tests this trait is
```
grynspan, line 61: This implies that the default turns off _all_ custom test execution for the test, which of course isn't correct. How about:

```suggestion
  /// instances of this type will not perform any custom test execution for tests they are applied to.
```

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
...
+  /// applied to.
+  var customTestExecutor: CustomTestExecutor? { get }
+}
+
+/// A protocol that allows customizing the execution of a test function (and
```
grynspan, line 66: This comment becomes high-level documentation, so it would be good to add some explanation of what this means, what it's for… basically distill the pitch down into a couple of paragraphs.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: Sure. Briefly: not all work before/after a test is simply mutating state in that process. It may involve writing to the filesystem, or interacting with other system-wide resources. (Clearly, parallelism would need to be considered in such scenarios.)

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
...
+  /// Execute a function for the specified test and/or test case.
+  ///
+  /// - Parameters:
+  ///   - function: The function to perform. If `test` represents a test suite,
+  ///     this function encapsulates running all the tests in that suite. If
+  ///     `test` represents a test function, this function is the body of that
+  ///     test function (including all cases if it is parameterized.)
+  ///   - test: The test under which `function` is being performed.
+  ///   - testCase: The test case, if any, under which `function` is being
+  ///     performed. This is `nil` when invoked on a suite.
```
grynspan, line 79:
```suggestion
  ///     performed. When invoked on a suite, the value of this argument is `nil`.
```

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: Also: Users can still use this facility to mutate global or static state, as long as it complies with Swift concurrency rules

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
...
+  ///
+  /// - Throws: Whatever is thrown by `function`, or an error preventing
+  ///   execution from running correctly.
```
grynspan, line 82:
```suggestion
  ///   execution from running correctly. An error thrown from this function is recorded as an issue associated with `test`. If an error is thrown before `function` is called, the corresponding test will not run.
```

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
...
+  ///
+  /// This function is called for each ``Trait`` on a test suite or test
```
grynspan, line 84: Can we rewrite this without the passive voice? Something like "When the testing library is preparing to run a test, it checks all traits associated with that test and [...]"

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,68 @@ public protocol Trait: Sendable {
...
+
+extension Trait where CustomTestExecutor == Self {
+  public var customTestExecutor: CustomTestExecutor? {
+    self
+  }
+}
+
+extension Never: CustomTestExecuting {
+  public func execute(_ function: @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws {
+    fatalError("Unreachable codepath: Never cannot be instantiated.")
```
grynspan, line 104: Do we even need this `fatalError()` given this is unreachable and the function doesn't need to return anything? 🤔

Sources/Testing/Testing.docc/Traits/Trait.md
stmontgomery: It's because it has multiple overloads in extensions

Sources/Testing/Traits/Trait.swift
```diff
@@ -70,45 +132,14 @@ extension Trait {
   public var comments: [Comment] {
     []
   }
+
+  public var customTestExecutor: CustomTestExecutor? {
```
grynspan, line 136: We don't want the default implementation to return `nil` if a type _other than `Never` is used. We should limit it to:

```swift
extension Trait where CustomTestExecutor == Never {
  ...
}
```

Sources/Testing/Testing.docc/Traits/Trait.md
stmontgomery: Yes, there are no (new) DocC diagnostics

grynspan: Ah, okay.

Sources/Testing/Traits/Trait.swift
stmontgomery: Good things to note

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,376 @@
...
+  /// However, if the value of this property is an instance of another type
+  /// conforming to ``CustomTestExecuting``, that instance will be used to
+  /// perform custom test execution instead.  Otherwise, the default value of
+  /// this property is `nil` (with the default type `Never?`), meaning that
+  /// custom test execution will not be performed for tests this trait is
+  /// applied to.
+  var customTestExecutor: CustomTestExecutor? { get }
+}
+
+extension Trait {
```
grynspan, line 266: This extension is no longer part of the proposal.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,376 @@
...
+  // ...
+
+  // The default implementation, which returns `nil`.
+  public var customTestExecutor: CustomTestExecutor? { get }
+}
+
+extension Trait where CustomTestExecutor == Self {
```
grynspan, line 273: I would consider removing this extension (and not have the one for `where CustomTestExecutor == Never` either) from the .md file. Their existence should be inferred from the documentation.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,376 @@
...
+  // Returns `self`.
+  public var customTestExecutor: CustomTestExecutor? { get }
+}
+
+extension Never: CustomTestExecuting {
+  public func execute(_ function: @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws
```
grynspan, line 279: The extension here can be empty; mentioning that it contains the protocol requirements is redundant.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: I think including the extensions may help clarify the mechanism by which those default conformances work, so I'll leave them in but adjust them to update the latest code changes

stmontgomery:
@swift-ci please test

Documentation/Proposals/NNNN-custom-test-execution-traits.md
grynspan: We have plenty of counter-examples too where a noun is used instead: `Collection`, `Sequence`, `Actor`… maybe it really just is `CustomTestExecutor`.

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test Windows

Documentation/Proposals/NNNN-custom-test-execution-traits.md
grynspan: Can we make sure to add this example to the discussion?

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: > We have plenty of counter-examples too where a noun is used instead: `Collection`, `Sequence`, `Actor`… maybe it really just is `CustomTestExecutor`.

The API Design Guidelines [address](https://www.swift.org/documentation/api-design-guidelines/#protocols-describing-what-is-should-read-as-nouns) that other common naming convention for protocols in the bullet above the one I quoted:

> * **Protocols that describe _what something is_ should read as nouns** (e.g. `Collection`).

So I think this adheres to the recommendations well: the `Trait` protocol best describes what the type _is_, and the `CustomTestExecuting` protocol describes one optional _capability_ a type (usually, a trait) can have.

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -6,6 +6,12 @@
 * Implementation: [swiftlang/swift-testing#733](https://github.com/swiftlang/swift-testing/pull/733), [swiftlang/swift-testing#86](https://github.com/swiftlang/swift-testing/pull/86)
 * Review: ([pitch](https://forums.swift.org/...))
```
grynspan, line 7: Can you update this link? Thanks!

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,397 @@
...
+extension Trait where Self == MockAPICredentialsTrait {
+  static var mockAPICredentials: Self {
+    Self()
+  }
+}
+```
+
+## Source compatibility
+
+The proposed APIs are purely additive.
```
grynspan, line 331: Might be worth a note here that this change will obsolete the `SPIAwareTrait` and `CustomExecutionTrait` experimental SPI.

stmontgomery: Noted!

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,125 @@ public protocol Trait: Sendable {
...
+  ///
+  /// This method is invoked once for the test its associated trait is applied
+  /// to, and then once for each test case in that test, if applicable. If a
+  /// test is skipped, this method is not invoked for that test or its cases.
+  ///
+  /// Issues recorded by this method are associated with `test`.
+  func execute(_ function: @Sendable () async throws -> Void, for test: Test, testCase: Test.Case?) async throws
+}
+
+extension Trait where Self: TestExecuting {
```
grynspan, line 145: I think you want this to be `TestTrait`?

Sources/Testing/Traits/Trait.swift
```diff
@@ -72,43 +191,14 @@ extension Trait {
   }
 }
 
+extension Trait where TestExecutor == Never {
```
grynspan, line 194: This one should stay `Trait` so it applies to both test- and suite-trait types.

Sources/Testing/Traits/Trait.swift
stmontgomery: The compiler doesn't allow that, because then it's ambiguous for types which conform to both `TestTrait` and `SuiteTrait`

Sources/Testing/Traits/Trait.swift
```diff
@@ -41,6 +41,125 @@ public protocol Trait: Sendable {
...
+  /// applied to that test (including those inherited from containing suites)
+  /// and asks each for its test executor (if any) by calling
+  /// ``Trait/executor(for:testCase:)-26qgm``. It then calls this method
+  /// on all non-`nil` instances, giving each an opportunity to perform
+  /// arbitrary work before or after invoking `function`.
+  ///
+  /// This method should either invoke `function` once before returning or throw
+  /// an error if it is unable to perform its custom logic successfully.
+  ///
+  /// This method is invoked once for the test its associated trait is applied
```
grynspan, line 137: I think this paragraph is now out of date? At a minimum, it should probably refer to the other function.

Sources/Testing/Traits/Trait.swift
stmontgomery: The way I have it now, the extension on `Trait` acts as the "default" implementation, and the `SuiteTrait` refines it conditionally.

grynspan: Perhaps it _should_ be ambiguous for types that conform to both, or we should provide an additional default for them that executes once per suite or test?

stmontgomery: I don't think we want to break the current ability to have a trait type conform to both `SuiteTrait` and `TestTrait`. Do you think something is problematic about the current approach?

jakepetroules:
@stmontgomery Per our conversation offline you mentioned that order of execution of custom execution traits is well-defined to be outer-to-inner, left-to-right -- it would be great if we could explicitly add a note about that in the documentation so that folks are aware of the expected behavior!

stmontgomery:
> @stmontgomery Per our conversation offline you mentioned that order of execution of custom execution traits is well-defined to be outer-to-inner, left-to-right -- it would be great if we could explicitly add a note about that in the documentation so that folks are aware of the expected behavior!

I've added language specifying this ordering to the updated proposal and PR

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: Where, specifically, do you recommend we place this example code snippet?

stmontgomery:
@swift-ci please test

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,499 @@
...
+
+## Future directions
+
+Some test authors have expressed interest in allowing custom traits to access
+the instance of a suite type for `@Test` instance methods, so the trait could
+inspect or mutate the instance. Currently, only instance-level members of a
+suite type (including `init`, `deinit`, and the test function itself) can access
+`self`, so this would grant traits applied to an instance test method access to
+the instance as well. This is certainly interesting, but poses several technical
+challenges that puts it out of scope of this proposal.
```
grynspan, line 431: Mind adding a quick note about a task-local trait here? Danke.

Sources/Testing/Traits/Trait.swift
```diff
@@ -72,43 +192,14 @@ extension Trait {
   }
 }
 
+extension Trait where TestScopeProvider == Never {
```
grynspan, line 195: These extensions will show up (at least minimally) in DocC, so we might want to consider formally documenting them with full markup since they have some interesting behaviours.

(Not a blocker for the PR, just noting.)

Documentation/Proposals/NNNN-custom-test-execution-traits.md
```diff
@@ -0,0 +1,499 @@
...
+  func scopeProvider(for test: Test, testCase: Test.Case?) -> TestScopeProvider?
+}
+
+extension Trait where Self: TestScoping {
+  // Returns `nil` if `testCase` is `nil`, else `self`.
+  public func scopeProvider(for test: Test, testCase: Test.Case?) -> Self?
+}
+
+extension SuiteTrait where Self: TestScoping {
+  // If `test` is a suite, returns `nil` if `isRecursive` is `true`, else `self`.
```
grynspan, line 368: The comments here literally describe the current implementations but don't really explain _why_ they are what they are. _Why_ does this function return `nil` if `isRecursive` is `true`? etc.

stmontgomery: I'll add DocC, per your other comment. As far as this proposal document is concerned though, I feel like that is explained adequately in the documentation of one of the earlier decls so this is just a very succinct comment about what the implementation does.

briancroom (approved):

Sources/Testing/Traits/Trait.swift
stmontgomery: Added documentation for these implementations in extensions

Documentation/Proposals/NNNN-custom-test-execution-traits.md
stmontgomery: Added this as a paragraph within Future Directions

stmontgomery:
@swift-ci please test

stmontgomery:
Resolves #36

Mcrich23:
@stmontgomery the proposal document no longer exists, but I would love to use this! How do I?

grynspan:
It moved [here](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0007-test-scoping-traits.md).

PR #736, opened 2024-09-27
Inherit isolation in `#expect(exitsWith:)`.

grynspan:
This PR ensures the body of _the implementation of_ `#expect(exitsWith:)` inherits isolation from the caller. The actual body closure passed to the macro cannot inherit isolation as it runs in a separate process, but the glue code we emit needs to inherit so that if the caller is actor-isolated, the code compiles cleanly. Without this change, the following test fails to compile:

```swift
@MainActor @Test func f() async {
  await #expect(exitsWith: .failure) { /* ... */ }
  // 🛑 ^ ^ ^ sending main actor-isolated value of type '() -> [Comment]' with later accesses to nonisolated context risks causing data races
}
```

### Checklist:

### DIFF
--- a/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
+++ b/Sources/Testing/Expectations/ExpectationChecking+Macro.swift
@@ -1142,15 +1142,15 @@ public func __checkClosureCall<R>(
 @_spi(Experimental)
 public func __checkClosureCall(
   exitsWith expectedExitCondition: ExitCondition,
-  performing body: @convention(thin) () async throws -> Void,
+  performing body: @convention(thin) () -> Void,
   expression: __Expression,
   comments: @autoclosure () -> [Comment],
   isRequired: Bool,
+  isolation: isolated (any Actor)? = #isolation,
   sourceLocation: SourceLocation
 ) async -> Result<Void, any Error> {
   await callExitTest(
     exitsWith: expectedExitCondition,
-    performing: { try await body() },
     expression: expression,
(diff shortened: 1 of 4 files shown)


Sources/Testing/Expectations/ExpectationChecking+Macro.swift
```diff
@@ -1142,15 +1142,15 @@ public func __checkClosureCall<R>(
 @_spi(Experimental)
 public func __checkClosureCall(
   exitsWith expectedExitCondition: ExitCondition,
-  performing body: @convention(thin) () async throws -> Void,
+  performing body: @convention(thin) () -> Void,
```
grynspan, line 1145: This makes the type checker a bit happier as it gets motion sick dealing with `@convention(thin)`.

grynspan:
@swift-ci test

Sources/Testing/Expectations/ExpectationChecking+Macro.swift
stmontgomery: Does this restrict the body closure of an exit test to not be async or throws?

stmontgomery (approved):

Sources/Testing/Expectations/ExpectationChecking+Macro.swift
grynspan: No, just the synthesized wrapper around the emitted type. This closure isn't actually ever called.

PR #737, opened 2024-09-27
Set `RLIMIT_CORE=1` instead of `=0` on Linux.

grynspan:
The Linux kernel [special-cases](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/coredump.c#n610) `RLIMIT_CORE=1` to mean "no, really, don't write a core file even if it'd be to a pipe." Which sounds like a niche use case but Ubuntu 24 by default enables Apport for crash reporting and Apport works by setting the core pattern to write to a pipe.

This resolves a performance issue on Linux where exit tests would always take almost exactly one second instead of a few microseconds (unless you explicitly turned off Apport.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -51,10 +51,18 @@ public struct ExitTest: Sendable, ~Copyable {
       EXCEPTION_DEFAULT,
       THREAD_STATE_NONE
     )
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci test

stmontgomery (approved):

grynspan:
@swift-ci test Windows

grynspan:
@swift-ci test Windows

PR #739, opened 2024-09-29
Fix macro expansion failure for a non-copyable suite type with a `test` method.

grynspan:
If a non-copyable suite type has a test function whose name starts with `test`, we emit a call to `__invokeXCTestCaseMethod()` that's a no-op (but we can't tell during macro expansion if it will be.) This function needs to be updated to accept a non-copyable generic type.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

PR #747, opened 2024-10-04
Simplify the configuration option for synchronous test isolation.

grynspan:
This PR replaces `Configuration.isMainActorIsolationEnforced` with a new property `Configuration.defaultIsolationContext` which can be set to any actor, not just the main actor. This property is then used as the isolation context for synchronous test functions that are not otherwise actor-isolated.

The typical use case for this property would be to set it to `MainActor.shared` so that synchronous test functions run on the main actor, but a testing tool might want to instead run them all on another actor or serial executor for performance (or other) reasons.

This change isn't just speculative: it allows us to simplify the macro-generated thunk for synchronous test functions so that it does not need to emit the body of the thunk twice (once for the main actor, once for non-isolated.) This change wasn't possible before `#isolation` was introduced, but we'd already added `Configuration.isMainActorIsolationEnforced` when that happened so the timing didn't line up.

By rewriting this part of the macro expansion, we're able to toss a bunch of support code that isn't needed anymore. Yay!

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -105,14 +105,13 @@ public struct Configuration: Sendable {
   /// By default, the value of this property allows for a single iteration.
   public var repetitionPolicy: RepetitionPolicy = .once
 
-  // MARK: - Main actor isolation
+  // MARK: - Isolation context for synchronous tests
 
-#if !SWT_NO_GLOBAL_ACTORS
-  /// Whether or not synchronous test functions need to run on the main actor.
+  /// The isolation context to use for synchronous test functions.
   ///
-  /// This property is available on platforms where UI testing is implemented.
-  public var isMainActorIsolationEnforced = false
-#endif
+  /// If the value of this property is `nil`, synchronous test functions run in
+  /// an unspecified isolation context.
+  public var defaultIsolationContext: (any Actor)? = nil
 
   // MARK: - Time limits
 
@@ -232,4 +231,24 @@ public struct Configuration: Sendable {
 
   /// The test case filter to which test cases should be filtered when run.
   public var testCaseFilter: TestCaseFilter = { _, _ in true }
+}
+
+// MARK: - Deprecated
+
+extension Configuration {
+#if !SWT_NO_GLOBAL_ACTORS
+  @available(*, deprecated, message: "Set defaultIsolationContext instead.")
+  public var isMainActorIsolationEnforced: Bool {
+    get {
+      defaultIsolationContext === MainActor.shared
+    }
+    set {
+      if newValue {
+        defaultIsolationContext = MainActor.shared
+      } else {
+        defaultIsolationContext = nil
+      }
+    }
+  }
+#endif
 }
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -111,7 +111,7 @@ public struct Configuration: Sendable {
   ///
   /// If the value of this property is `nil`, synchronous test functions run in
   /// an unspecified isolation context.
-  public var defaultIsolationContext: (any Actor)? = nil
+  public var defaultSynchronousIsolationContext: (any Actor)? = nil
 
   // MARK: - Time limits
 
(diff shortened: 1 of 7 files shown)


grynspan:
@swift-ci test

briancroom (approved):

grynspan:
I probably need to make the closure sendable. That sounds like a problem for Monday Jonathan.

grynspan:
Saturday Jonathan swooped in and rethinked/rethunked it.

grynspan:
@swift-ci test

grynspan:
The line diff has shrunk, but mostly because I've added a bunch of comments trying to explain these shenanigans.

Sources/Testing/Running/Configuration.swift
```diff
@@ -105,14 +105,13 @@ public struct Configuration: Sendable {
...
-#if !SWT_NO_GLOBAL_ACTORS
-  /// Whether or not synchronous test functions need to run on the main actor.
+  /// The isolation context to use for synchronous test functions.
   ///
-  /// This property is available on platforms where UI testing is implemented.
-  public var isMainActorIsolationEnforced = false
-#endif
+  /// If the value of this property is `nil`, synchronous test functions run in
+  /// an unspecified isolation context.
+  public var defaultIsolationContext: (any Actor)? = nil
```
stmontgomery, line 114: I'd find it helpful if the name of this property indicated that it only applies to _synchronous_ functions, not async ones. (Because asynchronous test functions could, hypothetically, honor this too, right? Not that we want them to.) Maybe drop the word "context" to make space for the word "synchronous"?
```suggestion
  public var defaultSynchronousIsolation: (any Actor)? = nil
```

Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
```diff
@@ -325,8 +325,6 @@ struct TestDeclarationMacroTests {
       ("@Test @_unavailableFromAsync @MainActor func f() {}", nil, "MainActor.run"),
       ("@Test @available(*, noasync) func f() {}", nil, "__requiringTry"),
       ("@Test @_unavailableFromAsync func f() {}", nil, "__requiringTry"),
-      ("@Test(arguments: []) func f(i: borrowing Int) {}", nil, "copy"),
-      ("@Test(arguments: []) func f(_ i: borrowing Int) {}", nil, "copy"),
```
stmontgomery, line 329: Are these removals because we're no longer supporting `borrowing` parameters?

Sources/Testing/Running/Configuration.swift
grynspan: The naming for this concept is a bit vague on the Swift side, I think, but in general it's an "isolation context", so we could write `defaultSynchronousIsolationContext`?

Sources/Testing/Test+Macro.swift
```diff
@@ -504,58 +504,13 @@ extension Test {
...
-/// - Warning: This function is used to implement the `@Test` macro. Do not call
+/// - Warning: This property is used to implement the `@Test` macro. Do not call
 ///   it directly.
-@inlinable public func __ifMainActorIsolationEnforced<R>(
-  _: @Sendable () async throws -> R,
-  else body: @Sendable () async throws -> R
-) async throws -> R where R: Sendable {
-  try await body()
+public var __defaultIsolationContext: (any Actor)? {
+  Configuration.current?.defaultIsolationContext ?? #isolation
```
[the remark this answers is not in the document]
grynspan: So we can potentially avoid an extra hop onto the shared thread pool if the configuration doesn't specify a context. (In practice it's always going to be `nil` right now because the caller is nonisolated but that could change in the future.)

Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
grynspan: No, it's because we no longer need to copy the arguments (which is what these cases were looking for.)

stmontgomery (approved):

grynspan:
@swift-ci test

grynspan:
@swift-ci test

silvgabriel:
@grynspan 
sorry for commenting on this thread but how can I set a value for the `defaultSynchronousIsolationContext` property when using the Testing framework??

PR #750, opened 2024-10-07
Simplify the thunk names we generate for test functions.

grynspan:
Parameterized test functions with the same name but different argument types currently get the same generated names from swift-syntax, so we do some additional decorating to ensure uniqueness. This results in very long symbol names that `swift-demangle` has trouble with.

This PR changes the decorating formula. Previously, we would include a copy of the test function's signature (stripped of whitespace, Unicode, and non-identifier-friendly ASCII) and, if the identifier contained non-ASCII characters, the CRC32 of the identifier for further uniqueness. This change drops the copy of the identifier name and always includes the CRC32. Collisions are only possible for fully-qualified test function names that are _identical_, and I naïvely judge the risk of those collisions to be sufficiently low that we can make this change.

For `ZipTests.allElementsEqual😀(i:j:)` we currently generate a thunk named:

```
$s12TestingTests03ZipB0V0022allElementsEqual_oxFJo4TestfMp_43funcallElementsEqual__i_Int_j_Int__5bcb7b35fMu_
```

This change would change it to:

```
$s12TestingTests03ZipB0V0022allElementsEqual_oxFJo4TestfMp_9Z5bcb7b35fMu_
```

Which demangles to:

```
$s12TestingTests03ZipB0V0022allElementsEqual_oxFJo4TestfMp_9Z5bcb7b35fMu_ ---> unique name #1 of Z5bcb7b35 in peer macro @Test expansion #1 of allElementsEqual😀 in TestingTests.ZipTests
```

The benefit of the change is that the generated names are shorter and easier to read when expanding a macro, and play better with the demangler. There may also be some benefit on Windows where the linker has a 65KB symbol name cap, although we're not exporting these symbols so probably not.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/Additions/MacroExpansionContextAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/MacroExpansionContextAdditions.swift
@@ -62,44 +62,18 @@ extension MacroExpansionContext {
       .tokens(viewMode: .fixedUp)
       .map(\.textWithoutBackticks)
       .joined()
+    let crcValue = crc32(identifierCharacters.utf8)
+    let suffix = String(crcValue, radix: 16, uppercase: false)
 
-    // Strip out any characters in the function's signature that won't play well
-    // in a generated symbol name.
-    let identifier = String(
-      identifierCharacters.map { character in
-        if character.isLetter || character.isWholeNumber {
-          return character
-        }
-        return "_"
-      }
-    )
-
-    // If there is a non-ASCII character in the identifier, we might be
-    // stripping it out above because we are only looking for letters and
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

grynspan:
@stmontgomery previously gave me a thumbs-up on the concept of the change, though not the diff itself.

stmontgomery:
> This results in very long symbol names that `swift-demangle` has trouble with.

Is there some bug there that needs to be reported?

stmontgomery (approved):

PR #765, opened 2024-10-15
Delete reference to simpleTypeSpecifier

meg-gupta:
Update as per swift-syntax changes in https://github.com/swiftlang/swift-syntax/pull/2876

### DIFF
--- a/Sources/TestingMacros/TestDeclarationMacro.swift
+++ b/Sources/TestingMacros/TestDeclarationMacro.swift
@@ -86,7 +86,12 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
       let invalidSpecifierKeywords: [TokenKind] = [.keyword(.inout), .keyword(.isolated), .keyword(._const),]
       if let parameterType = parameter.type.as(AttributedTypeSyntax.self) {
         for specifier in parameterType.specifiers {
-          if invalidSpecifierKeywords.contains(specifier.specifier.tokenKind) {
+          for token in specifier.tokens(viewMode: .sourceAccurate) {
+            if invalidSpecifierKeywords.contains(token.tokenKind) {
+              diagnostics.append(.specifierNotSupported(token, on: parameter, whenUsing: testAttribute))
+              break
+            }
+          }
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
It looks like this will break compatibility with the swift-syntax-600.0.x tag that the package depends on. Is there a way we can stage this change such that it will remain compatible with earlier tags of swift-syntax? Alternatively, could we first get this change included in a swift-syntax tag, then update the package dependency to reference that tag and then land this change?

grynspan:
Should be possible to do something like:

```swift
#if !canImport(SwiftSyntax, _version: N)
// old code here
#endif
```

Where `N` is whatever tag of swift-syntax deleted this symbol.

Sources/TestingMacros/TestDeclarationMacro.swift
```diff
@@ -86,9 +86,6 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
       let invalidSpecifierKeywords: [TokenKind] = [.keyword(.inout), .keyword(.isolated), .keyword(._const),]
       if let parameterType = parameter.type.as(AttributedTypeSyntax.self) {
         for specifier in parameterType.specifiers {
-          guard case let .simpleTypeSpecifier(specifier) = specifier else {
-            continue
-          }
           if invalidSpecifierKeywords.contains(specifier.specifier.tokenKind) {
```
grynspan, line 89:
```suggestion
          for token in specifier.tokens(viewMode: .sourceAccurate) {
            if invalidSpecifierKeywords.contains(token.tokenKind) {
              diagnostics.append(.specifierNotSupported(token, on: parameter, whenUsing: testAttribute))
              break
            }
          }
```

meg-gupta:
@swift-ci please test

grynspan:
Based on a side discussion, I think we can close this PR?

meg-gupta:
Closing this PR. Reworked https://github.com/swiftlang/swift-syntax/pull/2876 to avoid api break

Issue #714, opened 2024-09-19
Add attachments


Arideno:
### Description

The XCTest framework has an attachment feature, which is beneficial for snapshot tests to visualize the differences. It would be advantageous if this feature were incorporated into SwiftTesting as well.

grynspan:
Oddly timely as I'm working on a proposal for this right now!

Issue #478, opened 2024-06-13
In the swift program it works perfectly, don't you also have a test for Python?


MAFLIXD:
### Description

This same test exists but what can I use for other languages ​​like Python?

stmontgomery:
I'm sorry I don't understand what this issue is describing, can you elaborate? Closing for now, but if you can provide more information we can consider reopening.

PR #769, opened 2024-10-17
Split `SWT_NO_EXIT_TESTS` into `SWT_NO_EXIT_TESTS` and `SWT_NO_PROCESS_SPAWNING`.

grynspan:
This PR separates out our process-spawning code to be guarded by `SWT_NO_PROCESS_SPAWNING` instead of `SWT_NO_EXIT_TESTS`. We do this so that we can potentially use process spawning on platforms where exit tests are not supported for some other reason (such as the iOS/Android sandboxes) but process spawning is still internally possible.

There are a few use cases we have for spawning processes that don't involve exit tests:

- Calling out to `tar` to compress attachments (see #714)
- Running non-Swift scripts in their interpreters (see #478)
- Multi-process parallelism (the XCTest model)

I took the opportunity to clean up WaitFor.swift a bit and rearrange code so that the "new platform, dunno who lives here" case should compile (although not function) out-of-the-box.

### Checklist:


grynspan:
@swift-ci test

briancroom (approved):

PR #772, opened 2024-10-18
Update documentation to explain misconfigurations of `SWT_` conditions.

grynspan:
This PR was supposed to be part of #769, but I must have blown away a local commit because it didn't make it into that PR.

This PR updates Porting.md to explain when a developer might encounter an error due to mismatched `SWT_` conditions. It ensures that exit tests require process spawning to be enabled/implemented by adding a dependency between `SWT_NO_PROCESS_SPAWNING` and `SWT_NO_EXIT_TESTS`.

This PR also tweaks which bits of `ExitTest` are exposed on platforms that do not support exit tests at all. The type needs to be exposed so that `ExitTest.Result` is exposed so that we can correctly expose `#expect(exitsWith:)` and `#require(exitsWith:)` while marking them unavailable, as opposed to hiding the macros and not giving a developer clear diagnostics.

### Checklist:

### DIFF
--- a/Documentation/Porting.md
+++ b/Documentation/Porting.md
@@ -58,6 +58,16 @@ These warnings may be emitted by our internal C++ module (`_TestingInternals`)
 or by our library module (`Testing`). Both indicate areas of our code that need
 platform-specific attention.
 
+> [!NOTE]
+> Rarely, you may encounter errors of a similar form:
+> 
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci test

briancroom (approved):

stmontgomery (approved):

grynspan:
@swift-ci test Linux

PR #766, opened 2024-10-15
Handle signals on Windows in exit tests.

grynspan:
This PR adds support for handling signals as distinct exit conditions on Windows. Previously, we didn't support them because Windows itself has minimal support. However, through the judicious use of "stuffing a bunch of bits into a 32-bit field", we are able to propagate raised signals out of the child process and detect them in the parent process.

We do this by installing our own signal handlers for all signals supported on Windows, then masking the caught signal against an `NTSTATUS` severity and facility that are unlikely to be reported by the system in practice. In the parent process, we look for exit codes that match these values and extract the signal from them when found. Because the namespace for exit codes on Windows is shared with uncaught [VEH/SEH exceptions](https://learn.microsoft.com/en-us/cpp/cpp/structured-exception-handling-c-cpp) (which are propagated to the parent process as `NTSTATUS` codes), there is the potential for conflict with some hypothetical _real_ exception code, but I'm taking a gamble here that my choice of `NTSTATUS` facility is unique.

(Bonus points for there not being convenient macros to construct an `NTSTATUS` code anywhere in the Windows SDK I can find.)

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #776, opened 2024-10-21
Avoid using `NTSTATUS` near `exit()`.

grynspan:
On Windows, `NTSTATUS` is a typedef of `long`. On 32-bit Windows, `long` is imported into Swift as `Int` but on 64-bit Windows,it's imported as `Int32`. These types are layout-equivalent but not implicitly convertible, so we get errors on 32-bit Windows when using an `NTSTATUS` (i.e. `Int`) where a `CInt` (i.e. `Int32`) is expected.

This PR changes the types of the constants introduced in #766 to be explicitly `CInt` so that they are usable on both flavours of Windows.

### Checklist:


grynspan:
@swift-ci test

dschaefer2 (approved):

plemarquand (approved):

PR #779, opened 2024-10-22
Make the result of an exit test optional with `#expect`.

grynspan:
This PR changes the semantics of `#expect(exitsWith:)` such that it only returns a value on success, and returns `nil` on failure. This is consistent with general Swift semantics.

This means that if you have an exit test that fails to exit with the expected exit condition, you won't get a value back, but that's intentional now because your test will have already failed. This makes it easier to reason about properties we add to the result structure in the future: they'll always be valid if you have an instance of the structure because you can't get a half-baked instance back by mistake.

This change made me realize that our custom comparison operators on `ExitCondition` do not compare against `nil` correctly, so I've moved them to `Optional<ExitCondition>`. (I know @stmontgomery is interested in revisiting whether they should be operators at all, but let's leave that discussion for a separate PR or thread.)

I also renamed the structure `ExitTestArtifacts` for two reasons:

1. To get it out from under `ExitTest` so that that type does not need to be API, and
2. To give it an even uglier name so we are forced to debate what to call it when the feature comes up for review.

### Checklist:


grynspan:
@swift-ci test

stmontgomery:
Offline, I questioned how this would affect the amount of diagnostic info made available in case an `#expect(exitsWith:)` fails. Since the result value would now be `nil` in that case, a test author may not be given as direct access to that content in the outer test function body. I think besides that concern, the changes in this PR do make sense, so @grynspan and I brainstormed and thought we could continue to surface actionable details about the exit test's outcome as attachments (which are another emerging feature). I'd encourage moving forward with this PR as is, as long as we can plan to pursue this attachments enhancement as follow-on work.

stmontgomery (approved):

grynspan:
>  we could continue to surface actionable details about the exit test's outcome as attachments (which are another emerging feature).

This will have to wait for a few more PRs to land but yes, I think this is a good way to surface the stdio logs for an exit test that fails while preserving the nullability semantics that otherwise make sense for `#expect()` and `#require()`.

PR #787, opened 2024-10-28
Ensure the identifier of a "seen" object is only removed if the object's children were recursively reflected

### LINKED ISSUE #785: A cyclic graph of 3 or more items causes an infinite recursion
### Description

Initializing an `argument` that contains 3 or more cyclical classes causes an infinite recursion within `__Expression.Value`'s private init:
```
private init(
        _reflecting subject: Any,
        label: String?,
        seenObjects: inout [ObjectIdentifier: AnyObject]
    )
```

I tracked it down to the attempt to prevent cyclical dependencies.  Using the sample code provided in Steps to Reproduce, I was able to _maybe_ fix this by removing the `defer` and adding a `filter` onto the children:
```swift
    private init(
        _reflecting subject: Any,
        label: String?,
        seenObjects: inout [ObjectIdentifier: AnyObject]
    ) {
        ...

        if shouldIncludeChildren, !mirror.children.isEmpty || isCollection {
            children = mirror.children.**filter { child in
                !seenObjects.contains(ObjectIdentifier(child as AnyObject))
            }**.map { child in
                Self(_reflecting: child.value, label: child.label, seenObjects: &seenObjects)
            }
        }
    }
```
But I'm unsure if this has any unintended side effects (specifically, the comment callout about multiple references to the same object in the same subject).  Logging a `Value(reflecting: a)` in the `makeA()` method returns me a `Value` instance that looks to properly stop traversing future references of the same type after it traverses first one.

### Expected behavior

Cyclical graphs of any complexity should resolve their reflection correctly

### Actual behavior

Once 3 classes are introduced into the graph, the initializer enters an infinite loop

### Steps to reproduce

1. Attempt to run the following Test:
```swift
class A {
    private var c: C!
    private var c2: C!
    private var b: B!

    func setup(b: B, c: C) {
        self.b = b
        self.c = c
        c2 = c
    }
}

class B {
    private var a: A!
    private var c: C!

    func setup(a: A, c: C) {
        self.a = a
        self.c = c
    }
}

class C {
    private var a: A!
    private var b: B!

    func setup(a: A, b: B) {
        self.a = a
        self.b = b
    }
}

@Suite
struct TestingAppTests {
    @Test(arguments: [makeA()]) func example(foo _: A) async throws {
        // Write your test here and use APIs like `#expect(...)` to check expected conditions.
        #expect(true)
    }

    static func makeA() -> A {
        let a = A()
        let b = B()
        let c = C()

        a.setup(b: b, c: c)
        b.setup(a: a, c: c)
        c.setup(a: a, b: b)

        return a
    }
}
```
2. Observe the process crashes after the return of `makeA()` and before the main test begins

Xcode's version of swift-testing

swift-driver version: 1.115 Apple Swift version 6.0.2 (swiftlang-6.0.2.1.2 clang-1600.0.26.4)
Darwin <private>.local 24.2.0 Darwin Kernel Version 24.2.0: Tue Oct 15 18:15:36 PDT 2024; root:xnu-11215.60.364.501.5~3/RELEASE_ARM64_T6000 arm64

stmontgomery:
This fixes a crash which can occur if an object has certain cyclic references and it's included in an expectation expression which fails.

The bug is that the identifier of a "seen" object passed to `Expression.Value.init(_reflecting:label:seenObjects:)` should only be removed from the tracking dictionary if the object's children were recursively reflected. If recursion did not occur, that indicates that the object has been seen so it should be left in the tracking dictionary to prevent subsequent recursion. I added new unit tests to validate this.

Fixes #785

### Result:

The scenarios described in #785 no longer crash.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/Expression.swift
+++ b/Sources/Testing/SourceAttribution/Expression.swift
@@ -190,9 +190,9 @@ public struct __Expression: Sendable {
     ///     This is used to halt further recursion if a previously-seen object
     ///     is encountered again.
     private init(
-        _reflecting subject: Any,
-        label: String?,
-        seenObjects: inout [ObjectIdentifier: AnyObject]
+      _reflecting subject: Any,
+      label: String?,
+      seenObjects: inout [ObjectIdentifier: AnyObject]
     ) {
       let mirror = Mirror(reflecting: subject)
 
@@ -214,7 +214,7 @@ public struct __Expression: Sendable {
       // `type(of:)`, which is inexpensive. The object itself is stored as the
       // value in the dictionary to ensure it is retained for the duration of
       // the recursion.
-      var objectIdentifierTeRemove: ObjectIdentifier?
+      var objectIdentifierToRemove: ObjectIdentifier?
       var shouldIncludeChildren = true
       if mirror.displayStyle == .class, type(of: subject) is AnyObject.Type {
         let object = subject as AnyObject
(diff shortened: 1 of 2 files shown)


Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -214,24 +214,25 @@ public struct __Expression: Sendable {
       // `type(of:)`, which is inexpensive. The object itself is stored as the
       // value in the dictionary to ensure it is retained for the duration of
       // the recursion.
-      var objectIdentifierTeRemove: ObjectIdentifier?
+      var objectIdentifierToRemove: ObjectIdentifier?
```
stmontgomery, line 217: Drive by typo and style fixes here and above

suzannaratcliff (approved):

Issue #162, opened 2023-12-15
Value not shown in error message when calling async variant of #expect


tevelee:
### Description

`#expect` works differently when `await` keyword is inside its body. `await #expect(expr)` shows a breakdown of the actual value, while `#expect(await expr)` does not. I would like to know why the expectation failed even if the result is computed from an async expression.

![image](https://github.com/apple/swift-testing/assets/443078/d06098a7-98eb-4589-ad5e-69e57374ba05)

### Expected behavior

`#expect(await Array(stream) == [1,2])` throws the error `testAll(): Expectation failed: (Array(stream) → [1, 2, 3]) == ([1,2] → [1, 2])`

### Actual behavior

`#expect(await Array(stream) == [1,2])` throws the error `testAll(): Expectation failed: await Array(stream) == [1,2]`

### swift-testing version/commit hash

3fa4ea0bfcbfc48102ca5d81f03a00debd0d4372

### Swift & OS version (output of `swift --version && uname -a`)

swift-driver version: 1.87.3 Apple Swift version 5.9.2 (swiftlang-5.9.2.2.56 clang-1500.1.0.2.5)
Target: arm64-apple-macosx14.0
Darwin MBP 23.1.0 Darwin Kernel Version 23.1.0: Mon Oct  9 21:27:24 PDT 2023; root:xnu-10002.41.9~6/RELEASE_ARM64_T6000 arm64

grynspan:
This is expected and is a constraint of macros in Swift. The presence of the await keyword signals to the compiler that some subexpression of the expectation is asynchronous, but it is impossible to know exactly which subexpression from the AST alone. This means it is not possible to correctly decompose the expression as we can with synchronous, non-throwing ones.

tevelee:
@grynspan To be clear, I'm not asking to decompose the macro expressions at compile time. I'm asking for the error message to show the actual vs expected values at runtime when the assertion fails. This does not happen in the async case.

grynspan:
[...]

PR #790, opened 2024-10-30
Do not perform full expansion of `#expect()` when `try` or `await` is present.

### LINKED ISSUE #783: `#expect` doesn't expect that there may be a sub-expression that has `try` on the right side of binary operator.
### Description

Although we can't put `try` on the right side of `==`, the right expression can contain a sub-expression that has `try`:

```swift
func nonThrowingFunc() -> Int { 0 }
func throwingFunc() throws -> Int { 0 }
func anotherNonThrowingFunc(_ integer: Int) -> Int { integer }

print(nonThrowingFunc() == try throwingFunc()) // ❌ error: 'try' cannot appear to the right of a non-assignment operator
print(nonThrowingFunc() == anotherNonThrowingFunc(try throwingFunc())) // ✅ OK
```

However, `#expect` seems not to expect that there may be a `try` sub-expression on the right side of `==`:

```swift
import Testing

struct Foo: Equatable {
  init() {}
  init(_ integer: Int) {}
}

@Test func myTest() throws {
  #expect(try #require(Int("0")) == 0) // ✅ OK
  #expect((Foo() == Foo(try #require(Int("0")))) == true) // ✅ OK
  #expect(Foo() == Foo(try #require(Int("0")))) // ❌ error: call can throw, but it is executed in a non-throwing autoclosure
}
```

Expanded function named `__checkBinaryOperation` doesn't have an `@autoclosure` that `throws`.

### Expected behavior

No error at compile time.

### Actual behavior

Described above.

### Steps to reproduce

Try to compile the code above.

Testing Library Version: 94 (arm64e-apple-macos13.0)

swift-driver version: 1.115 Apple Swift version 6.0 (swiftlang-6.0.0.9.10 clang-1600.0.26.2)
Darwin yockow.local 24.0.0 Darwin Kernel Version 24.0.0: Tue Sep 24 23:39:07 PDT 2024; root:xnu-11215.1.12~1/RELEASE_ARM64_T6000 arm64

grynspan:
For reasons that have been documented at length (see #162), we aren't able to correctly expand conditions on `#expect()` or `#require()` that have effects (`try` or `await`.) We aren't currently detecting all possible patterns that expand incorrectly. This PR causes us to back out of the full expansion if the `try` or `await` keyword is present _anywhere_ inside an expectation condition expression.

Resolves #783.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/ConditionArgumentParsing.swift
+++ b/Sources/TestingMacros/Support/ConditionArgumentParsing.swift
@@ -312,17 +312,6 @@ private func _exprFromOptionalChainedExpr(_ expr: some ExprSyntaxProtocol) -> (E
 ///
 /// - Returns: An instance of ``Condition`` describing `expr`.
 private func _parseCondition(from expr: FunctionCallExprSyntax, for macro: some FreestandingMacroExpansionSyntax, in context: some MacroExpansionContext) -> Condition {
-  // If the member function call involves the `try` or `await` keywords, assume
-  // we cannot expand it. This check cannot handle expressions like
-  // `try #expect(a.b(c))` where `b()` is throwing because the `try` keyword
-  // is outside the macro expansion. SEE: rdar://109470248
-  let containsTryOrAwait = expr.tokens(viewMode: .sourceAccurate).lazy
-    .map(\.tokenKind)
-    .contains { $0 == .keyword(.try) || $0 == .keyword(.await) }
-  if containsTryOrAwait {
-    return Condition(expression: expr)
-  }
-
   // We do not support function calls with trailing closures because the
   // transform required to forward them requires more information than is
   // available solely from the syntax tree.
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci test

Sources/TestingMacros/Support/ConditionArgumentParsing.swift
```diff
@@ -477,6 +466,22 @@ private func _parseCondition(negating expr: ExprSyntax, isParenthetical: Bool, f
 ///
 /// - Returns: An instance of ``Condition`` describing `expr`.
 private func _parseCondition(from expr: ExprSyntax, for macro: some FreestandingMacroExpansionSyntax, in context: some MacroExpansionContext) -> Condition {
+  // Handle closures with a single expression in them (e.g. { $0.foo() })
```
grynspan, line 469: Moved this up so that closure "expansion" (not really but sure why not) still happens. If we ever actually start doing fun things with closures, we'll probably need to move this back down?

grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

Tests/TestingTests/MiscellaneousTests.swift
```diff
@@ -123,6 +123,15 @@ struct SendableTests: Sendable {
 
   @Test(.hidden, arguments: FixtureData.stringReturningClosureArray)
   func parameterizedAcceptingFunction(f: @Sendable () -> String) {}
+
+  @Test(.hidden) func throwKeywordInExpectation() throws {
```
briancroom, line 127: These are test cases that would have failed compile prior to this fix?

grynspan: Yep.

briancroom (approved):

PR #792, opened 2024-10-31
Update install to ensure the build directory is not added to the rpath

bnbarham:
This seems to be fine with the existing CMake, but in 3.29 (possibly earlier) the rpath of libTesting still has the build directory. Make sure to remove any existing build rpath.


grynspan:
@stmontgomery Can you review?

stmontgomery (approved):
Thanks!

briancroom (approved):

PR #797, opened 2024-11-02
Ensure the `Bool` overload of `__checkBinaryOperation()` is still preferred.

grynspan:
We noticed that in some conditions, boolean arguments to `#expect()` would compile to use the _optional_ overload of `__checkBinaryOperation()`, which would result in a pass when a failure should have occurred because the boolean value never evaluated to `nil`. This PR should resolve that issue.

Resolves rdar://138964155.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

briancroom (approved):
Thanks for the fix and new test!

PR #796, opened 2024-11-02
Attachments! [Take 2]

grynspan:
This PR introduces a new experimental feature, attachments. With this feature, you can "attach" values that conform to a new `Test.Attachable` protocol to a test. With the right command-line incantation (TBD), Swift Testing will automatically write attachments to disk for you.

> [!NOTE]
> This PR does not teach Xcode or VS Code how to handle attachments produced by
> Swift Testing, nor does it add the necessary command-line arguments to the
> `swift test` command-line tool.

This PR is one of a series that I'll be posting to build out this feature.

As always, keep in mind that symbols marked `@_spi(Experimental)` are subject to change or removal without notice.

Resolves #714.
Resolves rdar://88648735.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #800, opened 2024-11-03
WASI: Access to `EEXIST` through a stub getter function

kateinoigakukun:
Build fix for WASI

### Motivation:

Use of errno values in Swift without libc overlay is not supported with wasi-libc.

https://ci.swift.org/job/oss-swift-pr-test-crosscompile-wasm-ubuntu-20_04/1784/console

```
/home/build-user/swift-testing/Sources/Testing/Attachments/Test.Attachment.swift:249:61: error: cannot find 'EEXIST' in scope
247 |           result = preferredPath
248 |           break
249 |         } catch let error as CError where error.rawValue == EEXIST {
    |                                                             `- error: cannot find 'EEXIST' in scope
250 |           // Try again with a new suffix.
251 |           continue
```

### Modifications:

This change introduces a new function `swt_EEXIST` to get the value of `EEXIST` because it is a complex macro in wasi-libc and cannot be imported directly into Swift.

### Result:

WASI build will be repaired.

### Checklist:

### DIFF
--- a/Sources/Testing/Attachments/Test.Attachment.swift
+++ b/Sources/Testing/Attachments/Test.Attachment.swift
@@ -246,7 +246,7 @@ extension Test.Attachment {
           file = try FileHandle(atPath: preferredPath, mode: "wxb")
           result = preferredPath
           break
-        } catch let error as CError where error.rawValue == EEXIST {
+        } catch let error as CError where error.rawValue == swt_EEXIST() {
           // Try again with a new suffix.
           continue
         } catch where usingPreferredName {
--- a/Sources/_TestingInternals/include/Stubs.h
+++ b/Sources/_TestingInternals/include/Stubs.h
@@ -154,6 +154,14 @@ static int swt_siginfo_t_si_status(const siginfo_t *siginfo) {
 #endif
 #endif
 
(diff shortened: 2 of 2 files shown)


kateinoigakukun:
@swift-ci test

grynspan (approved):

grynspan:
Fixes issue introduced in #796.

grynspan:
@kateinoigakukun Seems like the Musl Swift overlay should (re)define these error codes so they're visible to Swift?

kateinoigakukun:
@grynspan The complex errno code definitions exist only in wasi-libc, not in musl. And we already define [these error codes in WASILibc overlay module](https://github.com/swiftlang/swift/blob/e089c5850ed6d11b61353a0a76e2e124746471ef/stdlib/public/Platform/WASILibc.swift.gyb#L67) but not used them here to follow [the porting policy](https://github.com/swiftlang/swift-testing/blob/main/Documentation/Porting.md#including-system-headers)

grynspan:
Fair enough. Just mentioning it in case it wasn't already handled and other projects could run into trouble. :)

PR #801, opened 2024-11-04
Specify the module to find the ABI entry point function in (in our own tests).

grynspan:
The order in which Xcode loads the Apple-vendored copy of Swift Testing vs. a developer-supplied XCTest bundle is unspecified. Our tests assume that the "current" copy of Swift Testing (that is, the copy they linked against) is the first one loaded, which is not always true when running tests in Xcode. The modified tests end up finding the wrong entry point functions when called within Xcode if the load order has changed.

This PR explicitly specifies that we are looking for the entry point functions in the module the tests are linked against. The change is applied across all platforms that support dynamic linking for the sake of consistency as any of them might have linked a copy of Swift Testing from the toolchain _and_ the package itself.

Resolves rdar://139140302.

### Checklist:


grynspan:
@swift-ci please test

grynspan:
@swift-ci test

briancroom (approved):

PR #803, opened 2024-11-04
Test cleanup: Remove an obsolete fixture suite and add .hidden to a few tests

stmontgomery:
A few small cleanups to the project's unit tests:

- Remove the obsolete fixture suite type `IndependentlyRunnableTests`. The test which used to reference it has been deleted.
- Add the `.hidden` trait to a few eligible `@Test` functions.

### Checklist:

### DIFF
--- a/Tests/TestingTests/ExitTestTests.swift
+++ b/Tests/TestingTests/ExitTestTests.swift
@@ -439,7 +439,7 @@ private import _TestingInternals
 }
 
 #if false // intentionally fails to compile
-@Test(arguments: 100 ..< 200)
+@Test(.hidden, arguments: 100 ..< 200)
 func sellIceCreamCones(count: Int) async throws {
   try await #require(exitsWith: .failure) {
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

Tests/TestingTests/ExitTestTests.swift
```diff
@@ -439,7 +439,7 @@ private import _TestingInternals
 }
 
 #if false // intentionally fails to compile
-@Test(arguments: 100 ..< 200)
```
grynspan, line 442: To be clear, this test doesn't ever compile. It's useful when testing exit tests to verify what happens if you attempt to close over some state.

grynspan (approved):

Tests/TestingTests/ExitTestTests.swift
stmontgomery: Yeah, understood. Despite being compiled out, without this trait it can still be included in the indexer's test discovery.

PR #806, opened 2024-11-04
Fix a crash when describing a confirmation failure with a range.

### LINKED ISSUE #805: Crash when a call to `confirmation(expectedCount: 1...) { ... }` fails
### Description

I'm seeing a crash during issue stringification when running a test which fails due to a particular range-based confirmation failing.

### Expected behavior

Expected an ordinary test failure, not a crash.

### Actual behavior

A crash is occurring inside `Issue.Kind.description.getter`, specifically, at `Issue.swift:203` in my checkout.

### Steps to reproduce

Run the following test:

```
@Test func confirmedTooFewTimes() async {
  await confirmation(expectedCount: 1...) { (thingHappened) async in }
}
```

759ce021552dc3939255f4cede093605404b58a8

_No response_

grynspan:
I was using `relative(to:)` to convert any arbitrary `RangeExpression` to one with an upper and lower bound, but that function may produce invalid ranges for expressions like `1...` that cause crashes.

This PR switches to a different, somewhat sillier implementation that should not run into this issue.

Resolves #805.
Resolves rdar://139222774.

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -175,11 +175,9 @@ extension Issue: CustomStringConvertible, CustomDebugStringConvertible {
 /// In the future, when our minimum deployment target supports casting a value
 /// to a constrained existential type ([SE-0353](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0353-constrained-existential-types.md#effect-on-abi-stability)),
 /// we can remove this protocol and cast to `RangeExpression<Int>` instead.
-private protocol _RangeExpressionOverIntValues: RangeExpression where Bound == Int {}
+private protocol _RangeExpressionOverIntValues: RangeExpression & Sequence where Bound == Int, Element == Int {}
 extension ClosedRange<Int>: _RangeExpressionOverIntValues {}
 extension PartialRangeFrom<Int>: _RangeExpressionOverIntValues {}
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci test

stmontgomery (approved):

Sources/Testing/Issues/Issue.swift
```diff
@@ -175,11 +175,9 @@ extension Issue: CustomStringConvertible, CustomDebugStringConvertible {
 /// In the future, when our minimum deployment target supports casting a value
 /// to a constrained existential type ([SE-0353](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0353-constrained-existential-types.md#effect-on-abi-stability)),
 /// we can remove this protocol and cast to `RangeExpression<Int>` instead.
-private protocol _RangeExpressionOverIntValues: RangeExpression where Bound == Int {}
+private protocol _RangeExpressionOverIntValues: RangeExpression & Sequence where Bound == Int, Element == Int {}
```
stmontgomery, line 178: Would it be stronger / more precise to constrain such that `Element == Bound`?
```suggestion
private protocol _RangeExpressionOverIntValues: RangeExpression & Sequence where Bound == Int, Element == Bound {}
```

briancroom (approved):

PR #807, opened 2024-11-04
Add support for platform-specific unconditional availability.

grynspan:
This PR adds support for the following variant of `@available`:

```swift
@available(macOS, unavailable)
@Test func f() {}
```

Previously, we were simply ignoring these availability attributes.

Resolves rdar://137776333.

### Checklist:

### DIFF
--- a/Sources/Testing/Traits/ConditionTrait+Macro.swift
+++ b/Sources/Testing/Traits/ConditionTrait+Macro.swift
@@ -92,9 +92,16 @@ extension Trait where Self == ConditionTrait {
     _ condition: @escaping @Sendable () -> Bool
   ) -> Self {
     // TODO: Semantic capture of platform name/version (rather than just a comment)
-    Self(
+    let message: Comment = if let message {
+      message
+    } else if let version {
+      "Obsolete as of \(_description(ofPlatformName: platformName, version: version))"
+    } else {
+      "Unavailable on \(_description(ofPlatformName: platformName, version: nil))"
+    }
+    return Self(
       kind: .conditional(condition),
-      comments: [message ?? "Obsolete as of \(_description(ofPlatformName: platformName, version: version))"],
+      comments: [message],
       sourceLocation: sourceLocation
     )
   }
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci test

Sources/Testing/Traits/ConditionTrait+Macro.swift
```diff
@@ -92,9 +92,16 @@ extension Trait where Self == ConditionTrait {
     _ condition: @escaping @Sendable () -> Bool
   ) -> Self {
     // TODO: Semantic capture of platform name/version (rather than just a comment)
-    Self(
+    let message: Comment = if let message {
+      message
+    } else if let version {
+      "Obsolete as of \(_description(ofPlatformName: platformName, version: version))"
```
stmontgomery, line 98: I know this is the existing behavior, but I wonder if this text would align better with the "obsolete**d**" term in source:
```suggestion
      "Obsoleted in \(_description(ofPlatformName: platformName, version: version))"
```

stmontgomery:
We should consider adding this to some DocC article as a recommended workflow for disabling a test on an entire platform, as an alternative to `.enabled(if: ...)` in that situation.

stmontgomery (approved):

Sources/Testing/Traits/ConditionTrait+Macro.swift
grynspan: `"Obsoleted as of macOS 99.0"`

vs.

`"Obsoleted in macOS 99.0"`

grynspan: I don't feel strongly, but as it's old code I'm going to leave it as-is for now. We can open another PR if you like.

stmontgomery: It's really "Obsolete" vs. "Obsolete**d**" that I was focused on. The latter would more closely match the language keyword which is spelled `obsoleted`.

PR #809, opened 2024-11-05
Eagerly move attachable values to the heap.

grynspan:
This PR modifies `Test.Attachment.init(...)` so that, if the attachable value passed to it is of a large or non-bitwise-copyable value type, we store it in a heap-allocated container object.

Why do we want to do this? Attachable values have arbitrary types we don't really know anything about, and may be very expensive to copy if they are large or if they contain many references to other objects. Since attachments are sendable and need to pass through many layers in the system, the cost of making copies for these attachable values can add up quickly. So, instead of copying them repeatedly, if we store them in a refcounted container and pass _that_ value around, we only infer a single copy cost.

There are four modes for attachable value storage after this PR:

```mermaid
flowchart TB;
    S@{ shape: diam, label: "Sendable?" };
    C@{ shape: diam, label: "Copyable?" };
    AO@{ shape: diam, label: "AnyObject?" };
    BC@{ shape: diam, label: "BitwiseCopyable (POD)?" };
    SMOL@{ shape: diam, label: "Fits in an existential box?" };

    MO@{ shape: das, label: "In a heap-allocated box\n(eagerly serialized)" };
    EX@{ shape: das,  label: "Inline in an existential box" };
    H@{ shape: das,  label: "In a heap-allocated box" };
    RC@{ shape: das,  label: "Directly on the heap" };

    S-- Yes ---C
        C-- Yes ---AO
            AO-- Yes ---RC
            AO-- No ---BC
                BC-- Yes ---SMOL
                BC-- No ---H
                    SMOL-- Yes ---EX
                    SMOL-- No ---H
        C-- No ---MO
    S-- No ---MO
```

### Checklist:

### DIFF
--- a/Sources/Testing/Attachments/Test.Attachment.swift
+++ b/Sources/Testing/Attachments/Test.Attachment.swift
@@ -68,7 +68,11 @@ extension Test {
       named preferredName: String? = nil,
       sourceLocation: SourceLocation = #_sourceLocation
     ) {
-      self.attachableValue = attachableValue
+      if _shouldMoveValueToHeap(attachableValue) {
+        self.attachableValue = _HeapAllocatedAttachableProxy(rawValue: attachableValue)
+      } else {
+        self.attachableValue = attachableValue
+      }
       self.preferredName = preferredName ?? Self.defaultPreferredName
       self.sourceLocation = sourceLocation
     }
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci test

stmontgomery (approved):

vanvoorden:
Ahh… interesting! Something on a similar topic I was investigating was the impact that migrating from pass-by-value (for large data structures) to pass-by-reference would have on binary size. This might not be so critical for *test* code (makes a bigger impact for production code)… but was there any attempt to benchmark or test for binary size impact or compile time improvements from reducing the amount of code `swift-testing` has to generate behind the scenes to pass very complex value types?

grynspan:
I would not expect this change to affect the binary size of test code in any statistically significant way. Swift Testing is built with library evolution enabled, so the internal `_HeapAllocatedAttachableProxy` class is not visible to client code.

For embedded targets (which we do not support yet), attachments will probably always be eagerly serialized and this class won't be necessary.

vanvoorden:
> Swift Testing is built with library evolution enabled, so the internal _HeapAllocatedAttachableProxy class is not visible to client code. Client code therefore cannot optimize based on whether or not it's used, nor can it create static specializations of the class that are emitted into any binaries.

https://forums.swift.org/t/what-does-island-suffix-in-symbol-name-mean/73105

Hmm… what I was investigating was if (accounting for the small increase in instructions and binary size from code landed in this diff to support heap boxing) we could still see a net negative reduction in binary size when product engineers build their test code… if the product engineer was previously attempting to pass-by-value very large data structures and the amount of code Swift generated behind the scenes to pass those very large data structures was no longer necessary.

It's not a high-priority blocker for me… but something I was trying to investigate and if there was any more potential insight from what was just shipped here that would be great!

grynspan:
I would be genuinely surprised if this has any non-negligible impact on the size of production binaries or test binaries when Swift Testing is used from the toolchain or Xcode.

PR #811, opened 2024-11-06
Provide an opt-out conditional for lazy attachment encoding.

grynspan:
This PR provides a compile-time conditional, `SWT_NO_LAZY_ATTACHMENTS`, which when set causes Swift Testing's new experimental attachments feature to always eagerly encode/serialize/whatever attachments even if they're copyable and sendable.

We anticipate using this functionality in Embedded Swift in the future. Note that support for Embedded Swift is not (yet!) a planned feature.

### "Why not make Attachment generic?"

I've got a separate branch that tries to do this. It has a significantly more complex implementation and is brittle because we need to type-erase the attachable value in order to pass it into the event-handling machinery (which entails having a bunch of special-casing since `any Test.Attachable` does not conform to `Test.Attachable`.) `Attachment` is _already_ meant to represent a type-erased `Attachable` value.

### Checklist:

### DIFF
--- a/Sources/Testing/Attachments/Test.Attachment.swift
+++ b/Sources/Testing/Attachments/Test.Attachment.swift
@@ -21,22 +21,29 @@ extension Test {
   /// instance of ``Test/Attachment`` with that value and, optionally, a
   /// preferred filename to use when writing to disk.
   public struct Attachment: Sendable {
+#if !SWT_NO_LAZY_ATTACHMENTS
+    /// Storage for ``attachableValue``.
+    private var _attachableValue: any Attachable & Sendable /* & Copyable rdar://137614425 */
+
     /// The value of this attachment.
     ///
     /// The type of this property's value may not match the type of the value
     /// originally used to create this attachment.
-    public var attachableValue: any Attachable & Sendable /* & Copyable rdar://137614425 */
-
-    /// The source location where the attachment was initialized.
-    ///
-    /// The value of this property is used when recording issues associated with
-    /// the attachment.
-    public var sourceLocation: SourceLocation
-
-    /// The default preferred name to use if the developer does not supply one.
-    package static var defaultPreferredName: String {
-      "untitled"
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci test

Sources/Testing/Attachments/Test.Attachment.swift
```diff
@@ -129,19 +148,22 @@ extension Test.Attachment {
...
     _ attachableValue: borrowing some Test.Attachable & ~Copyable,
     named preferredName: String? = nil,
     sourceLocation: SourceLocation = #_sourceLocation
   ) {
+    let preferredName = preferredName ?? Self.defaultPreferredName
     var proxyAttachable = _AttachableProxy()
     proxyAttachable.estimatedAttachmentByteCount = attachableValue.estimatedAttachmentByteCount
 
     // BUG: the borrow checker thinks that withErrorRecording() is consuming
     // attachableValue, so get around it with an additional do/catch clause.
```
stmontgomery, line 164: I'm wondering if we ought to propagate caught Errors in this initializer? Unlike the other one, where encoding the attachment is deferred, and it does make sense to catch it and record an Issue, this one might be better to propagate and let the test author decide how to handle it.

stmontgomery (approved):
Had one tangential comment, but this PR looks good

Sources/Testing/Attachments/Test.Attachment.swift
grynspan: I thought about it, but my reasoning was that:

1. Errors adding attachments are probably going to be pretty rare;
2. You'd have to `try` for non-sendable or move-only attachments, but not if the attachment is sendable and copyable, which is inconsistent and potentially confusing; and
3. An error that occurs when creating an attachment is _probably_ not a test failure, but an infrastructure failure, so it should not cause a test to halt execution—which it would unless you wrote `try?`, in which case we lose the error.

PR #812, opened 2024-11-06
Provide an environment flag to disable error backtrace capturing.

### LINKED ISSUE #791: enable opting-out of thrown Error tracing
### Description

follow up from Slack: https://swift-open-source.slack.com/archives/C06QXFKB4ER/p1730243185369769

running tests with Testing, even in `-c release` mode, results in an >100x slowdown when using certain libraries such as [swift-json](https://github.com/tayloraswift/swift-json) due to the way it parses UTF-8 text internally. this in turn is because [swift-grammar](https://github.com/tayloraswift/swift-grammar), which swift-json’s parser is implemented in, throws an Error on nearly every input character. 

### Expected behavior

there should be a way to locally disable Error tracing so we can use these libraries within test runners.

### Actual behavior

it takes many minutes to parse a large JSON file within a test runner, which should only take a couple seconds.

### Steps to reproduce

https://github.com/tayloraswift/swift-json/pull/80 demonstrates the problem

6.0.1

_No response_

grynspan:
In performance-sensitive environments that throw many errors, the cost of capturing a backtrace for each one may be prohibitive. This PR allows disabling backtrace capture in environments where this is a problem.

Resolves #791.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/Backtrace.swift
+++ b/Sources/Testing/SourceAttribution/Backtrace.swift
@@ -353,24 +353,25 @@ extension Backtrace {
   /// This value is named oddly so that it shows up clearly in symbolicated
   /// backtraces.
   private static let __SWIFT_TESTING_IS_CAPTURING_A_BACKTRACE_FOR_A_THROWN_ERROR__: Void = {
-    if Environment.flag(named: "SWT_ERROR_BACKTRACING_ENABLED") == false {
-      // Backtraces have been explicitly disabled in the current environment.
-      return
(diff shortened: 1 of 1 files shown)


stmontgomery (changes requested):
I think we should have a way to more granularly control whether the `_CFErrorSetCallStackCaptureEnabled` hook is enabled, independent of whether the broader `Error` backtrace capture feature is enabled.

grynspan:
@swift-ci test

stmontgomery (approved):

briancroom (approved):

PR #813, opened 2024-11-08
Refine the Foundation error backtracing guard

briancroom:
Follow-on from https://github.com/swiftlang/swift-testing/pull/812:

* Move the check slightly to ensure that it is applied in all scenarios.
* Disable the behavior by default due to it causing problems for some tests which perform equality checks on NSError objects.


grynspan (approved):

stmontgomery (approved):

briancroom:
@swift-ci please test

PR #814, opened 2024-11-08
Make `Test.Attachment` generic.

grynspan:
This PR makes `Test.Attachment` generic over its attachable value type. It gains conditional conformance to `Copyable` and `Sendable` depending on the attachable value and if you call `attach()` on a move-only or non-sendable attachment, will eagerly serialize the attachable value at that point (rather than during initialization.)

There are a few benefits here:

1. Callers can statically know the type of the attachable value in an attachment rather than needing to always deal with an existential box;
2. We can add associated types to `Test.Attachable` that will be readily accessible in `withUnsafeBufferPointer(for:_:)` again without needing an existential; and
3. When we eventually add support for image attachments, we won't need a bunch of additional initializers or intermediate box types or what-have-you; and
4. For Embedded Swift or other environments where existentials are problematic, we can eagerly serialize all attachments and pass a consistent type (~~`Test.Attachment<[UInt8]>`~~ `Test.Attachment<Test.AnyAttachable>`) to the event handler.

There are also some drawbacks:

1. Because conformance to `Copyable` and `Sendable` is conditional, we lose a bit of flexibility if you have a non-sendable `Test.Attachment` instance or whatnot;
2. We still need a lazy, type-erased attachment type that can be passed to the event handler. I played around with `Test.Attachment<Any>` but that causes as many problems as it solves.

   ~~We end up with `Test.Attachment<any Test.Attachable & Sendable & Copyable>` but, because that's an existential type that doesn't conform to itself, the generic parameter `AttachableValue` is not constrained to `Test.Attachable`. We only provide initializers for types that do conform though (plus the existential one internally) so in practice it's not a huge issue.~~

   I replaced `any Test.Attachable & Sendable & Copyable` with a dedicated type-erasing box type, `Test.AnyAttachable`, that wraps the existential. In Embedded Swift, it wraps a `[UInt8]` instead.

3. There is some code duplication necessary (i.e. multiple implementations of `attach()` ~~and `write()`~~.)
4. Non-final classes cannot conform to `Test.Attachable`. You can work around this with a box type that's generic over the class and conforms to `Test.Attachable`: the `Test.AttachableContainer` protocol now exists to make this easier.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/Attachments/Test.Attachment.swift
```diff
@@ -20,30 +20,13 @@ extension Test {
...
-    ///
-    /// The type of this property's value may not match the type of the value
-    /// originally used to create this attachment.
-    public var attachableValue: some Test.Attachable & Sendable & Copyable {
-      _attachableValue
-    }
-#endif
+    public var attachableValue: AttachableValue
 
     /// The path to which the this attachment was written, if any.
```
plemarquand, line 31:
```suggestion
    /// The path to which the attachment was written, if any.
```

Sources/Testing/Attachments/Test.Attachment.swift
```diff
@@ -93,98 +73,134 @@ extension Test.Attachment {
...
+extension Test.Attachment where AttachableValue: Test.Attachable & Sendable & Copyable {
+  /// Attach this instance to the current test.
+  ///
+  /// - Parameters:
+  ///   - sourceLocation: The source location of the call to this function.
+  ///
+  /// An attachment can only be attached once.
+  @_documentation(visibility: private)
+  public consuming func attach(sourceLocation: SourceLocation = #_sourceLocation) {
+    let attachmentCopy = Test.Attachment<any Test.Attachable & Sendable /* & Copyable rdar://137614425 */>(self)
```
plemarquand, line 109: This posts an event but doesn't actually attach? Should it match the logic here: https://github.com/swiftlang/swift-testing/pull/814/files#diff-6feef57224f4a66aab636c27007b774e7a9bcf464b4383ba0976f88baca2fab3R134 ?

grynspan: This logic is correct. The difference from line 134 is that that one is just calling into this function, having gotten a copy of the attachment with a sendable/copyable value in it.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Package.swift
```diff
@@ -124,6 +124,7 @@ extension Array where Element == PackageDescription.SwiftSetting {
     availabilityMacroSettings + [
       .unsafeFlags(["-require-explicit-sendable"]),
       .enableUpcomingFeature("ExistentialAny"),
+      .enableExperimentalFeature("SuppressedAssociatedTypes"),
```
stmontgomery, line 127: We'll need to match this in the CMake rules too

Sources/Testing/Attachments/Test.AttachableContainer.swift
```diff
@@ -0,0 +1,37 @@
...
+  ///
+  /// To attach an attachable value to a test report or test run output, use it
+  /// to initialize a new instance of ``Test/Attachment``, then call
+  /// ``Test/Attachment/attach(sourceLocation:)``. An attachment can only be
+  /// attached once.
+  ///
+  /// A type can conform to this protocol if it represents another type that
+  /// cannot directly conform to ``Test/Attachable``, such as a non-final class
+  /// or a type declared in a third-party module.
+  public protocol AttachableContainer<AttachableValue>: Attachable, ~Copyable {
```
stmontgomery, line 25: Maybe "Value" would be more succinct?
```suggestion
  public protocol AttachableContainer<Value>: Attachable, ~Copyable {
```

Package.swift
grynspan: I thought I did. Oops.

Sources/Testing/Attachments/Test.AttachableContainer.swift
grynspan: It's a very generic term—and if this protocol is applied to some other type that conforms to other protocols, there's risk of conflict. I'd prefer to keep it a bit wordier to avoid those conflicts.

Sources/Testing/Events/Event.swift
```diff
@@ -198,13 +223,16 @@ public struct Event: Sendable {
   ///     ``Test/Case/current``.
   ///   - instant: The instant at which the event occurred. The default value
   ///     of this argument is `.now`.
+  ///   - sourceLocation: The source location at which the event occurred, if
+  ///   	known, or `nil` if that information is not applicable.
```
stmontgomery, line 227: Funky whitespace

grynspan: That's what I get for using my personal computer.

Sources/Testing/Events/Event.swift
```diff
@@ -168,6 +168,28 @@ public struct Event: Sendable {
   /// The instant at which the event occurred.
   public var instant: Test.Clock.Instant
 
+  /// Storage for ``sourceLocation``.
+  private var _sourceLocation: SourceLocation?
```
stmontgomery, line 172: I'm a little wary of the perf implications of this additional storage on `Event`.

I see you propagated the new `sourceLocation:` parameter on `.expectationChecked` events, but the `Expectation` type already has a non-Optional source location property, so I think there should already be a value for those particular events to use.

I'm wondering if we really need this, or could we instead propagate the source location of an attachment via the payload of the `.valueAttachmentFailed` enum case? Alternatively, perhaps this could be a new property on `Event.Context`?

grynspan: Actually, I can just go ahead and put the storage back on `Attachment` and remove it from `Event`.

grynspan:
@swift-ci test

PR #815, opened 2024-11-09
Fix how we check if a range expression covers one value.

grynspan:
My previous fix, #806, had the wrong logic to determine if a confirmation's expected count as-a-`RangeExpression` had exactly one value, so we were printing sub-optimal messages on test failures. Specifically, this:

```swift
await confirmation(expectedCount: 1) { // transformed to 1...1
  // fail to confirm
}
```

Would report "1...1 time(s)" instead of "1 time".

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

PR #819, opened 2024-11-11
Add support for attachments to the Foundation cross-import overlay.

grynspan:
This PR adds experimental support for attachments to some types in Foundation via the (non-functional) cross-import overlay. @stmontgomery is working on setting up said overlay so that it can actually be used; until then, the changes here are speculative only.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+Encodable&NSSecureCoding.swift
```diff
@@ -0,0 +1,22 @@
...
+//
+
+#if canImport(Foundation)
+@_spi(Experimental) public import Testing
+public import Foundation
+
+@_spi(Experimental)
+extension Attachable where Self: Encodable & NSSecureCoding {
+  @_documentation(visibility: private)
+  public func withUnsafeBufferPointer<R>(for attachment: borrowing Attachment<Self>, _ body: (UnsafeRawBufferPointer) throws -> R) throws -> R {
```
stmontgomery, line 18: Is it worth adding a (non-documentation) comment explaining why this is needed? It's not obvious to me

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+Encodable.swift
```diff
@@ -0,0 +1,97 @@
...
+    data = try plistEncoder.encode(attachableValue)
+  case .default:
+    // The default format is JSON.
+    fallthrough
+  case .json:
+    // We cannot use our own JSON encoding wrapper here because that would
+    // require it be exported with (at least) package visibility which would
+    // create a visible external dependency on Foundation in the main testing
+    // library target.
+    data = try JSONEncoder().encode(attachableValue)
```
stmontgomery, line 46: (Not for this PR, but for consideration as work progresses, in case you hadn't already thought of this…)

I think it'd be nice if there were some affordance which allowed users to _optionally_ customize the `JSONEncoder` and modify its encoding options.

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+NSSecureCoding.swift
```diff
@@ -0,0 +1,78 @@
...
+  ///
+  /// The testing library uses this function when writing an attachment to a
+  /// test report or to a file on disk. The encoding used depends on the path
+  /// extension specified by the value of `attachment`'s ``Testing/Attachment/preferredName``
+  /// property:
+  ///
+  /// | Extension | Encoding Used | Encoder Used |
+  /// |-|-|-|
+  /// | `".xml"` | XML property list | [`NSKeyedArchiver`](https://developer.apple.com/documentation/foundation/nskeyedarchiver) |
+  /// | None, `".plist"` | Binary property list | [`NSKeyedArchiver`](https://developer.apple.com/documentation/foundation/nskeyedarchiver) |
```
stmontgomery, line 42: Not sure whether it's correct per se, but I've certainly seen a lot of XML property lists which use the `.plist` file extension. From a user's perspective, to make inspecting the contents easier with my chosen text editor I definitely prefer XML. (Of course, binary property lists have their place too—primarily when meant for machine consumption.)

I'm wondering whether we really _need_ to have the `.plist` extension encode as binary, or whether it could use XML too? I'm trying to optimize this for consumption by human test authors. I realize you can just switch to `.xml` as a workaround, but it seems easily to forget or overlook, and people may prefer `.plist`.

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
```diff
@@ -0,0 +1,161 @@
...
+          }
+        }
+      } catch {
+        continuation.resume(throwing: error)
+      }
+    }
+
+    try self.init(contentsOf: temporaryURL, options: [.mappedIfSafe])
+#else
+    throw CocoaError(.featureUnsupported, userInfo: [NSLocalizedDescriptionKey: "This platform does not support attaching directories to tests."])
```
stmontgomery, line 156: Slightly pedantic, but it's only the creation of the attachment which is unsupported here, not _attaching_ it to a test, right?

```suggestion
    throw CocoaError(.featureUnsupported, userInfo: [NSLocalizedDescriptionKey: "This platform does not support creating attachments with the contents of a directory."])
```

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+NSSecureCoding.swift
```diff
@@ -0,0 +1,78 @@
...
+      // NSKeyedArchiver.archivedData(withRootObject:requiringSecureCoding:)
+      // that is Swift-safe (throws errors instead of exceptions) and lets the
+      // caller specify the output format. Work around this issue by decoding
+      // the archive re-encoding it manually.
+      if propertyListFormat != .binary {
+        let plist = try PropertyListSerialization.propertyList(from: data, format: nil)
+        data = try PropertyListSerialization.data(fromPropertyList: plist, format: propertyListFormat, options: 0)
+      }
+    case .json:
+      throw CocoaError(.propertyListWriteInvalid, userInfo: [NSLocalizedDescriptionKey: "An instance of \(type(of: self)) cannot be encoded as JSON. Specify a property list format instead."])
```
stmontgomery, line 72: Do you expect anyone to encounter the code path of this `throw` in practice? (Not requesting a change, just for my understanding)

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
```diff
@@ -0,0 +1,161 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if canImport(Foundation)
+@_spi(Experimental) @_spi(ForSwiftTestingOnly) public import Testing
+public import Foundation
+
+#if SWT_TARGET_OS_APPLE && canImport(UniformTypeIdentifiers)
```
stmontgomery, line 15: Is `#if SWT_TARGET_OS_APPLE` alone sufficient? Although we do need runtime availability checks, the `UniformTypeIdentifiers` module should always be available at build-time on Apple platforms, right?

```suggestion
#if SWT_TARGET_OS_APPLE
```

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
```diff
@@ -0,0 +1,161 @@
...
+      preferredName = {
+#if SWT_TARGET_OS_APPLE && canImport(UniformTypeIdentifiers)
+        if #available(_uttypesAPI, *), let tgz = UTType.tgz {
+          return (preferredName as NSString).appendingPathExtension(for: tgz)
+        }
+#endif
+        return (preferredName as NSString).appendingPathExtension("tgz") ?? preferredName
+      }()
+
+      try await self.init(Data(compressedContentsOfDirectoryAt: url), named: preferredName, sourceLocation: sourceLocation)
```
stmontgomery, line 93: Am I interpreting correctly that for directories, it unconditionally _compresses_ the directory's contents? Could we make the compression step an optional parameter, but enabled by default? (Thinking of scenarios where you're attaching a directory containing files which are already highly compressed.) Maybe disabling that would use `.tar` instead of `.tgz`?

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+Encodable&NSSecureCoding.swift
```diff
@@ -0,0 +1,22 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if canImport(Foundation)
+@_spi(Experimental) public import Testing
+public import Foundation
+
+@_spi(Experimental)
+extension Attachable where Self: Encodable & NSSecureCoding {
+  @_documentation(visibility: private)
```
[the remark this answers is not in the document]
grynspan: Duplicate initializer that only exists to disambiguate for the compiler.

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+NSSecureCoding.swift
grynspan: If you say `named: "something.json"` you can wind up here, yeah.

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
grynspan: This is consistent with what XCTest does here: https://developer.apple.com/documentation/xctest/xctattachment/2887210-init

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
grynspan: No comment. :)

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+NSSecureCoding.swift
grynspan: This choice is intentional as there is no other way to specify which subformat of property list you want. A future PR will introduce a `contentType` property on Apple platforms (the non-Apple platform situation is TBD) that we would then use to refine this logic.

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+Encodable.swift
grynspan: That'll be part of the `Metadata` PR.

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+Encodable&NSSecureCoding.swift
grynspan: https://github.com/swiftlang/swift-testing/blob/6f204f56452cabf69e2c759f6083d8d9ea76632b/Sources/Overlays/_Testing_Foundation/Attachments/Attachable%2BEncodable.swift#L83

grynspan: Will add an additional comment here.

Package.swift
```diff
@@ -96,6 +96,7 @@ let package = Package(
       dependencies: [
```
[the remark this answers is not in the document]
grynspan: `// Cross-import overlays (not supported by Swift Package Manager)`

grynspan:
@swift-ci test

Sources/Overlays/_Testing_Foundation/Attachments/Attachable+NSSecureCoding.swift
stmontgomery: Ok great

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
stmontgomery: This can be for a future PR, but I'd like to request we revisit that and consider making it configurable. In general, "XCTest did it this way" isn't the most compelling justification 😛

stmontgomery (approved):

PR #820, opened 2024-11-12
Defer the recursive propagation of suite traits until after filtering

stmontgomery:
This PR includes some optimizations to the testing library's planning logic (in `Runner.Plan`) to avoid unnecessary work in common cases, and remove unnecessary nodes from the test graph once they're no longer needed.

### Motivation

The testing library's planning phase includes taking all runtime-discovered `Test` instances and organizing them into a tree-like data structure using the internal `Graph` type. It traverses this graph several times to do things like only run certain tests the user indicated (by applying a filter), propagate eligible traits from suites to their children (by recursively applying traits), and implement other features.

Currently, the step of recursively propagating eligible suite traits to their children must occur _before_ applying a test filter. This is because a test filter may require knowledge of trait information -- specifically, the `.tags` and `.hidden` traits. Even though test filtering only needs these specific traits, the planning phase currently propagates _all_ recursive suite traits down to _all_ tests, even though filtering may end up removing the majority of tests from the graph entirely. This can lead to wasted effort for any tests which were filtered out, and I anticipate this may become more pronounced in the future if/when we enhance the trait system to support things like consolidating same-typed traits with children.

Trait propagation is done by inserting all inherited traits from the parent at the beginning of the child's trait list (to ensure deterministic ordering), so this requires shifting the elements in an array. And this work is only needed if the test filter being applied includes a query of tests' `.tags` trait. (The `.hidden` trait is semantically simpler and easier to handle: it's recursive, but it's a simple Boolean flag which cannot be overridden by children, so it doesn't need to be "merged" with children.)

There is an opportunity to optimize this flow and completely skip trait propagation in the common case where a test filter doesn't care about tags, and allow propagation to happen after the filtering stage. Additionally, to reduce memory usage and simplify the graph for easier debugging, the filtering logic could be modified to "prune" the graph by removing branches which have zero children.

### Modifications

- Add a property to `Configuration.TestFilter` indicating whether the filter requires knowledge of trait information.
- Consult this property when applying a filter to a test graph. When necessary, temporarily convert the passed-in graph to a modified graph and recursively propagate tags to it.
- Modify `Configuration.TestFilter` to prune unnecessary children from the graph before finishing.
- Modify planning logic in `Runner.Plan` to move the recursive trait propagation step to after filering.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Configuration.TestFilter.swift
+++ b/Sources/Testing/Running/Configuration.TestFilter.swift
@@ -317,7 +317,7 @@ extension Configuration.TestFilter.Operation {
         lhs.apply(to: testGraph),
         rhs.apply(to: testGraph)
       ).mapValues { _, value in
-        op.functionValue()(value.0, value.1)
+        op(value.0, value.1)
       }
     }
   }
@@ -334,7 +334,7 @@ extension Configuration.TestFilter {
   func apply(to testGraph: Graph<String, Test?>) throws -> Graph<String, Test?> {
     var result: Graph<String, Test?>
 
-    if _kind.requiresFilterItemConversion {
+    if _kind.requiresTraitPropagation {
       // Convert the specified test graph to a graph of filter items temporarily
       // while performing filtering, and apply inheritance for the properties
       // which are relevant when performing filtering (e.g. tags).
@@ -427,21 +427,23 @@ extension Configuration.TestFilter {
     /// This operator is equivalent to `||`.
     case or
 
-    /// The equivalent of this instance as a callable function.
-    fileprivate func functionValue<T>() -> @Sendable (T?, T?) -> T? where T: _FilterableItem {
+    /// Evaluate this combination operator with two optional operands.
+    ///
+    /// - Parameters:
+    ///   - lhs: The left-hand argument
+    ///   - rhs: The right-hand argument.
+    ///
+    /// - Returns: The combined result of evaluating this operator.
+    fileprivate func callAsFunction<T>(_ lhs: T?, _ rhs: T?) -> T? where T: _FilterableItem {
       switch self {
       case .and:
-        return { lhs, rhs in
-          if lhs != nil && rhs != nil {
-            lhs
-          } else {
-            nil
-          }
+        if lhs != nil && rhs != nil {
+          lhs
+        } else {
+          nil
         }
       case .or:
-        return { lhs, rhs in
-          lhs ?? rhs
-        }
+        lhs ?? rhs
       }
     }
   }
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -365,7 +428,7 @@ extension Configuration.TestFilter {
     case or
 
     /// The equivalent of this instance as a callable function.
```
grynspan, line 430: Nitpick: add more markup here now that it's a function.

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -422,3 +485,60 @@ extension Configuration.TestFilter {
...
+
+extension Configuration.TestFilter.Kind {
+  /// Whether this kind of test filter requires knowledge of test traits.
+  ///
+  /// If the value of this property is `true`, the values of a test graph must
+  /// be converted to `FilterItem` and have trait information recursively
+  /// propagated before the filter can be applied, or else the results may be
+  /// inaccurate. This facilitates a performance optimization where trait
+  /// propagation can be skipped for filters which don't require such knowledge.
+  fileprivate var requiresFilterItemConversion: Bool {
```
grynspan, line 499: I don't have a better name for this property, but I wonder if we can come up with one that doesn't just describe its own technical effects. `requiresRecursiveBlahBlah` or somethin'.

(Not critical for this PR, but worth bikeshedding if we think this property or pattern is going to be useful in other places too.)

Sources/Testing/Running/Configuration.TestFilter.swift
```diff
@@ -422,3 +485,60 @@ extension Configuration.TestFilter {
...
+    self
+  }
+}
+
+/// An item representing a test and its filtering-related properties.
+///
+/// Instances of this type are needed when applying a test graph to a kind of
+/// filter for which the value of the `requiresFilterItemConversion` property
+/// is `true`.
+fileprivate struct FilterItem: _FilterableItem {
```
grynspan, line 535: How is this type fundamentally different from just using a test directly, other than caching `tags`?

grynspan (approved):

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: Oh, I can enhance it to use `callAsFunction()`!

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: Renamed to `requiresTraitPropagation`. Not going to sweat this one too much further though, especially since it's `fileprivate`

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Configuration.TestFilter.swift
stmontgomery: It doesn't just _cache_ tags, it collects all the tags for a test including those inherited from parents.

PR #821, opened 2024-11-13
Hoist `Attachment` and `Attachable` out of `Test`.

grynspan:
I initially placed these types inside `Test` out of concern that they could conflict with symbols in second- or third-party modules that are being tested. After discussing with my colleagues, we've decided to hoist the types up to the top level of the Testing module.

There's no general solution for name conflicts in Swift other than providing the module name at point-of-use, and by and large the language, toolchain, and Apple's frameworks just sort of accept this possibility? So I was maybe overthinking it. If we run into trouble down the line before release, we can reevaluate this decision.

### Checklist:


grynspan:
Preemptively declaring this as reviewed by @briancroom and @stmontgomery as they already signed off on these renames privately. 😈

grynspan:
@swift-ci test

PR #824, opened 2024-11-14
[WIP] Add an associated `AttachmentMetadata` type to `Attachable`.

grynspan:
This PR adds a new associated type to `Attachable` that can be used to supply additional metadata to an attachment that is specific to that type. Metadata is always optional and the default type is `Never` (i.e. by default there is no metadata.)

The `Encodable` and `NSSecureCoding` conformances in the Foundation cross-import overlay have been updated such that this type equals a new structure that describes the format to use as well as options to pass to the JSON encoder (if one is used) and user info to pass to the plist or JSON encoders (if used.) We must use this type even for types that conform only to `NSSecureCoding`, otherwise we get compile-time errors about the type being ambiguous if a type conforms to both protocols and to `Attachable`.

### Checklist:

### DIFF
--- a/Sources/Testing/Attachments/Attachment.swift
+++ b/Sources/Testing/Attachments/Attachment.swift
@@ -26,6 +26,12 @@ public struct Attachment<AttachableValue>: ~Copyable where AttachableValue: Atta
   /// Storage for ``attachableValue-7dyjv``.
   fileprivate var _attachableValue: AttachableValue
 
+  /// Metadata associated with this attachment.
+  ///
+  /// The value of this property is `nil` if you passed `nil` when initializing
+  /// this instance, or if you set it to `nil` at a later time.
+  public var metadata: AttachableValue.AttachmentMetadata?
+
   /// The path to which the this attachment was written, if any.
   ///
   /// If a developer sets the ``Configuration/attachmentsPath`` property of the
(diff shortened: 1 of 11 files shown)


grynspan:
@swift-ci test

Sources/Testing/Attachments/Attachment.swift
```diff
@@ -23,6 +23,12 @@ public struct Attachment<AttachableValue>: ~Copyable where AttachableValue: Atta
   /// Storage for ``attachableValue-7dyjv``.
   fileprivate var _attachableValue: AttachableValue
 
+  /// Metadata associated with this attachment.
+  ///
+  /// The value of this property is `nil` if you passed `nil` when initializing
+  /// this instance, or if you set it to `nil` at a later time.
+  public var metadata: AttachableValue.AttachmentMetadata?
+
   /// The path to which the this attachment was written, if any.
```
plemarquand, line 32:
```suggestion
  /// The path to which this attachment was written, if any.
```

Tests/TestingTests/AttachmentTests.swift
```diff
@@ -424,6 +431,53 @@ struct AttachmentTests {
...
     }
   }
+
+  @Test("Attach Codable-conformant type with metadata")
+  func codableWithMetadata() async throws {
+    let attachableValue = MyCodableAttachable(string: "abc123")
+    let attachment = Attachment(
+      attachableValue,
+      metadata: .init(
+        format: .propertyListFormat(.xml)
```
plemarquand, line 441: Could you add a similar test that checks the happy path of .json metadata serialization

grynspan:
Not planning to move forward with this feature.

PR #827, opened 2024-11-16
Add a CoreGraphics cross-import overlay with support for attaching `CGImage`s.

grynspan:
This PR adds a new cross-import overlay target with Apple's Core Graphics framework that allows attaching a `CGImage` as an attachment in an arbitrary image format (PNG, JPEG, etc.)

Because `CGImage` is imported into Swift as a non-final class, it cannot conform directly to `Attachable`, so an `AttachableContainer` type acts as a proxy. This type is not meant to be used directly, so its name is underscored. Initializers on `Attachment` are provided so that this abstraction is almost entirely transparent to test authors.

A new protocol, `AttachableAsCGImage`, is introduced that abstracts away the relationship between the attached image and Core Graphics; in the future, I intend to make additional image types like `NSImage` and `UIImage` conform to this protocol too.

Example usage:

```swift
let sparklyDiamonds: CGImage = ...
let attachment = Attachment(image, named: "sparkly-diamonds", as: .tiff, encodingQuality: 0.75)
...
attachment.attach()
```

The code in this PR is, by definition, specific to Apple's platforms. In the future, I'd be interested in adding Windows/Linux equivalents (`HBITMAP`? Whatever Gnome/KDE/Qt use?) but that's beyond the scope of this PR.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -51,6 +51,7 @@ let package = Package(
       name: "TestingTests",
       dependencies: [
         "Testing",
+        "_Testing_CoreGraphics",
         "_Testing_Foundation",
       ],
       swiftSettings: .packageSettings
@@ -91,6 +92,14 @@ let package = Package(
     ),
 
     // Cross-import overlays (not supported by Swift Package Manager)
+    .target(
+      name: "_Testing_CoreGraphics",
+      dependencies: [
+        "Testing",
+      ],
+      path: "Sources/Overlays/_Testing_CoreGraphics",
+      swiftSettings: .packageSettings
+    ),
     .target(
       name: "_Testing_Foundation",
       dependencies: [
--- a/Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
+++ b/Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
@@ -39,7 +39,7 @@ extension Attachment {
   fileprivate init<T>(
     attachableValue: T,
     named preferredName: String?,
-    as contentType: (any Sendable)?,
+    contentType: (any Sendable)?,
     encodingQuality: Float,
     sourceLocation: SourceLocation
   ) where AttachableValue == _AttachableImageContainer<T> {
@@ -111,7 +111,7 @@ extension Attachment {
     encodingQuality: Float = 1.0,
     sourceLocation: SourceLocation = #_sourceLocation
   ) where AttachableValue == _AttachableImageContainer<T> {
-    self.init(attachableValue: attachableValue, named: preferredName, as: contentType, encodingQuality: encodingQuality, sourceLocation: sourceLocation)
+    self.init(attachableValue: attachableValue, named: preferredName, contentType: contentType, encodingQuality: encodingQuality, sourceLocation: sourceLocation)
   }
 
   /// Initialize an instance of this type that encloses the given image.
@@ -141,7 +141,7 @@ extension Attachment {
     encodingQuality: Float = 1.0,
     sourceLocation: SourceLocation = #_sourceLocation
   ) where AttachableValue == _AttachableImageContainer<T> {
-    self.init(attachableValue: attachableValue, named: preferredName, as: nil, encodingQuality: encodingQuality, sourceLocation: sourceLocation)
+    self.init(attachableValue: attachableValue, named: preferredName, contentType: nil, encodingQuality: encodingQuality, sourceLocation: sourceLocation)
   }
 }
 #endif
--- a/Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
+++ b/Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
@@ -1,0 +1,147 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if SWT_TARGET_OS_APPLE && canImport(CoreGraphics)
+@_spi(ForSwiftTestingOnly) @_spi(Experimental) public import Testing
+
+public import UniformTypeIdentifiers
+
+extension Attachment {
+  /// Initialize an instance of this type that encloses the given image.
+  ///
+  /// - Parameters:
+  ///   - attachableValue: The value that will be attached to the output of
+  ///     the test run.
+  ///   - preferredName: The preferred name of the attachment when writing it
+  ///     to a test report or to disk. If `nil`, the testing library attempts
+  ///     to derive a reasonable filename for the attached value.
+  ///   - contentType: The image format with which to encode `attachableValue`.
+  ///     If this type does not conform to [`UTType.image`](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype-swift.struct/image),
+  ///     the result is undefined. Pass `nil` to let the testing library decide
+  ///     which image format to use.
+  ///   - encodingQuality: The encoding quality to use when encoding the image.
+  ///     If the image format used for encoding (specified by the `contentType`
+  ///     argument) does not support variable-quality encoding, the value of
+  ///     this argument is ignored.
+  ///   - sourceLocation: The source location of the call to this initializer.
+  ///     This value is used when recording issues associated with the
+  ///     attachment.
+  ///
+  /// This is the designated initializer for this type when attaching an image
+  /// that conforms to ``AttachableAsCGImage``.
+  fileprivate init<T>(
+    attachableValue: T,
+    named preferredName: String?,
+    as contentType: (any Sendable)?,
+    encodingQuality: Float,
+    sourceLocation: SourceLocation
+  ) where AttachableValue == _AttachableImageContainer<T> {
+    var imageContainer = _AttachableImageContainer(image: attachableValue, encodingQuality: encodingQuality)
+
+    // Update the preferred name to include an extension appropriate for the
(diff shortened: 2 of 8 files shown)


grynspan:
@swift-ci test

Package.swift
```diff
@@ -51,6 +51,7 @@ let package = Package(
       name: "TestingTests",
       dependencies: [
         "Testing",
+        "_Testing_CoreGraphics",
```
stmontgomery, line 54: It can be done via a separate PR, but just a reminder to add this new target/module to CMake too, following the recent pattern for the Foundation cross-import overlay.

Sources/Overlays/_Testing_CoreGraphics/Attachments/ImageAttachmentError.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if SWT_TARGET_OS_APPLE && canImport(CoreGraphics)
+/// A type representing errors that can occur when attaching an image.
```
stmontgomery, line 12:
```suggestion
/// A type representing an error that can occur when attaching an image.
```

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
```diff
@@ -0,0 +1,163 @@
...
+///
+/// Three reasons:
+///
+/// 1. Several image classes are not marked `Sendable`, which means that as far
+///    as Swift is concerned, they cannot be safely passed to Swift Testing's
+///    event handler (primarily because `Event` is `Sendable`.) So we would have
+///    to eagerly serialize them, which is unnecessarily expensive if we know
+///    they're actually concurrency-safe.
+/// 2. We would have no place to store metadata such as the encoding quality
+/// 	 (although in the future we may introduce a "metadata" associated type to
```
stmontgomery, line 28: nit: funky indentation

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
```diff
@@ -0,0 +1,163 @@
...
+/// attached indirectly.
+///
+/// You do not need to use this type directly. Instead, initialize an instance
+/// of ``Attachment`` using an instance of an image type that conforms to
+/// ``AttachableAsCGImage``. The following system-provided image types conform
+/// to the ``AttachableAsCGImage`` protocol and can be attached to a test:
+///
+/// - [`CGImage`](https://developer.apple.com/documentation/coregraphics/cgimage)
+@_spi(Experimental)
+public struct _AttachableImageContainer<ImageClass>: Sendable where ImageClass: AttachableAsCGImage {
```
stmontgomery, line 51: Any reason "Class" needs to be in the generic type parameter `ImageClass`, or could it just be `Image`?

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
```diff
@@ -0,0 +1,163 @@
...
+
+/// A wrapper type for image types such as `CGImage` and `NSImage` that can be
+/// attached indirectly.
+///
+/// You do not need to use this type directly. Instead, initialize an instance
+/// of ``Attachment`` using an instance of an image type that conforms to
+/// ``AttachableAsCGImage``. The following system-provided image types conform
+/// to the ``AttachableAsCGImage`` protocol and can be attached to a test:
+///
+/// - [`CGImage`](https://developer.apple.com/documentation/coregraphics/cgimage)
```
stmontgomery, line 49: I see the protocol docs for `AttachableAsCGImage` list the known conforming types, maybe this should just link there rather than repeating it here?

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
```diff
@@ -0,0 +1,163 @@
...
+//
+
+#if SWT_TARGET_OS_APPLE && canImport(CoreGraphics)
+@_spi(Experimental) public import Testing
+private import CoreGraphics
+
+private import ImageIO
+import UniformTypeIdentifiers
+
+/// ## Why can't images directly conform to Attachable?
```
stmontgomery, line 18: Silly question, but is there any benefit to this being a `///` (doc comment) instead of `//` since it's not attached to any declaration? Very useful comment, regardless

Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
```diff
@@ -0,0 +1,147 @@
...
+  ///
+  /// - [`CGImage`](https://developer.apple.com/documentation/coregraphics/cgimage)
+  @_spi(Experimental)
+  public init<T>(
+    _ attachableValue: T,
+    named preferredName: String? = nil,
+    encodingQuality: Float = 1.0,
+    sourceLocation: SourceLocation = #_sourceLocation
+  ) where AttachableValue == _AttachableImageContainer<T> {
+    self.init(attachableValue: attachableValue, named: preferredName, as: nil, encodingQuality: encodingQuality, sourceLocation: sourceLocation)
```
stmontgomery, line 144: The `as: nil` argument here is a little unclear, which I guess is because the `contentType` property is Optional. I wonder if it'd be more clearly labeled `contentType:`?

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
grynspan: I think at one point I had it constrained to `AnyObject`.

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
grynspan: Just for the formatting.

Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
grynspan: Sure, can rename that one. It's an internal-only initializer.

Package.swift
grynspan: Will address separately.

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
grynspan: Going to leave it as-is. If somebody is reading the documentation for this type, they're presumably looking in the source, not the compiled DocC archive, so linking around is harder.

grynspan:
@swift-ci test

Sources/Overlays/_Testing_CoreGraphics/Attachments/ImageAttachmentError.swift
```diff
@@ -0,0 +1,39 @@
...
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if SWT_TARGET_OS_APPLE && canImport(CoreGraphics)
+/// A type representing an error that can occur when attaching an image.
+package enum ImageAttachmentError: Error, CustomStringConvertible {
```
grynspan, line 13: Needs to be SPI for now.

Sources/Overlays/_Testing_CoreGraphics/Attachments/ImageAttachmentError.swift
```diff
@@ -0,0 +1,39 @@
...
+  /// The image could not be converted to an instance of `CGImage`.
+  case couldNotCreateCGImage
+
+  /// The image destination could not be created.
+  case couldNotCreateImageDestination
+
+  /// The image could not be converted.
+  case couldNotConvertImage
+
+  package var description: String {
```
grynspan, line 26: Needs to be SPI.

Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
```diff
@@ -0,0 +1,147 @@
...
+  /// The following system-provided image types conform to the
+  /// ``AttachableAsCGImage`` protocol and can be attached to a test:
+  ///
+  /// - [`CGImage`](https://developer.apple.com/documentation/coregraphics/cgimage)
+  @_spi(Experimental)
+  @available(_uttypesAPI, *)
+  public init<T>(
+    _ attachableValue: T,
+    named preferredName: String? = nil,
+    as contentType: UTType?,
```
stmontgomery, line 110:
```suggestion
    contentType: UTType?,
```

grynspan: No, that one's on purpose so that it reads when used:

```swift
Attachment(something, as: .png)
```

Which is consistent with other uses of `UTType` in Apple API.

stmontgomery: It's just that it can be unclear when it's `as: nil`. Do the other Apple APIs accept Optional `UTType` values?

grynspan: Yes, but also there's an overload that doesn't take a content type at all, and I'd expect people to use that instead of a literal `nil`. The optionality here is so that you can pass in an optional value rather than needing to manually branch with `if`.

PR #828, opened 2024-11-17
Documentation: Update CMake.md to use the ABI entry point

ADKaster:
The SwiftPM entry point is unstable and the new ABIv0 entry point has already been added to the library.

### Motivation:

Using the SwiftPM entry point when building tests from a CMake project as recommended in the documentation is outdated and unwise.

### Modifications:

Replace the example with one using the new ABIv0 entry point.

### Result:

CMake projects should stop relying on the SwiftPM entry point.

### Checklist:


ADKaster:
cc @compnerd

ADKaster:
Hm. Chatting with Jonathan on Slack some more, it seems like the fact that this works with the Swift OSS toolchain on linux is a toolchain bug: we shouldn't be able to access the private module from user code.

Another alternative (with -enable-experimental-feature Extern) looks like this:

```swift
import Foundation

typealias EntryPoint = @convention(thin) @Sendable (_ configurationJSON: UnsafeRawBufferPointer?, _ recordHandler: @escaping @Sendable (_ recordJSON: UnsafeRawBufferPointer) -> Void) async throws -> Bool

@_extern(c, "swt_abiv0_getEntryPoint")
func swt_abiv0_getEntryPoint() -> UnsafeRawPointer

@main struct Runner {
    static func main() async throws {
        let configurationJSON: UnsafeRawBufferPointer? = nil
        let recordHandler: @Sendable (UnsafeRawBufferPointer) -> Void = { _ in }

        let entryPoint = unsafeBitCast(swt_abiv0_getEntryPoint(), to: EntryPoint.self)

        if try await entryPoint(configurationJSON, recordHandler) {
            exit(EXIT_SUCCESS)
        } else {
            exit(EXIT_FAILURE)
        }
    }
}
```

Documentation/CMake.md
```diff
@@ -59,21 +59,28 @@ endif()
 ## Add an entry point
 
 You must include a source file in your test executable target with a
-`@main` entry point. The following example uses the SwiftPM entry point:
+`@main` entry point.
 
 ```swift
-import Testing
+import Foundation
+@_spi(ForToolsIntegrationOnly) import Testing
```
grynspan, line 66: So, yes-and-no. As we talked about in Slack, it's a bug that any `@_spi` content is available in the Swift toolchain right now and it's going to get removed in the future.

For the moment, code should either use `dlsym()/`GetProcAddress()` or, if a little more adventurous, use the experimental `@_extern(c)` to call `swt_abiv0_getEntryPoint()`, then load the SwiftCC entry point from there. (`@_silgen_name` could be used too, but technically the calling conventions are incompatible, so _caveat codor_.)

We're still looking at how to expose these symbols in a way that makes it easy for test tools/harness authors but doesn't make the symbols visible to test authors by default. We're looking at adding a dedicated target in the package (if we can do so without requiring building the whole library from source) or creating a separate `swift-testing-tools-support` package, among other options.

Documentation/CMake.md
```diff
@@ -59,21 +59,28 @@ endif()
...
 
 @main struct Runner {
-    static func main() async {
-        await Testing.__swiftPMEntryPoint() as Never
+    static func main() async throws {
+        let configurationJSON: UnsafeRawBufferPointer? = nil
```
grynspan, line 70: You can leave this line out and just pass `nil` if you're not actually going to use it. :)

grynspan:
@ADKaster Did you want to proceed with this PR?

Documentation/CMake.md
ADKaster: I've replaced the spi call with an explicit extern(c) for the symbol. I mentioned in a note above the example code that you could do the same with a C header file and a module.modulemap to fwd declare it.

Documentation/CMake.md
ADKaster: Not quite, even when just passing nil Swift 6 gives me an isolation warning. marking this as `nonisolated(unsafe)` makes that go away.

ADKaster:
@grynspan I've updated the PR to address your two comments and other discussions we had on slack. I'm not 100% sure on the formatting, but the example code as-is does work.

Documentation/CMake.md
grynspan: Calloo callay. Okay, that's fine then.

Documentation/CMake.md
```diff
@@ -59,21 +59,36 @@ endif()
...
 
 You must include a source file in your test executable target with a
-`@main` entry point. The following example uses the SwiftPM entry point:
+`@main` entry point. The example main below requires the experimental
+`Extern` feature. The declaration of `swt_abiv0_getEntryPoint` could
+also be written in a C header file with its own `module.modulemap`.
 
 ```swift
-import Testing
+import Foundation
```
grynspan, line 67: Shouldn't need this import for the barebones implementation below. Only needed if you want to use `JSONEncoder`/`JSONDecoder`.

ADKaster: Is it not needed for `exit(EXIT_SUCCESS)`? Or is there a non-foundation way to do the standard Unix process exit dance.

grynspan: `exit()` and `EXIT_SUCCESS` are defined in the system's C module, which may be `Darwin`, `Glibc`, `ucrt`… uh… `Musl`?

Foundation reexports the C module for the current system. It seems wrong to me (but I can be convinced otherwise) to tell developers to include a full Foundation dependency just for `exit()`.

Depending on the dev's specific requirements, they really just need the process to terminate with zero (the default!) or non-zero. If they're using Swift Argument Parser, it has built-in functionality for this. If they're using _just_ the Swift stdlib and nothing else, they could call `fatalError()` on failure. I don't know what's the best choice for this example code.

ADKaster: Eh.. I'm not sure that I know what the best choice is either. When a test binary is executed by ctest, as in my use case, returning 0 or 1 from main() is important for the test harness to understand whether the test passed or failed. Anything more than that on the ctest side is up to whatever @etcwilde comes up with for https://gitlab.kitware.com/cmake/cmake/-/issues/26452

I feel like adding a 
```
#if canImport(GlibC)
#import GlibC
#elseif canImport(YourCLibModule)
#import YourCLibModule
#endif
```
chain to the top of the example would get in the way of what it's actually trying to show that you need :shrug: 

I'll change it to whatever anyone feels strongly about though, as in theory more comprehensive support for the CMake use case could come soon :tm: to make this example obsolete-ish.

grynspan: It may be sufficient to just show `exit(EXIT_SUCCESS)` without importing anything. Developers who are going down this path are presumably skilled enough to figure out how to make it compile…? 😇

ADKaster: Sure... "compilable example code left as an exercise for the reader" is definitely one approach. Not my first choice, but it's not my library :). I've pushed up that change then.

grynspan: Again, anybody who's digging down into the CMake scripts is on their own and I have to assume has some esoteric environment—we can't even guarantee Foundation is present! (We _can_ guarantee libc is present because we won't build without it, at least.)

Documentation/CMake.md
```diff
@@ -59,21 +59,34 @@ endif()
...
+        }
     }
 }
 ```
 
-> [!WARNING]
-> The entry point is expected to change to an entry point designed for other
-> build systems prior to the initial stable release of Swift Testing.
+For more information on the input configuration and output records of the ABI entry
+point, refer to the [ABI documentation](ABI/JSON.md)
```
stmontgomery, line 89: nit
```suggestion
point, refer to the [ABI documentation](ABI/JSON.md).
```

stmontgomery (approved):

Documentation/CMake.md
```diff
@@ -59,21 +59,34 @@ endif()
...
+        nonisolated(unsafe) let configurationJSON: UnsafeRawBufferPointer? = nil
+        let recordHandler: @Sendable (UnsafeRawBufferPointer) -> Void = { _ in }
+
+        let entryPoint = unsafeBitCast(swt_abiv0_getEntryPoint(), to: EntryPoint.self)
+
+        if try await entryPoint(configurationJSON, recordHandler) {
+            exit(EXIT_SUCCESS)
+        } else {
+            exit(EXIT_FAILURE)
+        }
```
stmontgomery, line 83: The entry point function itself will only throw an error if some problem was encountered during the act of running the tests. (An example of this might be a system error encountered while discovering or planning the set of tests to run.) It will _not_ throw an error if a test successfully ran, but failed.

Typically, a test runner executable is expected to exit with a nonzero status if at least one test failed, so ideally it would be better for this logic to implement a real record handler, interpret the records it receives, detect if one represents an issue, and set a flag whenever at least one issue is received so it can `exit(EXIT_FAILURE)`.

I realize that's more complicated to document here, and we don't necessarily need to go that far in this PR. But I do think it would be helpful to at least leave a code comment or something indicating that a well-behaved test runner executable should implement such logic.

stmontgomery (changes requested):
I'd like to see at least a code comment related to failure tracking leading to a non-zero exit status. Bonus points for actually implementing that logic in the example code!

Documentation/CMake.md
grynspan: It does return `false` on failure-during-testing though, if any unknown issues were recorded. :)

stmontgomery: Great, that addresses my concern

stmontgomery (approved):

grynspan:
@ADKaster Can you update us on the state of this PR?

Documentation/CMake.md
```diff
@@ -59,21 +59,34 @@ endif()
 ## Add an entry point
 
 You must include a source file in your test executable target with a
-`@main` entry point. The following example uses the SwiftPM entry point:
+`@main` entry point. The example main below requires the experimental
+`Extern` feature. The declaration of `swt_abiv0_getEntryPoint` could
+also be written in a C header file with its own `module.modulemap`.
```
grynspan, line 64: Since the function is exported, a developer could also use `dlsym()` or `GetProcAddress()` or equivalent, so a note about that is worthwhile (the implementation does not need to be provided here.)

ADKaster: Though actually, one can never go wrong with a POSIX.1-2024 reference https://pubs.opengroup.org/onlinepubs/9799919799/functions/dlsym.html

Documentation/CMake.md
```diff
@@ -59,21 +59,38 @@ endif()
 ## Add an entry point
 
 You must include a source file in your test executable target with a
-`@main` entry point. The following example uses the SwiftPM entry point:
+`@main` entry point. The example main below requires the experimental
+`Extern` feature. The function `swt_abiv0_getEntryPoint` is exported
+from the swift-testing dylib. As such, its declaration could instead
+be written in a C header file with its own `module.modulemap`, or
+the runtime address could be obtained via
+[`dlsym()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/dlsym.html) or
```
grynspan, line 67: You already changed the URL, but for future reference, official Linux documentation is on kernel.org, and man7.org is technically one (important-to-Linux) engineer's personal website. (kernel.org just links out to man7.org, but still...) 😄

grynspan:
Okay, if you're satisfied with the changes you've made here, we can merge them any time. If you don't have a **Squash and merge** button in GitHub, @stmontgomery can do it.

ADKaster:
Yeah I'm all set with this. There's other sections of this file don't seem quite right, but that's for a separate PR. I don't have merge permissions though.

stmontgomery:
Hi, let me catch up with this discussion and PR this morning and I will hopefully merge it

stmontgomery (approved):
Thank you @ADKaster. While reviewing this I was reminded about a semi-related issue you filed, https://github.com/swiftlang/swift-testing/issues/1039. If/when we resolve or improve those install locations, it may warrant further improvements to this document, but that can be done separately of course.

PR #825, opened 2024-11-15
Configure CMake to build and install the Foundation cross-import overlay

stmontgomery:
This augments the project's CMake rules to begin building and installing the Foundation cross-import overlay (`_Testing_Foundation`), including the associated `.swiftcrossimport` directory which causes the overlay to be applied to clients.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

Resolves rdar://139877808

PR #829, opened 2024-11-19
Fix a non-portable target reference in CMake helper function

stmontgomery:
This fixes an oversight in a helper function in the project’s CMake rules: `_swift_testing_install_target()` has a hard-coded reference to the target named ”Testing” but it should instead refer to whatever target was passed in via the `module` argument to the function.

Discovered while working on #825.

### Checklist:


stmontgomery:
@swift-ci please test

PR #837, opened 2024-11-26
Don't install the binary Swift module when building for Apple platforms

stmontgomery:
This modifies the CMake rules to stop installing binary .swiftmodule files when building for Apple platforms.

### Motivation:

Apple platforms have a stable ABI and modules for those platforms should instead use the textual .swiftinterface which is already installed by CMake. The binary .swiftmodule permits access to [SPI](https://github.com/swiftlang/swift-testing/blob/main/Documentation/SPI.md) declarations from the testing library, and these are not intended to be exposed in distribution builds. Although this PR only removes access to these on macOS, since it's the only platform which has a textual .swiftinterface currently, we intend to investigate ways to match this behavior for other platforms in the future.

### Checklist:

Resolves rdar://136083081


stmontgomery:
@swift-ci please test

briancroom (approved):

etcwilde (approved):

PR #843, opened 2024-12-09
[NFC] Update _TestingInternals C macro for better compatibility

Kyle-Ye:
Update _TestingInternals C macro for better compatibility

### Motivation:

Better cross platform support

### Modifications:

- Add a `#else` clause if some compiler feature is not supported. 
- Add `SWT_ENUM` macro to avoid use `enum __attribute__((enum_extensibility(open)))` directly.

### Result:

_[After your change, what will change.]_

### Checklist:

### DIFF
--- a/Sources/_TestingInternals/include/Defines.h
+++ b/Sources/_TestingInternals/include/Defines.h
@@ -11,8 +11,13 @@
 #if !defined(SWT_DEFINES_H)
 #define SWT_DEFINES_H
 
+#if __has_feature(assume_nonnull)
 #define SWT_ASSUME_NONNULL_BEGIN _Pragma("clang assume_nonnull begin")
 #define SWT_ASSUME_NONNULL_END _Pragma("clang assume_nonnull end")
+#else
+#define SWT_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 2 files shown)


grynspan:
This code is always and only compiled with clang, which supports this macro.

Kyle-Ye:
Is `SWT_ENUM` needed here? I think we'd better not use `enum __attribute__((enum_extensibility(open)))` directly in a header file.

grynspan:
The `NS_ENUM` macro exists so that Foundation's headers (and other headers in Apple's SDKs) can be used with various compilers and language modes. Since we only ever build this module using clang-with-Swift, it isn't necessary to add a macro.

If we find ourselves writing this attribute frequently, abstracting it into a macro may be helpful.

Kyle-Ye:
Got it. The root reason is that currently the policy is we do not support other compiler other than Clang for swift project (swift-testing included)

grynspan:
It's a bit more complicated than that: early phases of the Swift toolchain build rely on the local C++ compiler (which could be clang, GCC, or MSVC++, or possibly some other compiler.) But the phases in which Swift Testing are built are always built with clang.

PR #849, opened 2024-12-10
Make `Attachment` conform to `CustomStringConvertible`.

grynspan:
This PR adds `CustomStringConvertible` conformance to `Attachment`. Note that if the attachable value is a move-only type, the conformance is not available in the stdlib, but we still provide a `description` property.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

PR #854, opened 2024-12-12
Add a `preferredName(for:basedOn:)` member to `Attachable` to allow customizing filenames.

grynspan:
This PR introduces a new optional member of `Attachable`, `preferredName(for:basedOn:)`, that we use when writing the corresponding attachment to disk/test reports/etc. in order to allow attachable types to customize the name independently of what the user specifies.

For example:

```swift
let a = Attachment(x)
let b = Attachment(y, named: "hello")
```

In both the attachments created above, the file name is incompletely specified. `a` has a default name, and both `a` and `b` have no path extension (which is important for the OS to correctly recognize the produced file's type.) By adding this new function to `Attachable`, we give `x` and `y` the opportunity to say "this is JPEG data" or "this is plain text" (and so forth.)

The new function is implemented by `_AttachableImageContainer`. I've also created `_AttachableURLContainer` to represent files mapped from disk for attachment (instead of directly passing them around as `Data`.)

Third-party conforming types will generally want to use Foundation's `NSString` or `URL` API to append path extensions (etc.)

> [!NOTE]
> Attachments remain experimental.

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
+++ b/Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
@@ -15,10 +15,6 @@ private import CoreGraphics
 private import ImageIO
 import UniformTypeIdentifiers
 
-#if canImport(CoreServices_Private)
-private import CoreServices_Private
-#endif
-
 /// ## Why can't images directly conform to Attachable?
 ///
 /// Three reasons:
@@ -174,12 +170,6 @@ extension _AttachableImageContainer: AttachableContainer {
   public borrowing func preferredName(for attachment: borrowing Attachment<Self>, basedOn suggestedName: String) -> String {
     if #available(_uttypesAPI, *) {
       return (suggestedName as NSString).appendingPathExtension(for: computedContentType)
-    } else {
-#if canImport(CoreServices_Private)
-      if let result = _UTTypeCreateSuggestedFilename(suggestedName as CFString, typeIdentifier)?.takeRetainedValue() {
-        return result as String
-      }
-#endif
     }
 
     return suggestedName
--- a/Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
+++ b/Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
@@ -15,6 +15,10 @@ private import CoreGraphics
 private import ImageIO
 import UniformTypeIdentifiers
 
+#if canImport(CoreServices_Private)
+private import CoreServices_Private
+#endif
+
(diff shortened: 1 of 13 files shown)


grynspan:
@swift-ci test

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageContainer.swift
```diff
@@ -159,5 +170,19 @@ extension _AttachableImageContainer: AttachableContainer {
       try body(UnsafeRawBufferPointer(start: data.bytes, count: data.length))
     }
   }
+
+  public borrowing func preferredName(for attachment: borrowing Attachment<Self>, basedOn suggestedName: String) -> String {
+    if #available(_uttypesAPI, *) {
+      return (suggestedName as NSString).appendingPathExtension(for: computedContentType)
+    } else {
+#if canImport(CoreServices_Private)
+      if let result = _UTTypeCreateSuggestedFilename(suggestedName as CFString, typeIdentifier)?.takeRetainedValue() {
```
grynspan, line 179: Probably a bad idea to do this, but it's there if we need it…

Sources/Testing/Attachments/Attachable.swift
```diff
@@ -64,6 +64,22 @@ public protocol Attachable: ~Copyable {
...
   /// the image.
   borrowing func withUnsafeBufferPointer<R>(for attachment: borrowing Attachment<Self>, _ body: (UnsafeRawBufferPointer) throws -> R) throws -> R
+
+  /// Generate a preferred name for the given attachment.
+  ///
+  /// - Parameters:
+  ///   - suggestedName: A suggested name to use as the basis of the preferred
+  ///     name. This string was provided by the developer when they initialized
+  ///     `attachment`.
+  ///   - attachment: The attachment that needs to be named.
```
stmontgomery, line 74: Want to flip the order of the parameters to match the signature?

Sources/Testing/Attachments/Attachment.swift
```diff
@@ -42,14 +42,24 @@ public struct Attachment<AttachableValue>: ~Copyable where AttachableValue: Atta
     "untitled"
   }
 
+  /// Storage for ``preferredName``.
+  fileprivate var _preferredName: String?
```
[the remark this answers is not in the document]
stmontgomery: Never mind, of course; it's because it's a shadow var for storage. Disregard

Sources/Testing/Attachments/Attachable.swift
grynspan: Oops, yes, I had moved stuff around after talking to some cow-orkers.

grynspan:
@swift-ci test

stmontgomery (approved):

PR #858, opened 2024-12-13
Change default behavior of exit tests to always succeed.

grynspan:
Exit tests simulate calling `main()` (or some similar thing), but if you don't explicitly exit from them, they force a failure. After much mulling and some discussion with colleagues a while back, we should change the behavior so that, if a test doesn't otherwise terminate, it acts as if `main()` returned naturally—that is, by exiting with `EXIT_SUCCESS` rather than by forcing `EXIT_FAILURE`. This behavior is more consistent with the feature's theory of operation.

For example:

```swift
await #expect(exitsWith: .success) {
  assert(2 > 1) // this is true and doesn't assert, so this test should pass, right?
}
```

Exit tests remain an experimental feature.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -158,11 +158,9 @@ extension ExitTest {
       _errorInMain(error)
     }
 
-    // Run some glue code that terminates the process with an exit condition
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci test

briancroom (approved):

PR #869, opened 2024-12-19
Add a cross-import overlay with AppKit to allow attaching `NSImage`s.

grynspan:
This PR adds on to the Core Graphics cross-import overlay added in #827 to allow attaching instances of `NSImage` to a test.

`NSImage` is a more complicated animal because it is not `Sendable`, but we don't want to make a (potentially very expensive) deep copy of its data until absolutely necessary. So we check inside the image to see if its contained representations are known to be safely copyable (i.e. copies made with `NSCopying` do not share any mutable state with their originals.) If it looks safe to make a copy of the image by calling `copy()`, we do so; otherwise, we try to make a deep copy of the image.

Due to how Swift implements polymorphism in protocol requirements, and because we don't really know what they're doing, subclasses of `NSImage` just get a call to `copy()` instead of deep introspection.

`UIImage` support will be implemented in a separate PR.

> [!NOTE]
> Attachments remain an experimental feature.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -54,6 +54,7 @@ let package = Package(
       name: "TestingTests",
       dependencies: [
         "Testing",
+        "_Testing_AppKit",
         "_Testing_CoreGraphics",
         "_Testing_Foundation",
       ],
@@ -95,6 +96,15 @@ let package = Package(
     ),
 
     // Cross-import overlays (not supported by Swift Package Manager)
+    .target(
+      name: "_Testing_AppKit",
+      dependencies: [
+        "Testing",
+        "_Testing_CoreGraphics",
+      ],
+      path: "Sources/Overlays/_Testing_AppKit",
+      swiftSettings: .packageSettings
+    ),
     .target(
       name: "_Testing_CoreGraphics",
       dependencies: [
--- a/Sources/Overlays/_Testing_AppKit/Attachments/NSImage+AttachableAsCGImage.swift
+++ b/Sources/Overlays/_Testing_AppKit/Attachments/NSImage+AttachableAsCGImage.swift
@@ -1,0 +1,95 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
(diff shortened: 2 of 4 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

briancroom (approved):

grynspan:
@swift-ci test Linux

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Package.swift
```diff
@@ -190,6 +191,15 @@ let package = Package(
...
 
     // Cross-import overlays (not supported by Swift Package Manager)
+    .target(
+      name: "_Testing_AppKit",
+      dependencies: [
+        "Testing",
+        "_Testing_CoreGraphics",
+      ],
+      path: "Sources/Overlays/_Testing_AppKit",
+      swiftSettings: .packageSettings
```
stmontgomery, line 201: ` + .enableLibraryEvolution()` is encouraged now, and the `_Testing_CoreGraphics` target/overlay has adopted it below

Sources/Overlays/_Testing_AppKit/Attachments/NSImage+AttachableAsCGImage.swift
```diff
@@ -0,0 +1,95 @@
...
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if SWT_TARGET_OS_APPLE && canImport(AppKit)
+public import AppKit
+@_spi(ForSwiftTestingOnly) @_spi(Experimental) public import _Testing_CoreGraphics
```
stmontgomery, line 13: `@_spi(ForSwiftTestingOnly)` has been phased out since this PR was originally posted, and we can use `package` now.

stmontgomery (commented):
I know it's already merged but I had just a couple small comments that are relevant now but weren't when the PR was first posted. Hopefully we can do a fix-up PR or incorporate them into some follow-up work.

Package.swift
grynspan: Fixed in #1196.

Sources/Overlays/_Testing_AppKit/Attachments/NSImage+AttachableAsCGImage.swift
grynspan: Fixed in #1196.

PR #816, opened 2024-11-11
Work around Windows build failure due to SwiftPM regression.

grynspan:
Works around a link-time bug in SwiftPM that appears to have been introduced in https://github.com/swiftlang/swift-package-manager/pull/8049.

Works around https://github.com/swiftlang/swift-package-manager/issues/8111 until https://github.com/swiftlang/swift-package-manager/pull/8112 is in the build.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #872, opened 2024-12-19
Remove the workaround for swift-package-manager-#8111.

grynspan:
Remove the workaround for https://github.com/swiftlang/swift-package-manager/issues/8111. The issue has been resolved and Windows correctly exports symbols from dynamic library targets.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,6 +28,7 @@ let package = Package(
   products: [
     .library(
       name: "Testing",
+      type: .dynamic,
       targets: ["Testing"]
     ),
   ],
--- a/Package.swift
+++ b/Package.swift
@@ -28,7 +28,7 @@ let package = Package(
   products: [
     .library(
       name: "Testing",
-      type: .dynamic,
+      type: .dynamic, // needed so Windows exports ABI entry point symbols
       targets: ["Testing"]
     ),
   ],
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci test

Package.swift
```diff
@@ -28,6 +28,7 @@ let package = Package(
   products: [
     .library(
       name: "Testing",
+      type: .dynamic,
```
stmontgomery, line 31: The workaround in #816 was just the compilation guard in the test file below. Is changing this to `.dynamic` actually necessary?

grynspan: Yep, otherwise the symbols aren't exported. @dschaefer2 has more info.

stmontgomery: Can you please add a comment capturing why this is needed? I anticipate it may be forgotten since the current motivation for adding it is a platform-specific workaround

grynspan: Added.

grynspan:
@swift-ci test

stmontgomery (approved):

xedin:
@grynspan @stmontgomery I wonder if this caused Source Compatibility Suite regressions - https://ci.swift.org/job/swift-main-source-compat-suite-debug/989/artifact/swift-source-compat-suite/

grynspan:
I suppose it's possible, but looking at the failures, they don't seem related. It looks like the testing module simply isn't being built, whereas our change here (combined with the follow-up 840c82289e2c0ecc98605c7fc9fda787053c95b8 patch) just makes the library build as a DLL on Windows. There should be absolutely no visible change on non-Windows systems.

xedin:
I see, so it must be something about ScanDependencies or something related to that.

grynspan:
I'm not willing to rule this change out _entirely_ since the follow-up commit might have been delayed making it into a toolchain. But I would definitely look at ScanDependencies first if only to rule it out.

xedin:
Sounds good, thank you!

PR #874, opened 2024-12-20
Don't use `@_semantics`, use our own attribute instead.

grynspan:
This PR replaces our use of `@_semantics` with a custom attribute macro. `@_semantics` is reserved for use by the standard library and runtime and, while useful, can be emulated on our side of the module barrier without much difficulty.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/Test+Macro.swift
```diff
@@ -484,6 +484,27 @@ extension Test {
...
+///
+/// The use cases for this macro are subject to change over time as the needs of
+/// the testing library change. The implementation of this macro in the
+/// TestingMacros target determines how different arguments are handled.
+///
+/// - Note: This macro has compile-time effects _only_ and should not affect a
+///   compiled test target.
+///
+/// - Warning: This macro is used to implement the `@Test` macro. Do not use it
+///   directly.
```
stmontgomery, line 503: Strictly speaking, isn't it used to implement by `#expect` macros, not `@Test`? Maybe this should be more vague

Sources/Testing/Test+Macro.swift
```diff
@@ -484,6 +484,27 @@ extension Test {
...
+@attached(peer) public macro __testing(
```
stmontgomery, line 504: Does this _need_ to be `public` initially? I suspect that's the direction this is headed regardless, but just wondering if it's only needed by our own test code right now

Sources/TestingMacros/PragmaMacro.swift
```diff
@@ -0,0 +1,79 @@
...
+/// Attributes conditionally applied with `#if` are ignored.
+func pragmas(on node: some WithAttributesSyntax) -> [AttributeSyntax] {
+  node.attributes
+    .compactMap { attribute in
+      if case let .attribute(attribute) = attribute {
+        return attribute
+      }
+      return nil
+    }.filter { attribute in
+      attribute.attributeNameText == "__testing"
```
stmontgomery, line 54: Is this robust enough to accept a fully-qualified name with module prefix like `@Testing.__testing` ?

stmontgomery: (And conversely, to reject attributes with the same name which could hypothetically exist in other modules?)

Sources/Testing/Test+Macro.swift
grynspan: Yes, no, maybe, both :)

Right now it's actually used to implement one of our tests only.

Sources/TestingMacros/PragmaMacro.swift
grynspan: FBOFW that isn't actually supported by the language, so… no. But I can code it defensively if you're concerned.

grynspan: Actually, no, I don't think we care: realistically nobody else is going to write `@__testing(semantics: "...")`. If somebody really _does_ run into trouble here, we can address it in the future.

Sources/Testing/Test+Macro.swift
grynspan: We can make it internal for now—there's a public version coming almost immediately after this PR though.

grynspan: See https://github.com/swiftlang/swift-testing/pull/745/commits/ddfdbc9a751f167dd4c6c99cb4c949739f813182

Sources/TestingMacros/PragmaMacro.swift
grynspan: Added fully-qualified name check. If somebody else makes a macro named `@__testing(semantics:)`, it'll trip as ambiguous before getting to our code, so I don't care too much. :)

grynspan:
@swift-ci test

Sources/Testing/Test+Macro.swift
stmontgomery:
```suggestion
/// - Warning: This macro is used to implement other macros declared by the testing
///   library. Do not use it directly.
```

stmontgomery (approved):

grynspan:
@swift-ci test

PR #875, opened 2024-12-21
Mark `#expect(_:throws:)` and `#require(_:throws:)` as to-be-deprecated.

### LINKED ISSUE #873: Undeprecate the various `throws:` expectation/requirements
### Description

Various methods were deprecated in https://github.com/swiftlang/swift-testing/pull/780. We maintain a number of libraries, usually for the last three Swift versions, so having test methods deprecated is a little tedious as we either need to live with the deprecation warnings or use `#if compiler(...)` in our tests to use the most appropriate API. It'd be great if the deprecated APIs weren't actually deprecated and lived for few more Swift versions. 

### Expected behavior

_No response_

### Actual behavior

_No response_

### Steps to reproduce

_No response_

_No response_

_No response_

grynspan:
These symbols should go through a transition period where they are documented as to-be-deprecated rather than being deprecated outright. My mistake.

With this change applied, Xcode treats these macros as deprecated (deemphasizing them in autocomplete, for instance) and DocC marks them as deprecated in an unspecified Swift version, but no warning is emitted if they are used. This is consistent with API marked `API_TO_BE_DEPRECATED` or `deprecated: 100000.0` in Apple's SDKs.

> [!NOTE]
> There is a bug in the DocC compiler that emits a diagnostic of the form:
>
> > ⚠️ 'require(_:sourceLocation:performing:throws:)' isn't unconditionally deprecated
>
> This issue is being tracked already with rdar://141785948.

Resolves #873.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
+++ b/Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
@@ -48,8 +48,8 @@ running your test if the code doesn't throw the expected error.
 
 To check that the code under test throws an error of any type, pass
 `(any Error).self` as the first argument to either
-``expect(throws:_:sourceLocation:performing:)-1xr34`` or
-``require(_:_:sourceLocation:)-5l63q``:
+``expect(throws:_:sourceLocation:performing:)-1hfms`` or
+``require(throws:_:sourceLocation:performing:)-7n34r``:
 
 ```swift
 @Test func cannotAddToppingToPizzaBeforeStartOfList() {
--- a/Documentation/Proposals/0006-return-errors-from-expect-throws.md
+++ b/Documentation/Proposals/0006-return-errors-from-expect-throws.md
@@ -105,7 +105,7 @@ is not statically available. The problematic overloads will also be deprecated:
 -) where E: Error & Equatable
 +) -> E where E: Error & Equatable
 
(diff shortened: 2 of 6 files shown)


grynspan:
@swift-ci test

Sources/Testing/Testing.docc/testing-for-errors-in-swift-code.md
```diff
@@ -48,8 +48,8 @@ running your test if the code doesn't throw the expected error.
 
 To check that the code under test throws an error of any type, pass
 `(any Error).self` as the first argument to either
-``expect(throws:_:sourceLocation:performing:)-1xr34`` or
```
grynspan, line 51: @heckj FYI fixed some typos here that I noticed when building the DocC bundle.

heckj: With any luck these hashes can be replaced in 6.1+ (somewhere downstream) with annotations that are more directly related to signatures and types. It doesn't take much of a change to throw off the hashes into something new, which is really annoying (to me)

grynspan: That's a question for the DocC folks, at any rate.

briancroom (approved):

PR #881, opened 2025-01-03
Remove the `introduced: 6.0` argument from the to-be-deprecated attributes on  `#expect(performing:throws:)` and `#require(performing:throws:)`.

grynspan:
The `introduced: 6.0` argument is causing the compiler to error out when building in earlier language modes. It isn't syntactically necessary and our DocC bundle provides the equivalent documentation info. So just drop it.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@briancroom reviewed externally but didn't click the green button.

Merging as self-reviewed!

PR #892, opened 2025-01-06
Use the Toolhelp32 API to enumerate loaded Win32 modules.

grynspan:
This PR replaces our call to `EnumProcessModules()` with one to `CreateToolhelp32Snapshot(TH32CS_SNAPMODULE)`. Three reasons:

1. `EnumProcessModules()` requires us to specify a large, fixed-size buffer to contain all the `HMODULE` handles;
2. `EnumProcessModules()` does not own any references to the handles it returns, meaning that a module can be unloaded while we are iterating over them (while `CreateToolhelp32Snapshot()` temporarily bumps the refcounts of the handles it produces); and
3. `CreateToolhelp32Snapshot()` lets us produce a lazy sequence of `HMODULE` values rather than an array, letting us write somewhat Swiftier code that uses it.

The overhead of using `CreateToolhelp32Snapshot()` was negligible (below the noise level when measuring).

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/WinSDKAdditions.swift
+++ b/Sources/Testing/Support/Additions/WinSDKAdditions.swift
@@ -50,4 +50,56 @@ let STATUS_SIGNAL_CAUGHT_BITS = {
 
   return result
 }()
+
+// MARK: - HMODULE members
+
+extension HMODULE {
+  /// A helper type that manages state for ``HMODULE/all``.
+  private final class _AllState {
+    /// The toolhelp snapshot.
+    var snapshot: HANDLE?
+
+    /// The module iterator.
+    var me = MODULEENTRY32W()
+
+    deinit {
+      if let snapshot {
+        CloseHandle(snapshot)
+      }
+    }
+  }
+
+  /// All modules loaded in the current process.
+  ///
+  /// - Warning: It is possible for one or more modules in this sequence to be
+  ///   unloaded while you are iterating over it. To minimize the risk, do not
+  ///   discard the sequence until iteration is complete. Modules containing
+  ///   Swift code can never be safely unloaded.
+  static var all: some Sequence<Self> {
+    sequence(state: _AllState()) { state in
+      if let snapshot = state.snapshot {
+        // We have already iterated over the first module. Return the next one.
+        if Module32NextW(snapshot, &state.me) {
+          return state.me.hModule
+        }
+      } else {
+        // Create a toolhelp snapshot that lists modules.
+        guard let snapshot = CreateToolhelp32Snapshot(DWORD(TH32CS_SNAPMODULE), 0) else {
+          return nil
+        }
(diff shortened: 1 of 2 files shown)


Sources/Testing/Support/Additions/WinSDKAdditions.swift
```diff
@@ -50,4 +50,56 @@ let STATUS_SIGNAL_CAUGHT_BITS = {
...
+  static var all: some Sequence<Self> {
+    sequence(state: _AllState()) { state in
+      if let snapshot = state.snapshot {
+        // We have already iterated over the first module. Return the next one.
+        if Module32NextW(snapshot, &state.me) {
+          return state.me.hModule
+        }
+      } else {
+        // Create a toolhelp snapshot that lists modules.
+        guard let snapshot = CreateToolhelp32Snapshot(DWORD(TH32CS_SNAPMODULE), 0) else {
```
compnerd, line 87: I'd pull out `TH32CS_SNAPMODULE` as an extension:

```
var TH32CS_SNAPMODULE: DWORD {
  DWORD(Win32.TH32CS_SNAPMODULE)
}
```

Sources/Testing/Support/Additions/WinSDKAdditions.swift
```diff
@@ -50,4 +50,56 @@ let STATUS_SIGNAL_CAUGHT_BITS = {
...
+        // Initialize the iterator for use by the resulting sequence and return
+        // the first module.
+        state.me.dwSize = DWORD(MemoryLayout.stride(ofValue: state.me))
+        if Module32FirstW(snapshot, &state.me) {
+          return state.me.hModule
+        }
+      }
+
+      // Reached the end of the iteration.
+      return nil
```
compnerd, line 101: I think that it might be nicer to use `guard` statements for the `Module32FirstW` and `Module32NextW` rather than falling through the "end of iteration".

compnerd (approved):

Sources/Testing/Support/Additions/WinSDKAdditions.swift
grynspan: Add it to the pile. There's a few dozen (?) of these that I would love to go in and change, but I would much rather we teach the importer how to better-handle Windows headers.

Sources/Testing/Support/Additions/WinSDKAdditions.swift
grynspan: I had that in an intermediate stage of this PR and found it harder to reason about because the Windows enumeration API isn't _quite_ what `sequence()` expects, but I can take another pass at it.

grynspan: Gave it another shot and it just turned into the same thing but with a final `return state.me.hModule` at the end instead, so… eh. It's fine.

PR #880, opened 2025-01-02
Store test content in a custom metadata section.

grynspan:
This PR uses the experimental symbol linkage margers feature in the Swift compiler to emit metadata about tests (and exit tests) into a dedicated section of the test executable being built. At runtime, we discover that section and read out the tests from it.

This has several benefits over our current model, which involves walking Swift's type metadata table looking for types that conform to a protocol:

1. We don't need to define that protocol as public API in Swift Testing,
1. We don't need to emit type metadata (much larger than what we really need) for every test function,
1. We don't need to duplicate a large chunk of the Swift ABI sources in order to walk the type metadata table correctly, and
1. Almost all the new code is written in Swift, whereas the code it is intended to replace could not be fully represented in Swift and needed to be written in C++.

This change will be necessary to support Embedded Swift because there is no type metadata section emitted for embedded targets.

The change also opens up the possibility of supporting generic types in the future because we can emit metadata without needing to emit a nested type (which is not always valid in a generic context.) That's a "future direction" and not covered by this PR specifically.

I've defined a layout for entries in the new `swift5_tests` section that should be flexible enough for us in the short-to-medium term and which lets us define additional arbitrary test content record types. The layout of this section is covered in depth in the new [TestContent.md](https://github.com/swiftlang/swift-testing/blob/main/Documentation/ABI/TestContent.md) article.

This functionality is only available if a test target enables the experimental `"SymbolLinkageMarkers"` feature and only if Swift Testing is used as a package (not as a toolchain component.) We continue to emit protocol-conforming types for now—that code will be removed if and when the experimental feature is properly supported (modulo us adopting relevant changes to the feature's API.)

## See Also

https://github.com/swiftlang/swift-testing/issues/735
https://github.com/swiftlang/swift-testing/issues/764
https://github.com/swiftlang/swift/issues/76698
https://github.com/swiftlang/swift/pull/78411

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

Issue #735, opened 2024-09-26
Stop calling runtime-internal function to enumerate tests on ELF platforms


grynspan:
Currently, on ELF only, we need to call an internal runtime function `swift_enumerateAllMetadataSections()` in order to find sections that contain our metadata. That's because ELF binaries generally don't preserve section header information. I'd love for our platform-specific implementation to look like:

```swift
template <typename SectionEnumerator>
static void enumerateTypeMetadataSections(const SectionEnumerator& body) {
  dl_iterate_phdr([] (struct dl_phdr_info *info, size_t size, void *context) -> int {
    const SectionEnumerator& body = *reinterpret_cast<SectionEnumerator *>(context);

    bool stop = false;
    if (auto sb = /* ??? */(info)) {
      body(sb, &stop);
    }
    return stop;
  }, const_cast<SectionEnumerator *>(&body));
}
```

grynspan:
Potentially addressed by https://github.com/swiftlang/swift/issues/76698.

grynspan:
Tracked internally as rdar://136789701.

grynspan:
Should be possible to resolve this once https://github.com/swiftlang/swift/pull/78411 lands.

grynspan:
Alas, not gonna happen. ELF does what ELF wants.

PR #893, opened 2025-01-06
Discover test content stored in the test content section of loaded images.

grynspan:
This PR implements discovery, _but not emission_, of test content that has been added to a loaded image's test content metadata section at compile time.

Loading this data from a dedicated section has several benefits over our current model, which involves walking Swift's type metadata table looking for types that conform to a protocol:

1. We don't need to define that protocol as public API in Swift Testing,
1. We don't need to emit type metadata (much larger than what we really need) for every test function,
1. We don't need to duplicate a large chunk of the Swift ABI sources in order to walk the type metadata table correctly, and
1. Almost all the new code is written in Swift, whereas the code it is intended to replace could not be fully represented in Swift and needed to be written in C++.

The change also opens up the possibility of supporting generic types in the future because we can emit metadata without needing to emit a nested type (which is not always valid in a generic context.) That's a "future direction" and not covered by this PR specifically.

I've defined a layout for entries in the new `swift5_tests` section that should be flexible enough for us in the short-to-medium term and which lets us define additional arbitrary test content record types. The layout of this section is covered in depth in the new [TestContent.md](Documentation/ABI/TestContent.md) article.

This PR does **not** include the code necessary to _emit_ test content records into images at compile time. That part of the change is covered by #880 and requires a new language feature to control which section data is emitted to. An experimental version of that language feature is currently available under the `"SymbolLinkageMarkers"` label.

Because there is no test content in the test content section yet, this PR does not remove the "legacy" codepaths that discover tests in the type metadata section.

> [!NOTE]
> This change is experimental.

### See Also

https://github.com/swiftlang/swift-testing/pull/880 https://github.com/swiftlang/swift-testing/issues/735 https://github.com/swiftlang/swift/issues/76698
https://github.com/swiftlang/swift/pull/78411

### Checklist:

### DIFF
--- a/Documentation/ABI/TestContent.md
+++ b/Documentation/ABI/TestContent.md
@@ -28,7 +28,7 @@ section in built test products:
 | macOS, iOS, watchOS, tvOS, visionOS | Mach-O | `__DATA_CONST,__swift5_tests` |
 | Linux, FreeBSD, OpenBSD, Android | ELF | `swift5_tests` |
 | WASI | WebAssembly | `swift5_tests` |
-| Windows | PE/COFF | `.sw5test`[^windowsPadding] |
+| Windows | PE/COFF | `.sw5test$B`[^windowsPadding] |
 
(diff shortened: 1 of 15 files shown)


grynspan:
@swift-ci test

briancroom (approved):

PR #895, opened 2025-01-07
Add an attribute that emits a warning.

grynspan:
This PR adds an attribute (an overload of `@__testing()`) that emits a warning. This allows us to emit compile-time warnings from contexts where only attributes are semantically valid.

We need this in particular for test content records because their section info is comprised of a big `IfConfigDecl` node and platforms that aren't covered (i.e. the `#else` clause) don't have a way to emit a diagnostic that says "we need to fix this platform's Swift Testing port!" like we can do in other contexts with `#warning()`:

```swift
#if os(macOS) || os(iOS) || os(watchOS) || os(tvOS) || os(visionOS)
@_section("__DATA_CONST,__swift5_tests")
#elseif os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android) || os(WASI)
@_section("swift5_tests")
#elseif os(Windows)
@_section(".sw5test$B")
#else
// ⚠️ #warning isn't valid here, what do we do!?
#endif
@_used
private static let record: __TestContentRecord = (...)
```

### Checklist:


grynspan:
@swift-ci test

Sources/TestingMacros/PragmaMacro.swift
```diff
@@ -27,6 +29,17 @@ public struct PragmaMacro: PeerMacro, Sendable {
     providingPeersOf declaration: some DeclSyntaxProtocol,
     in context: some MacroExpansionContext
   ) throws -> [DeclSyntax] {
+    if case let .argumentList(arguments) = node.arguments,
+       arguments.first?.label?.textWithoutBackticks == "warning" {
+      let targetNode = Syntax(declaration)
+      let messages = arguments
```
stmontgomery, line 35: Surprised to not see `.lazy` in this chain :)

grynspan: We don't use `.lazy` as much in the macro target, you know that! :P

Tests/TestingMacrosTests/PragmaMacroTests.swift
```diff
@@ -18,12 +18,24 @@ import SwiftSyntax
...
+  @Test func warningGenerated() throws {
+    let sourceCode = """
+    @__testing(warning: "abc123")
+    let x = 0
+    """
+
+    let (_, diagnostics) = try parse(sourceCode)
+    #expect(diagnostics.count == 1)
+    #expect(diagnostics[0].message == "abc123")
+    #expect(diagnostics[0].diagMessage.severity == .warning)
```
stmontgomery, line 39: Can the test also validate which syntax node the diagnostic was associated with?

Sources/TestingMacros/PragmaMacro.swift
stmontgomery: Hmm, seems just as useful there as elsewhere to me

stmontgomery (approved):

briancroom (approved):

Tests/TestingMacrosTests/PragmaMacroTests.swift
grynspan: Done.

Sources/TestingMacros/PragmaMacro.swift
stmontgomery: Sure. Not something we need to fix here, but I don't personally see a reason to _avoid_ our general coding style of using `.lazy` in the macro implementation.

grynspan: Our general style has been _not_ to use it in the macro implementation. :)

PR #898, opened 2025-01-08
Use `Never` instead of `Void` as the fallback lock/mutex type.

grynspan:
When using WASI without multithreading, or when using a platform to which Swift Testing has been incompletely ported, we don't know what type to use as a lock/mutex, so we use `Void`. However, turns out that produces a diagnostic:

> warning: UnsafeMutablePointer<Void> has been replaced by UnsafeMutableRawPointer

So use `Never` instead. (Yes, this will produce a pointer to an uninitialized instance of `Never`. If this affects you, please read Porting.md!)

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

PR #901, opened 2025-01-08
Remove SPIAwareTrait and adopt TestScoping in its place

stmontgomery:
This is a follow-on from #733 which introduced the Test Scoping Traits feature. As that proposal indicated, the `SPIAwareTrait` SPI protocol is no longer necessary, so this PR removes it and adopts `TestScoping` in its place.

### Motivation:

Adopting Test Scoping Traits for the purpose which `SPIAwareTrait` previously served simplifies a lot of runner and planning logic.

### Modifications:

- Delete `SPIAwareTrait` and related SPI.
- Adopt `TestScoping` in `ParallelizationTrait` instead.
- Changed `ParallelizationTrait` to _not_ always have the value `true` for `isRecursive`. Test Scoping Traits makes this no longer necessary: the invocation of children of any suites which have `.serialized` is scoped to have parallelization disabled.
- Removed tracking of `isParallelizationEnabled` on the `Runner.Plan.Action.RunOptions` type. For now, I kept the `RunOptions` struct, since it could be useful in the future and it's SPI.
- Removed obsolete logic in `Runner` for propagating ancestor steps, and refactored slightly to make many things `static`. This is to help ensure certain functions don't access the `self.configuration` property when instead they should be accessing `Configuration.current`.

### Result:

No functional change, but implementation is simplified.

### Checklist:


stmontgomery:
@swift-ci please test

Sources/Testing/Running/Runner.swift
```diff
@@ -103,21 +103,18 @@ extension Runner {
...
+  ///     determine whether parallelization is enabled. The default value is the
+  ///     current configuration, if any, or else the default configuration.
   ///   - body: The function to invoke.
   ///
   /// - Throws: Whatever is thrown by `body`.
-  private func _forEach<E>(
+  private static func _forEach<E>(
     in sequence: some Sequence<E>,
-    for step: Plan.Step?,
+    configuration: Configuration = .current ?? .init(),
```
grynspan, line 115: Do we ever actually explicitly pass `configuration`? If not, maybe just make it a local in the function(s) that use it?

Sources/Testing/Traits/ParallelizationTrait.swift
```diff
@@ -16,31 +16,28 @@
 /// test suite, this trait causes that suite to run its contained test functions
 /// and sub-suites serially instead of in parallel.
 ///
-/// This trait is recursively applied: if it is applied to a suite, any
-/// parameterized tests or test suites contained in that suite are also
-/// serialized (as are any tests contained in those suites, and so on.)
+/// If this trait is applied to a suite, any test functions or test suites
```
grynspan, line 19: It's important to note that parameterized test functions are serialized (as opposed to just function A, function B, function C get serialized.)

Sources/Testing/Traits/ParallelizationTrait.swift
```diff
@@ -16,31 +16,28 @@
...
-@_spi(ForToolsIntegrationOnly)
-extension ParallelizationTrait: SPIAwareTrait {
-  public func prepare(for test: Test, action: inout Runner.Plan.Action) async throws {
-    if case var .run(options) = action {
-      options.isParallelizationEnabled = false
-      action = .run(options: options)
+extension ParallelizationTrait: TestScoping {
+  public func provideScope(for test: Test, testCase: Test.Case?, performing function: @Sendable () async throws -> Void) async throws {
+    guard var configuration = Configuration.current else {
+      throw SystemError(description: "There is no current Configuration when attempting to provide scope for test '\(test.name)'")
```
grynspan, line 36: `Configuration` isn't API, so this may be a confusing message. Why not just do `var configuration = Configuration.current ?? .init()` instead?

Sources/Testing/Traits/SPIAwareTrait.swift
```diff
@@ -14,7 +14,8 @@
 /// This protocol refines ``Trait`` in various ways that require the use of SPI.
 /// Ideally, such requirements will be promoted to API when their design
 /// stabilizes.
-@_spi(Experimental) @_spi(ForToolsIntegrationOnly)
+@available(*, deprecated, message: "The SPIAwareTrait protocol is deprecated and no longer used.")
```
grynspan, line 17: Was it intentional that you removed `@_spi(Experimental)`?

FWIW we don't need to deprecate SPI and can just delete it.

Sources/Testing/Traits/ParallelizationTrait.swift
stmontgomery: This really would indicate a bug in our implementation of the testing library, so here at least, I wanted to throw an error instead of attempting to recover. The same argument could be made for the usages in `Runner.swift`, but making those `throws` felt more awkward so I did opt to use `?? .init()`.

Sources/Testing/Traits/SPIAwareTrait.swift
stmontgomery: The `SPI.md` document says that `@_spi(ForToolsIntegrationOnly)` will go through a deprecation period. In retrospect, I don't actually think `SPIAwareTrait` needed to have been added to that SPI group, since tools don't really need it AFAIK, but I was trying to honor that guideline. I'd _like_ to delete it outright.

grynspan: Sorry, we don't need to deprecate _experimental_ SPI.

It was added to that group for the benefit of Xcode (although I don't recall if Xcode ended up using it at any point.)

Sources/Testing/Traits/ParallelizationTrait.swift
stmontgomery: I'll try to tweak the wording more. What I was attempting to clarify with this re-wording is that regular, non-parameterized test functions within a suite which has `.serialized` are also serialized alongside any sub-suites which are their peers. I found the existing wording a bit unclear about the handling of non-parameterized test functions.

Sources/Testing/Running/Runner.swift
stmontgomery: Can do

Sources/Testing/Traits/ParallelizationTrait.swift
grynspan: If it would indicate a bug in our code, let's be sure to include the "file a bug" boilerplate in the message.

Sources/Testing/Traits/SPIAwareTrait.swift
stmontgomery: But per the guideline in that document, we _do_ need to deprecate `ForToolsIntegrationOnly` decls, right? This protocol is in both SPI groups.

grynspan: I wrote that document. 🙂 The intent is that the rule about `Experimental` takes precedence. Experimental stuff may be modified or deleted at any time regardless of other SPI groups it belongs to.

Sources/Testing/Traits/ParallelizationTrait.swift
grynspan: I recall this being difficult to get right the last time too—might want to ping @iamleeg & friends for assistance?

Sources/Testing/Traits/SPIAwareTrait.swift
stmontgomery: Let's clarify that intent in the document because I do find it ambiguous. I can do that separately.

In the mean time, I'm pleased to go ahead and remove this stuff outright

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -15,21 +15,7 @@ extension Runner {
...
-        /// ``Configuration/isParallelizationEnabled`` property of the
-        /// configuration passed during initialization has a value of `true`.
-        ///
-        /// Traits such as ``Trait/serialized`` applied to individual tests may
-        /// affect whether or not that test is parallelized.
-        ///
-        /// ## See Also
-        ///
-        /// - ``ParallelizationTrait``
-        public var isParallelizationEnabled: Bool
```
grynspan, line 31: This _wasn't_ marked `Experimental`, sadly.

Sources/Testing/Running/Runner.swift
```diff
@@ -188,30 +184,30 @@ extension Runner {
     defer {
       if let step = stepGraph.value {
         if shouldSendTestEnded {
-          Event.post(.testEnded, for: (step.test, nil), configuration: configuration)
+          Event.post(.testEnded, for: (step.test, nil), configuration: _configuration)
```
grynspan, line 187: If these events are all pulling the configuration from the `current` task-local, there's no need to specify it as an argument since that's the default.

Sources/Testing/Running/Runner.swift
```diff
@@ -188,30 +184,30 @@ extension Runner {
...
         }
-        Event.post(.planStepEnded(step), for: (step.test, nil), configuration: configuration)
+        Event.post(.planStepEnded(step), for: (step.test, nil), configuration: _configuration)
       }
     }
 
     if let step = stepGraph.value, case .run = step.action {
       await Test.withCurrent(step.test) {
-        _ = await Issue.withErrorRecording(at: step.test.sourceLocation, configuration: configuration) {
+        _ = await Issue.withErrorRecording(at: step.test.sourceLocation, configuration: _configuration) {
```
grynspan, line 195: Should we update `withErrorRecording` to just pull the `current` task-local value? Do we ever use it in a way that needs to take any other value?

Sources/Testing/Running/Runner.swift
```diff
@@ -298,13 +281,13 @@ extension Runner {
   ///
   /// If parallelization is supported and enabled, the generated test cases will
   /// be run in parallel using a task group.
-  private func _runTestCases(_ testCases: some Sequence<Test.Case>, within step: Plan.Step) async throws {
+  private static func _runTestCases(_ testCases: some Sequence<Test.Case>, within step: Plan.Step) async throws {
     // Apply the configuration's test case filter.
     let testCases = testCases.lazy.filter { testCase in
-      configuration.testCaseFilter(testCase, step.test)
+      _configuration.testCaseFilter(testCase, step.test)
```
grynspan, line 287: For perf, can we or should we capture `_configuration` outside the closure here?

Sources/Testing/Running/Runner.swift
stmontgomery: I'm hesitant to change all these `Event.post()` calls without a better understanding of what perf cost it might have. But I think I can safely capture the current configuration once in a local variable for the functions which need it, and that'll further reduce diffs needed here

Sources/Testing/Running/Runner.swift
stmontgomery: Given where I landed with other changes I don't think I'll need this right now

stmontgomery: But I agree in principle we could make that utility infer the current configuration, just not sure it's needed for this PR

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Runner.swift
grynspan: You're already paying the cost of the task-local here anyway.

grynspan (approved):

Issue #903, opened 2025-01-10
Allow evaluating `ConditionTrait` programmatically


Uncommon:
### Description

I'm working on a testing library for writing structured tests like RSpec or Quick, with the intention of making it compatible with Swift Testing and XCTest rather than being a stand-alone framework. Right now I'm looking into whether I can use Swift Testing traits in my library's test elements; being able to just re-use existing traits could be really cool.

At first glance it was promising - `Bug` for example just contains the properties describing the bug, and then conforms to `TestTrait` via extension, making it easy to also use for other purposes.

But then there's `ConditionTrait`, where the `Kind` enum and `kind` property are not public, and the logic for evaluating the condition happens in `prepare(for:)` even though that method doesn't use the given test object. Is there a reason these elements are hidden? Or could they be made public, and the evaluation logic extracted to a separate function? Just the latter would probably be sufficient.

### Expected behavior

`ContitionTrait` can be used/evaluated directly

### Actual behavior

Evaluation logic is inaccessible because of non-public members

grynspan:
`Kind`/`kind` are simply implementation details of the type and do not need to be made public for the type to be usable.

What specifically do you need them for? Are you trying to create your own variations of `ConditionTrait`? You can create a custom factory function for that. For instance, if you want a condition trait that skips a test when the sky is blue:

```swift
extension Trait where Self == ConditionTrait {
  static var disabledIfSkyIsBlue: Self {
    .disabled(if: Earth.current.sky.color == .blue)
  }
}
```

Uncommon:
Mainly I want to evaluate the condition without having to give it a `Test` (which it doesn't really need anyway), so that I can use it in the test elements of my library. I could do that if I had access to `kind`, or if there were a method that just gave me a true/false result.

Uncommon:
Yes, exactly.

To give more context about what I have in mind, a test body using my library can look like below, where `Describe`, `BeforeAny`, and `It` are types used by a result builder:

```swift
Describe("Some feature") {
    BeforeAny {
        // do some setup
    }
    BeforeAny(.enabled(if: /* some check */)) {
        // condition-specific setup
    }
    It("behaves correctly") {
        // test some things
    }
    It("does more things", .projectSpecificTrait) {
        // other tests
    }
}
```

[...]

PR #909, opened 2025-01-14
Add ConditionTrait.evaluate()

Uncommon:
Add `ConditionTrait.evaluate()` so that a condition can be evaluated independent of a `Test` object.

### Motivation:

Currently, the only way a `ConditionTrait` is evaluated is inside the `prepare(for:)` method. This makes it difficult and awkward for third-party libraries to utilize these traits because evaluating a condition would require creating a dummy `Test` to pass to that method.

### Modifications:

Add `ConditionTrait.evaluate()`, and `ConditionTrait.Evaluation` enum for the return value.

### Result:

Public API allows for evaluating a `ConditionTrait` in any context.

### Checklist:


Uncommon:
I'm trying to decide what tests to add. Are existing tests enough, since I pretty much just moved existing logic to a different place?

Sources/Testing/Traits/ConditionTrait.swift
```diff
@@ -79,19 +87,26 @@ public struct ConditionTrait: TestTrait, SuiteTrait {
...
   /// The source location where this trait was specified.
   public var sourceLocation: SourceLocation
-
-  public func prepare(for test: Test) async throws {
-    let result: Bool
-    var commentOverride: Comment?
-
+  
+  /// Returns the result of evaluating the condition.
+  public func evaluate() async throws -> Evaluation {
```
grynspan, line 92: Perhaps just have this function return a tuple of `isEnabled: Bool, comment: Comment?)`

Uncommon: Do you have a strong preference? I like the enum better because with the tuple version the `comment` field is meaningless when `isEnabled` is true.

grynspan: I would prefer a tuple here.

The enum constrains us in the future if we decide we _do_ want to emit some comment on success. (We haven't planned such a thing, but we could conceivably want that.) Adding an associated value to an enum case is a potentially source-breaking change.

Whereas if we changed what `evaluate()` returns, we could do it by deprecating it and replacing it with a function that returns something else or has a different name or whatever.

Sources/Testing/Traits/ConditionTrait.swift
```diff
@@ -19,13 +19,8 @@
...
 /// - ``Trait/disabled(_:sourceLocation:_:)``
 public struct ConditionTrait: TestTrait, SuiteTrait {
-  /// An enumeration identifying the result of evaluating the condition.
-  public enum Evaluation: Sendable {
-    /// The condition succeeded.
-    case success
-    /// The condition failed, potentially returning a `Comment`.
-    case failure(Comment?)
-  }
+  /// The result of evaluating the condition.
```
grynspan, line 22: This typealias is not necessary, but if you do feel strongly about including it, we'll want the tuple to include labels for both elements so we can document it fully. Something like:

```swift
/// The result of evaluating the condition.
///
/// - Parameters:
///   - wasMet: Whether or not the condition was met.
///   - comment: Optionally, a comment describing the result of evaluating the condition.
public typealias Evaluation = (_ wasMet: Bool, comment: Comment?)
```

(Feel free to tweak the wording—that was just off the top of my head.)

Sources/Testing/Traits/ConditionTrait.swift
```diff
@@ -79,19 +82,22 @@ public struct ConditionTrait: TestTrait, SuiteTrait {
 
   /// The source location where this trait was specified.
   public var sourceLocation: SourceLocation
-
-  public func prepare(for test: Test) async throws {
-    let result: Bool
-    var commentOverride: Comment?
-
+  
+  /// Returns the result of evaluating the condition.
```
grynspan, line 86:
```suggestion
  /// Evaluate this instance's underlying condition.
  ///
  /// - Returns: The result of evaluating this instance's underlying condition.
  ///
```

You may also want to include a discussion paragraph after `- Returns:` if there is additional information for the reader to consume. In particular, it may be worth noting that the result of this function is not cached.

Sources/Testing/Traits/ConditionTrait.swift
Uncommon: I do like having it to avoid repeating the definition. I'll add the other label.

Sources/Testing/Traits/ConditionTrait.swift
```diff
@@ -19,6 +19,9 @@
 /// - ``Trait/disabled(if:_:sourceLocation:)``
 /// - ``Trait/disabled(_:sourceLocation:_:)``
 public struct ConditionTrait: TestTrait, SuiteTrait {
+  /// The result of evaluating the condition.
+  public typealias Evaluation = (Bool, comment: Comment?)
```
stmontgomery, line 23: This typealias will need `@_spi(Experimental)` too

stmontgomery, line 23: "Evaluation**Result**" may be more clear?

Uncommon:
I plan to add some tests soon, at which point I think it will be appropriate to take the PR out of draft mode.

Documentation/Proposals/NNNN-evaluate-condition.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Proposed solution
+
+The proposal is to add a `ConditionTrait.evaluate()` method which returns the
+result of the evaluation. The existing `prepare(for:)` method is updated to call
+`evaluate()` so that the logic is not duplicated.
+
+## Detailed design
+
+The signature is `evaluate() async throws -> Evaluation`, where `Evaluation` is
```
stmontgomery, line 29: Looks like you did take the suggestion to rename the typealias to `EvaluationResult`, so these references to the old name can be updated.

Uncommon: Done.

Documentation/Proposals/NNNN-evaluate-condition.md
```diff
@@ -0,0 +1,57 @@
...
+a `typealias` for the tuple already used for the callback in `Kind.conditional`,
+containing a boolean result and an optional comment.
```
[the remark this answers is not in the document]
Uncommon: Added.

Tests/TestingTests/RunnerTests.swift
```diff
@@ -332,6 +332,23 @@ final class RunnerTests: XCTestCase {
     let test2 = Test(.disabled(if: Bool.random())) { }
     XCTAssertTrue(test2.traits.compactMap { $0 as? ConditionTrait }.allSatisfy { !$0.isConstant })
   }
+  
+  func testEvaluateConditionTrait() async throws {
```
stmontgomery, line 336: We now have a test file named `ConditionTraitTests`, whose tests are implemented using Swift Testing instead of XCTest which this file still uses for legacy reasons. I think this test would make more sense in that new file.

Sources/Testing/Traits/ConditionTrait.swift
```diff
@@ -26,7 +26,6 @@ public struct ConditionTrait: TestTrait, SuiteTrait {
     ///
     /// - Parameters:
     ///   - body: The function to call. The result of this function determines
-    ///     if the condition is satisfied or not.
```
stmontgomery, line 29: Looks like this line may have been deleted by accident?

stmontgomery (commented):
Thanks @Uncommon! Just a couple more comments but this looks close to landing. We can land this before the associated Swift evolution proposal review begins (assuming it moves forward), since the additions are `@_spi`.

Tests/TestingTests/RunnerTests.swift
Uncommon: I had to add `@_spi(Experimental)` to the import in that file, which seems like it should be fine.

stmontgomery (approved):

stmontgomery:
@swift-ci please test

stmontgomery:
Fixes #903
Fixes rdar://142690932

PR #912, opened 2025-01-15
Don't capture arguments to `#expect(exitsWith:)` during macro expansion.

grynspan:
Other than the body closure, the arguments to `#expect(exitsWith:)` are not guaranteed to be literals or otherwise context-free, but we capture them during macro expansion. This can result in obscure errors if the developer writes an exit test with an argument that is computed:

```swift
let foo = ...
await #expect(exitsWith: self.exitCondition(for: foo)) {
  // 🛑 error: closure captures 'foo' before it is declared
  // 🛑 error: enum declaration cannot close over value 'self' defined in outer scope
  // (and other possible errors)
  ...
}
```

This PR removes the macro's compile-time dependency on its `exitCondition` and `sourceLocation` arguments and introduces an `ID` type to identify exit tests (instead of using their source locations as unique IDs.) The `ID` type is meant to be a UUID, but for the moment I've just strung together two 64-bit integers instead to avoid a dependency on Foundation or platform API.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -69,6 +69,7 @@ public struct __ExitTest: Sendable, ~Copyable {
     }
   }
 
+  /// A value that uniquely identifies this instance.
   public var id: ID
 
   /// The body closure of the exit test.
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -52,16 +52,24 @@ public typealias ExitTest = __ExitTest
 @available(*, unavailable, message: "Exit tests are not available on this platform.")
 #endif
 public struct __ExitTest: Sendable, ~Copyable {
-  /// The expected exit condition of the exit test.
-  @_spi(ForToolsIntegrationOnly)
-  public var expectedExitCondition: ExitCondition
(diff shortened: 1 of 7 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -52,16 +52,25 @@ public typealias ExitTest = __ExitTest
...
 public struct __ExitTest: Sendable, ~Copyable {
-  /// The expected exit condition of the exit test.
-  @_spi(ForToolsIntegrationOnly)
-  public var expectedExitCondition: ExitCondition
+  /// A type whose instances uniquely identify instances of `__ExitTest`.
+  public struct ID: Sendable, Equatable, Codable {
+    /// An underlying UUID (stored as two `UInt64` values to avoid relying on
+    /// `UUID` from Foundation or any platform-specific interfaces.)
+    private var _lo: UInt64
+    private var _hi: UInt64
```
stmontgomery, line 60: Any reason not to store them as a tuple, matching the initializer parameter?

grynspan: Tuples don't conform to `Codable`, so we'd have to manually implement codable conformance, and I am way too lazy for that.

stmontgomery (approved):

PR #913, opened 2025-01-15
[WIP] Add an experimental `.serialized(.globally)` trait.

grynspan:
This PR adds a variant of the `.serialized` that causes annotated tests and suites to run in serial _after_ all other test content has run, in effect allowing them to opt out of parallelization entirely.

This PR introduces a new concept to a runner plan, "stages." The names of stages are arbitrary, but the point is that they run in a fixed order and different work is performed in each stage. Test _functions_ always run in a single stage, while test _suites_ may get smeared across multiple stages. Normal test work is performed in the first stage, with a second stage that runs everything in it serially. Hence, any test that gets put in that second stage runs in a globally serialized order.

"Globally" here does not refer to multi-process parallelization, which is a whole 'nother animal. If we ever add support for that, we'll presumably want tests in this stage to be serialized even between processes.

Stages aren't API or SPI, although we could potentially promote it to either in the future.

Resolves rdar://135288463.

### Checklist:


spnkr:
Yes!! Controlling concurrent/serialized execution of tests in swift testing is not great at the moment. It's my biggest pain point actually.

grynspan:
@spnkr This is just an experimental branch and does not imply we're actually going to make a feature out of it, sorry. :)

PR #914, opened 2025-01-16
Refactor test content discovery to produce a sequence.

grynspan:
This PR does away with `enumerateTestContent {}` and replaces it with `discover()`, a function that returns a sequence of `TestContentRecord` instances that the caller can then inspect. This simplifies the logic in callers by letting them use sequence operations/monads like `map()`, and simplifies the logic in the implementation by doing away with intermediate tuples and allowing it to just map section bounds to sequences of `TestContentRecord` values.

Evaluation of the accessor functions for test content records is now lazier, allowing us to potentially leverage the `context` field of a test content record to avoid calling accessors for records we know won't match due to the content of that field. (This functionality isn't currently needed, but could be useful for future test content types.)

### Checklist:

### DIFF
--- a/Sources/Testing/Discovery.swift
+++ b/Sources/Testing/Discovery.swift
@@ -120,7 +120,7 @@ extension TestContentRecord where T: TestContent & ~Copyable {
   ///
   /// If this function is called more than once on the same instance, a new
   /// value is created on each call.
-  func load<R>(withHint hint: T.TestContentAccessorHint? = nil) -> R? where R == T.TestContentAccessorResult, R: ~Copyable {
+  func load(withHint hint: T.TestContentAccessorHint? = nil) -> T.TestContentAccessorResult? {
     guard let accessor = _record.accessor.map(swt_resign) else {
       return nil
     }
@@ -150,7 +150,7 @@ extension TestContent where Self: ~Copyable {
   /// - Returns: A sequence of instances of ``TestContentRecord``. Only test
   ///   content records matching this ``TestContent`` type's requirements are
   ///   included in the sequence.
-  static func allTestContentRecords() -> some Sequence<TestContentRecord<Self>> {
+  static func allTestContentRecords() -> LazySequence<some Sequence<TestContentRecord<Self>>> {
     SectionBounds.all(.testContent).lazy.flatMap { sb in
       sb.buffer.withMemoryRebound(to: __TestContentRecord.self) { records in
         records.lazy
--- a/Sources/Testing/Discovery.swift
+++ b/Sources/Testing/Discovery.swift
@@ -68,8 +68,9 @@ protocol TestContent: ~Copyable {
 
 /// A type describing a test content record of a particular (known) type.
 ///
-/// Instances of this type can be created by calling ``TestContent/discover()``
-/// on a type that conforms to ``TestContent``.
+/// Instances of this type can be created by calling
+/// ``TestContent/allTestContentRecords()`` on a type that conforms to
+/// ``TestContent``.
 ///
 /// This type is not part of the public interface of the testing library. In the
 /// future, we could make it public if we want to support runtime discovery of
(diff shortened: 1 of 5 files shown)


grynspan:
@swift-ci test

grynspan:
Note the rendering of the diff is _terrible_. Might help to just look at one side or the other.

Sources/Testing/Test+Discovery.swift
```diff
@@ -22,59 +22,51 @@ extension Test: TestContent {
...
       // the legacy and new mechanisms, but we can set an environment variable
       // to explicitly select one or the other. When we remove legacy support,
       // we can also remove this enumeration and environment variable check.
-      enum DiscoveryMode {
-        case tryBoth
-        case newOnly
-        case legacyOnly
-      }
-      let discoveryMode: DiscoveryMode = switch Environment.flag(named: "SWT_USE_LEGACY_TEST_DISCOVERY") {
+      let (useNewMode, useLegacyMode) = switch Environment.flag(named: "SWT_USE_LEGACY_TEST_DISCOVERY") {
```
grynspan, line 34: This change makes it easier to disable this code in #880.

Tests/TestingTests/MiscellaneousTests.swift
```diff
@@ -615,43 +615,45 @@ struct MiscellaneousTests {
       0xABCD1234,
       0,
       { outValue, hint in
-        if let hint, hint.loadUnaligned(as: TestContentAccessorHint.self) != expectedHint {
+        if let hint, hint.load(as: TestContentAccessorHint.self) != expectedHint {
```
grynspan, line 618: The hint is always well-aligned.

grynspan:
@swift-ci test

Sources/Testing/Discovery.swift
```diff
@@ -128,45 +138,23 @@ extension TestContent where Self: ~Copyable {
...
-      if let accessor = record.accessor, let result = _callAccessor(accessor, withHint: hint) {
-        // Call the callback.
-        body(imageAddress, result, record.context, &stop)
-        if stop {
-          break
-        }
+  /// - Returns: A sequence of instances of ``TestContentRecord``. Only test
+  ///   content records matching this ``TestContent`` type's requirements are
+  ///   included in the sequence.
+  static func discover() -> some Sequence<TestContentRecord<Self>> {
```
stmontgomery, line 152: Can you talk about the choice of name? I might suggest a property instead of a function and naming it something like `allRecords` or `allTestContentRecords`. At the call site, `discover()` feels more like an imperative instruction rather than a call which returns a value to me.

Tests/TestingTests/MiscellaneousTests.swift
```diff
@@ -615,43 +615,45 @@ struct MiscellaneousTests {
...
-      DiscoverableTestContent.enumerateTestContent { _, value, context, _ in
-        if value == DiscoverableTestContent.expectedResult && context == DiscoverableTestContent.expectedContext {
-          found()
-        }
-      }
-    }
-
-    await confirmation("Can find a test record with matching hint") { found in
+    // Check the type of the test record sequence (it should be lazy.)
+    let allRecords = DiscoverableTestContent.discover()
```
stmontgomery, line 631: Not strictly related to this PR, but it's occurring to me seeing this that it might make sense to migrate these tests (and their fixtures) related to test discovery into a distinct new test file and suite.

Sources/Testing/Discovery.swift
grynspan: Changed.

Tests/TestingTests/MiscellaneousTests.swift
grynspan: I had the same thought, but I am going to wait until the test content emission PR has landed and there are more tests we can think of to write.

grynspan:
@swift-ci test

stmontgomery (approved):

grynspan:
@swift-ci test

grynspan:
@swift-ci test

PR #915, opened 2025-01-16
Limit the amount of value reflection data expectation checking collects by default

stmontgomery:
This modifies the internal logic which recursively collects reflection data about values passed to `#expect(...)` and similar expectation APIs so that it imposes an artificial limit on the amount of data collected.

### Motivation:

Some users have attempted to use `#expect` with instances of values with large collections or with deep value hierarchies involving many sub-properties, etc. In these situations they have noticed that the automatic value reflection is adding a noticeable amount of overhead to the overall test execution time, and would like to reduce that by default (near term), as well as have some ability to control that so the behavior could be tweaked for particular tests (in the future).

### Modifications:

- Add some new `Configuration` properties controlling data collection behaviors and setting default limits.
- Consult these new configuration properties in the relevant places in `Expression.Value`.
- Make `Expression.Value.init(reflecting:)` optional and return `nil` when value reflection is disabled in the configuration.
- Add a new initializer `Expression.Value.init(describing:)` as a lighter-weight alternative to `init(reflecting:)` which only forms a string description of its subject, and adopt this as a fallback in places affected by the new optional in the previous bullet.
- Add representative new unit tests.

### Checklist:

Fixes rdar://138208832

### DIFF
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -286,7 +286,7 @@ public struct Configuration: Sendable {
     /// this maximum count. After this maximum is reached, all subsequent
     /// elements are omitted and a single placeholder child is added indicating
     /// the number of elements which have been truncated.
-    public var maximumCollectionCount: Int = 10
+    public var maximumCollectionCount = 10
 
     /// The maximum depth of children that can be included in the reflection of
     /// a checked expectation value.
@@ -303,7 +303,7 @@ public struct Configuration: Sendable {
     ///   Since optionals are common, the default value of this property is
     ///   somewhat larger than it otherwise would be in an attempt to make the
     ///   defaults useful for real-world tests.
-    public var maximumChildDepth: Int = 10
+    public var maximumChildDepth = 10
   }
 }
 
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -259,43 +259,52 @@ public struct Configuration: Sendable {
   /// The test case filter to which test cases should be filtered when run.
   public var testCaseFilter: TestCaseFilter = { _, _ in true }
 
-  // MARK: - Expectation checking
+  // MARK: - Expectation value reflection
 
-  /// Whether or not reflection of values in expressions checked by expectations
-  /// is enabled.
+  /// The options to use when reflecting values in expressions checked by
+  /// expectations, or `nil` if reflection is disabled.
   ///
-  /// When the value of this property is `false`, value reflection is disabled
-  /// and values will only be represented using `String(describing:)` instead of
-  /// using `Mirror` to recursively reflect their substructure.
-  public var isValueReflectionEnabled: Bool = true
+  /// When the value of this property is a non-`nil` instance, values checked by
+  /// expressions will be reflected using `Mirror` and the specified options
+  /// will influence how that reflection is formed. Otherwise, when its value is
+  /// `nil`, value reflection will not use `Mirror` and instead will use
+  /// `String(describing:)`.
+  ///
+  /// The default value of this property is an instance of ``ValueReflectionOptions-swift.struct``
+  /// with its properties initialized to their default values.
+  public var valueReflectionOptions: ValueReflectionOptions? = .init()
 
-  /// The maximum number of elements that can included in a single child
-  /// collection when reflecting a value checked by an expectation.
-  ///
-  /// When ``Expression/Value/init(reflecting:)`` is reflecting a value and it
-  /// encounters a child value which is a collection, it consults the value of
-  /// this property and only includes the children of that collection up to this
-  /// maximum count. After this maximum is reached, all subsequent elements are
-  /// omitted and a single placeholder child is added indicating the number of
-  /// elements which have been truncated.
-  public var maximumValueReflectionCollectionCount: Int = 10
+  /// A type describing options to use when forming a reflection of a value
+  /// checked by an expectation.
+  public struct ValueReflectionOptions: Sendable {
+    /// The maximum number of elements that can included in a single child
+    /// collection when reflecting a value checked by an expectation.
+    ///
+    /// When ``Expression/Value/init(reflecting:)`` is reflecting a value and it
+    /// encounters a child value which is a collection, it consults the value of
+    /// this property and only includes the children of that collection up to
+    /// this maximum count. After this maximum is reached, all subsequent
+    /// elements are omitted and a single placeholder child is added indicating
+    /// the number of elements which have been truncated.
+    public var maximumCollectionCount: Int = 10
 
-  /// The maximum depth of children that can be included in the reflection of a
-  /// checked expectation value.
-  ///
-  /// When ``Expression/Value/init(reflecting:)`` is reflecting a value, it
-  /// recursively reflects that value's children. Before doing so, it consults
-  /// the value of this property to determine the maximum depth of the children
-  /// to include. After this maximum depth is reached, all children at deeper
-  /// levels are omitted and the ``Expression/Value/isTruncated`` property is
-  /// set to `true` to reflect that the reflection is incomplete.
-  ///
-  /// - Note: `Optional` values contribute twice towards this maximum, since
-  ///   their mirror represents the wrapped value as a child of the optional.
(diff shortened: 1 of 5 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

briancroom (approved):

Sources/Testing/Running/Configuration.swift
```diff
@@ -258,6 +258,44 @@ public struct Configuration: Sendable {
...
+
+  // MARK: - Expectation checking
+
+  /// Whether or not reflection of values in expressions checked by expectations
+  /// is enabled.
+  ///
+  /// When the value of this property is `false`, value reflection is disabled
+  /// and values will only be represented using `String(describing:)` instead of
+  /// using `Mirror` to recursively reflect their substructure.
+  public var isValueReflectionEnabled: Bool = true
```
grynspan, line 270:
```suggestion
  public var isValueReflectionEnabled = true
```

Sources/Testing/Running/Configuration.swift
```diff
@@ -258,6 +258,44 @@ public struct Configuration: Sendable {
...
+  /// The maximum number of elements that can included in a single child
+  /// collection when reflecting a value checked by an expectation.
+  ///
+  /// When ``Expression/Value/init(reflecting:)`` is reflecting a value and it
+  /// encounters a child value which is a collection, it consults the value of
+  /// this property and only includes the children of that collection up to this
+  /// maximum count. After this maximum is reached, all subsequent elements are
+  /// omitted and a single placeholder child is added indicating the number of
+  /// elements which have been truncated.
+  public var maximumValueReflectionCollectionCount: Int = 10
```
grynspan, line 281:
```suggestion
  public var maximumValueReflectionCollectionCount = 10
```

Sources/Testing/Running/Configuration.swift
```diff
@@ -258,6 +258,44 @@ public struct Configuration: Sendable {
...
+  /// to include. After this maximum depth is reached, all children at deeper
+  /// levels are omitted and the ``Expression/Value/isTruncated`` property is
+  /// set to `true` to reflect that the reflection is incomplete.
+  ///
+  /// - Note: `Optional` values contribute twice towards this maximum, since
+  ///   their mirror represents the wrapped value as a child of the optional.
+  ///   Since optionals are common, the default value of this property is
+  ///   somewhat larger than it otherwise would be in an attempt to make the
+  ///   defaults useful for real-world tests.
+  public var maximumValueReflectionChildDepth: Int = 10
```
grynspan, line 298:
```suggestion
  public var maximumValueReflectionChildDepth = 10
```

Sources/Testing/Running/Configuration.swift
```diff
@@ -258,6 +258,44 @@ public struct Configuration: Sendable {
 
   /// The test case filter to which test cases should be filtered when run.
   public var testCaseFilter: TestCaseFilter = { _, _ in true }
+
+  // MARK: - Expectation checking
+
+  /// Whether or not reflection of values in expressions checked by expectations
+  /// is enabled.
+  ///
+  /// When the value of this property is `false`, value reflection is disabled
```
grynspan, line 267: It may be easier for the reader to understand if you express this concept in terms of "when the value is `true`".

Sources/Testing/Running/Configuration.swift
```diff
@@ -258,6 +258,44 @@ public struct Configuration: Sendable {
 
   /// The test case filter to which test cases should be filtered when run.
   public var testCaseFilter: TestCaseFilter = { _, _ in true }
+
+  // MARK: - Expectation checking
+
+  /// Whether or not reflection of values in expressions checked by expectations
```
grynspan, line 264: Rather than 3 separate properties, how about a `ValueReflectionOptions` structure?

Sources/Testing/Running/Configuration.swift
```diff
@@ -258,6 +258,44 @@ public struct Configuration: Sendable {
...
+  /// checked expectation value.
+  ///
+  /// When ``Expression/Value/init(reflecting:)`` is reflecting a value, it
+  /// recursively reflects that value's children. Before doing so, it consults
+  /// the value of this property to determine the maximum depth of the children
+  /// to include. After this maximum depth is reached, all children at deeper
+  /// levels are omitted and the ``Expression/Value/isTruncated`` property is
+  /// set to `true` to reflect that the reflection is incomplete.
+  ///
+  /// - Note: `Optional` values contribute twice towards this maximum, since
```
grynspan, line 293: Perhaps we should special-case optional values, then?

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -167,14 +177,32 @@ public struct __Expression: Sendable {
...
+
     /// Initialize an instance of this type describing the specified subject and
     /// its children (if any).
     ///
     /// - Parameters:
-    ///   - subject: The subject this instance should describe.
-    init(reflecting subject: Any) {
+    ///   - subject: The subject this instance should reflect.
+    init?(reflecting subject: Any) {
+      let configuration = Configuration.current ?? .init()
```
grynspan, line 199: Is this going to be called frequently? Perhaps it should be an argument the caller can pass, and then the outermost caller in the operation can pass `.current` so we only hit the task local once?

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -236,24 +280,19 @@ public struct __Expression: Sendable {
...
-      }
-
       if shouldIncludeChildren && (!mirror.children.isEmpty || isCollection) {
-        self.children = mirror.children.map { child in
-          Self(_reflecting: child.value, label: child.label, seenObjects: &seenObjects)
+        var children: [Self] = []
+        for (index, child) in mirror.children.enumerated() {
+          if isCollection && index >= configuration.maximumValueReflectionCollectionCount {
+            isTruncated = true
+            let message = "(\(mirror.children.count - index) elements (out of \(mirror.children.count) total) omitted for brevity.)"
```
grynspan, line 288:
```suggestion
            let message = "(\(mirror.children.count - index) of \(mirror.children.count) elements omitted for brevity.)"
```

To avoid the double parentheses. Unsure if the trailing period is appropriate.

grynspan (changes requested):

Sources/Testing/SourceAttribution/Expression.swift
stmontgomery: Good call, I will apply this to the recursive initializer below

Sources/Testing/Running/Configuration.swift
stmontgomery: I've taken that suggestion, and made the options property `Optional`, with `nil` indicating that reflection is disabled.

grynspan: Makes sense.

Sources/Testing/Running/Configuration.swift
stmontgomery: I like having the types of properties with `public` or greater access level state their type explicitly, to avoid the possibility of the inferred type changing without our knowledge and causing a API breakage. I see now that we have not done this with some trivial properties in `Configuration`, but now that I've moved most of these into a nested struct I've used this style within that new struct at least. I'd like to apply this coding style more broadly in a subsequent PR.

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Configuration.swift
stmontgomery: I've thought about that before but not sure it's worthwhile in general. I mainly just wanted to call attention to one major consideration for how the default value for this property should be determined.

Sources/Testing/Running/Configuration.swift
grynspan: We pretty much never do this for variables to which we assign values. We only specify types on variables that will be assigned values later. Consistency would be good (either way.)

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Configuration.swift
stmontgomery: Already merged, but I did strip off the explicit types from these properties for now, for consistency!

PR #916, opened 2025-01-17
Fix the build for WASI

tshortli:
Avoids a compiler error when `SWT_NO_DYNAMIC_LINKING` is true.

### Motivation:

Fixes the broken build, which is blocking Swift compiler PR testing.

### Modifications:

Makes `imageAddress` a computed property when `SWT_NO_DYNAMIC_LINKING` is true so that it can be marked `@available(*, unavailable)`.


tshortli:
@swift-ci please test

stephentyrone (approved):

ktoso:
Merging to unblock Swift CI.

real fix coming by @grynspan i hear

PR #918, opened 2025-01-17
Don't allocate a buffer on the heap when enumerating type metadata.

grynspan:
This PR eliminates the need for us to allocate a large buffer to contain type metadata pointers we discover in type metadata sections using the legacy test discovery mechanism.

Why do I keep opening these PRs? Because for each line of C++ I remove from the Swift toolchain, I get a cookie! 🍪

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):
Nice!

PR #921, opened 2025-01-21
[TestingMacros] Remove reference to implicit backtracing import.

al45tair:
The `_Backtracing` module was never implicitly imported by default, so this wasn't necessary anyway, but adding it is a blocker for SE-0419.

rdar://143310300

### DIFF
--- a/Sources/TestingMacros/CMakeLists.txt
+++ b/Sources/TestingMacros/CMakeLists.txt
@@ -108,8 +108,7 @@ target_sources(TestingMacros PRIVATE
   TestingMacrosMain.swift)
 
 target_compile_options(TestingMacros PRIVATE
-  "SHELL:-Xfrontend -disable-implicit-string-processing-module-import"
-  "SHELL:-Xfrontend -disable-implicit-backtracing-module-import")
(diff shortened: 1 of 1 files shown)


al45tair:
> Do we need any SwiftPM manifest changes?

I think this was the only reference I could find. I only went looking after my Windows build kept unexpectedly failing because of something using this option.

grynspan:
I suppose I'm curious _why_ this caused a failure.

grynspan:
@swift-ci test

al45tair:
> I suppose I'm curious _why_ this caused a failure.

To be fair, I don't know that it's specifically this that's causing the failure, but _something_ is passing the `-disable-implicit-backtracing-support` argument, so I'm trying to eliminate references to that (because it's being removed).

grynspan:
@swift-ci test macOS

grynspan:
CI is still down, I think. Please do not merge until all three jobs have completed successfully. Thanks!

grynspan:
@swift-ci test

stmontgomery (approved):

PR #922, opened 2025-01-21
Use `swift_getErrorValue()` unless Objective-C is available.

grynspan:
Tweak the logic for when we use the runtime-internal `swift_getErrorValue()` function in our `_swift_willThrow` handler so that we use it if the runtime does not support Objective-C interop. Currently, we toggle based on whether we're on an Apple platform, but this isn't quite right.

Resolves rdar://143328537.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

briancroom (approved):

PR #924, opened 2025-01-24
Cleaning recently changed docs

iamleeg:
Improving the style and explanation of recent additions to the documentation.

### Motivation:

The recent changes to test traits are great, and come with some very nice documentation. I've done what I can to improve the style and the wording to make it easier to understand, and to make the style consistent with other documentation in the project.

### Modifications:

Changed the documentation for `Trait` and associated types.

### Result:

There are no behavioral changes in this PR, but I hope that the existing behavior is easier to use with better documentation.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/Traits.md
+++ b/Sources/Testing/Testing.docc/Traits.md
@@ -10,14 +10,14 @@ See https://swift.org/LICENSE.txt for license information
 See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
 
-Add traits to tests to annotate them or customize their behavior.
+Annotate tests and suites, and customize their behavior.
 
 ## Overview
 
 Pass built-in traits to test functions or suite types to comment, categorize, 
-classify, and modify runtime behaviors. You can also use the ``Trait``, ``TestTrait``, 
-and ``SuiteTrait`` protocols to create your own types that customize the 
-behavior of test functions.
+classify, and modify the runtime behavior of test functions. Implement the
+``Trait``, ``TestTrait``, and ``SuiteTrait`` protocols to create your own
+types that customize the behavior of test functions.
 
 ## Topics
 
--- a/Sources/Testing/Testing.docc/Traits/Trait.md
+++ b/Sources/Testing/Testing.docc/Traits/Trait.md
@@ -20,17 +20,15 @@ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 - ``Trait/disabled(if:_:sourceLocation:)``
 - ``Trait/disabled(_:sourceLocation:_:)``
 
-### Limiting the running time of tests
+### Controlling how tests are run
 
 - ``Trait/timeLimit(_:)-4kzjp``
- 
-### Running tests serially or in parallel
-
 - ``Trait/serialized``
-
-### Categorizing tests
+ 
+### Categorizing tests and adding information
 
 - ``Trait/tags(_:)``
+- ``Trait/comments``
 
 ### Associating bugs
 
@@ -38,16 +36,9 @@ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 - ``Trait/bug(_:id:_:)-10yf5``
 - ``Trait/bug(_:id:_:)-3vtpl``
 
-### Adding information to tests
-
-- ``Trait/comments``
-
-### Preparing internal state
-
-- ``Trait/prepare(for:)-3s3zo``
-
-### Providing custom execution scope for tests
+### Run code before and after test functions
 
 - ``TestScoping``
 - ``Trait/scopeProvider(for:testCase:)-cjmg``
 - ``Trait/TestScopeProvider``
+- ``Trait/prepare(for:)-3s3zo``
--- a/Sources/Testing/Traits/Bug.swift
+++ b/Sources/Testing/Traits/Bug.swift
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-/// A type representing a bug report tracked by a test.
+/// A type that represents a bug report tracked by a test.
 ///
 /// To add this trait to a test, use one of the following functions:
 ///
@@ -16,7 +16,7 @@
 /// - ``Trait/bug(_:id:_:)-10yf5``
 /// - ``Trait/bug(_:id:_:)-3vtpl``
 public struct Bug {
-  /// A URL linking to more information about the bug, if available.
+  /// A URL that links to more information about the bug, if available.
   ///
   /// The value of this property represents a URL conforming to
   /// [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt).
@@ -59,42 +59,42 @@ extension Bug: TestTrait, SuiteTrait {
 }
 
 extension Trait where Self == Bug {
-  /// Construct a bug to track with a test.
+  /// Constructs a bug to track with a test.
   ///
   /// - Parameters:
-  ///   - url: A URL referring to this bug in the associated bug-tracking
+  ///   - url: A URL that refers to this bug in the associated bug-tracking
   ///     system.
   ///   - title: Optionally, the human-readable title of the bug.
   ///
-  /// - Returns: An instance of ``Bug`` representing the specified bug.
+  /// - Returns: An instance of ``Bug`` that represents the specified bug.
   public static func bug(_ url: _const String, _ title: Comment? = nil) -> Self {
     Self(url: url, title: title)
   }
 
-  /// Construct a bug to track with a test.
+  /// Constructs a bug to track with a test.
   ///
   /// - Parameters:
-  ///   - url: A URL referring to this bug in the associated bug-tracking
+  ///   - url: A URL that refers to this bug in the associated bug-tracking
   ///     system.
   ///   - id: The unique identifier of this bug in its associated bug-tracking
   ///     system.
   ///   - title: Optionally, the human-readable title of the bug.
   ///
-  /// - Returns: An instance of ``Bug`` representing the specified bug.
+  /// - Returns: An instance of ``Bug`` that represents the specified bug.
   public static func bug(_ url: _const String? = nil, id: some Numeric, _ title: Comment? = nil) -> Self {
     Self(url: url, id: String(describing: id), title: title)
   }
 
-  /// Construct a bug to track with a test.
+  /// Constructs a bug to track with a test.
   ///
   /// - Parameters:
-  ///   - url: A URL referring to this bug in the associated bug-tracking
+  ///   - url: A URL that refers to this bug in the associated bug-tracking
   ///     system.
   ///   - id: The unique identifier of this bug in its associated bug-tracking
   ///     system.
   ///   - title: Optionally, the human-readable title of the bug.
   ///
-  /// - Returns: An instance of ``Bug`` representing the specified bug.
+  /// - Returns: An instance of ``Bug`` that represents the specified bug.
   public static func bug(_ url: _const String? = nil, id: _const String, _ title: Comment? = nil) -> Self {
     Self(url: url, id: id, title: title)
   }
--- a/Sources/Testing/Traits/Comment+Macro.swift
+++ b/Sources/Testing/Traits/Comment+Macro.swift
@@ -9,13 +9,13 @@
 //
 
 extension Trait where Self == Comment {
-  /// Construct a comment related to a test from a single-line source code
+  /// Constructs a comment related to a test from a single-line source code
   /// comment near it.
   ///
   /// - Parameters:
   ///   - comment: The comment about the test.
   ///
-  /// - Returns: An instance of ``Comment`` containing the specified comment.
+  /// - Returns: An instance of ``Comment`` that contains the specified comment.
   ///
   /// - Warning: This function is used to implement the `@Test` macro. Do not
   ///   call it directly.
@@ -23,13 +23,13 @@ extension Trait where Self == Comment {
     Self(rawValue: comment, kind: .line)
   }
 
-  /// Construct a comment related to a test from a source code block comment
+  /// Constructs a comment related to a test from a source code block comment
   /// near it.
   ///
   /// - Parameters:
   ///   - comment: The comment about the test.
   ///
-  /// - Returns: An instance of ``Comment`` containing the specified comment.
+  /// - Returns: An instance of ``Comment`` that contains the specified comment.
   ///
   /// - Warning: This function is used to implement the `@Test` macro. Do not
   ///   call it directly.
@@ -37,14 +37,13 @@ extension Trait where Self == Comment {
     Self(rawValue: comment, kind: .block)
   }
 
-  /// Construct a comment related to a test from a single-line
-  /// [Markup](https://github.com/swiftlang/swift/blob/main/docs/DocumentationComments.md)
+  /// Constructs a comment related to a test from a single-line markup
   /// comment near it.
   ///
   /// - Parameters:
   ///   - comment: The comment about the test.
   ///
-  /// - Returns: An instance of ``Comment`` containing the specified comment.
+  /// - Returns: An instance of ``Comment`` that contains the specified comment.
   ///
   /// - Warning: This function is used to implement the `@Test` macro. Do not
   ///   call it directly.
@@ -52,14 +51,13 @@ extension Trait where Self == Comment {
     Self(rawValue: comment, kind: .documentationLine)
   }
 
-  /// Construct a comment related to a test from a
-  /// [Markup](https://github.com/swiftlang/swift/blob/main/docs/DocumentationComments.md)
+  /// Constructs a comment related to a test from a markup
   /// block comment near it.
   ///
   /// - Parameters:
   ///   - comment: The comment about the test.
   ///
-  /// - Returns: An instance of ``Comment`` containing the specified comment.
+  /// - Returns: An instance of ``Comment`` that contains the specified comment.
   ///
   /// - Warning: This function is used to implement the `@Test` macro. Do not
   ///   call it directly.
--- a/Sources/Testing/Traits/Comment.swift
+++ b/Sources/Testing/Traits/Comment.swift
@@ -8,22 +8,21 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-/// A type representing a comment related to a test.
+/// A type that represents a comment related to a test.
 ///
-/// This type may be used to provide context or background information about a
+/// Use this type to provide context or background information about a
 /// test's purpose, explain how a complex test operates, or include details
 /// which may be helpful when diagnosing issues recorded by a test.
 ///
 /// To add a comment to a test or suite, add a code comment before its `@Test`
 /// or `@Suite` attribute. See <doc:AddingComments> for more details.
 ///
-/// - Note: This type is not intended to reference bugs related to a test.
-///   Instead, use ``Trait/bug(_:_:)``, ``Trait/bug(_:id:_:)-10yf5``, or
-///   ``Trait/bug(_:id:_:)-3vtpl``.
+/// - Note: To reference bugs related to a test, use ``Trait/bug(_:_:)``,
+///   ``Trait/bug(_:id:_:)-10yf5``, or ``Trait/bug(_:id:_:)-3vtpl``.
 public struct Comment: RawRepresentable, Sendable {
-  /// The single comment string contained in this instance.
+  /// The single comment string that this comment contains.
   ///
-  /// To obtain the complete set of comments applied to a test, see
+  /// To get the complete set of comments applied to a test, see
   /// ``Test/comments``.
   public var rawValue: String
 
@@ -31,42 +30,41 @@ public struct Comment: RawRepresentable, Sendable {
     self.rawValue = rawValue
   }
 
-  /// An enumeration describing the possible kind of a comment.
+  /// An enumeration that describes a comment's kind.
   @_spi(ForToolsIntegrationOnly)
   public enum Kind: Sendable {
-    /// This comment came from a single-line comment in the test's source code
-    /// starting with `//`.
+    /// This comment is in a single-line comment in the test's source code
+    /// that starts with `//`.
     case line
 
-    /// This comment came from a block comment in the test's source code
-    /// starting with `/*` and ending with `*/`.
+    /// This comment is in a block comment in the test's source code
+    /// that starts with `/*` and ends with `*/`.
     case block
 
-    /// This comment came from a single-line [Markup](https://github.com/swiftlang/swift/blob/main/docs/DocumentationComments.md)
-    /// comment in the test's source code starting with `///`.
+    /// This comment is in a single-line [Markup](https://github.com/swiftlang/swift/blob/main/docs/DocumentationComments.md)
+    /// comment in the test's source code that starts with `///`.
     case documentationLine
 
-    /// This comment came from a block [Markup](https://github.com/swiftlang/swift/blob/main/docs/DocumentationComments.md)
-    /// comment in the test's source code starting with `/**` and ending with
+    /// This comment is in a block [Markup](https://github.com/swiftlang/swift/blob/main/docs/DocumentationComments.md)
+    /// comment in the test's source code that starts with `/**` and ends with
     /// `*/`.
     case documentationBlock
 
-    /// This comment came from an explicit call to ``Trait/comment(_:)``.
+    /// This comment comes from an explicit call to ``Trait/comment(_:)``.
     case trait
 
-    /// This comment was initialized from a string literal.
+    /// This comment is initialized from a string literal.
     case stringLiteral
   }
 
-  /// The kind of this comment, if known.
+  /// This comment's kind.
   ///
-  /// If this instance was created with a call to ``init(rawValue:)``, the value
-  /// of this property is `nil`. Otherwise, it can be used to determine which
-  /// kind of comment is represented.
+  /// For comments you create by calling ``init(rawValue:)``, the value
+  /// of this property is `nil`. Otherwise, it represents the comment's kind.
   @_spi(ForToolsIntegrationOnly)
   public var kind: Kind?
 
-  /// Initialize an instance of this type.
+  /// Initialize a comment from a string value.
   ///
   /// - Parameters:
   ///   - rawValue: The string value of the comment.
@@ -137,13 +135,13 @@ extension Test {
     traits.flatMap(\.comments)
   }
 
-  /// The complete set of comments about this test from all traits of a certain
+  /// The complete set of comments about this test from all traits of the given
   /// type.
   ///
   /// - Parameters:
-  ///   - traitType: The type of ``Trait`` whose comments should be returned.
+  ///   - traitType: The type of ``Trait`` for which comments should be returned.
   ///
-  /// - Returns: The comments found for the specified test trait type.
+  /// - Returns: The comments about this test for the specified test trait type.
   @_spi(Experimental)
   public func comments<T>(from traitType: T.Type) -> [Comment] where T: Trait {
     traits.lazy
--- a/Sources/Testing/Traits/ConditionTrait+Macro.swift
+++ b/Sources/Testing/Traits/ConditionTrait+Macro.swift
@@ -35,8 +35,8 @@ extension Trait where Self == ConditionTrait {
     return "\(platformName) \(version.major).\(minorVersion).\(patchVersion)"
   }
 
-  /// Create a trait controlling availability of a test based on an
-  /// `@available()` attribute applied to it.
+  /// Create a trait that controls a test's availability based on a
+  /// platform availability attribute.
   ///
   /// - Parameters:
   ///   - platformName: The name of the platform specified in the `@available()`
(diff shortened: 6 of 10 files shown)


Sources/Testing/Testing.docc/Traits.md
```diff
@@ -10,14 +10,14 @@ See https://swift.org/LICENSE.txt for license information
...
+Annotate tests and suites, and customize their behavior.
 
 ## Overview
 
 Pass built-in traits to test functions or suite types to comment, categorize, 
-classify, and modify runtime behaviors. You can also use the ``Trait``, ``TestTrait``, 
-and ``SuiteTrait`` protocols to create your own types that customize the 
-behavior of test functions.
+classify, and modify the runtime behavior of test functions. Implement the
+``Trait``, ``TestTrait``, and ``SuiteTrait`` protocols to create your own
```
stmontgomery, line 19: > Implement the `Trait`, `TestTrait`, and `SuiteTrait` protocols…

This is pre-existing, but I think we could reasonable omit the base `Trait` protocol from that list. It's the common parent of the other two protocols, and you typically don't conform to it directly, but instead choose one or both of the child protocols to implement conformance to.

Sources/Testing/Testing.docc/Traits.md
```diff
@@ -10,14 +10,14 @@ See https://swift.org/LICENSE.txt for license information
...
+types that customize the behavior of test functions.
```
stmontgomery, line 20: Another (pre-existing) reference here to test _functions_ specifically.

Sources/Testing/Testing.docc/Traits/Trait.md
```diff
@@ -20,34 +20,25 @@ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
...
-### Adding information to tests
-
-- ``Trait/comments``
-
-### Preparing internal state
-
-- ``Trait/prepare(for:)-3s3zo``
-
-### Providing custom execution scope for tests
+### Run code before and after test functions
```
stmontgomery, line 39:
```suggestion
### Run code before and after a test or suite
```
perhaps?

Sources/Testing/Traits/TimeLimitTrait.swift
```diff
@@ -10,17 +10,16 @@
...
   ///
-  /// This type is intended for use specifically for specifying test timeouts
-  /// with ``TimeLimitTrait``. It is used instead of Swift's built-in `Duration`
-  /// type because test timeouts do not support high-precision, arbitrarily
-  /// short durations. The smallest allowed unit of time is minutes.
+  /// Use this type to specify a test timeout with ``TimeLimitTrait``.
+  /// Use this type instead of Swift's built-in `Duration`
+  /// type because the testing library doesn't support high-precision,
+  /// arbitrarily short durations for test timeouts.
+  /// The smallest allowed unit of time is minutes.
```
stmontgomery, line 22: Purely stylistic, but it seems like we could re-flow the sentences here to condense these lines

Sources/Testing/Traits/TimeLimitTrait.swift
```diff
@@ -54,35 +52,34 @@ public struct TimeLimitTrait: TestTrait, SuiteTrait {
...
-  /// The time limit amount specified by `timeLimit` may be reduced if the
-  /// testing library is configured to enforce a maximum per-test limit. When
-  /// such a maximum is set, the effective time limit of the test this trait is
-  /// applied to will be the lesser of `timeLimit` and that maximum. This is a
-  /// policy which may be configured on a global basis by the tool responsible
-  /// for launching the test process. Refer to that tool's documentation for
-  /// more details.
+  /// The testing library can use a shorter time limit than that specified by
+  /// `timeLimit` may be reduced if it's configured to enforce a maximum
+  /// per-test limit. When you configure a maximum per-test limit, the time
```
stmontgomery, line 75: I think the first sentence of this paragraph was only partly reworded and needs another look

Sources/Testing/Traits/TimeLimitTrait.swift
```diff
@@ -97,26 +94,25 @@ extension Trait where Self == TimeLimitTrait {
...
   ///
-  /// The time limit amount specified by `timeLimit` may be reduced if the
-  /// testing library is configured to enforce a maximum per-test limit. When
-  /// such a maximum is set, the effective time limit of the test this trait is
-  /// applied to will be the lesser of `timeLimit` and that maximum. This is a
-  /// policy which may be configured on a global basis by the tool responsible
-  /// for launching the test process. Refer to that tool's documentation for
-  /// more details.
+  /// The testing library can use a shorter time limit than that specified by
+  /// `timeLimit` may be reduced if it's configured to enforce a maximum
```
stmontgomery, line 107: (Same rewording comment here)

Sources/Testing/Traits/TimeLimitTrait.swift
```diff
@@ -185,11 +181,10 @@ extension TimeLimitTrait.Duration {
...
 @available(_clockAPI, *)
 extension Test {
-  /// The maximum amount of time the cases of this test may run for.
+  /// The maximum amount of time this test's cases may run for.
   ///
-  /// Time limits are associated with tests using this trait:
-  ///
-  /// - ``Trait/timeLimit(_:)-4kzjp``
+  /// Associate time limit with tests by using the ``Trait/timeLimit(_:)-4kzjp``
+  /// trait.
```
stmontgomery, line 187:
```suggestion
  /// Associate a time limit with tests using ``Trait/timeLimit(_:)-4kzjp``.
```

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,79 @@
...
-  /// custom scope for the tests it's applied to.
+  /// The default type is `Never`, which can't be instantiated. The
+  /// ``scopeProvider(for:testCase:)-cjmg`` method for any trait with
+  /// `Never` as its test scope provider type must return `nil`, meaning that
+  /// the trait doesn't provide a custom scope for tests it's applied to.
   associatedtype TestScopeProvider: TestScoping = Never
 
-  /// Get this trait's scope provider for the specified test and/or test case,
-  /// if any.
+  /// Get this trait's scope provider for the specified test or test case.
```
stmontgomery, line 56: The "and/or" was there because, there is _always_ a `Test` passed, it's just the `Test.Case` parameter which is optional. Any ideas for a better way to phrase that? Saying "test **or** test case" makes it sound like it's one or the other.

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,79 @@
...
+  /// - Returns: A value conforming to ``Trait/TestScopeProvider`` which you
+  ///   use to provide custom scoping for `test` and/or `testCase`, or `nil` if
+  ///   the trait doesn't provide any custom scope for the test or test case.
   ///
   /// If this trait's type conforms to ``TestScoping``, the default value
   /// returned by this method depends on `test` and/or `testCase`:
   ///
-  /// - If `test` represents a suite, this trait must conform to ``SuiteTrait``.
+  /// - If `test` represents a suite, this trait needs to conform to
+  /// ``SuiteTrait``.
```
stmontgomery, line 72: Indentation

Sources/Testing/Traits/Comment.swift
```diff
@@ -8,65 +8,63 @@
...
+  /// To get the complete set of comments applied to a test, see
   /// ``Test/comments``.
   public var rawValue: String
 
   public init(rawValue: String) {
     self.rawValue = rawValue
   }
 
-  /// An enumeration describing the possible kind of a comment.
+  /// An enumeration that describes a comment's kind.
```
grynspan, line 33: This is not API, so don't worry about rewriting it.

Sources/Testing/Traits/Comment.swift
```diff
@@ -8,65 +8,63 @@
...
+    /// This comment comes from an explicit call to ``Trait/comment(_:)``.
     case trait
 
-    /// This comment was initialized from a string literal.
+    /// This comment is initialized from a string literal.
     case stringLiteral
   }
 
-  /// The kind of this comment, if known.
+  /// This comment's kind.
```
grynspan, line 60: SPI, so no need to rewrite.

Sources/Testing/Traits/ConditionTrait+Macro.swift
```diff
@@ -35,8 +35,8 @@ extension Trait where Self == ConditionTrait {
     return "\(platformName) \(version.major).\(minorVersion).\(patchVersion)"
   }
 
-  /// Create a trait controlling availability of a test based on an
-  /// `@available()` attribute applied to it.
+  /// Create a trait that controls a test's availability based on a
```
grynspan, line 38: Symbols starting with two underscores are not part of the API surface and don't need revising.

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,79 @@
...
-  ///   they should not have any custom scope.
+  /// - Returns: A value conforming to ``Trait/TestScopeProvider`` which you
+  ///   use to provide custom scoping for `test` and/or `testCase`, or `nil` if
+  ///   the trait doesn't provide any custom scope for the test or test case.
   ///
   /// If this trait's type conforms to ``TestScoping``, the default value
   /// returned by this method depends on `test` and/or `testCase`:
   ///
-  /// - If `test` represents a suite, this trait must conform to ``SuiteTrait``.
+  /// - If `test` represents a suite, this trait needs to conform to
```
stmontgomery, line 71: > If `test` represents a suite, this trait must conform to `SuiteTrait`.

What I meant to convey here is that, if `test` is a suite, that _implies_ this trait conforms to `SuiteTrait`. It's not something the developer needs to do; it's something which must already be true. Maybe a clearer wording would be
```diff
-- If `test` represents a suite, this trait must conform to ``SuiteTrait`..
+- If `test` represents a suite, this trait conforms to ``SuiteTrait`.

**Review by grynspan — CHANGES_REQUESTED** [review]
Please revert changes to symbols marked `@_spi` or `__` as they aren't part of the API surface and won't be published. Let's narrow the scope of the changes so we can focus on the ones that will be visible to developers.

File: Sources/Testing/Traits/ParallelizationTrait.swift [hunk 14]
```diff
@@ -8,20 +8,21 @@
...
-/// A type that affects whether or not a test or suite is parallelized.
+/// A type that defines whether the testing library runs this test serially
+/// or in parallel.
 ///
-/// When added to a parameterized test function, this trait causes that test to
-/// run its cases serially instead of in parallel. When added to a
-/// non-parameterized test function, this trait has no effect.
+/// When you add this trait to a parameterized test function, that test runs its
+/// cases serially instead of in parallel. This trait has no effect when you
+/// apply it to a non-parallelized test function.
```
grynspan, line 16: **parameterized**, not **parallelized**

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,79 @@
...
   ///   If the value of this suite trait's ``SuiteTrait/isRecursive`` property
-  ///   is `true`, then this method returns `nil`; otherwise, it returns `self`.
-  ///   This means that by default, a suite trait will _either_ provide its
-  ///   custom scope once for the entire suite, or once per-test function it
-  ///   contains.
+  ///   is `true`, then this method returns `nil`, and the suite trait
+  ///   provides its custom scope once for the entire suite.
+  ///   If the value of ``SuiteTrait/isRecursive`` is `false`, this method
+  ///   returns `self`, and the suite trait provides its custom scope once for
+  ///   each test function the test suite contains.
```
stmontgomery, line 78: I like how you reworded to simplify the sentence structure, but it looks like the two "outcomes" are reversed and need to be swapped.  I.e.
```suggestion
  ///   is `true`, then this method returns `nil`, and the suite trait
  ///   provides its custom scope once for each test function the
  ///   test suite contains. If the value of ``SuiteTrait/isRecursive``
  ///   is `false`, this method returns `self`, and the suite trait
  ///   provides its custom scope once for the entire suite.
```
[P1]

Sources/Testing/Traits/TimeLimitTrait.swift
```diff
@@ -54,35 +52,34 @@ public struct TimeLimitTrait: TestTrait, SuiteTrait {
...
+  /// due to variability in testing environments. You express the duration in
+  /// minutes, with a minimum duration of one minute.
   ///
-  /// When this trait is associated with a test, that test must complete within
-  /// a time limit of, at most, `timeLimit`. If the test runs longer, an issue
-  /// of kind ``Issue/Kind/timeLimitExceeded(timeLimitComponents:)`` is
-  /// recorded. This timeout is treated as a test failure.
+  /// When you associate this trait with a test, that test must complete within
+  /// a time limit of, at most, `timeLimit`. If the test runs longer, the
+  /// testing library records a
```
grynspan, line 69: Nitpick: whitespace seems wonky here?

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,79 @@
...
-/// functions and to test suites. Developers can define their own traits by
-/// creating types that conform to ``TestTrait`` and/or ``SuiteTrait``.
+/// functions and to test suites. Define your own traits by
+/// creating types that conform to ``TestTrait`` and/or ``SuiteTrait``:
 ///
-/// When creating a custom trait type, the type should conform to ``TestTrait``
-/// if it can be added to test functions, ``SuiteTrait`` if it can be added to
-/// test suites, and both ``TestTrait`` and ``SuiteTrait`` if it can be added to
-/// both test functions _and_ test suites.
+/// - term ``TestTrait``: Conform to this type in traits that you add to test
```
grynspan, line 18: _wat?_

Sources/Testing/Traits/Trait.swift
```diff
@@ -136,43 +139,45 @@ public protocol TestScoping: Sendable {
...
   /// ``Trait/scopeProvider(for:testCase:)-cjmg``.
   public func scopeProvider(for test: Test, testCase: Test.Case?) -> Self? {
     testCase == nil ? nil : self
   }
 }
 
 extension SuiteTrait where Self: TestScoping {
-  /// Get this trait's scope provider for the specified test and/or test case,
-  /// if any.
+  /// Get this trait's scope provider for the specified test and/or test case.
```
stmontgomery, line 169: This one still has the "and/or", but it's been removed other places and I had an earlier comment asking about a clearer wording in general.

Sources/Testing/Testing.docc/Traits.md
iamleeg: I reworded this to "your tests" because we just mentioned suites and functions in the previous sentence.

Sources/Testing/Traits/ConditionTrait.swift
```diff
@@ -75,9 +76,10 @@ public struct ConditionTrait: TestTrait, SuiteTrait {
     }
   }
 
+  /// The comments that the testing library associates with this trait.
   public var comments: [Comment]
```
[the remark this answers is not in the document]
iamleeg: You're right this doesn't need documenting, I just got into a groove :)

Sources/Testing/Traits/TimeLimitTrait.swift
iamleeg: I re-worded this to avoid saying "Use this type" twice and it ended up just as long, but see what you think.

Sources/Testing/Traits/TimeLimitTrait.swift
iamleeg: OK had another go at this paragraph, see what you think.

Sources/Testing/Traits/Trait.swift
iamleeg: Right, good point. I avoided and/or because it's difficult for readers of English as a second language. I've gone with "test, and optional test case". I'm not sure I'm entirely happy with that but maybe it helps advance the discussion :).

Sources/Testing/Traits/Trait.swift
iamleeg: Ah, got it. I've written "also conforms to `SuiteTrait`" so it isn't surprising that someone reading the docs for one protocol gets told that the type actually conforms to another protocol.

Sources/Testing/Traits/Trait.swift
iamleeg: Whoops, sorry!

Sources/Testing/Traits/TimeLimitTrait.swift
iamleeg: It's because the link ``Issue/Kind/timeLimitExceeded(timeLimitComponents:)`` is really long and forces a wrap. I've got the line-limit marker set to 80 columns, and configured Xcode to use a fixed-width font for documentation markup, and I'm not sure I can do better than this layout.

Sources/Testing/Traits/Trait.swift
iamleeg: I don't understand the question. This syntax is a [term list](https://www.swift.org/documentation/docc/formatting-your-documentation-content#Add-Bulleted-Numbered-and-Term-Lists).

Sources/Testing/Testing.docc/Traits.md
```diff
@@ -10,14 +10,14 @@ See https://swift.org/LICENSE.txt for license information
 See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 -->
 
-Add traits to tests to annotate them or customize their behavior.
+Annotate tests and suites, and customize their behavior.
```
grynspan, line 13: I don't think we're entirely consistent here, but I've usually been writing "test functions and [test] suites". We should decide what phrasing we prefer and apply it consistently.

Sources/Testing/Traits/Bug.swift
```diff
@@ -59,42 +59,42 @@ extension Bug: TestTrait, SuiteTrait {
 }
 
 extension Trait where Self == Bug {
-  /// Construct a bug to track with a test.
+  /// Constructs a bug to track with a test.
```
grynspan, line 62: Now that I think about it… does Swift normally use "construct" here or "initialize" or…? I ask because Swift doesn't have _constructors_ like C++ or some other languages.

Sources/Testing/Traits/Comment+Macro.swift
```diff
@@ -9,57 +9,55 @@
 //
 
 extension Trait where Self == Comment {
-  /// Construct a comment related to a test from a single-line source code
+  /// Constructs a comment related to a test from a single-line source code
```
grynspan, line 12: These functions are not API. You don't need to worry yourself with 'em. :)

Sources/Testing/Testing.docc/Traits.md
iamleeg: I think "test functions and suites" is fine, we don't need the repetition of "test functions and test suites", I'll use that.

Sources/Testing/Traits/Bug.swift
grynspan: Not directly relevant to the PR, but we should make sure the guidelines cover factory functions like these.

Sources/Testing/Traits/Trait.swift
grynspan: I didn't know that was even a thing. Huh.

Sources/Testing/Traits/ConditionTrait.swift
```diff
@@ -197,17 +195,16 @@ extension Trait where Self == ConditionTrait {
     Self(kind: .conditional { !(try condition()) }, comments: Array(comment), sourceLocation: sourceLocation)
   }
 
-  /// Construct a condition trait that causes a test to be disabled if it
-  /// returns `true`.
+  /// Construct a condition trait that disables a test if its value is true.
```
grynspan, line 198:
```suggestion
  /// Constructs a condition trait that disables a test if its value is true.
```

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,80 @@
...
-  /// custom scope for the tests it's applied to.
+  /// The default type is `Never`, which can't be instantiated. The
+  /// ``scopeProvider(for:testCase:)-cjmg`` method for any trait with
+  /// `Never` as its test scope provider type must return `nil`, meaning that
+  /// the trait doesn't provide a custom scope for tests it's applied to.
   associatedtype TestScopeProvider: TestScoping = Never
 
-  /// Get this trait's scope provider for the specified test and/or test case,
-  /// if any.
+  /// Get this trait's scope provider for the specified test, and optional test
```
grynspan, line 56: Comma splice.

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,80 @@
...
-  ///   they should not have any custom scope.
+  /// - Returns: A value conforming to ``Trait/TestScopeProvider`` which you
+  ///   use to provide custom scoping for `test` and/or `testCase`, or `nil` if
+  ///   the trait doesn't provide any custom scope for the test or test case.
   ///
   /// If this trait's type conforms to ``TestScoping``, the default value
   /// returned by this method depends on `test` and/or `testCase`:
   ///
-  /// - If `test` represents a suite, this trait must conform to ``SuiteTrait``.
+  /// - If `test` represents a suite, this trait also conforms to
```
grynspan, line 72: The conformance to `SuiteTrait` isn't automatic, hence "must"

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,80 @@
...
+  ///   ``SuiteTrait``.
   ///   If the value of this suite trait's ``SuiteTrait/isRecursive`` property
-  ///   is `true`, then this method returns `nil`; otherwise, it returns `self`.
-  ///   This means that by default, a suite trait will _either_ provide its
-  ///   custom scope once for the entire suite, or once per-test function it
-  ///   contains.
+  ///   is `true`, then this method returns `nil`, and the suite trait
```
grynspan, line 75: This discussion only applies to suite traits (it's saying "if the suite is recursive, then this function is recursively called for tests in the suite.")

Sources/Testing/Traits/Trait.swift
```diff
@@ -12,75 +12,80 @@
...
+  ///   provides its custom scope once for each test function the test suite
+  ///   contains. If the value of ``SuiteTrait/isRecursive`` is `false`, this
+  ///   method returns `self`, and the suite trait provides its custom scope
+  ///   once for the entire test suite.
   /// - Otherwise `test` represents a test function. If `testCase` is `nil`,
```
grynspan, line 80: "Otherwise" here means "if the trait does not conform to `SuiteTrait`" but it's so far removed from that introductory sentence that I'm not sure it's clear.

Sources/Testing/Traits/Trait.swift
```diff
@@ -92,41 +97,40 @@ public protocol Trait: Sendable {
...
   /// this method returns `nil` by default. This means that instances of this
-  /// trait will not provide a custom scope for tests to which they're applied.
+  /// trait don't provide a custom scope for tests to which they're applied.
   func scopeProvider(for test: Test, testCase: Test.Case?) -> TestScopeProvider?
 }
 
-/// A protocol that allows providing a custom execution scope for a test
-/// function (and each of its cases) or a test suite by performing custom code
-/// before or after it runs.
+/// A protocol that tells the test runner to run custom code
```
grynspan, line 104: Nitpick: line is short?

Sources/Testing/Traits/Trait.swift
```diff
@@ -92,41 +97,40 @@ public protocol Trait: Sendable {
...
 ///
-/// Types conforming to this protocol may be used in conjunction with a
+/// Use types that conform to this protocol in conjunction with a
 /// ``Trait``-conforming type by implementing the
-/// ``Trait/scopeProvider(for:testCase:)-cjmg`` method, allowing custom traits
-/// to provide custom scope for tests. Consolidating common set-up and tear-down
-/// logic for tests which have similar needs allows each test function to be
-/// more succinct with less repetitive boilerplate so it can focus on what makes
-/// it unique.
+/// ``Trait/scopeProvider(for:testCase:)-cjmg`` method, so your custom traits
```
grynspan, line 109: Awkward phrasing: "Use types [...], so your custom traits [...]". Can we revise?

Sources/Testing/Traits/Trait.swift
```diff
@@ -92,41 +97,40 @@ public protocol Trait: Sendable {
...
+/// provide custom scope for tests. Do this to consolidate common set-up and
```
grynspan, line 110: "Do this" what? "This" needs a noun. :)

Sources/Testing/Traits/Trait.swift
```diff
@@ -136,43 +140,46 @@ public protocol TestScoping: Sendable {
...
   /// ``Trait/scopeProvider(for:testCase:)-cjmg``.
   public func scopeProvider(for test: Test, testCase: Test.Case?) -> Self? {
     testCase == nil ? nil : self
   }
 }
 
 extension SuiteTrait where Self: TestScoping {
-  /// Get this trait's scope provider for the specified test and/or test case,
-  /// if any.
+  /// Get this trait's scope provider for the specified test, and optional test
```
grynspan, line 170: Misplaced Oxford comma?

Sources/Testing/Traits/Trait.swift
```diff
@@ -136,43 +140,46 @@ public protocol TestScoping: Sendable {
...
-  ///     if any. When `test` represents a suite, the value of this argument is
+  ///   - test: The test for which the testing library requests a
+  ///     scope provider.
+  ///   - testCase: The test case for which the testing library requests a scope
+  ///     provider, if any. When `test` represents a suite, the value of this argument is
   ///     `nil`.
   ///
-  /// This default implementation is used when this trait type conforms to
-  /// ``TestScoping`` and its return value is discussed in
+  /// The testing library uses the default implementation of this method when
```
grynspan, line 160: We specifically use _this_ implementation when those constraints are met. How about:

```swift
/// The testing library uses this implementation of
/// ``Trait/scopeProvider(for:testCase:)-cjmg`` when the trait type conforms to
/// ``TestScoping``.
```

And drop "For information this [...]" since we'll have added a link?

Sources/Testing/Traits/Trait.swift
```diff
@@ -136,43 +140,46 @@ public protocol TestScoping: Sendable {
...
-  ///
-  /// This default implementation is used when this trait type conforms to
-  /// ``TestScoping`` and its return value is discussed in
+  ///   - test: The test for which the testing library requests a scope
+  ///     provider.
+  ///   - testCase: The test case for which the testing library requests a scope
+  ///     provider, if any. When `test` represents a suite, the value of this
+  ///     argument is `nil`.
+  ///
+  /// The testing library uses the default implementation of this method when
```
grynspan, line 180: This is wrong (but also was before.) How about:

```swift
/// The testing library uses this implementation of
/// ``Trait/scopeProvider(for:testCase:)-cjmg`` when the trait type conforms to
/// both ``SuiteTrait`` and ``TestScoping``.
```

And delete "For information on [...]" again?

Sources/Testing/Traits/Trait.swift
```diff
@@ -217,18 +227,19 @@ extension Trait {
...
+  ///   - test: The test for which the testing library requests a
+  ///     scope provider.
+  ///   - testCase: The test case for which the testing library requests a scope
+  ///     provider, if any. When `test` represents a suite, the value of this argument is
   ///     `nil`.
   ///
-  /// This default implementation is used when this trait type's associated
-  /// ``Trait/TestScopeProvider`` type is the default value of `Never`, and its
-  /// return value is discussed in ``Trait/scopeProvider(for:testCase:)-cjmg``.
+  /// The testing library uses the default implementation when this trait type's
```
grynspan, line 239: How about:

```swift
/// The testing library uses this implementation of
/// ``Trait/scopeProvider(for:testCase:)-cjmg`` when the trait type's associated
/// ``Trait/TestScopeProvider`` type is `Never`.
```

And again, drop "For information on [...]"?

Sources/Testing/Traits/Trait.swift
iamleeg: This is the opposite of @stmontgomery's suggestion:

> What I meant to convey here is that, if test is a suite, that implies this trait conforms to SuiteTrait. It's not something the developer needs to do; it's something which must already be true. Maybe a clearer wording would be
> 
> -- If `test` represents a suite, this trait must conform to ``SuiteTrait`..
> +- If `test` represents a suite, this trait conforms to ``SuiteTrait`.

I can only satisfy one of these, and I think @stmontgomery is correct that if you get here and `test` is a suite, then the trait _already_ conforms to `SuiteTrait` and the developer doesn't need to do anything.

Sources/Testing/Traits/Trait.swift
iamleeg: OK I've had a go at this, see what you think.

Sources/Testing/Traits/Trait.swift
grynspan: This documentation is going to be viewed in isolation, and somebody who reaches it may not realize they're looking at documentation for the _specific_ case where the type already conforms to `SuiteTrait`. Hence we need to make it clear that the described behaviour of this symbol only applies under certain conditions.

Sources/Testing/Traits/TimeLimitTrait.swift
grynspan: FWIW, we usually just let the long links overflow. The point of wrapping is to give us a sense of how long the documentation will be when rendered, and the link will be rendered shorter anyway.

Sources/Testing/Traits/Trait.swift
grynspan: Looks good, thanks!

grynspan (approved):

PR #929, opened 2025-01-27
Remove Stubs.cpp and reimplement what it has in Swift.

grynspan:
This PR removes Stubs.cpp, which currently houses some thunks for functions that are conditionally unavailable in glibc, and replaces it with runtime function lookups in Swift. Is there potentially a one-time non-zero performance cost? Yes. Is that performance cost prohibitive given that the functions are only looked up once and then cached? No.

These functions won't get called on Linux if `SWT_NO_DYNAMIC_LINKING` is defined but we don't currently support that combination anyway. Even if you're using a statically-linked Swift standard library, we'd expect Linux to still support calling `dlsym()`.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

briancroom (approved):

PR #931, opened 2025-01-27
Introduce a severity level for issues, and a 'warning' severity

stmontgomery:
This introduces the concept of severity to the `Issue` type, represented by a new enum `Issue.Severity` with two cases: `.warning` and `.error`. Error is the default severity for all issues, matching current behavior, but warning is provided as a new option which does not cause the test the issue is associated with to be marked as a failure.

In this PR, these are [SPI](https://github.com/swiftlang/swift-testing/blob/main/Documentation/SPI.md) but they could be considered for promotion to public API eventually. Additional work would be needed to permit test authors to record issues with severity < `.error`, since APIs like `Issue.record()` are not being modified at this time to allow customizing the severity.

### Motivation:

There are certain situations where a problem may arise during a test that doesn't necessarily affect its outcome or signal an important problem, but is worth calling attention to. A specific example use case I have in mind is to allow the testing library to record a warning issue about problems with the arguments passed to a parameterized test, such as having duplicate arguments.

### Modifications:

- Introduce `Issue.Severity` as an SPI enum.
- Introduce an SPI property `severity` to `Issue` with default value `.error`.
- Modify entry point logic to exit with `EXIT_SUCCESS` if all issues recorded had severity < `.error`.
- Modify console output formatting logic and data structures to represent warning issues sensibly.

### Checklist:

- [x] Add new tests


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.Symbol.swift
```diff
@@ -34,7 +34,11 @@ extension Event {
 
     /// A warning or caution symbol to use when the developer should be aware of
     /// some condition.
-    case warning
+    ///
+    /// - Parameters:
+    ///   - warningIssueCount: The number of issues with warning severity
+    ///     recorded for the test. The default value is `0`.
+    case warning(warningIssueCount: Int = 0)
```
grynspan, line 41: This symbol's just a ⚠️ symbol. I don't think you want to change it.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -56,8 +56,11 @@ extension Event {
         /// The instant at which the test started.
         var startInstant: Test.Clock.Instant
 
-        /// The number of issues recorded for the test.
-        var issueCount = 0
```
grynspan, line 60: Would it be simpler to have:

```swift
var issueCount = [Issue.Severity: Int]()
```

?

Sources/Testing/Events/Recorder/Event.Symbol.swift
```diff
@@ -22,8 +22,8 @@ extension Event {
     /// The symbol to use when a test passes.
     ///
     /// - Parameters:
-    ///   - knownIssueCount: The number of known issues encountered by the end
-    ///     of the test.
+    ///   - knownIssueCount: The number of known issues recorded for the test.
+    ///     The default value is `0`.
```
grynspan, line 26: The only reason there's an associated value here is because, at the time I was building this out, it felt weird to have `case pass` and `case passWithKnownIssues` as separate cases. Doesn't really translate to `case warning`.

Sources/Testing/Issues/Issue.swift
```diff
@@ -244,6 +288,9 @@ extension Issue {
     /// The kind of issue this value represents.
     public var kind: Kind.Snapshot
 
+    @_spi(Experimental)
+    public var severity: Severity
```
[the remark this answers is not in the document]
grynspan: I don't think we can/should add this to `Snapshot` as Xcode 16.0 does not know how to handle severity and will report them all as errors. Rather, we may need to just suppress issues with `severity < .error`.

We _do_ need to add this to `ABIv0.EncodedIssue`, but severity is effectively ABI-breaking (because now we have issues that aren't known but don't cause a test to fail.) So that may require a new ABI revision.

Documentation/ABI/JSON.md
```diff
@@ -197,10 +197,20 @@ sufficient information to display the event in a human-readable format.
   "runEnded" ; additional event kinds may be added in the future
 
 <issue> ::= {
+<!--
```
grynspan, line 200: I haven't bothered to add other underscored stuff prior to review. We can leave it out for now and add it as part of your proposal when you make it.

I _think_ this is going to require a new ABI revision though.

stmontgomery:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -121,8 +119,8 @@ extension Event.HumanReadableOutputRecorder {
     guard let graph else {
       return (0, 0, 0, 0, "")
     }
-    let errorIssueCount = graph.compactMap(\.value?.errorIssueCount).reduce(into: 0, +=)
-    let warningIssueCount = graph.compactMap(\.value?.warningIssueCount).reduce(into: 0, +=)
+    let errorIssueCount = graph.compactMap { $0.value?.issueCount[.error] }.reduce(into: 0, +=)
```
grynspan, line 122: I believe this can be expressed as a keypath `compactMap(\.value?.issueCount[.error])` if you like.

Sources/Testing/Issues/Issue.swift
```diff
@@ -164,7 +197,7 @@ extension Issue: CustomStringConvertible, CustomDebugStringConvertible {
     let joinedComments = comments.lazy
       .map(\.rawValue)
       .joined(separator: "\n")
-    return "\(kind): \(joinedComments)"
+    return "\(severity): \(kind): \(joinedComments)"
```
grynspan, line 200: Suggest only including severity if it varies from the default (i.e. `!= .error`.)

Sources/Testing/Issues/Issue.swift
```diff
@@ -484,7 +535,7 @@ extension Issue.Snapshot: CustomStringConvertible, CustomDebugStringConvertible
...
```
grynspan, line 538: As above.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -400,13 +417,19 @@ extension Event.HumanReadableOutputRecorder {
...
         symbol = .pass(knownIssueCount: 1)
-        known = " known"
+        introducer = "a known"
       } else {
-        symbol = .fail
-        known = "n"
+        switch issue.severity {
+        case .warning:
+          symbol = .warning
+          introducer = "a warning"
```
grynspan, line 428: "a warning issue" sounds awkward to me. Perhaps just "a warning"—developers understand the concept of warnings from dev tools.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -114,27 +115,36 @@ extension Event.HumanReadableOutputRecorder {
...
+    let warningIssueCount = graph.compactMap { $0.value?.issueCount[.warning] }.reduce(into: 0, +=)
     let knownIssueCount = graph.compactMap(\.value?.knownIssueCount).reduce(into: 0, +=)
-    let totalIssueCount = issueCount + knownIssueCount
+    let totalIssueCount = errorIssueCount + warningIssueCount + knownIssueCount
 
     // Construct a string describing the issue counts.
-    let description = switch (issueCount > 0, knownIssueCount > 0) {
-    case (true, true):
+    let description = switch (errorIssueCount > 0, warningIssueCount > 0, knownIssueCount > 0) {
+    case (true, true, true):
```
grynspan, line 129: Combinatoric explosion incoming if we add another severity level like `.note`.

Does this string end up double-counting known issues of severity `.warning`?

Sources/Testing/Issues/Issue.swift
stmontgomery: When I was considering this change, I actually concluded that it _would_ be useful to always include the severity. One reason I find it helpful is that often, people search for the string `error:` in text to find problems. Although we don't directly output this description string to the console, I think it would be generally useful for it to include "error" for issues with that severity level.

Thinking about it more now, perhaps I should go _further_ and include that word in our actual console output, for the same reason?

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
stmontgomery: > Combinatoric explosion incoming if we add another severity level like .note

I think there will be a reasonable way to refactor when the next dimension arrives, and for now I think it's still manageable.

> Does this string end up double-counting known issues of severity .warning

- [ ] Add a test validate the output for a _known_ warning issue, and adjust the implementation if needed.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
stmontgomery: I like this better, but it seems to have led to a new compiler crash in CI:

> https://ci.swift.org/job/pr-swift-testing-linux-5.10/1630/console

May need to use less sugar on the subscript assignment or some other workaround

Sources/Testing/Issues/Issue.swift
grynspan: That's fair. Where does this string show up? Does Xcode use it (I genuinely can't remember)? Does VS Code?

Does the string need to be title-cased? `kind` is, but now it's not going to be at the head of the string, so should it be lowercased?

Could we reorder the bits here so they fit better? Perhaps:

```swift
"\(kind) (\(severity)): \(joinedComments)"
```

Producing something like:

> An API was misused (warning): blah blah

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: Yes, I agree for now this is manageable, but let's keep it in the back of our minds.

Please make sure test coverage covers all these cases!

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -355,15 +365,22 @@ extension Event.HumanReadableOutputRecorder {
...
         CollectionOfOne(
           Message(
             symbol: .fail,
             stringValue: "\(_capitalizedTitle(for: test)) \(testName) failed after \(duration)\(issues.description)."
           )
         ) + _formattedComments(for: test)
+      } else if issues.warningIssueCount > 0 {
+        [
+          Message(
+           symbol: .warning,
```
grynspan, line 378: Thinking about it, I am not sure we want to use this symbol. It stands out quite a bit (especially when using SF Symbols) and we've been using it in a few places for _infrastructure issues_, i.e. issues of kind `.system`.

We might want to instead render it the same as `.fail`, but yellow instead of red. Or we might want a different symbol entirely.

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/Running/Configuration.swift
```diff
@@ -178,14 +178,35 @@ public struct Configuration: Sendable {
...
+  public struct EventHandlingOptions: Sendable {
+    /// Whether or not events of the kind ``Event/Kind-swift.enum/issueRecorded(_:)``
+    /// containing issues with warning (or lower) severity should be delivered
+    /// to the event handler of the configuration these options are applied to.
+    ///
+    /// By default, events matching this criteria are delivered to event
+    /// handlers unless the entry point specifies an event stream version which
+    /// predates the introduction of warning issues, in which case this is
+    /// disabled to maintain legacy behavior.
+    public var isWarningIssueEventDeliveryEnabled: Bool = true
```
grynspan, line 192:
```suggestion
    public var isWarningIssueRecordedEventEnabled: Bool = true
```

Sources/Testing/Running/Configuration.swift
```diff
@@ -178,14 +178,35 @@ public struct Configuration: Sendable {
...
+
+    /// Whether or not events of the kind
+    /// ``Event/Kind-swift.enum/expectationChecked(_:)`` should be delivered to
+    /// the event handler of the configuration these options are applied to.
+    ///
+    /// By default, events of this kind are not delivered to event handlers
+    /// because they occur frequently in a typical test run and can generate
+    /// significant back-pressure on the event handler.
+    public var isExpectationCheckedEventDeliveryEnabled: Bool = false
```
grynspan, line 201:
```suggestion
    public var isExpectationCheckedEventEnabled: Bool = false
```

Sources/Testing/Running/Configuration.swift
```diff
@@ -178,14 +178,35 @@ public struct Configuration: Sendable {
...
-  /// this configuration's ``eventHandler`` closure.
+  /// A type describing options to use when delivering events to this
+  /// configuration's event handler
+  public struct EventHandlingOptions: Sendable {
+    /// Whether or not events of the kind ``Event/Kind-swift.enum/issueRecorded(_:)``
+    /// containing issues with warning (or lower) severity should be delivered
+    /// to the event handler of the configuration these options are applied to.
+    ///
+    /// By default, events matching this criteria are delivered to event
+    /// handlers unless the entry point specifies an event stream version which
```
grynspan, line 189: "unless" onward is not necessary here, or maybe needs to be reworded—the details of the event stream schema are mostly orthogonal to `Configuration` because anybody who's using `Configuration` directly is _not_ using the JSON event stream.

Should we just default to `false` and require opt-in? That makes it simpler for all existing code paths that don't understand warnings yet.

Sources/Testing/Events/Event.swift
```diff
@@ -306,6 +303,26 @@ extension Event {
...
+  /// options in this instance.
+  ///
+  /// - Parameters:
+  ///   - event: The event to consider handling.
+  ///
+  /// - Returns: Whether or not the event should be handled or suppressed.
+  fileprivate func shouldHandleEvent(_ event: borrowing Event) -> Bool {
+    switch event.kind {
+    case let .issueRecorded(issue):
+      issue.severity > .warning || isWarningIssueEventDeliveryEnabled
```
grynspan, line 317: Nitpick, but is this strictly the correct logic if we added some new severity later?

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -547,6 +547,15 @@ public func configurationForEntryPoint(from args: __CommandLineArguments_v0) thr
   configuration.exitTestHandler = ExitTest.handlerForEntryPoint()
 #endif
 
+  switch args.eventStreamVersion {
+  case .some(...0):
```
grynspan, line 551: See my comment on `Configuration`—I like this approach, but maybe we switch the default around.

Sources/Testing/ABI/EntryPoints/ABIEntryPoint.swift
```diff
@@ -47,7 +47,7 @@ extension ABIv0 {
   /// callback.
   public static var entryPoint: EntryPoint {
     return { configurationJSON, recordHandler in
-      try await Testing.entryPoint(
+      try await _entryPoint(
```
grynspan, line 50: Just a drive-by fix I assume? :)

stmontgomery: Yep

Sources/Testing/Running/Configuration.swift
stmontgomery: I like that (and the other suggested property name)

Sources/Testing/Running/Configuration.swift
stmontgomery: After an offline discussion, I ended up making this `false` but for a slightly different reason: Because this is still an experimental feature. The ABI version ≤ 0 check which disables it will remain in place though, because even if/when this `Configuration` setting changes to `true`, we'll still want existing ABI versions to maintain the legacy behavior.

stmontgomery:
@swift-ci please test

Sources/Testing/Issues/Issue.swift
stmontgomery: Moved severity later in the string, but kept it lowercase

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
stmontgomery: I've included coverage for all these code paths in new and adjusted unit tests

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
stmontgomery: Ideally, I'd like to use a symbol which features an exclamation point `!`. I don't see a diamond-with-exclamation point in SF Symbols, so that's one challenge.

stmontgomery:
@swift-ci please test

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -270,6 +270,13 @@ public struct __CommandLineArguments_v0: Sendable {
   /// The value(s) of the `--skip` argument.
   public var skip: [String]?
 
+  /// Whether or not to include tests with the `.hidden` trait when constructing
+  /// a test filter based on these arguments.
+  ///
+  /// This property is intended for use in testing the testing library itself.
+  /// It is not parsed as a command-line argument.
+  public var includeHiddenTests: Bool?
```
grynspan, line 278: Maybe make it `internal` then?

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -296,9 +310,11 @@ extension __CommandLineArguments_v0: Codable {
     case eventStreamVersion
     case filter
     case skip
+    case includeHiddenTests
```
grynspan, line 313: If we don't ever parse it, it doesn't need to be encoded/decoded, right?

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -278,6 +285,13 @@ public struct __CommandLineArguments_v0: Sendable {
 
   /// The value of the `--experimental-attachments-path` argument.
   public var experimentalAttachmentsPath: String?
+
+  /// The value(s) of the `--enable-experimental-feature` argument(s),
+  /// indicating the name(s) of experimental features to enable.
+  ///
+  /// The value of this property is not yet populated by parsing command-line
+  /// arguments, but may be in the future.
+  public var experimentalFeatures: [String]?
```
grynspan, line 294: This is a broader feature that's not directly related to this PR. I think we should leave it out for now and plan to talk to the SwiftPM folks about building this feature formally (especially if we think we're going to use the same terminology, which I think is a good idea in general.)

grynspan (changes requested):
I see where you're going with the experimental features stuff, but it deserves its own PR and feature proposal.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: We use the filled diamond-with-X symbol for errors—we could use the hollow diamond-with-X for warnings? There's also a diamond-with-question-mark, or various symbols with exclamation marks that aren't diamonds.

Don't forget we also need a separate icon for Unicode-without-SF-Symbols, and then possibly a different one for Windows (which has a reduced character set in Terminal compared to Linux and macOS.)

Sources/Testing/Issues/Issue.swift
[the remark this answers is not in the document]
grynspan: We can't encode `severity` here non-optionally: imagine you use a newer Xcode that recognizes it (but still uses the snapshot stuff instead of ABIv0), but you use an older swift-syntax package that doesn't encode the value? The issue will fail to decode with a missing key.

Tests/TestingTests/ABIEntryPointTests.swift
```diff
@@ -54,7 +54,7 @@ struct ABIEntryPointTests {
   ) async throws -> CInt {
 #if !SWT_NO_DYNAMIC_LINKING
     // Get the ABI entry point by dynamically looking it up at runtime.
-    let copyABIEntryPoint_v0 = try withTestingLibraryImageAddress { testingLibrary in
+    let copyABIEntryPoint_v0 = try _withTestingLibraryImageAddress { testingLibrary in
```
grynspan, line 57: This code is in a test file. We don't otherwise bother with underscores in the test target.

grynspan:
Something that appears to be missing: a way for a developer to record a warning. Something like:

```swift
extension Issue {
  @_spi(Experimental)
  public static func warn(_ message: Comment, blah blah)
```

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
stmontgomery: It has to be encoded because the ABI entry point unit tests use it, and they round-trip through the encoding representation

stmontgomery:
> Something that appears to be missing: a way for a developer to record a warning.

Yeah, I consider that out of scope for this PR. In the description I wrote:

> In this PR, these are [SPI](https://github.com/swiftlang/swift-testing/blob/main/Documentation/SPI.md) but they could be considered for promotion to public API eventually. Additional work would be needed to permit test authors to record issues with severity < .error, since APIs like Issue.record() are not being modified at this time to allow customizing the severity.

My immediate motivation is to allow the testing library itself to create and record issues with warning severity. We can consider exposing this as SPI/API in subsequent PRs.

Sources/Testing/Issues/Issue.swift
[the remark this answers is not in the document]
stmontgomery: Good call; I modified the decoding path to use `decodeIfPresent()` and fall back to `.error`.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
stmontgomery: Yeah, I included it to offer a practical example of how I think it would be beneficial to formalize the concept of a "feature" within the testing library, but I expected to revert that and simplify things since it's out of scope here. I'll follow up with this idea in a separate PR.

stmontgomery:
@swift-ci please test

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: That means it would be possible for external clients to set it, which we don't want.

Do we actually need the ABI entry point tests to use it, or do we just need the ability to set up an event handler to listen for warnings? Because there's two layers there: the ABI (JSON-based) event stream and the slightly higher-level all-Swift entry point, and the latter can also be set up to emit JSON if we want without requiring a JSON blob for its configuration.

stmontgomery: Ah yes, I can switch these tests to use the direct Swift entry point and avoid en/decoding these two new internal properties.

stmontgomery:
@swift-ci please test

Documentation/ABI/JSON.md
```diff
@@ -210,6 +210,12 @@ sufficient information to display the event in a human-readable format.
   "fail" | "difference" | "warning" | "details"
 ```
 
+<!--
```
grynspan, line 213: Not critical for this PR, but let's think about starting to draft the change list for ABIv1.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -278,6 +287,13 @@ public struct __CommandLineArguments_v0: Sendable {
 
   /// The value of the `--experimental-attachments-path` argument.
   public var experimentalAttachmentsPath: String?
+
+  /// Whether or not the experimental warning issue severity feature should be
+  /// enabled.
+  ///
+  /// This property is intended for use in testing the testing library itself.
+  /// It is not parsed as a command-line argument.
+  var isWarningIssueRecordedEventEnabled: Bool?
```
grynspan, line 296: We can make this `public` the same way `public var experimentalAttachmentsPath: String?` is in the line above; the name should include "experimental" to reflect that it's not a finalized feature yet.

Actually, it should probably be `@_spi(Experimental)`, as should `experimentalAttachmentsPath`.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -547,6 +566,22 @@ public func configurationForEntryPoint(from args: __CommandLineArguments_v0) thr
   configuration.exitTestHandler = ExitTest.handlerForEntryPoint()
 #endif
 
+  // Warning issues (experimental).
+  if args.isWarningIssueRecordedEventEnabled == true {
+    configuration.eventHandlingOptions.isWarningIssueRecordedEventEnabled = true
+  } else {
+    switch args.eventStreamVersion {
```
grynspan, line 573: _For now_, until we've proposed the change list for ABIv1, I think we should just leave this switch out. It'll be easy to add back in v1 (actually, there's a whole different way we can do it that may be more scalable, I'll discuss with you in DMs.)

I say this because somebody passing `1` here is either from the future or is providing what is currently invalid input.

Sources/Testing/Events/Event.swift
```diff
@@ -306,6 +303,26 @@ extension Event {
   }
 }
 
+extension Configuration.EventHandlingOptions {
```
grynspan, line 306: Nitpick: move this to the same file as `handleEvent()`?

Tests/TestingTests/RunnerTests.swift
```diff
@@ -459,7 +459,7 @@ final class RunnerTests: XCTestCase {
 #endif
 
     var configuration = Configuration()
-    configuration.deliverExpectationCheckedEvents = true
+    configuration.eventHandlingOptions.isExpectationCheckedEventEnabled = true
```
grynspan, line 462: As annoying as it is, can we add some stub test here somewhere that uses `deliverExpectationCheckedEvents` so we can get code coverage of it? Gotta keep those numbers up.

Tests/TestingTests/EventRecorderTests.swift
```diff
@@ -183,15 +183,20 @@ struct EventRecorderTests {
   @Test(
     "Issue counts are summed correctly on test end",
     arguments: [
-      ("f()", false, (total: 5, expected: 3)),
-      ("g()", false, (total: 2, expected: 1)),
-      ("PredictablyFailingTests", true, (total: 7, expected: 4)),
+      ("f()", ".* Test f\\(\\) failed after .+ seconds with 5 issues \\(including 3 known issues\\)\\."),
```
grynspan, line 186: Can we use regex literals?

Sources/Testing/Running/Configuration.swift
```diff
@@ -178,14 +178,35 @@ public struct Configuration: Sendable {
...
+  }
+
+  /// The options to use when delivering events to this configuration's event
+  /// handler.
   ///
-  /// By default, events of this kind are not delivered to event handlers
-  /// because they occur frequently in a typical test run and can generate
-  /// significant backpressure on the event handler.
-  public var deliverExpectationCheckedEvents: Bool = false
+  /// The default value of this property is an instance of ``EventHandlingOptions-swift.struct``
```
grynspan, line 207: Nitpick: the default is the default—probably don't need to call that out. :)

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -267,7 +277,8 @@ extension Event.HumanReadableOutputRecorder {
         if issue.isKnown {
           testData.knownIssueCount += 1
         } else {
-          testData.issueCount += 1
+          let issueCount = testData.issueCount[issue.severity] ?? 0
```
grynspan, line 280:
```suggestion
          let issueCount = testData.issueCount[issue.severity, default: 0]
```

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedIssue.swift
```diff
@@ -16,6 +16,19 @@ extension ABIv0 {
   /// assists in converting values to JSON; clients that consume this JSON are
   /// expected to write their own decoders.
   struct EncodedIssue: Sendable {
+    /// An enumeration representing the level of severity of a recorded issue.
+    ///
+    /// For descriptions of individual cases, see ``Issue/Severity-swift.enum``.
+    enum Severity: String {
```
grynspan, line 22: FWIW we could just use the `Issue.Severity` type here; we only need a separate type if it will encode in a different way (which I don't think is the case here?)

In any event, if you're keeping this enum, please mark it `Sendable`.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: I realize I might be giving conflicting advice here.

_If_ this is something we want to be able to set from the command line, we _should_ name it `experimentalWhatever` and include the ability to decode it so that it can be passed in a JSON blob.

If it is _only_ for testing purposes, then it's fine as-is.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: See also my comment here: https://github.com/swiftlang/swift-testing/pull/931/files#r1944900283

I realize I'm being confusing, sorry.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
stmontgomery: Right now, I don't need or want this property to be controllable outside the project, so I think it's appropriate as-is. I expect to formalize some mechanism to opt-in in a subsequent PR.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
stmontgomery: Replied there, and in the mean time we need this so the tests can opt-in.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
stmontgomery: I tired that initially but it caused an unrelated compiler crash in `main` snapshot toolchains

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedIssue.swift
stmontgomery: I'm a bit wary of sharing types like that since I fear it might cause us to overlook something when evaluating whether changes have any ABI impact. I can certainly mark it `Sendable` though.

Tests/TestingTests/EventRecorderTests.swift
stmontgomery: `Regex` isn't `Sendable` :(

Documentation/ABI/JSON.md
stmontgomery: I had that thought as well; I've taken an initial stab at that in `JSON.md`.

Tests/TestingTests/RunnerTests.swift
stmontgomery: Added

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
stmontgomery: After further consideration I think I don't really want or need to alter the symbol for the _test ended_ or _run ended_ events; if the only issues recorded were warnings, I think we can use the standard `.pass` (e.g. green diamond with checkmark) symbol. However, I think for _issue recorded_ events, the current `.warning` (e.g. yellow triangle with exclamation mark) is the most appropriate, especially since it's so recognizable as a warning. I plan to revert usage of the `.warning` symbol in this file except for the `.issueRecorded` event.

I think the content of the message itself will further help to distinguish these things. Warnings associated with specific tests will have a format like "Test 'name' passed with N warnings" format to them, whereas system wranings will be formatted quite differently.

Documentation/ABI/JSON.md
```diff
@@ -210,7 +210,22 @@ sufficient information to display the event in a human-readable format.
...
+<issue-severity> ::= "warning" | "error" ; additional severity levels may be
+                                         ; added in the future
+-->
+
 <!--
   ["testID": <test-id>,
     ["testCase": <test-case>]]
 -->
+
+## Version history
```
grynspan, line 224: I was actually planning to rename the existing file "ABIv0.md" or similar and create a new one to cover ABIv1. I was suggesting we sit down, write a list of what we want to do for v1, and draft it in an entirely separate PR.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: Aren't the tests passing `isWarningIssueRecordedEventEnabled`?

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: You filed an issue, yes? :)

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
grynspan: We could instead use a yellow X (since we use a grey X for known issues)? I really would like us to avoid conflating these issues with existing uses of `.warning` if we can.

stmontgomery:
@swift-ci please test

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
stmontgomery: This `switch` case doesn't actually _enable_ the feature for ABI**v1**, it ensures that it remains _disabled_ for ABI**v0**. We definitely want that safeguard, even if we decide not to enable this feature in ABIv1.

Documentation/ABI/JSON.md
stmontgomery: I've broken out the "pending changes" into a standalone document, with a clear indication that everything there is tentative.

Sources/Testing/ABI/v0/Encoded/ABIv0.EncodedMessage.swift
```diff
@@ -25,7 +25,8 @@ extension ABIv0 {
       case `default`
       case skip
       case pass
-      case passWithKnownIssue
+      case passWithWarning
+      case passWithKnownIssue = "_passWithKnownIssue"
```
grynspan, line 29: Swapped these by mistake?

Sources/Testing/Events/Recorder/Event.Symbol.swift
```diff
@@ -22,9 +22,11 @@ extension Event {
...
     ///
     /// - Parameters:
-    ///   - knownIssueCount: The number of known issues encountered by the end
-    ///     of the test.
-    case pass(knownIssueCount: Int = 0)
+    ///   - warningIssueCount: The number of warnings recorded for the test.
+    ///     The default value is `0`.
+    ///   - knownIssueCount: The number of known issues recorded for the test.
+    ///     The default value is `0`.
+    case pass(warningIssueCount: Int = 0, knownIssueCount: Int = 0)
```
grynspan, line 29: Adding an associated value is an ABI-breaking change.

stmontgomery:
@swift-ci please test Linux

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: As of right now, it's dead code since the default value is always `false` though.

Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -152,13 +152,14 @@ extension Configuration {
...
-    if let configuration, configuration.deliverExpectationCheckedEvents {
+    if let configuration, configuration.eventHandlingOptions.isExpectationCheckedEventEnabled {
       Self._deliverExpectationCheckedEventsCount.decrement()
     }
   }
 
   /// An atomic counter that tracks the number of "current" configurations that
-  /// have set ``deliverExpectationCheckedEvents`` to `true`.
+  /// have set ``EventHandlingOptions/isExpectationCheckedEventEnabled`` to
+  /// `true`.
```
grynspan, line 162: Mind deleting the "On older platforms" blurb here? It's not accurate (I was trying to use `Atomic` but ran into technical issues.)

Sources/Testing/Running/Configuration.swift
```diff
@@ -178,14 +178,32 @@ public struct Configuration: Sendable {
...
+  public struct EventHandlingOptions: Sendable {
+    /// Whether or not events of the kind ``Event/Kind-swift.enum/issueRecorded(_:)``
+    /// containing issues with warning (or lower) severity should be delivered
+    /// to the event handler of the configuration these options are applied to.
+    ///
+    /// By default, events matching this criteria are not delivered to event
+    /// handlers since this is an experimental feature.
+    ///
+    /// - Warning: Warning issues are not yet an approved feature.
+    public var isWarningIssueRecordedEventEnabled: Bool = false
```
grynspan, line 192: Please make this property `@_spi(Experimental)`. The structure is fine though.

Documentation/ABI/Pending-ABIv1.md
```diff
@@ -0,0 +1,20 @@
+# Pending ABIv1 Changes
```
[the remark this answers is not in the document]
stmontgomery: https://github.com/swiftlang/swift-testing/pull/946

Documentation/ABI/JSON.md
stmontgomery: Now in #946

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test macOS

Tests/TestingTests/EventRecorderTests.swift
grynspan, line 186: If you use `#"this"#` syntax, you don't need double-backslashes.

Tests/TestingTests/EventRecorderTests.swift
```diff
@@ -204,28 +209,13 @@ struct EventRecorderTests {
...
-      Optionally("s")
-      ")."
-    }
-    let match = try #require(
-      buffer
-        .split(whereSeparator: \.isNewline)
-        .compactMap(testFailureRegex.wholeMatch(in:))
-        .first
+    let expectedSuffixRegex = try Regex(expectedPattern)
+    _ = try #require(buffer
```
grynspan, line 213: Use `#expect()` and `!= nil`?

grynspan (approved):
A couple of test tweaks, but otherwise :shipit:

stmontgomery:
@swift-ci please test

hjyamauchi:
It seems like this PR uncovered a compiler crash error: https://github.com/swiftlang/swift/issues/79304 

I don't think it is this PR's fault but it just uncovered a latent bug in the compiler.

The crashes are seen on the CIs:
https://ci.swift.org/job/oss-swift-incremental-RA-macos-apple-silicon/8156/consoleText
https://ci-external.swift.org/job/swift-main-windows-toolchain/1055/consoleText

I'm not sure what's the best way forward, but I wonder if we could work around like editing `issueCounts` code somehow or even temporarily reverting this PR until the compiler issue is fixed.

PR #932, opened 2025-01-28
Missing an explicit include of `<mach-o/dyld.h>` on Apple platforms.

grynspan:
On some Apple platforms (e.g. iOS), the set of includes we have in `_TestingInternals` doesn't transitively include dyld.h, causing a build failure when building for those platforms. So explicitly include it.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

grynspan:
Change won't affect Windows, so I'm merging this before Windows CI completes. 😇

PR #934, opened 2025-01-28
Implement `CommandLine.executablePath` on all Apple platforms.

grynspan:
This change just makes sure `CommandLine.executablePath` is implemented consistently across Apple platforms and doesn't hit the unimplemented/warning path on e.g. iOS. (We're not using this property there, but the source file it lives in gets compiled regardless.)

### Checklist:


stmontgomery (approved):

grynspan:
@swift-ci test

PR #936, opened 2025-01-31
Update `Comment` to use `CustomTestStringConvertible` during string interpolation.

grynspan:
This PR ensures that `Comment`, when constructed from a string literal with interpolations, stringifies interpolated values using `String.init(descriptionForTest:)` rather than the default behaviour (i.e. `String.init(description:)`.)

For example:

```swift
enum E: CustomTestStringConvertible {
  case a
  case b
  case c

  var testDescription: String {
    switch self {
    case .a:
      "Once I was the King of Spain"
    case .b:
      "Now I eat humble pie"
    case .c:
      "Now I vacuum the turf at SkyDome™"
    }
  }
}

 #expect(1 == 2, "\(E.a) / \(E.b) / \(E.a) / \(E.c)")
```

Before:

> 🛑 Expectation failed: 1 == 2 - .a / .b / .a / .c

After:

> 🛑 Expectation failed: 1 == 2 - Once I was the King of Spain / Now I eat
> humble pie / Once I was the King of Spain / Now I vacuum the turf at SkyDome™

### Checklist:

### DIFF
--- a/Sources/Testing/Traits/Comment.swift
+++ b/Sources/Testing/Traits/Comment.swift
@@ -77,15 +77,50 @@ public struct Comment: RawRepresentable, Sendable {
   }
 }
 
-// MARK: - ExpressibleByStringLiteral, ExpressibleByStringInterpolation, CustomStringConvertible
+// MARK: - ExpressibleByStringLiteral, CustomStringConvertible
 
-extension Comment: ExpressibleByStringLiteral, ExpressibleByStringInterpolation, CustomStringConvertible {
+extension Comment: ExpressibleByStringLiteral, CustomStringConvertible {
   public init(stringLiteral: String) {
     self.init(rawValue: stringLiteral, kind: .stringLiteral)
   }
 
   public var description: String {
     rawValue
+  }
+}
+
+// MARK: - ExpressibleByStringInterpolation
+
+extension Comment: ExpressibleByStringInterpolation {
+  public init(stringInterpolation: StringInterpolation) {
+    self.init(rawValue: stringInterpolation.rawValue)
+  }
+
+  /// The string interpolation handler type for ``Comment``.
+  @_documentation(visibility: private)
+  public struct StringInterpolation: StringInterpolationProtocol, Sendable {
+    /// Storage for the string constructed by this instance.
+    ///
+    /// `DefaultStringInterpolation` in the Swift standard library also simply
+    /// accumulates its result in a string.
+    @usableFromInline var rawValue: String = ""
+
+    @inlinable public init(literalCapacity: Int, interpolationCount: Int) {}
+
+    @inlinable public mutating func appendLiteral(_ literal: String) {
+      rawValue += literal
+    }
+
+    @inlinable public mutating func appendInterpolation(_ value: some Any) {
+      rawValue += String(describingForTest: value)
+    }
+
+    @inlinable public mutating func appendInterpolation(_ value: (some Any)?) {
+      // This overload is provided so that the compiler does not emit warnings
+      // about optional values in string interpolations (which we are fine with
+      // when constructing Comment instances.)
+      rawValue += String(describingForTest: value)
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci test

Sources/Testing/Traits/Comment.swift
```diff
@@ -89,6 +89,41 @@ extension Comment: ExpressibleByStringLiteral, ExpressibleByStringInterpolation,
...
+
+    @inlinable public mutating func appendInterpolation(_ value: some Any) {
+      rawValue += String(describingForTest: value)
+    }
+
+    @inlinable public mutating func appendInterpolation(_ value: (some Any)?) {
+      // This overload is provided so that the compiler does not emit warnings
+      // about optional values in string interpolations (which we are fine with
+      // when constructing Comment instances.)
+      rawValue += String(describingForTest: value)
```
stmontgomery, line 122: Not sure about this, might be too magical and surprising compared to how string literal interpolation usually works. Is there a strong reason to include it?

Tests/TestingTests/Traits/CommentTests.swift
```diff
@@ -57,6 +57,15 @@ struct CommentTests {
   func explicitlyNilComment() {
     #expect(true as Bool, nil as Comment?)
   }
+
+  @Test("String interpolation")
+  func stringInterpolation() {
+    let value1: Int? = 123
+    let value2: Int? = nil
+    let value3: Any.Type = Int.self
+    let comment: Comment = "abc\(value1)def\(value2)ghi\(value3)"
```
stmontgomery, line 66: I'd include some type which explicitly conforms to `CustomTestStringConvertible`, and customizes the value of its `testDescription` property to differ from `description`.

Sources/Testing/Traits/Comment.swift
grynspan: My expectation here (heh) is that if you pass an optional value in a testing context, you want to see its value. Also the warning's fix-it is emitted by the compiler and tells you to use `String(describing:)`, not `String(describingForTest:)`, which is wrong.

Tests/TestingTests/Traits/CommentTests.swift
grynspan: `Int?` conforms to `CustomTestStringConvertible`.

stmontgomery: Ok, that's somewhat subtle but I see.

Sources/Testing/Traits/Comment.swift
stmontgomery: I wouldn't say it's _wrong_ for the compiler to suggest that; users are welcome to call `String(describing:)` themselves. I think `String(describingForTest:)` is primarily for the testing library to call on values users provide, and (to a lesser extent) for use by test authors when writing helpers/utilities.

I think this violates the principle of least surprise. Regardless of what we think of the compiler's fix-it, users are generally accustomed to needing to handle optionals in string interpolations, and I don't know of a reason to deviate from that precedent here. It makes it seem as if the rules are more lax just because it's a test.

grynspan: The compiler emits that diagnostic because the default string representation of an optional value may either be `"Optional(x)"` (where `x` is the wrapped value's description) or `nil`, and neither of those values is likely expected by a developer who wants to just print `x`. In our case, that doesn't hold because optional values already have special handling in the library.

I would expect the following code to "just work" rather than emitting any diagnostics, for instance:

```swift
let x: Int? = f()
#expect(x == 10, "\(x) should be 10!")
```

stmontgomery (approved):
Great enhancement overall. I'm still not [convinced](https://github.com/swiftlang/swift-testing/pull/936/files#r1937919065) about the merit of the overload which suppresses warnings for optionals, and worry it could be surprising or confusing. But it's admittedly a minor concern and I don't want to stand in the way of the broader PR.

Tests/TestingTests/Traits/CommentTests.swift
grynspan: I can add additional testing with a local type too.

grynspan:
@swift-ci test

stmontgomery (approved):

PR #938, opened 2025-02-03
Avoid using `String` while holding an internal libobjc lock.

grynspan:
`String` can inadvertently touch libobjc when we implicitly cast it to a C string (when calling `getsectiondata()`), which can result in a deadlock if that operation needs to acquire one of libobjc's internal locks.

Switch to `StaticString` which we can guarantee never touches libobjc.

Resolves rdar://144093524.

### Checklist:


grynspan:
@swift-ci test

briancroom (approved):

PR #939, opened 2025-02-04
Add a `type` argument to the ABI accessor function signature.

grynspan:
We access the test content record section directly instead of using some (non-existent) type-safe Swift API. On some platforms (Darwin in particular), it is possible for two copies of the testing library to be loaded at runtime, and for them to have incompatible definitions of types like `Test`. That means one library's `@Test` macro could produce a test content record that is then read by the other library's discovery logic, resulting in a stack smash or worse.

We can resolve this issue by adding a `type` argument to the accessor function we define for test content records; the body of the accessor can then compare the value of this argument against the expected Swift type of its result and, if they don't match, bail early.

The new argument is defined as a pointer _to_ a Swift type rather than as a Swift type directly because `@convention(c)` functions cannot directly reference Swift types. The value of the new argument can be safely ignored if the type of the test content record's value is and always has been `@frozen`.

### Checklist:

### DIFF
--- a/Documentation/ABI/TestContent.md
+++ b/Documentation/ABI/TestContent.md
@@ -41,10 +41,16 @@ Regardless of platform, all test content records created and discoverable by the
 testing library have the following layout:
 
 ```swift
+typealias Accessor = @convention(c) (
+  _ outValue: UnsafeMutableRawPointer,
+  _ type: UnsafeRawPointer,
+  _ hint: UnsafeRawPointer?
+) -> CBool
+
 typealias TestContentRecord = (
   kind: UInt32,
   reserved1: UInt32,
-  accessor: (@convention(c) (_ outValue: UnsafeMutableRawPointer, _ hint: UnsafeRawPointer?) -> CBool)?,
+  accessor: Accessor?,
   context: UInt,
   reserved2: UInt
 )
@@ -54,10 +60,16 @@ This type has natural size, stride, and alignment. Its fields are native-endian.
 If needed, this type can be represented in C as a structure:
 
 ```c
+typedef bool (* SWTAccessor)(
+  void *outValue,
+  const void *type,
+  const void *_Nullable hint
+);
+
 struct SWTTestContentRecord {
   uint32_t kind;
   uint32_t reserved1;
-  bool (* _Nullable accessor)(void *outValue, const void *_Null_unspecified hint);
+  SWTAccessor _Nullable accessor;
   uintptr_t context;
   uintptr_t reserved2;
 };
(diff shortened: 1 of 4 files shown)


Sources/Testing/Discovery.swift
```diff
@@ -24,11 +44,27 @@ private import _TestingInternals
...
+/// - Parameters:
+///   - lhsAddress: A pointer to (not a bitcast) a Swift type.
+///   - rhs: Another Swift type.
+///
+/// - Returns: Whether or not the type at `lhsAddress` and `rhs` are equivalent.
+///
+/// - Warning: This function is used to implement the `@Test` macro. Do not use
+///   it directly.
+public func __type(at lhsAddress: UnsafeRawPointer, is rhs: (some ~Copyable).Type) -> Bool {
+  let lhs = TypeInfo(describing: lhsAddress.load(as: (any ~Copyable.Type).self))
```
grynspan, line 63: This function is currently necessary because `==` is not defined for move-only metatypes (rdar://134276458), but once it is, we can just use it directly in the accessor function bodies.

Documentation/ABI/TestContent.md
```diff
@@ -105,42 +117,59 @@ If `accessor` is `nil`, the test content record is ignored. The testing library
...
 
-The second argument to this function, `hint`, is an optional input that can be
+The second argument to this function, `type`, is a pointer to the type[^mightNotBeSwift]
+(not a bitcast Swift type) of the value expected to be written to `outValue`.
+This argument helps to prevent memory corruption if two copies of Swift Testing
+or a third-party library are inadvertently loaded into the same process. If the
+value at `type` does not match the test content record's expected type, the
+accessor function must return `false` and must not modify `outValue`.
+
+<!-- TODO: discuss this argument's value in Embedded Swift (no metatypes) -->
```
grynspan, line 127: In Embedded Swift, we can just say "the value of this argument is unspecified."

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

grynspan:
> LGTM. Is it important that the mechanism to distinguish records be the type of `outValue`, in particular? Or could it be something else? Just curious what led to that choice

The implementation actually does use a different type for `Test` for technical reasons (non-nominal types). Making it the same type by default simplifies the discovery machinery, because it means it has a reasonable default value to pass for the argument.

If/when we make the `TestContent` protocol public, I don't know if we'll want to expose that bit too, but we'll see.

PR #942, opened 2025-02-05
Use `os_unfair_lock` on Darwin when available.

grynspan:
This PR refactors the `Locked` type to be generic over the type of lock it uses, allowing us to adopt `os_unfair_lock` in most places, but also allowing us to use other lock types as needed (i.e. `pthread_mutex_t` in WaitFor.swift if libdispatch is unavailable.)

Under `-O`, locking and unlocking is fully inlined, while initialization (which is relatively rare) is partially, but not completely, inlined.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/WaitFor.swift
+++ b/Sources/Testing/ExitTests/WaitFor.swift
@@ -80,7 +80,7 @@ func wait(for pid: consuming pid_t) async throws -> ExitCondition {
 }
 #elseif SWT_TARGET_OS_APPLE || os(Linux) || os(FreeBSD) || os(OpenBSD)
 /// A mapping of awaited child PIDs to their corresponding Swift continuations.
-private let _childProcessContinuations = Locked<[pid_t: CheckedContinuation<ExitCondition, any Error>]>()
+private let _childProcessContinuations = LockedWith<pthread_mutex_t, [pid_t: CheckedContinuation<ExitCondition, any Error>]>()
 
 /// A condition variable used to suspend the waiter thread created by
 /// `_createWaitThread()` when there are no child processes to await.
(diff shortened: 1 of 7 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/Support/Locked.swift
```diff
@@ -198,4 +185,9 @@ extension Locked {
   init<K, V>() where T == Dictionary<K, V> {
     self.init(rawValue: [:])
   }
+
+  /// Initialize an instance of this type with a raw value of `nil`.
```
stmontgomery, line 189:
```suggestion
  /// Initialize an instance of this type with a raw value of `[]`.
```

Sources/Testing/Support/Locked.swift
```diff
@@ -21,67 +52,26 @@ internal import _TestingInternals
...
-#elseif os(WASI)
-  // No locks on WASI without multithreaded runtime.
-  typealias PlatformLock = Never
-#else
-#warning("Platform-specific implementation missing: locking unavailable")
-  typealias PlatformLock = Never
-#endif
+struct LockedWith<L, T>: RawRepresentable where L: Lockable {
+  /// The type of the underlying lock that guards instances of this type.
+  typealias UnderlyingLock = L
```
stmontgomery, line 57: Is this typealias used anywhere?

grynspan: In tests in particular, yes.

grynspan:
@swift-ci test

stmontgomery (approved):

grynspan:
@swift-ci test

Sources/Testing/ExitTests/WaitFor.swift
```diff
@@ -80,7 +80,7 @@ func wait(for pid: consuming pid_t) async throws -> ExitCondition {
 }
 #elseif SWT_TARGET_OS_APPLE || os(Linux) || os(FreeBSD) || os(OpenBSD)
 /// A mapping of awaited child PIDs to their corresponding Swift continuations.
-private let _childProcessContinuations = Locked<[pid_t: CheckedContinuation<ExitCondition, any Error>]>()
+private let _childProcessContinuations = LockedWith<pthread_mutex_t, [pid_t: CheckedContinuation<ExitCondition, any Error>]>()
```
stmontgomery, line 83: Why do you need to specifically use LockedWith and specify pthread_mutex here instead of Locked?

grynspan: Because we explicitly need a `pthread_mutex_t` so we can use a `pthread_cond_t` with it. In the future, we might want to e.g. change the Linux default lock to a futex, but this can't change.

stmontgomery (approved):

PR #945, opened 2025-02-07
Add a (internal-only) `CustomIssueRepresentable` protocol.

grynspan:
This PR adds a (for now internal-only) protocol that lets us transform arbitrary errors to arbitrary issues (rather than always emitting them as `.errorCaught`). This protocol replaces the special-casing we were doing for `SystemError` and `APIMisuseError`.

This change also causes us to emit an API misuse issue if a developer passes an error of type `ExpectationFailedError` to `Issue.record()` as we can reasonably assume that such an error was already recorded correctly as `.expectationFailed`.

### Checklist:


grynspan:
@swift-ci test

Sources/Testing/Support/CustomIssueRepresentable.swift
```diff
@@ -0,0 +1,86 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// A protocol that provides instances of conforming types with the ability to
+/// record themselves as test issues.
+///
+/// When a type conforms to this protocol, values of that type can be passed to
+/// ``Issue/record(_:_:)``. The testing library then calls the
+/// ``represent(as:)`` function and passes it an instance of ``Issue`` that will
```
stmontgomery, line 16:
```suggestion
/// ``customize(_:)`` function and passes it an instance of ``Issue`` that will
```

Sources/Testing/Support/CustomIssueRepresentable.swift
```diff
@@ -0,0 +1,86 @@
...
+/// types to be represented as issues, we will need to add an overload of
+/// `Issue.record()` that takes `some CustomIssueRepresentable`.
+protocol CustomIssueRepresentable: Error {
+  /// Customize the issue that will represent this value.
+  ///
+  /// - Parameters:
+  ///   - issue: The issue to customize. The function consumes this value.
+  ///
+  /// - Returns: A customized copy of `issue`.
+  func customize(_ issue: consuming Issue) -> Issue
```
stmontgomery, line 32: It could be useful to return Optional, so that an error could elect to suppress recording an issue representing itself

grynspan: That would require us to change `Issue.record()` to return an optional value, which would be API- and ABI-breaking. If we made this a public feature, we might want to do that, but it's not helpful for the current use case. Make sense?

grynspan:
@swift-ci test

Sources/Testing/Support/CustomIssueRepresentable.swift
stmontgomery: Sounds good. If we do move forward with promoting this to public, I do think we ought to consider it

stmontgomery (approved):

Sources/Testing/Support/CustomIssueRepresentable.swift
grynspan: Fair enough.

PR #947, opened 2025-02-11
Suppress warning about `#require(nonOptional)` in some cases.

grynspan:
When using `try #require()` to unwrap an optional value, we emit a compile-time warning if the value is not actually optional. However, there is a bug in the type checker (https://github.com/swiftlang/swift/issues/79202) that triggers a false positive when downcasting an object (of `class` type), e.g.:

```swift
class Animal {}
class Duck: Animal {}
let beast: Animal = Duck()
let definitelyADuck = try #require(beast as? Duck)
// ⚠️ '#require(_:_:)' is redundant because 'beast as? Duck' never equals 'nil'
```

This change suppresses the warning we emit if the expression contains certain syntax tokens (namely `?`, `nil`, or `Optional`) on the assumption that their presence means the test author is expecting an optional value and we've hit a false positive.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/ConditionMacro.swift
+++ b/Sources/TestingMacros/ConditionMacro.swift
@@ -324,7 +324,24 @@ public struct NonOptionalRequireMacro: RefinedConditionMacro {
     in context: some MacroExpansionContext
   ) throws -> ExprSyntax {
     if let argument = macro.arguments.first {
+#if !SWT_FIXED_137943258
+      // Silence this warning if we see a token (`?`, `nil`, or "Optional") that
+      // might indicate the test author expects the expression is optional.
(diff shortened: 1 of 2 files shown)


grynspan:
@swift-ci test

briancroom:
Is there an additional test you could add which would be an expected failure right now, but pass once https://github.com/swiftlang/swift/issues/79202 is resolved?

briancroom (approved):

grynspan:
> Is there an additional test you could add which would be an expected failure right now, but pass once https://github.com/swiftlang/swift/issues/79202 is resolved?

No, because I don't have a way to trigger the mismatch except during macro expansion, which is faked (different macro overloads get different names.)

briancroom (approved):

PR #960, opened 2025-02-14
Add an overview of test serialization to the XCTest migration guide

iamleeg:
Describe test serialization in the article about migrating tests from XCTest.

### Motivation:

Because XCTest runs tests in a suite serially by default, tests that people migrate from XCTest may encounter issues if they run in parallel.

### Modifications:

Add an example of serializing a test suite to the migration guide, along with links to the docs about test parallelization.

### Result:

The documentation on migrating tests from XCTest includes guidance on serializing tests within a suite.

### Checklist:

- [X] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [X] If public symbols are renamed or modified, DocC references should be updated.

PR #963, opened 2025-02-18
Corrects use of XCTAssertEqual macro in parallelization examples.

iamleeg:
Fixes a typo in the "before" example of adopting test serialization.

### Motivation:

The example of XCTest tests that need to run in series doesn't compile.

### Modifications:

Correct the code in the example.

### Result:

A documentation change that means the code snippet in the example is correct.

### Checklist:


stmontgomery (approved):

stmontgomery:
Follow-on to #960

PR #967, opened 2025-02-19
Avoid calling `unsafeBitCast(_:to:)` to cast C function pointers.

grynspan:
This PR replaces calls to `unsafeBitCast(_:to:)` (where the value being cast is a C function pointer) with calls to an internal function that checks that the value being cast is actually a C function pointer. This is intended to make the code in question more self-documenting (by making it clear what kind of cast is being performed.)

The function does then call `unsafeBitCast(_:to:)`, but it's in one place only instead of many places—the call sites are all now self-documenting.

Also fixed one place we're using `unsafeBitCast(_:to:)` to cast a metatype where, as of Swift 6.1, we don't need to anymore.

### Checklist:

### DIFF
--- a/Sources/Testing/Parameterization/TypeInfo.swift
+++ b/Sources/Testing/Parameterization/TypeInfo.swift
@@ -265,9 +265,11 @@ extension TypeInfo {
     }
     switch _kind {
     case let .type(type):
-      // _mangledTypeName() works with move-only types, but its signature has
-      // not been updated yet. SEE: rdar://134278607
+#if compiler(>=6.1)
+      return _mangledTypeName(type)
+#else
       return _mangledTypeName(unsafeBitCast(type, to: Any.Type.self))
+#endif
     case let .nameOnly(_, _, mangledName):
       return mangledName
     }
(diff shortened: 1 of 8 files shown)


grynspan:
@swift-ci test

Sources/Testing/Parameterization/TypeInfo.swift
```diff
@@ -265,9 +265,11 @@ extension TypeInfo {
     }
     switch _kind {
     case let .type(type):
-      // _mangledTypeName() works with move-only types, but its signature has
-      // not been updated yet. SEE: rdar://134278607
+#if compiler(>=6.1)
+      return _mangledTypeName(type)
```
stmontgomery, line 269: I see the `#if compiler()` guard, but does this not have differing API availability?

stmontgomery (approved):

Sources/Testing/Parameterization/TypeInfo.swift
grynspan: Nope, see https://github.com/swiftlang/swift/pull/76041

PR #970, opened 2025-02-21
Restrict use of a new overload of `contains()` to Apple platforms that have it.

grynspan:
We have one test that uses a newer overload of [`contains()`](https://developer.apple.com/documentation/swift/collection/contains(_:)) that doesn't exist on older Apple platforms. This PR ensures that our use of that method is constrained to platforms that have it.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

grynspan:
@swift-ci test Linux

PR #974, opened 2025-02-25
Change the return types of test content accessors and add `ExitTest.current`.

grynspan:
This PR changes the signatures of the accessor functions for the test content metadata section so that we're not using underscored types. Instead, a `Test.Generator` type replaces `Test._Record` and the bare function type we were using for `Test` and `ExitTest` is promoted to (still experimental) API instead of hiding behind underscores.

This change allows the types produced by accessor functions to always be nominal types, avoids the iffy substitution of `_Record` for `@Sendable () async -> Test` at runtime, and simplifies the internal `TestContent` protocol by eliminating the need for an associated type to stand in for the conforming type.

This change leaves `ExitTest` as a nominal, but empty, type. I've added `ExitTest.current` to make it useful so that test code, especially libraries, can detect if it's running inside an exit test or not. I will separately update the exit test proposal PR to include this type and property.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/ConditionMacro.swift
+++ b/Sources/TestingMacros/ConditionMacro.swift
@@ -435,7 +435,7 @@ extension ExitTestConditionMacro {
 
     // TODO: use UUID() here if we can link to Foundation
     let exitTestID = (UInt64.random(in: 0 ... .max), UInt64.random(in: 0 ... .max))
-    let exitTestIDExpr: ExprSyntax = "Testing.__ExitTest.ID(__uuid: (\(literal: exitTestID.0), \(literal: exitTestID.1)))"
+    let exitTestIDExpr: ExprSyntax = "(\(literal: exitTestID.0), \(literal: exitTestID.1))"
 
     var decls = [DeclSyntax]()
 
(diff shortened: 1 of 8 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/TestingMacros/ConditionMacro.swift
```diff
@@ -435,7 +435,7 @@ extension ExitTestConditionMacro {
 
     // TODO: use UUID() here if we can link to Foundation
     let exitTestID = (UInt64.random(in: 0 ... .max), UInt64.random(in: 0 ... .max))
-    let exitTestIDExpr: ExprSyntax = "Testing.__ExitTest.ID(__uuid: (\(literal: exitTestID.0), \(literal: exitTestID.1)))"
+    let exitTestIDExpr: ExprSyntax = "(\(literal: exitTestID.0), \(literal: exitTestID.1))"
```
stmontgomery, line 438: Is the switch to a tuple because the ID type is now `@_spi`?

grynspan: Yes, it reduces the number of symbols we need to make `public`. The current protocol-based discovery mechanism will go away once we get `@section`, and the PR for that should make it more clear how we consume this value.

stmontgomery (approved):

PR #975, opened 2025-02-25
Rename `ExitTestArtifacts` and split `ExitCondition` in twain.

grynspan:
This PR renames `ExitTestArtifacts` to `ExitTest.Result` and splits `ExitCondition` into two types: `ExitTest.Condition` which can be passed to `#expect(exitsWith:)` and `StatusAtExit` which represents the raw, possibly platform-specific status reported by the kernel when a child process terminates.

The latter type is not nested in `ExitTest` because it can be used independently of exit tests and we may want to use it in the future for things like multi-process parallelization, but if a platform supports spawning processes but not exit tests, nesting it in `ExitTest` would make it unavailable.

I considered several names for `StatusAtExit`:

- `ExitStatus`: too easily confusable with exit _codes_ such as `EXIT_SUCCESS`;
- `ProcessStatus`: we don't say "process" in our API surface elsewhere;
- `Status`: too generic
- `ExitReason`: "status" is a more widely-used term of art for this concept.

Foundation uses `terminationStatus` to represent the raw integer value and `Process.TerminationReason` to represent whether it's an exit code or signal. We don't use "termination" in Swift Testing's API anywhere.

I settled on `StatusAtExit` because it was distinct and makes it clear that it represents the status of a process _at exit time_ (as opposed to while running, e.g. `enum ProcessStatus { case running; case suspended; case terminated }`.

Updates to the exit tests proposal document will follow in a separate PR.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/CMakeLists.txt
```diff
@@ -31,10 +31,11 @@ add_library(Testing
   Events/Recorder/Event.JUnitXMLRecorder.swift
   Events/Recorder/Event.Symbol.swift
   Events/TimeValue.swift
-  ExitTests/ExitCondition.swift
   ExitTests/ExitTest.swift
-  ExitTests/ExitTestArtifacts.swift
+  ExitTests/ExitTest.Condition.swift
+  ExitTests/ExitTest.Result.swift
   ExitTests/SpawnProcess.swift
+  ExitTests/StatusAtExit.swift
```
stmontgomery, line 38: You mentioned in the PR description this type may be used independently of exit tests, so I'm wondering if the file ought to be placed in a different directory to emphasize that.

Sources/Testing/ExitTests/ExitTest.Condition.swift
```diff
@@ -0,0 +1,148 @@
...
+  ///
+  /// Values of this type are used to describe the conditions under which an
+  /// exit test is expected to pass or fail by passing them to
+  /// ``expect(exitsWith:observing:_:sourceLocation:performing:)`` or
+  /// ``require(exitsWith:observing:_:sourceLocation:performing:)``.
+  public struct Condition: Sendable {
+    /// An enumeration describing the possible requirements for an exit test.
+    private enum _Kind: Sendable, Equatable {
+      /// The exit test must exit with a particular exit status.
+      case statusAtExit(StatusAtExit)
```
stmontgomery, line 28: Even if you do go with the current type name (`StatusAtExit`), I'm not so sure you need to repeat the "AtExit" part in the name of this `case`
```suggestion
      case status(StatusAtExit)
```

Sources/Testing/ExitTests/ExitTest.Condition.swift
```diff
@@ -0,0 +1,148 @@
...
+
+      /// The exit test must exit with any failure.
+      case failure
+    }
+
+    /// The kind of requirement.
```
stmontgomery, line 34: You use the term "requirement" in many places but the actual type is named "condition". I get that it's helpful to use an alternate word occasionally in documentation to explain its purpose, but I wonder if it would be more clear to use the word condition in _most_ places, such as this one (and others throughout this PR too).
```suggestion
    /// The kind of condition.
```

Sources/Testing/ExitTests/ExitTest.Condition.swift
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// MARK: -
+
+@_spi(Experimental)
+#if SWT_NO_EXIT_TESTS
+@available(*, unavailable, message: "Exit tests are not available on this platform.")
+#endif
+extension ExitTest.Condition {
+  /// The process terminated successfully with exit code `EXIT_SUCCESS`.
```
stmontgomery, line 46: The phrasing of this documentation feels weird for a property. (Similar comment for `.failure`, `.exitCode(_:)`, and `.signal(_:)`.) Maybe something like this?
```suggestion
  /// A condition which matches when a process terminates successfully
  /// with exit code `EXIT_SUCCESS`.
```

Sources/Testing/ExitTests/ExitTest.Condition.swift
```diff
@@ -0,0 +1,148 @@
...
+#if SWT_NO_EXIT_TESTS
+@available(*, unavailable, message: "Exit tests are not available on this platform.")
+#endif
+extension ExitTest.Condition {
+  /// Check whether or not an exit test condition matches a given exit status.
+  ///
+  /// - Parameters:
+  ///   - other: Another value to compare.
+  ///
+  /// - Returns: Whether or not `self` and `other` are equal.
```
stmontgomery, line 133: References to `other` need to be updated

Sources/Testing/ExitTests/ExitTest.Condition.swift
```diff
@@ -0,0 +1,148 @@
...
+  ///
+  /// Two exit test requirements can be compared; if either instance is equal to
+  /// ``failure``, it will compare equal to any instance except ``success``.
+  func isApproximatelyEqual(to statusAtExit: StatusAtExit) -> Bool {
```
stmontgomery, line 137: Since this type is called "Condition", maybe `isSatisfied(by:)` would be more fitting? I find it confusing to be considering this a form of _equality_; I think it's really more like "this status _satisfies_ this condition", or "_meets_", or something like that.

Sources/Testing/ExitTests/ExitTest.Result.swift
```diff
@@ -0,0 +1,86 @@
...
+  ///
+  /// Both ``expect(exitsWith:observing:_:sourceLocation:performing:)`` and
+  /// ``require(exitsWith:observing:_:sourceLocation:performing:)`` return
+  /// instances of this type.
+  public struct Result: Sendable {
+    /// The exit condition the exit test exited with.
+    ///
+    /// When the exit test passes, the value of this property is equal to the
+    /// exit status reported by the process that hosted the exit test.
+    public var statusAtExit: StatusAtExit
```
stmontgomery, line 27: Since this type is (now) `ExitTest.Result`, it feels sort of implied that a status it holds would be the status _at exit_. Like another comment elsewhere, could we omit "AtExit" from the property name?
```suggestion
    public var status: StatusAtExit
```

Sources/Testing/ExitTests/ExitTest.Result.swift
```diff
@@ -0,0 +1,86 @@
...
+    /// it exited.
+    ///
+    /// The value of this property may contain any arbitrary sequence of bytes,
+    /// including sequences that are not valid UTF-8 and cannot be decoded by
+    /// [`String.init(cString:)`](https://developer.apple.com/documentation/swift/string/init(cstring:)-6kr8s).
+    /// Consider using [`String.init(validatingCString:)`](https://developer.apple.com/documentation/swift/string/init(validatingcstring:)-992vo)
+    /// instead.
+    ///
+    /// When checking the value of this property, keep in mind that the standard
+    /// output stream is globally accessible, and any code running in an exit
```
stmontgomery, line 65:
```suggestion
    /// error stream is globally accessible, and any code running in an exit
```

Sources/Testing/ExitTests/StatusAtExit.swift
```diff
@@ -0,0 +1,73 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+private import _TestingInternals
+
+/// An enumeration describing possible status a process will yield on exit.
+///
+/// You can cast an instance of this type to an instance of
+/// ``ExitTest/Condition`` using ``ExitTest/Condition/init(_:)``. That value
```
stmontgomery, line 16:
```suggestion
/// You can initialize an instance of ``ExitTest/Condition`` with an instance
/// of this type using ``ExitTest/Condition/init(_:)``. That value
```

Sources/Testing/Expectations/ExpectationChecking+Macro.swift
```diff
@@ -1148,18 +1148,18 @@ public func __checkClosureCall<R>(
 @_spi(Experimental)
 public func __checkClosureCall(
   identifiedBy exitTestID: (UInt64, UInt64),
-  exitsWith expectedExitCondition: ExitCondition,
-  observing observedValues: [any PartialKeyPath<ExitTestArtifacts> & Sendable],
+  exitsWith requirement: ExitTest.Condition,
```
stmontgomery, line 1151: This is one of the situations I was thinking of in another comment about using the term `condition` most of the time.

Also, I see other decls still use the argument name `expectedExitCondition`

Sources/Testing/ExitTests/StatusAtExit.swift
```diff
@@ -0,0 +1,73 @@
...
+/// can then be used to describe the condition under which an exit test is
+/// expected to pass or fail by passing it to
+/// ``expect(exitsWith:observing:_:sourceLocation:performing:)`` or
+/// ``require(exitsWith:observing:_:sourceLocation:performing:)``.
+@_spi(Experimental)
+#if SWT_NO_PROCESS_SPAWNING
+@available(*, unavailable, message: "Exit tests are not available on this platform.")
+#endif
+public enum StatusAtExit: Sendable {
```
stmontgomery, line 25: I read the PR description with your rationale for choosing this name—thanks for that. To me, the most compelling name is `ExitStatus` though. I don't really find it very worrisome that it will be confused with exit _code_, and if someone _is_ confused, the'll quickly realize when they try to use an instance of this type that they're not the same. Plus documentation can help. Naming is subjective of course but I just find the "Status**AtExit**" wording kind of clunky.

stmontgomery, line 25: (Separate from the name of the type) Maybe we should make this a `struct` with a `Kind` enum. I always worry about exposing enums as public API since they're less flexible if we need to add another stored value which is common to all cases. Just to help future-proof

stmontgomery (commented):
Overall I like these refinements! Several specific comments

Sources/Testing/ExitTests/ExitTest.Condition.swift
grynspan: It felt weird typing just `status` elsewhere. In any case, this is a private enum, so we don't need to worry too much about it.

Sources/Testing/ExitTests/ExitTest.Condition.swift
grynspan: Just a typo from an earlier revision that I didn't post. Will fix.

Sources/Testing/ExitTests/StatusAtExit.swift
grynspan: As I discussed with @natecook1000: this structure directly models the values reported by processes when they terminate. There is approximately zero chance that the platforms we support today are going to come up with a whole new kind of process exit status, and if they do, they'll need to report it via an entirely new mechanism as `siginfo_t`, `wait()`, etc. have no extensibility.

Sources/Testing/ExitTests/StatusAtExit.swift
grynspan: Used "convert" instead of "cast".

Sources/Testing/ExitTests/ExitTest.Condition.swift
grynspan: This is an internal function and when you look at the use case, I don't think it's terribly confusion. Do you feel strongly?

Sources/Testing/ExitTests/ExitTest.Result.swift
grynspan: If we renamed the type `ExitStatus`, I'd still want to call it `exitStatus`. `status` is just too generic for my comfort.

Sources/Testing/ExitTests/StatusAtExit.swift
grynspan: I'd like to throw this to the crowd and get feedback from other folks, if that's okay.

Sources/Testing/ExitTests/ExitTest.Condition.swift
stmontgomery: I certainly don't mind as much after realizing it's not API, but I do think a different phrase would be more clear. To me, if you have a condition you evaluate it against something. If you say one condition "is approximately equal" to something else, I would assume that something else would itself be a condition. But the parameter is a status, i.e. a subject with which you evaluate the condition.

stmontgomery (approved):

PR #980, opened 2025-02-27
Note in documentation when `#expect(throws:)` started returning a value.

grynspan:
DocC doesn't let us annotate our return type's availability separately from our macros' availability, and we can't include both the old symbols and new ones because it is ambiguous at compile time, so this PR explains in prose when the return types were added.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

PR #952, opened 2025-02-12
Introduce a severity level for issues, and a 'warning' severity (un-revert #931)

stmontgomery:
This un-reverts #950, effectively reintroducing the changes recently landed in #931.

The revert was needed because it revealed a latent bug in the Swift compiler, tracked by https://github.com/swiftlang/swift/issues/79304. I reproduced that failure and included a workaround in the second commit on this PR.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #986, opened 2025-02-28
JUnit XML recorder should ignore warning issues

stmontgomery:
This updates `Event.JUnitXMLRecorder` to ignore `Issue` instances whose `severity` is less than `.error` (such as `.warning`).

### Motivation:

The concept of issue severity was recently added in #931 (but was reverted and re-landed in #952), and that did not adjust the JUnit XML recorder logic. The JUnit XML schema we currently attempt to adhere to does not appear to have a way to represent non-fatal issues, so I think it would be best for now to ignore these issues.

### Modifications:

- Implement the fix and a validating unit test.
- (Drive-by) Fix a nearby test I noticed wasn't actually working as intended and wasn't properly validating the fix it was intended to.

### Checklist:

### DIFF
--- a/Tests/TestingTests/EventRecorderTests.swift
+++ b/Tests/TestingTests/EventRecorderTests.swift
@@ -467,15 +467,18 @@ struct EventRecorderTests {
   @Test("JUnitXMLRecorder counts issues without associated tests")
   func junitRecorderCountsIssuesWithoutTests() async throws {
     let issue = Issue(kind: .unconditional)
-    let event = Event(.issueRecorded(issue), testID: nil, testCaseID: nil)
     let context = Event.Context(test: nil, testCase: nil, configuration: nil)
 
-    let recorder = Event.JUnitXMLRecorder { string in
-      if string.contains("<testsuite") {
-        #expect(string.contains(#"failures=1"#))
+    await confirmation { wroteTestSuite in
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

Tests/TestingTests/EventRecorderTests.swift
```diff
@@ -467,15 +467,35 @@ struct EventRecorderTests {
   @Test("JUnitXMLRecorder counts issues without associated tests")
   func junitRecorderCountsIssuesWithoutTests() async throws {
     let issue = Issue(kind: .unconditional)
-    let event = Event(.issueRecorded(issue), testID: nil, testCaseID: nil)
     let context = Event.Context(test: nil, testCase: nil, configuration: nil)
 
-    let recorder = Event.JUnitXMLRecorder { string in
-      if string.contains("<testsuite") {
-        #expect(string.contains(#"failures=1"#))
```
stmontgomery, line 475: This is a drive-by fix to the pre-existing test above the new one I added below: This one wasn't actually working, since no `.runEnded` event was recorded, so this callback wasn't happening. Once I fixed that I noticed a secondary issue, which is that the `"failures=1"` string was missing double quotes around the number 1.

grynspan (approved):

PR #988, opened 2025-03-02
Diagnose when using a non-escapable type as suite.

grynspan:
This PR introduces a custom compile-time diagnostic when attempting to use a non-escapable type as a suite. For example:

```swift
@Suite struct NumberOfBeesTests: ~Escapable {
  @Test borrowing func countBees() { ... } // 🛑 Attribute 'Test' cannot be applied to a function within structure 'NumberOfBeesTests' because its conformance to 'Escapable' has been suppressed
}
```

Values with non-escapable type cannot currently be initialized nor returned from a function, and we need to be able to do both in order to correctly implement the `@Test` macro.

This change does not diagnose if `@Suite` is applied to such a type but does not contain any test functions, because we do compile successfully in that case and this sort of pattern remains valid:

```swift
@Suite struct MyTests: ~Escapable {
  @Suite struct EndToEndTests {
    ...
  }
}
```

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -171,6 +171,7 @@ extension Array where Element == PackageDescription.SwiftSetting {
       .unsafeFlags(["-require-explicit-sendable"]),
       .enableUpcomingFeature("ExistentialAny"),
       .enableExperimentalFeature("SuppressedAssociatedTypes"),
+      .enableExperimentalFeature("NonescapableTypes"),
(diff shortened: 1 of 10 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

stmontgomery (approved):

PR #999, opened 2025-03-06
build: update missed callsites of `get_swift_testing_install_lib_dir`

compnerd:
This has been replaced with a global variable. Update the callsites to use that instead.


compnerd:
@swift-ci please test

grynspan (approved):

PR #972, opened 2025-02-23
Show the number of test cases that ran at the end of a test

hrayatnia:
### Motivation:

To have the number of test cases at the end of the test in for parameterized tests, so the logs don’t easily get buried.

### Modifications:

Log the number of test cases for parameterized tests at the end of a test.

### Result:

If the user runs the parameterized tests, an extra piece of information (with * test cases) would get bound to the end log.

Resolves #943.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

Issue #943, opened 2025-02-05
Log the number of test cases run when a test function finishes


mesqueeb:
### Description

Given test logs can easily get burried, the final test logs printed are significant to understand how well a test ran.

TLDR;
- Currently: it's not clear from the final statement how many arguments were tested
- Desired: I suggest below some extra logs to better understand the total arguments tested:

### Desired
```
􀟈 Test aPileBringsHuggers(moveFn:) started.
􀟈 Passing 1 argument moveFn → (Function) to aPileBringsHuggers(moveFn:)
// ...
​􁁛 Argument in aPileBringsHuggers(moveFn:) passed after 2 seconds. 🆕
​􀟈 Passing 1 argument moveFn → (Function) to aPileBringsHuggers(moveFn:)
// ...
​􁁛 Argument in aPileBringsHuggers(moveFn:) passed after 2 seconds. 🆕
​􁁛 Test aPileBringsHuggers(moveFn:) with 2 arguments passed after 4.677 seconds. 🆕
```

Here above, I suggested the lines marked as 🆕
1. an extra log after each argument passes or fails, and
2. a total amount of arguments tested to the final log.

### Currently (example with 2 arguments tested)
```
􀟈 Test aPileBringsHuggers(moveFn:) started.
􀟈 Passing 1 argument moveFn → (Function) to aPileBringsHuggers(moveFn:)
// ...
​􀟈 Passing 1 argument moveFn → (Function) to aPileBringsHuggers(moveFn:)
// ...
​􁁛 Test aPileBringsHuggers(moveFn:) passed after 4.677 seconds.
```
Given that these logs easily get buried, I think something like the above would be way better.

### Steps to reproduce

Write a swift test with arguments and run it.

Relevant thread:
https://forums.swift.org/t/fr-test-with-arguments-and-closures-dx-suggestions-to-clarify-all-arguments-were-tested/77677/7

### swift-testing version/commit hash

not sure how to check

### Swift & OS version (output of `swift --version && uname -a`)

swift-driver version: 1.115.1 Apple Swift version 6.0.3 (swiftlang-6.0.3.1.10 clang-1600.0.30.1)
Target: arm64-apple-macosx15.0
Darwin Dendrocopus.local 24.2.0 Darwin Kernel Version 24.2.0: Fri Dec  6 19:02:41 PST 2024; root:xnu-11215.61.5~2/RELEASE_ARM64_T6030 arm64

grynspan:
[...]

PR #1000, opened 2025-03-06
Represent non-encodable test argument values in Test.Case.ID

### LINKED ISSUE #995: Represent non-encodable test argument values in Test.Case.ID
### Description

Currently `Test.Case.ID` only attempts to provide an identifier for arguments (or combinations of arguments, when there are multiple parameters) which are encodable, using one of the heuristics described in [Run selected test cases](https://swiftpackageindex.com/swiftlang/swift-testing/main/documentation/testing/parameterizedtesting#Run-selected-test-cases). It does not attempt to model or represent arguments which don't conform to one of those protocols, and this is limiting because it means there's no way to distinguish those arguments in reporting.

Even though arguments which don't conform to one of those protocols are not considered stable, and thus cannot be selectively re-run, it's nevertheless helpful to have a way to uniquely identify and refer to such test cases within the scope of a single run. This issue tracks expanding `Test.Case.ID` and its associated types to model this, including providing a way to distinguish stable vs. non-stable arguments.

This is a prerequisite to #671.

Tracked by rdar://119522099

### Additional information

_No response_

stmontgomery:
This expands `Test.Case.ID` to represent combinations of arguments which aren't fully encodable, and uniquely distinguishes test cases associated with the same parameterized test function which have otherwise identical IDs.

### Motivation:

It is possible to declare a parameterized test function with a collection of arguments that includes the same element more than once. As a trivial example:

```swift
@Test(arguments: [1, 1])
func repeatedArg(value: Int) { ... }
```

This can happen in more realistic scenarios if you dynamically construct a collection of arguments that accidentally or intentionally includes the same value more than once.

The testing library attempts to form a unique identifier for each argument passed to a parameterized test. It does so by checking whether the value conforms to `Encodable`, or one of the other protocols mentioned in the documentation (see [Run selected test cases](https://swiftpackageindex.com/swiftlang/swift-testing/main/documentation/testing/parameterizedtesting#Run-selected-test-cases)).

One problem with the current implementation is that if the value _doesn't_ conform to one of those known protocols, the testing library gives up and doesn't produce a unique identifier for that test case at all. Specifically, in that situation the `argumentIDs` property of `Test.Case.ID` will have a value of `nil`.

Another problem is that if an argument is passed more than once, the derived identifiers for each argument's test case will be the same and the result reporting will be ambiguous at best; or worse, the lack of a unique identifier could cause an integrated tool to misbehave or crash. (Current versions of Xcode 16 experience this issue non-deterministically for projects which have test parallelization enabled.) To solve this, there needs to be a way to deterministically distinguish test cases which, from the testing library's perspective, appear identical—either because they actually are the same value, or because they encode to the same representation.

### Modifications:

- Add a new `isStable` boolean property to `Test.Case.Argument.ID` representing whether or not the testing library was able to encode a stable representation of the argument value it identifies.
- Add a new SPI property named `discriminator` to `Test.Case` which distinguishes test cases associated with the same test function whose arguments are identical.
- Add a corresponding property to `Test.Case.ID` with the same name, `discriminator`.
- Change the `Test.Case.ID.argumentIDs` property to non-Optional, so that it always has a value even if one or more argument IDs is non-stable.
- Add a derived boolean property `isStable` to `Test.Case.ID` whose value is `true` iff all of its argument IDs are stable.
- Add `Hashable` conformance to `Test.Case`, since the reason for avoiding such conformance has been resolved.
- Add new tests.

### Checklist:

Resolves #995
Resolves rdar://119522099

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
@@ -124,9 +124,13 @@ extension ABI {
     var displayName: String
 
     init(encoding testCase: borrowing Test.Case) {
+      guard let arguments = testCase.arguments else {
+        preconditionFailure("Attempted to initialize an EncodedTestCase encoding a test case which is not parameterized: \(testCase)")
+      }
+
       // TODO: define an encodable form of Test.Case.ID
       id = String(describing: testCase.id)
-      displayName = testCase.arguments.lazy
+      displayName = arguments.lazy
         .map(\.value)
         .map(String.init(describingForTest:))
         .joined(separator: ", ")
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
@@ -125,7 +125,7 @@ extension ABI {
 
     init(encoding testCase: borrowing Test.Case) {
       guard let arguments = testCase.arguments else {
-        preconditionFailure("Attempted to initialize an EncodedTestCase encoding a test case which is not parameterized: \(testCase)")
+        preconditionFailure("Attempted to initialize an EncodedTestCase encoding a test case which is not parameterized: \(testCase). Please file a bug report at https://github.com/swiftlang/swift-testing/issues/new")
       }
 
       // TODO: define an encodable form of Test.Case.ID
--- a/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
@@ -171,8 +171,12 @@ extension Test.Case {
   /// - Parameters:
   ///   - includeTypeNames: Whether the qualified type name of each argument's
   ///     runtime type should be included. Defaults to `false`.
+  ///
+  /// - Returns: A string containing the arguments of this test case formatted
+  ///   for presentation, or an empty string if this test cases is
+  ///   non-parameterized.
   fileprivate func labeledArguments(includingQualifiedTypeNames includeTypeNames: Bool = false) -> String {
-    arguments.lazy
+    (arguments ?? []).lazy
       .map { argument in
         let valueDescription = String(describingForTest: argument.value)
 
@@ -494,14 +498,14 @@ extension Event.HumanReadableOutputRecorder {
       return result
 
     case .testCaseStarted:
-      guard let testCase = eventContext.testCase, testCase.isParameterized else {
+      guard let testCase = eventContext.testCase, testCase.isParameterized, let arguments = testCase.arguments else {
         break
       }
 
       return [
         Message(
           symbol: .default,
-          stringValue: "Passing \(testCase.arguments.count.counting("argument")) \(testCase.labeledArguments(includingQualifiedTypeNames: verbosity > 0)) to \(testName)"
+          stringValue: "Passing \(arguments.count.counting("argument")) \(testCase.labeledArguments(includingQualifiedTypeNames: verbosity > 0)) to \(testName)"
         )
       ]
 
--- a/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
@@ -176,7 +176,9 @@ extension Test.Case {
   ///   for presentation, or an empty string if this test cases is
   ///   non-parameterized.
   fileprivate func labeledArguments(includingQualifiedTypeNames includeTypeNames: Bool = false) -> String {
-    (arguments ?? []).lazy
+    guard let arguments else { return "" }
+
+    return arguments.lazy
       .map { argument in
         let valueDescription = String(describingForTest: argument.value)
 
--- a/Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
+++ b/Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
@@ -43,50 +43,76 @@ public protocol CustomTestArgumentEncodable: Sendable {
   func encodeTestArgument(to encoder: some Encoder) throws
 }
 
+/// Get the best encodable representation of a test argument value, if any.
+///
+/// - Parameters:
+///   - value: The value for which an encodable representation is requested.
+///
+/// - Returns: The best encodable representation of `value`, if one is available,
+///   otherwise `nil`.
+///
+/// For a description of the heuristics used to obtain an encodable
+/// representation of an argument value, see <doc:ParameterizedTesting>.
+func encodableArgumentValue(for value: some Sendable) -> (any Encodable)? {
+#if canImport(Foundation)
+  // Helper for opening an existential.
+  func customArgumentWrapper(for value: some CustomTestArgumentEncodable) -> some Encodable {
+    _CustomArgumentWrapper(rawValue: value)
+  }
+
+  return if let customEncodable = value as? any CustomTestArgumentEncodable {
+    customArgumentWrapper(for: customEncodable)
+  } else if let rawRepresentable = value as? any RawRepresentable, let encodableRawValue = rawRepresentable.rawValue as? any Encodable {
+    encodableRawValue
+  } else if let encodable = value as? any Encodable {
+    encodable
+  } else if let identifiable = value as? any Identifiable, let encodableID = identifiable.id as? any Encodable {
+    encodableID
+  } else {
+    nil
+  }
+#else
+  return nil
+#endif
+}
+
 extension Test.Case.Argument.ID {
   /// Initialize this instance with an ID for the specified test argument.
   ///
   /// - Parameters:
   ///   - value: The value of a test argument for which to get an ID.
+  ///   - encodableValue: An encodable representation of `value`, if any, with
+  ///     which to attempt to encode a stable representation.
   ///   - parameter: The parameter of the test function to which this argument
   ///     value was passed.
   ///
-  /// - Returns: `nil` if an ID cannot be formed from the specified test
-  ///   argument value.
-  ///
-  /// - Throws: Any error encountered while attempting to encode `value`.
+  /// If a representation of `value` can be successfully encoded, the value of
+  /// this instance's `bytes` property will be the the bytes of that encoded
+  /// JSON representation and the value of its `isStable` property will be
+  /// `true`. Otherwise, the value of its `bytes` property will be the bytes of
+  /// a textual description of `value` and the value of `isStable` will be
+  /// `false` to reflect that the representation is not considered stable.
   ///
   /// This function is not part of the public interface of the testing library.
   ///
   /// ## See Also
   ///
   /// - ``CustomTestArgumentEncodable``
-  init?(identifying value: some Sendable, parameter: Test.Parameter) throws {
+  init(identifying value: some Sendable, encodableValue: (any Encodable)?, parameter: Test.Parameter) {
 #if canImport(Foundation)
-    func customArgumentWrapper(for value: some CustomTestArgumentEncodable) -> some Encodable {
-      _CustomArgumentWrapper(rawValue: value)
+    if let encodableValue {
+      do {
+        self = .init(bytes: try Self._encode(encodableValue, parameter: parameter), isStable: true)
+        return
+      } catch {
+        // FIXME: Capture the error and propagate to the user, not as a test
+        // failure but as an advisory warning. A missing argument ID will
+        // prevent re-running the test case, but is not a blocking issue.
+      }
     }
-
-    let encodableValue: (any Encodable)? = if let customEncodable = value as? any CustomTestArgumentEncodable {
-      customArgumentWrapper(for: customEncodable)
-    } else if let rawRepresentable = value as? any RawRepresentable, let encodableRawValue = rawRepresentable.rawValue as? any Encodable {
-      encodableRawValue
-    } else if let encodable = value as? any Encodable {
-      encodable
-    } else if let identifiable = value as? any Identifiable, let encodableID = identifiable.id as? any Encodable {
-      encodableID
-    } else {
-      nil
-    }
-
-    guard let encodableValue else {
-      return nil
-    }
-
-    self = .init(bytes: try Self._encode(encodableValue, parameter: parameter))
-#else
-    nil
 #endif
+
+    self = .init(bytes: String(describingForTest: value).utf8, isStable: false)
   }
 
 #if canImport(Foundation)
--- a/Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
+++ b/Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
@@ -43,76 +43,57 @@ public protocol CustomTestArgumentEncodable: Sendable {
   func encodeTestArgument(to encoder: some Encoder) throws
 }
 
-/// Get the best encodable representation of a test argument value, if any.
-///
-/// - Parameters:
-///   - value: The value for which an encodable representation is requested.
-///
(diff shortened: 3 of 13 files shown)


stmontgomery:
@swift-ci please test

briancroom (approved):

stmontgomery:
@swift-ci please test

briancroom (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -124,9 +124,13 @@ extension ABI {
     var displayName: String
 
     init(encoding testCase: borrowing Test.Case) {
+      guard let arguments = testCase.arguments else {
+        preconditionFailure("Attempted to initialize an EncodedTestCase encoding a test case which is not parameterized: \(testCase)")
```
grynspan, line 128: Maybe add the "please file a bug" blurb here?

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -171,8 +171,12 @@ extension Test.Case {
   /// - Parameters:
   ///   - includeTypeNames: Whether the qualified type name of each argument's
   ///     runtime type should be included. Defaults to `false`.
+  ///
+  /// - Returns: A string containing the arguments of this test case formatted
+  ///   for presentation, or an empty string if this test cases is
+  ///   non-parameterized.
   fileprivate func labeledArguments(includingQualifiedTypeNames includeTypeNames: Bool = false) -> String {
-    arguments.lazy
+    (arguments ?? []).lazy
```
grynspan, line 179: Might be easier to read as something like `guard let arguments else { return ... }`

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
```diff
@@ -43,50 +43,76 @@ public protocol CustomTestArgumentEncodable: Sendable {
...
+  /// a textual description of `value` and the value of `isStable` will be
+  /// `false` to reflect that the representation is not considered stable.
   ///
   /// This function is not part of the public interface of the testing library.
   ///
   /// ## See Also
   ///
   /// - ``CustomTestArgumentEncodable``
-  init?(identifying value: some Sendable, parameter: Test.Parameter) throws {
+  init(identifying value: some Sendable, encodableValue: (any Encodable)?, parameter: Test.Parameter) {
```
grynspan, line 101:
```suggestion
  init(identifying value: some Sendable, encodableValue: (some Encodable)?, parameter: Test.Parameter) {
```

If the compiler allows it?

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
```diff
@@ -43,50 +43,76 @@ public protocol CustomTestArgumentEncodable: Sendable {
...
-    }
-
-    guard let encodableValue else {
-      return nil
+    if let encodableValue {
+      do {
+        self = .init(bytes: try Self._encode(encodableValue, parameter: parameter), isStable: true)
+        return
+      } catch {
+        // FIXME: Capture the error and propagate to the user, not as a test
```
grynspan, line 108: Are you planning to fix this soon? Do we need an issue filed?

Sources/Testing/Parameterization/Test.Case.Generator.swift
```diff
@@ -257,7 +257,30 @@ extension Test.Case {
 
 extension Test.Case.Generator: Sequence {
   func makeIterator() -> some IteratorProtocol<Test.Case> {
-    _sequence.lazy.map(_mapElement).makeIterator()
+    sequence(state: (
```
grynspan, line 260: I would recommend pulling the state structure out of the arguments list here so it's a bit easier to read. Something like:

```swift
let state = (
  ...
)
return sequence(state: state) { state in
  ...
}
```

Sources/Testing/Parameterization/Test.Case.swift
```diff
@@ -188,7 +331,7 @@ extension Test.Case {
     ///   - testCase: The original test case to snapshot.
     public init(snapshotting testCase: borrowing Test.Case) {
       id = testCase.id
-      arguments = testCase.arguments.map(Test.Case.Argument.Snapshot.init)
+      arguments = (testCase.arguments ?? []).map(Test.Case.Argument.Snapshot.init)
```
grynspan, line 334: Recommend using `if`/`else` here instead of `??` for readability.

Sources/Testing/Parameterization/Test.Case.swift
```diff
@@ -156,10 +277,32 @@ extension Test {
...
+    let isStable = try container.decodeIfPresent(type(of: isStable), forKey: .isStable) ?? true
+
+    let bytes = try container.decode(type(of: bytes), forKey: .bytes)
+    self.init(bytes: bytes, isStable: isStable)
+  }
+}
 
 // MARK: - Equatable, Hashable
 
+extension Test.Case: Hashable {
```
grynspan, line 296: Nitpick: explicitly state `Equatable`.

Is this `==` implementation expensive to call?

Sources/Testing/Parameterization/Test.Case.Generator.swift
```diff
@@ -257,7 +257,30 @@ extension Test.Case {
...
+      var testCase = _mapElement(element)
+
+      if testCase.isParameterized {
+        // Store the original, unmodified test case ID. We're about to modify a
+        // property which affects it, and we want to update state based on the
+        // original one.
+        let testCaseID = testCase.id
+
+        // Ensure test cases with identical IDs each have a unique discriminator.
+        let discriminator = state.testCaseIDs[testCaseID, default: 0]
```
grynspan, line 277: What's the benefit of maintaining a dictionary here? What if the discriminator were just the `enumerated()` index of the case? Yes, we'd end up skipping over some discriminators, but so what—they'll still be unique in this context, and it'll be much cheaper to compute and store them.

grynspan (changes requested):
Mostly minor feedback, but a couple of points we ought to clarify/address.

Sources/Testing/Parameterization/Test.Case.Generator.swift
stmontgomery: The goal is different than you may be thinking: I'm not trying to assign each test case the next-highest integer as the discriminator, which would be equivalent to the "enumeration index" of the collection; instead I'm trying to locate the test cases for each parameterized test function _which have identical arguments_, and for each such subset, assign each of those test cases a unique discriminator. Consider this example:

```swift
@Test(arguments: [9, 9, 8, 7]
func example(_: Int) {}
```

This will have 4 test cases, and with this PR's changes, they should be (in pseudo-code):

```
[
  Test.Case(arguments: [9], discriminator: 0),
  Test.Case(arguments: [9], discriminator: 1),
  Test.Case(arguments: [8], discriminator: 0), // back to 0 here!
  Test.Case(arguments: [7], discriminator: 0), // and here!
]
```

The end goal is to be able to detect duplicates of the same set of arguments. If we used `enumerated()` on the original collection that would be harder for clients to detect duplicates, and for the testing library itself to detect and report warnings for them (which I intend to do in a follow-on PR). I also worry that would tempt clients to abuse the `discriminator` value and interpret it as an index of the original collection, and that's explicitly not what I'm intending—particularly because the order may differ between successive runs (for example for hashed collections).

grynspan: I see—is there something we can do that's less heavyweight than a hash table?

Sources/Testing/Parameterization/Test.Case.swift
```diff
@@ -197,8 +340,8 @@ extension Test.Case.Argument {
   /// A serializable snapshot of a ``Test/Case/Argument`` instance.
   @_spi(ForToolsIntegrationOnly)
   public struct Snapshot: Sendable, Codable {
-    /// The ID of this parameterized test argument, if any.
-    public var id: Test.Case.Argument.ID?
+    /// The ID of this parameterized test argument.
+    public var id: Test.Case.Argument.ID
```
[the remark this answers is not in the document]
stmontgomery: I think I can actually achieve my goals without any modifications to `Test.Case.Argument.Snapshot`, so I'll revert both changes to this type.

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
stmontgomery: > error: Type 'any Encodable' cannot conform to 'Encodable'

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
stmontgomery: This is actually just relocating a comment from elsewhere in the codebase

Sources/Testing/Parameterization/Test.Case.swift
```diff
@@ -15,6 +15,29 @@ extension Test {
   /// Tests that are _not_ parameterized map to a single instance of
   /// ``Test/Case``.
   public struct Case: Sendable {
+    /// An enumeration describing the various kinds of test cases.
+    private enum _Kind: Sendable {
```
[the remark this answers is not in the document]
stmontgomery: Before I added this enum, I had added a bunch of `precondition`s to various places enforcing that the optionality of `arguments` and `discriminator` match each other — they should both be either `nil` or non-`nil`. On balance this felt cleaner since I could reduce the places which potentially alter these. And now, because the `arguments` and `discriminator` properties on `Test.Case` are derived from this, the enum isn't _re_-stating their values; it's now the source of truth.

Sources/Testing/Parameterization/Test.Case.swift
stmontgomery: > Is this == implementation expensive to call?

It primarily amounts to calling `==` on two `Array<UInt8>` operands. It does not trigger (re-)encoding each time, in case that's what you're wondering.

Sources/Testing/Parameterization/Test.Case.Generator.swift
stmontgomery: Not sure. The use of a dictionary doesn't seem that heavyweight to me in context. This isn't now causing values to be encoded where they wouldn't otherwise be, if that's the concern.

One small optimization I just realized I could make, though, is to specify a capacity for the dictionary based on the generator's `underestimatedCount`.

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test Linux

Sources/Testing/Parameterization/Test.Case.Generator.swift
grynspan: A concern I have is that this is going to scale at least linearly with the number of test cases in a test. As of right now, we _try_ to avoid keeping information about all test cases in memory during a run. If the inputs are an artisanally crafted array of values, then that's one thing. But if there's a dynamically generated sequence of inputs that's particularly long, we're now going to pay a memory cost to track _all_ of them.

Sources/Testing/Parameterization/Test.Case.swift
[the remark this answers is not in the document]
grynspan: Alrighty!

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
grynspan: Sigh.

Sources/Testing/Parameterization/Test.Case.swift
```diff
@@ -95,15 +192,39 @@ extension Test {
       parameters: [Parameter],
       body: @escaping @Sendable () async throws -> Void
     ) {
-      let arguments = zip(values, parameters).map { value, parameter in
-        Argument(value: value, parameter: parameter)
+      // Attempt to obtain an encodable representation of each value in order
+      // to construct a stable ID.
+      let encodingResult = values.reduce(into: ([any Encodable](), hasFailure: false)) { result, value in
```
grynspan, line 197: Perhaps something like this might be clearer or easier to read?

```swift
let encodables: [any Encodable] = values.compactMap { value in
  ...
}
let hasFailure = encodables.count != values.count
```

Sources/Testing/Parameterization/Test.Case.swift
```diff
@@ -156,10 +277,32 @@ extension Test {
...
+    let isStable = try container.decodeIfPresent(type(of: isStable), forKey: .isStable) ?? true
+
+    let bytes = try container.decode(type(of: bytes), forKey: .bytes)
+    self.init(bytes: bytes, isStable: isStable)
+  }
+}
 
 // MARK: - Equatable, Hashable
 
+extension Test.Case: Equatable, Hashable {
```
grynspan, line 296: This is publicly visible. Are you sure you want to commit to making instances of `Test.Case` equatable?

Sources/Testing/Parameterization/Test.Case.Generator.swift
stmontgomery: The dictionary only stores their IDs, not the test cases themselves, and I don't expect that to be onerous. I definitely share your view that we should not be holding on to test case instances themselves for the duration of the test run, or even for the lifetime of a `Test.Case.Generator`.

We already "accumulate" the IDs of `Test` instances in places like `Event.HumanReadableOutputRecorder` to keep track of events per-test. I imagine in the not-too-distant future we'll want to extend that pattern to test cases too, so that (for example) when running in verbose mode, our console output could log when test cases finish and how many failures each one had. That topic came up in https://github.com/swiftlang/swift-testing/issues/943, and was only partly solved by https://github.com/swiftlang/swift-testing/pull/972. One of my motivations for this PR is to allow us to properly implement that. I've been meaning to file a new issue to track that, so I just filed #1021.

Sources/Testing/Parameterization/Test.Case.swift
stmontgomery: Sure, that's not strictly needed here so I can remove it for now.

Sources/Testing/Parameterization/Test.Case.swift
stmontgomery: As I detailed in the comments, I want to bail early and stop attempting to get encodable versions of subsequent arguments if any earlier one failed. Getting an encodable value involves multiple conditional `as?` downcasts, and especially since you've mentioned some performance considerations here I wanted to be sure we're not doing unnecessary work. I know `reduce(into:)` can sometimes be less straightforward to read but I think my code comments explain everything adequately.

stmontgomery:
@swift-ci please test

Sources/Testing/Parameterization/Test.Case.swift
grynspan: Maybe we just do a plain old for-loop then?

grynspan (approved):

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test macOS

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

stmontgomery:
For anyone subscribed to this issue, an update:

Until recently there has been an Xcode bug associated with this issue which could cause a runtime crash in tests which have non-encodable or identical-looking arguments, but that has been resolved in Xcode 26 Beta! From the [release notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-26-release-notes):

> **Testing**
> - Fixed: If a parameterized Swift Testing test function includes two or more arguments that either don’t have a stable representation (e.g. they don’t conform to `Codable` or any other supported protocol) or have an identical stable representation, they will each now run even if parallelization is enabled. Previously this situation could crash, and most often affected arguments which had equal values for their `description` property. (121455205)

PR #1001, opened 2025-03-07
Fix link to Swift Forums

ZevEisenberg:
Fix broken link to Swift forums.

### Motivation:

The old link was a 404.

Old link: <https://forums.swift.org/c/related-projects/swift-testing>
New link: <https://forums.swift.org/c/development/swift-testing/103>

### Modifications:

I looked on the forum and tried to find the most canonical version of the link. When I remove the `/103`, it still loads with a 301 Moved Permanently. I'm not sure whether including the `/103` in the URL or not is more future-proof, but at least this way we prevent a redirect.

### Result:

The link works again.

### Checklist:


grynspan (approved):

grynspan:
@shahmishal Is there any way to get redirects working in the forums?

grynspan:
Docs change only, can skip CI.

stmontgomery (approved):
Thank you! You are correct: we _should_ include the trailing 103 digits in the URL, that is what allows Discourse forum URLs to properly redirect if a category is moved, and the problem is we didn't include those digits originally or the link wouldn't have broken. Lesson learned!

PR #1012, opened 2025-03-11
Rewrite static section bounds discovery in Swift.

grynspan:
This PR rewrites the static section bounds discovery code in Swift (replacing the current C++ implementation.) We are making use of `@_silgen_name` here to name the section bounds symbols that are defined by the linker, so we'll want to get the core team's approval before merging.

### Why do we need to rewrite this code in Swift?

In Embedded Swift, there is only one (statically linked) image in the process containing Swift code, and if the target is a microcontroller or similar, it probably doesn't even have a dynamic linker/loader in the first place for us to query. So we need, on Embedded Swift targets, to perform static discovery of exactly one test content section. (This is potentially something we can make configurable via the target's toolset in a future PR.)

The bounds of the test content section are currently defined in C++ using `__asm__` to reference linker-defined symbols. However, our C++ target can't reliably tell if it is being built for Embedded Swift. That knowledge is treated by the Swift toolchain as a Swift language feature and can be checked in Swift with `hasFeature(Embedded)`, but there's no C++ equivalent. If the implementation is written in Swift, we can know at compile time whether or not we're dynamically or statically linking to a test target (and therefore whether we need to do dynamic or static lookup of images, customized toolset aside.)

Rewriting this logic in Swift reduces our C++ code size (and our direct dependencies on the C++ STL) which is also a nice bonus.

### Checklist:

### DIFF
--- a/Sources/_TestDiscovery/SectionBounds.swift
+++ b/Sources/_TestDiscovery/SectionBounds.swift
@@ -308,8 +308,8 @@ private func _sectionBounds(_ kind: SectionBounds.Kind) -> EmptyCollection<Secti
 /// A type representing the upper or lower bound of a metadata section.
 ///
 /// This type uses the experimental `@_rawLayout` attribute to ensure that
-/// instances have fixed addresses. Use the `unsafeAddress` property to get the
-/// address of an instance of this type.
+/// instances have fixed addresses. Use the `..<` operator to get a buffer
+/// pointer from two instances of this type.
 ///
 /// On platforms that use static linkage and have well-defined bounds symbols,
 /// those symbols are imported into Swift below using `@_silgen_name`, another
--- a/Sources/_TestDiscovery/SectionBounds.swift
+++ b/Sources/_TestDiscovery/SectionBounds.swift
@@ -23,7 +23,7 @@ struct SectionBounds: Sendable {
 
   /// An enumeration describing the different sections discoverable by the
   /// testing library.
-  enum Kind: Int, Equatable, Hashable, CaseIterable {
+  enum Kind: Equatable, Hashable, CaseIterable {
     /// The test content metadata section.
     case testContent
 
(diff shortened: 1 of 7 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/_TestDiscovery/SectionBounds.swift
```diff
@@ -289,22 +291,68 @@ private func _sectionBounds(_ kind: SectionBounds.Kind) -> some Sequence<Section
...
+/// A type representing the upper or lower bound of a metadata section.
+///
+/// This type uses the experimental `@_rawLayout` attribute to ensure that
+/// instances have fixed addresses. Use the `..<` operator to get a buffer
+/// pointer from two instances of this type.
+///
+/// On platforms that use static linkage and have well-defined bounds symbols,
+/// those symbols are imported into Swift below using `@_silgen_name`, another
+/// experimental attribute.
+@_rawLayout(like: CChar) private struct _SectionBound: @unchecked Sendable, ~Copyable {
```
Azoy, line 322: I'd much rather not use this here and use `var` for the section begins and ends and label them `nonisolated(unsafe)` so that you can take a pointer to them.

grynspan: I had that in an earlier commit, but I've seen the compiler try to optimize away the stable address, which puts us in a bad place obviously.

Azoy: The optimizer cannot optimize away grabbing the pointer to an inout value.

Azoy: `withUnsafePointer(to: &_testContentSectionBegin)` will always be stable.

Azoy (approved):
Beautiful 👍

grynspan:
@swift-ci test

grynspan:
@swift-ci test

briancroom (approved):

stmontgomery (approved):

grynspan:
@swift-ci test

grynspan:
Waiting for a toolchain build to complete before merging.

jckarter (approved):
It seems unlikely that this would break.

grynspan:
@jckarter Famous last words.

PR #1014, opened 2025-03-11
Add withKnownIssue comments to known Issues

aroben:
Add withKnownIssue comments to known Issues

### Motivation:

The Comment passed to withKnownIssue() is currently only used when a known issue does not occur. When a known issue does occur, it is marked isKnown but does not contain any record of the withKnownIssue() comment, making it harder to understand why the issue was considered to be known, especially if there are multiple nested withKnownIssue() calls.

### Modifications:

When an issue is marked "known" by a `withKnownIssue()` call, the recorded issue will now have a new `knownIssueContext` property containing the comment passed to `withKnownIssue()` (if any). This comment will be included in the `messages` array of the `ABI.EncodedEvent` that represents the issue.

If the issue is recorded within multiple nested `withKnownIssue()` calls, `knownIssueContext` corresponds to the innermost matching call.

The `Issue.isKnown` setter is now deprecated and a no-op.

When an error is thrown within a `withKnownIssue()` call, the `Issue` passed to the issue matcher used to have the known issue comment in `Issue.comments`. That has now been removed; Issues for thrown errors now have no comments at all when passed to the issue matcher. This matches Issues for thrown errors that occur outside of `withKnownIssue()` calls.

### Checklist:


aroben:
@swift-ci test

aroben:
@swift-ci test

aroben:
@swift-ci test

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -9,13 +9,24 @@
...
 extension Issue {
-  /// The known issue matcher, as set by `withKnownIssue()`, associated with the
+  /// The known issue context, as set by `withKnownIssue()`, associated with the
   /// current task.
   ///
   /// If there is no call to `withKnownIssue()` executing on the current task,
   /// the value of this property is `nil`.
   @TaskLocal
-  static var currentKnownIssueMatcher: KnownIssueMatcher?
+  static var currentKnownIssueContext: KnownIssueContext?
```
stmontgomery, line 18: This task local could probably now be relocated to the new `KnownIssueContext` type and be renamed `current`.

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -9,13 +9,24 @@
...
+
+  /// Mark that an issue is a known issue.
+  /// - Parameter comment: The comment passed to the invocation of
+  ///   ``withKnownIssue`` that caused this issue to be considered "known".
+  mutating func markAsKnown(comment: Comment?) {
+    precondition(!isKnown)
+    isKnown = true
+    if let comment {
+      comments.insert(comment, at: 0)
```
stmontgomery, line 27: I think it may be useful and important for us to propagate the comment from `withKnownIssue` (if any) to the issue in a way that allows distinguishing it from the issue's _original_ comments (if it had any). If we don't, then clients who wish to display the issue might conflate the two categories of comments, which may be confusing to end users. We don't have a mechanism right now to distinguish the "original" comments from these, so I think it may be worthwhile to add a new field to `Issue` to support this, and to generally contain any other relevant details about the issue having been marked as "known". (If we had such a property, the current `isKnown` property could be derived from it.)

I'm not quite sure why this is inserting `comment` into the array at index 0, but if the reason is so that clients could find it and assume the first comment was the known issue comment, that feels fragile.

Sources/Testing/Issues/Issue+Recording.swift
aroben: Good idea, done.

Sources/Testing/Issues/Issue+Recording.swift
aroben: I agree it would be clearer to separate the known-issue data into a separate property. I'll look into doing that.

aroben:
@stmontgomery I took a crack at putting the known-issue data into a separate stored property and converting `isKnown` to a computed property. Some questions:

1. I left a setter in place for `isKnown` to try to preserve ABI compatibility. Does this seem important to you?
2. What do you think about the names `Issue.knownIssueContext` and `Issue.KnownIssueContext`?
3. I replaced `EncodedIssue.isKnown` with a new `EncodedIssue.knownIssueContext` and made `KnownIssueContext` conform to Codable. Does that seem like the right approach? Do I need to bump the ABI version or do anything else to handle compatibility?

aroben:
@swift-ci test

aroben:
@swift-ci test

aroben:
@stmontgomery ping :)

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -29,8 +29,10 @@ extension ABI {
     /// - Warning: Severity is not yet part of the JSON schema.
     var _severity: Severity
 
-    /// Whether or not this issue is known to occur.
-    var isKnown: Bool
+    /// An ``Issue/KnownIssueContext-swift.struct`` representing the
+    /// ``withKnownIssue(_:isIntermittent:sourceLocation:_:when:matching:)`` call
+    /// that matched this issue, if any.
+    var knownIssueContext: Issue.KnownIssueContext?
```
stmontgomery, line 35: The fields in these `ABI`-related structs are very intentional in their Codable serialization and need to avoid breaking serialization changes. See [ST-0002](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0002-json-abi.md) for some background. But that means I think we won't want to _remove_ the `isKnown` field — at least not without making a feature proposal and revising the published format, which probably isn't worthwhile. We can safely add information if it's useful (albeit with underscore prefixing for not-yet-public fields), but not remove things.

Sources/Testing/Issues/Issue.swift
```diff
@@ -115,9 +115,29 @@ public struct Issue: Sendable {
...
+  @_spi(ForToolsIntegrationOnly)
+  public var knownIssueContext: KnownIssueContext? = nil
+
   /// Whether or not this issue is known to occur.
   @_spi(ForToolsIntegrationOnly)
-  public var isKnown: Bool = false
+  public var isKnown: Bool {
+    get { knownIssueContext != nil }
+    set {}
+  }
```
stmontgomery, line 140: We may be able to get away with making this a `get`-only property since it's SPI.

Sources/Testing/Issues/Issue.swift
```diff
@@ -115,9 +115,29 @@ public struct Issue: Sendable {
   @_spi(ForToolsIntegrationOnly)
   public var sourceContext: SourceContext
 
+  /// A type representing a
+  /// ``withKnownIssue(_:isIntermittent:sourceLocation:_:when:matching:)`` call
+  /// that matched an issue.
+  @_spi(ForToolsIntegrationOnly)
+  public struct KnownIssueContext: Sendable, Codable {
```
stmontgomery, line 122: The word "issue" in the name of this type may be redundant now that it's nested within `Issue`. Does `Issue.KnownContext` sound strange? Will ponder it, but curious what you think

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,25 +8,53 @@
...
-///   current configuration set.
-private func _combineIssueMatcher(_ issueMatcher: @escaping KnownIssueMatcher, matchesCountedBy matchCounter: Locked<Int>) -> KnownIssueMatcher {
-  let oldIssueMatcher = Issue.currentKnownIssueMatcher
-  return { issue in
-    if issueMatcher(issue) || true == oldIssueMatcher?(issue) {
-      matchCounter.increment()
-      return true
+/// A type that represents an active `withKnownIssue()` call and any parent calls.
+///
+/// A stack of these is stored in `KnownIssueContext.current`. The stack is
```
stmontgomery, line 13:
```suggestion
/// A stack of these is stored in `Issue.KnownIssueContext.current`. The stack is
```

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,25 +8,53 @@
...
+/// mutated by calls to `withKnownIssue()`.
+struct KnownIssueScope: Sendable {
+  /// Determine if an issue is known to this scope or any of its ancestor
+  /// scopes.
+  ///
+  /// Returns `nil` if the issue is not known.
+  var match: @Sendable (Issue) -> Issue.KnownIssueContext?
```
stmontgomery, line 20: Maybe `matcher` (-er) instead?

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,25 +8,53 @@
...
+
+  /// Create a new ``KnownIssueScope`` by combining a new issue matcher with
+  /// any already-active scope.
+  ///
+  /// - Parameters:
+  ///   - parent: The context that should be checked next if `issueMatcher`
+  ///     fails to match an issue.
+  ///   - issueMatcher: A function to invoke when an issue occurs that is used
+  ///     to determine if the issue is known to occur.
+  ///   - comment: Any comment to be associated with issues matched by
```
stmontgomery, line 32: `- comment` param needs updated

Sources/Testing/Issues/KnownIssue.swift
aroben: Actually:
```suggestion
/// A stack of these is stored in `KnownIssueScope.current`. The stack is
```

Sources/Testing/Issues/KnownIssue.swift
aroben: I was thinking this name reads better at the call site:

```swift
if let context = scope.match(issue) {
```

Sources/Testing/Issues/Issue.swift
aroben: OK, I'll remove the setter.

Sources/Testing/Issues/Issue.swift
aroben: `Issue.KnownContext` does sound a little strange to me. I do agree the current name is a bit repetitive. I'd actually be tempted to un-nest this type and just call it `KnownIssue` to match `withKnownIssue`. I.e., `withKnownIssue` will apply a `KnownIssue` to any `Issue`s that are recorded by its closure.

aroben:
@swift-ci test

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -30,8 +30,17 @@ extension ABI {
...
     /// Whether or not this issue is known to occur.
+    ///
+    /// This should perhaps be deprecated in favor of `_knownIssueContext`.
     var isKnown: Bool
 
+    /// An ``Issue/KnownIssueContext-swift.struct`` representing the
+    /// ``withKnownIssue(_:isIntermittent:sourceLocation:_:when:matching:)`` call
+    /// that matched this issue, if any.
+    ///
+    /// - Warning: Severity is not yet part of the JSON schema.
```
stmontgomery, line 41: This comment about "Severity" looks copy/pasted from elsewhere but isn't relevant

Sources/Testing/Issues/Issue.swift
stmontgomery: Between those options I think I'd prefer to leave it nested with the name you have right now. We can consider it more deeply if/when we decide to elevate this to fully public API.

Sources/Testing/Issues/Issue.swift
```diff
@@ -115,9 +115,26 @@ public struct Issue: Sendable {
...
+    /// The comment that was passed to `withKnownIssue()`.
```
stmontgomery, line 123:
```suggestion
    /// The comment that was passed to `withKnownIssue()`, if any.
```

Sources/Testing/Issues/Issue.swift
```diff
@@ -115,9 +115,26 @@ public struct Issue: Sendable {
...
+    var comment: Comment?
+    /// The source location that was passed to `withKnownIssue()`.
```
stmontgomery, line 125:
```suggestion
    var comment: Comment?

    /// The source location that was passed to `withKnownIssue()`.
```

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -40,12 +68,12 @@ private func _combineIssueMatcher(_ issueMatcher: @escaping KnownIssueMatcher, m
 ///     function.
 ///   - sourceLocation: The source location to which the issue should be
 ///     attributed.
-private func _matchError(_ error: any Error, using issueMatcher: KnownIssueMatcher, comment: Comment?, sourceLocation: SourceLocation) throws {
+private func _matchError(_ error: any Error, using scope: KnownIssueScope, comment: Comment?, sourceLocation: SourceLocation) throws {
```
stmontgomery, line 71: The `- issueMatcher:` parameter in the documentation above can be updated

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -51,6 +60,7 @@ extension ABI {
       case .error: .error
       }
       isKnown = issue.isKnown
+      _knownIssueContext = issue.knownIssueContext
```
stmontgomery, line 63: It may be smarter to create a specific `ABI.EncodedKnownIssueContext` type for this, instead of making the actual `Issue.KnownIssueContext` type `Codable`. That's the pattern we have _mostly_ used for the data included in the ABI serialization types; there's a few exceptions but we're trying to move in that direction.

Having this property be underscore-prefixed is a good idea though regardless.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
aroben: Fixed.

Sources/Testing/Issues/Issue.swift
aroben: Fixed.

Sources/Testing/Issues/KnownIssue.swift
aroben: Fixed.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
aroben: Added `ABI.EncodedKnownIssueContext.

aroben:
@swift-ci please test

aroben:
@stmontgomery How is this looking now?

Sources/Testing/ABI/Encoded/ABI.EncodedKnownIssueContext.swift
stmontgomery, line 1: New files like this one need to be added to the CMake rules file

Sources/Testing/ABI/Encoded/ABI.EncodedKnownIssueContext.swift
```diff
@@ -0,0 +1,37 @@
...
+  ///
+  /// - Warning: Known issues are not yet part of the JSON schema.
+  struct EncodedKnownIssueContext<V>: Sendable where V: ABI.Version {
+    /// The comment that was passed to `withKnownIssue()`, if any.
+    var comment: Comment?
+
+    /// The source location that was passed to `withKnownIssue()`.
+    var sourceLocation: SourceLocation
+
+    init(encoding context: Issue.KnownIssueContext, in eventContext: borrowing Event.Context) {
```
stmontgomery, line 28: Looks like the `eventContext` parameter might be unused?

aroben: Correct, but I left the parameter here to match EncodedAttachment and EncodedError which also take this parameter but don't use it.

Looking around further, I see that there are various ABI.Encoded* types that *don't* take this parameter, so maybe this isn't as much of a convention as I thought. I can remove it here. Should I remove it from EncodedAttachment and EncodedError as well?

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,44 +8,71 @@
...
+      }
+      return matchedContext
     }
-    return false
   }
+
+  /// The known issue context, as set by `withKnownIssue()`, associated with the
+  /// current task.
+  ///
+  /// If there is no call to `withKnownIssue()` executing on the current task,
```
stmontgomery, line 54: Several places in this PR where you have the shorthand 
```
`withKnownIssue()`
```
in documentation comments, it looks strange to me because they're not real, double-backtick DocC symbol references. Can we audit those and for each one, either rephrase it to avoid mentioning the backticked API name, or else lean in and actually use a DocC symbol reference? I don't think each one needs to be a symbol reference; some (like this line) can just be rephrased more simply like:

> If the current task is not within a known issue context, the value of this property is `nil`.

Sources/Testing/Issues/KnownIssue.swift
stmontgomery: I'd find it nicer if we used a `typealias` for the function type, because then we could document the semantics of the matcher function itself and more clearly indicate the `- Return:` and other fields. E.g.

```swift
/// A function which ...
///
/// - Parameters:
///   - issue: ...
///
/// - Returns: ...
typealias Matcher = @Sendable (_ issue: Issue) -> Issue.KnownIssueContext?
```

Sources/Testing/ABI/Encoded/ABI.EncodedKnownIssueContext.swift
stmontgomery: Hmm, maybe @grynspan knows the reason behind having that parameter. These types aren't publicly exposed though so I'd lean towards removing this parameter and leaving the others unmodified since they're not relevant to this PR.

grynspan: It should not be necessary to create this type in order to support this change.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -30,8 +30,17 @@ extension ABI {
...
+    ///
+    /// This should perhaps be deprecated in favor of `_knownIssueContext`.
     var isKnown: Bool
 
+    /// An ``Issue/KnownIssueContext-swift.struct`` representing the
+    /// ``withKnownIssue(_:isIntermittent:sourceLocation:_:when:matching:)`` call
+    /// that matched this issue, if any.
+    ///
+    /// - Warning: Known issues are not yet part of the JSON schema.
+    var _knownIssueContext: EncodedKnownIssueContext<V>?
```
grynspan, line 42: This change does not appear necessary.

Sources/Testing/Issues/Issue.swift
```diff
@@ -115,9 +115,27 @@ public struct Issue: Sendable {
   @_spi(ForToolsIntegrationOnly)
   public var sourceContext: SourceContext
 
+  /// A type representing a
+  /// ``withKnownIssue(_:isIntermittent:sourceLocation:_:when:matching:)`` call
+  /// that matched an issue.
+  @_spi(ForToolsIntegrationOnly)
+  public struct KnownIssueContext: Sendable {
+    /// The comment that was passed to `withKnownIssue()`, if any.
+    var comment: Comment?
```
grynspan, line 124: I don't see why this comment can't be placed directly on the issue, which already has storage for comments.

grynspan (changes requested):
This is not the direction I'd go with this change.

Sources/Testing/ABI/Encoded/ABI.EncodedKnownIssueContext.swift
grynspan: The `eventContext` paramenter is inconsistently present because I was inconsistent in including it. These types don't currently need it and I don't recall having a change in the wings that _would_, but if I do I'd add it in a future PR. Doesn't need to be in any changes you make (but also this change doesn't appear necessary in the first place.)

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
grynspan: Even if we do want a `KnownIssueContext` type, this is not the best design for JSON output. The JSON stream's structure is meant to be clean/usable JSON, even if that doesn't exactly match the structure of Swift types in the library.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -30,8 +30,17 @@ extension ABI {
...
     /// Whether or not this issue is known to occur.
+    ///
+    /// This should perhaps be deprecated in favor of `_knownIssueContext`.
     var isKnown: Bool
 
+    /// An ``Issue/KnownIssueContext-swift.struct`` representing the
+    /// ``withKnownIssue(_:isIntermittent:sourceLocation:_:when:matching:)`` call
+    /// that matched this issue, if any.
+    ///
+    /// - Warning: Known issues are not yet part of the JSON schema.
```
grynspan, line 41: Well this is just untrue.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -30,8 +30,17 @@ extension ABI {
     var _severity: Severity
 
     /// Whether or not this issue is known to occur.
+    ///
+    /// This should perhaps be deprecated in favor of `_knownIssueContext`.
```
grynspan, line 34: No it should not. :)

Sources/Testing/Issues/Issue.swift
```diff
@@ -115,9 +115,27 @@ public struct Issue: Sendable {
...
+
+    /// The source location that was passed to `withKnownIssue()`.
+    var sourceLocation: SourceLocation
```
grynspan, line 127: It's not clear to me why or how this source location is valuable.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
aroben: Hm. I copied this comment from `var _severity` just above. I took it to mean that these properties are not part of the documented JSON *schema*, even though they are part of the JSON *output*. I'd be happy to remove this here (and from `_severity` if you'd like) if you think this is unhelpful.

Sources/Testing/ABI/Encoded/ABI.EncodedKnownIssueContext.swift
aroben: OK, I will remove this parameter if this type survives. I added the type at @stmontgomery's suggestion: https://github.com/swiftlang/swift-testing/pull/1014#discussion_r2008239304

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
aroben: Can you describe how you'd like to represent this info in the JSON?

Sources/Testing/Issues/Issue.swift
aroben: I was thinking that Xcode etc. could make it easy to navigate to the `withKnownIssue` call that matched this Issue. This seems helpful in the case of custom matcher functions, especially if you have multiple nested `withKnownIssue` calls. For example, when analyzing test results you might decide that a particular Issue should *not* be considered "known", and want to figure out which `withKnownIssue` call you need to adjust to address that.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
aroben: OK, I can remove this comment. I was thinking we wouldn't want redundant data to stick around forever, but you know the design here better than I do.

Sources/Testing/Issues/Issue.swift
aroben: > I don't think that's going to be very useful—Xcode/VSCode/SwiftPM won't know to display it with other comments, which leaves the original reported issue unresolved. (Do not assume Xcode/VSCode/SwiftPM will be updated to reflect changes we make in this repository.)

Do Xcode/VSCode/SwiftPM currently display all comments in the `Issue.comments` array, or just one of them?

If they do display all comments then I agree that putting the known-issue comment in the array is the fastest path to getting it displayed in those UIs. If not then those clients will need changes regardless of what approach we choose.

I'm a little confused by "don't assume they'll make changes"; does that mean there's no use in ever adding more information to swift-testing's JSON output because those implementations are already frozen? (Or, stepping back a bit, maybe they aren't even using the JSON output and would have to essentially reimplement these changes in some fashion?)

aroben (commented):
> This is not the direction I'd go with this change.

@grynspan For reference, I've pushed my initial version of this PR here: https://github.com/swiftlang/swift-testing/compare/main...aroben:swift-testing:known-issue-comments-v1

In that take, I just prepended the known-issue comment to `Issue.comments` and left it at that. As you saw, @stmontgomery encouraged me to pull it out into its own struct to make the difference between the Issue's comment and the known-issue comment explicit.

Maybe you two can sync up and agree on an approach here? I'm happy to do it either way.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
grynspan: As I wrote in some other comments, it's not even clear to me that we _would_ put it in the JSON.

Sources/Testing/Issues/Issue.swift
grynspan: > Do Xcode/VSCode/SwiftPM currently display all comments in the Issue.comments array, or just one of them?

There are a few different contexts where comments are presented, but currently yes the `comments` array is shown in full at least some of the time.

If we want to distinguish comments from `withKnownIssue()` from those on the original issue (before it was marked `isKnown = true`), I suspect we would benefit from changes to the `Comment` type itself followed by a JSON schema bump that tools could, in the future, if they so deign, represent in their UIs in some way. JSON schema changes that we anticipate promoting to supported/API/ABI should be discussed before being added. You, Stuart, and I can sync up about this privately.

> I'm a little confused by "don't assume they'll make changes"; does that mean there's no use in ever adding more information to swift-testing's JSON output because those implementations are already frozen?

It means that the shipping versions of these tools do not support the changes you've made, so the original issue would not be addressed. We should look for a resolution that works with the tools that exist today because we do not know if or when they might ever be updated to adopt these schema changes.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
grynspan: A{P,B}I is forever.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
grynspan: The concept of issue severity is not yet officially supported and is experimental only; it needs to be made a feature through the Swift Evolution process. Known issues are already API and _are_ part of the supported schema (via the `isKnown` field). The specific changes you've made, on the other hand, are what's not yet supported.

Sources/Testing/Issues/Issue.swift
grynspan: It is worth considering whether we should instead add a `sourceLocation` property to the `Comment` type instead. That's beyond the scope of this particular PR, I think, but it would then let us provide source location info for _all_ comments, not just this one, and in a way that's (ideally) more consistent.

Sources/Testing/Issues/Issue.swift
aroben: I wasn't assuming that this change on its own would magically make these comments show up in the various UI clients. But it sounds like you're saying that might actually be possible, and if so we should prefer an approach that works with the shipping versions of those tools.

Sources/Testing/Issues/Issue.swift
stmontgomery: It's not the source location of the _comment_ that is interesting to capture, it's the source location of the `withKnownIssue()` call. And there may not always be a `Comment` passed to `withKnownIssue()`; it may be `nil`, so this `sourceLocation` property allows us to always capture the location of the `withKnownIssue()` call which caused an issue to be considered "known".

aroben: Yeah, I can see the utility in that. Although I wonder if you'd ever want both: the place that constructed the `Comment` vs. the place that applied it to an `Issue`? I'm not certain that would be useful, but the two concepts do seem distinct to me, and a `sourceLocation` on `Comment` itself kind of implies the location of construction to me.

stmontgomery: Agree they're distinct. Personally I just find the location of a comment _less_ useful in comparison, and agreed, it's out of scope for this PR

Sources/Testing/Issues/Issue.swift
```diff
@@ -115,9 +115,27 @@ public struct Issue: Sendable {
...
+  /// A ``KnownIssueContext-swift.struct`` representing the
+  /// ``withKnownIssue(_:isIntermittent:sourceLocation:_:when:matching:)`` call
+  /// that matched this issue, if any.
+  @_spi(ForToolsIntegrationOnly)
+  public var knownIssueContext: KnownIssueContext? = nil
+
   /// Whether or not this issue is known to occur.
   @_spi(ForToolsIntegrationOnly)
-  public var isKnown: Bool = false
+  public var isKnown: Bool { knownIssueContext != nil }
```
grynspan, line 138: Please add a setter:

```swift
@available(*, deprecated, message: "Setting this property has no effect.")
set {}
```

Sources/Testing/ABI/Encoded/ABI.EncodedKnownIssueContext.swift
aroben: I removed this type.

Sources/Testing/Issues/KnownIssue.swift
aroben: Fixed.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
aroben: I removed this property.

aroben:
After discussion with @grynspan and @stmontgomery I've reverted the ABI changes. Instead, we append the known issue comment (if any) to `ABI.EncodedEvent.messages` for the `.issueRecorded` event.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
aroben: I removed this property.

Sources/Testing/ABI/Encoded/ABI.EncodedKnownIssueContext.swift
aroben: I've removed this new file.

Sources/Testing/Issues/KnownIssue.swift
aroben: Renamed to `matcher` and added the typealias.

aroben:
@swift-ci please test

aroben:
I wanted to see if the known issue comment now appears in `swift test`'s output, so I added a new suite to this repo:

```swift
struct TryIt {
  @Test func knownIssueTest() async throws {
    withKnownIssue("known issue comment") {
      Issue.record("issue comment")
    }
  }

  @Test func knownIssueTest_throws() async throws {
    withKnownIssue("known issue comment") {
      struct TheError: Error {}
      throw TheError()
    }
  }
}
```

Here's what I got:

```shellsession
$ SWT_SF_SYMBOLS_ENABLED=no xcrun swift test --filter TryIt      
Building for debugging...
[1/1] Write swift-version--1EBDF3DDE0C0C582.txt
Build complete! (0.16s)
Test Suite 'Selected tests' started at 2025-04-07 10:48:10.677.
Test Suite 'swift-testingPackageTests.xctest' started at 2025-04-07 10:48:10.678.
Test Suite 'swift-testingPackageTests.xctest' passed at 2025-04-07 10:48:10.678.
	 Executed 0 tests, with 0 failures (0 unexpected) in 0.000 (0.000) seconds
Test Suite 'Selected tests' passed at 2025-04-07 10:48:10.678.
	 Executed 0 tests, with 0 failures (0 unexpected) in 0.000 (0.001) seconds
◇ Test run started.
↳ Testing Library Version: feb96cf762e5732f8702e1dc9755fa2908bfb5ba (modified)
◇ Suite TryIt started.
◇ Test knownIssueTest_throws() started.
◇ Test knownIssueTest() started.
✘ Test knownIssueTest() recorded a known issue at EventRecorderTests.swift:660:19: Issue recorded
↳ issue comment
↳ known issue comment
✘ Test knownIssueTest() passed after 0.001 seconds with 1 known issue.
✘ Test knownIssueTest_throws() recorded a known issue at EventRecorderTests.swift:665:19: Caught error: TheError()
↳ known issue comment
↳ known issue comment
✘ Test knownIssueTest_throws() passed after 0.001 seconds with 1 known issue.
✘ Suite TryIt passed after 0.001 seconds with 2 known issues.
✘ Test run with 2 tests passed after 0.001 seconds with 2 known issues.
```

This looks decent, but you can see that `known issue comment` is duplicated for `knownIssueTest_throws()`. This is because `_matchError` puts the known issue comment directly in the `Issue.comments` array, but then we *also* include it in `Issue.knownIssueContext`, so we end up with two copies of the comment.

Here are some ways to handle this:
1. Leave it as-is
2. Remove the known issue comment from `Issue.comments`, leaving it only in `Issue.knownIssueContext`. 
    a. This seems right from a modeling perspective, but will in practice remove the known issue comment from Xcode's inline error annotation UI
3. Omit the known issue comment from `Issue.knownIssueContext` for thrown errors
    a. This feels bad, though I'm not sure it has any practical downsides.
4. Teach HumanReadableOutputRecorder to somehow detect this case and avoid the duplication
    a. We should be able to detect this reliably: `if case let .errorCaught = issue.kind, issue.isKnown`

I'm leaning toward 4 (and would add a test for it of course).

aroben:
> I'm leaning toward 4 (and would add a test for it of course).

I implemented this.

aroben:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -454,7 +454,11 @@ extension Event.HumanReadableOutputRecorder {
       }
       additionalMessages += _formattedComments(issue.comments)
       if let knownIssueComment = issue.knownIssueContext?.comment {
-        additionalMessages.append(_formattedComment(knownIssueComment))
+        if case .errorCaught = issue.kind {
```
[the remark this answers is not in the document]
aroben: Pushed an updated comment, how does it look to you?

aroben:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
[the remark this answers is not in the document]
grynspan: Seems like the correct fix might be to _remove_ the comment from `_matchError()` so that nothing needs to be special-cased?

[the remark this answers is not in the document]
grynspan: Note for compatibility, we would want to clear the `comments` property of `Issue` in `_matchError()` after calling the matcher function as there could be code out there that inspects it:

```swift
withKnownIssue {
  ...
} matching: { issue in
  issue.comments.first?.contains("abc") ?? false
}
```

(Something like that.)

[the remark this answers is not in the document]
aroben: > Seems like the correct fix might be to _remove_ the comment from `_matchError()` so that nothing needs to be special-cased?

We certainly could, but that has a downside that I mentioned in an earlier comment:

> 2. Remove the known issue comment from `Issue.comments`, leaving it only in `Issue.knownIssueContext`.
>    a. This seems right from a modeling perspective, but will in practice remove the known issue comment from Xcode's inline error annotation UI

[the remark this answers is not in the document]
aroben: (It also seems pretty weird that the `matcher` would get to see the comment in `Issue.comments` but then it would disappear from that array.)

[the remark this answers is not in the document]
grynspan: > but will in practice remove the known issue comment from Xcode's inline error annotation UI

This PR is intended to fix that problem in the general case, no?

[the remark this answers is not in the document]
aroben: Yes, but I'm trying not to introduce any regressions along the way.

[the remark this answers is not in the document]
grynspan: I appreciate that :) And that's a good goal—but we also need to be mindful of creating technical debt for ourselves.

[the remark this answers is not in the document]
aroben: OK, so do you think I should take approach 2 despite the regression in Xcode and the weirdness of `Issue.comments` changing after the `matcher` runs?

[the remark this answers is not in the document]
aroben: We discussed this and decided to remove the comment entirely from `_matchError`. This has a slight risk of causing incompatibilities because the `matcher` function will no longer see the comment there, but it's actually kind of weird that the matcher sees the comment anyway since it is always the same as the comment passed to `withKnownIssue`:

```swift
@Test func knownIssueCommentPassedToIssueMatcher() throws {
    struct E: Error {}
    try withKnownIssue("Known Issue Comment") {
        throw E()
    } matching: { issue in
        print("Comments passed to matcher:", issue.comments)
        // prints: Comments passed to matcher: ["Known Issue Comment"]
        // which seems unhelpful since we can see the comment a few lines up in the withKnownIssue call
        return true
    }
}```

File: Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift [hunk 23]
```diff
@@ -93,10 +93,20 @@ extension Event.HumanReadableOutputRecorder {
   /// - Parameters:
   ///   - comments: The comments that should be formatted.
   ///
-  /// - Returns: A formatted string representing `comments`, or `nil` if there
-  ///   are none.
+  /// - Returns: An array of formatted messages representing `comments`, or an
+  /// empty array if there are none.
```
stmontgomery, line 97:
```suggestion
  ///   empty array if there are none.
```

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,44 +8,83 @@
...
+struct KnownIssueScope: Sendable {
+  /// A function which determines if an issue matches a known issue scope or
+  /// any of its ancestor scopes.
+  ///
+  /// - Parameters:
+  ///   - issue: The issue being matched.
+  /// - Returns: A known issue context containing information about the known
+  /// issue, if the issue is considered "known" by this known issue scope or any
+  /// ancestor scope, or `nil` otherwise.
+  typealias Matcher = @Sendable (Issue) -> Issue.KnownIssueContext?
```
stmontgomery, line 25: Explicitly declare `issue` so that the `- Parameters:` entry refers to it:

```suggestion
  typealias Matcher = @Sendable (_ issue: Issue) -> Issue.KnownIssueContext?
```

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,44 +8,83 @@
...
+/// A stack of these is stored in `KnownIssueScope.current`.
+struct KnownIssueScope: Sendable {
+  /// A function which determines if an issue matches a known issue scope or
+  /// any of its ancestor scopes.
+  ///
+  /// - Parameters:
+  ///   - issue: The issue being matched.
+  /// - Returns: A known issue context containing information about the known
+  /// issue, if the issue is considered "known" by this known issue scope or any
+  /// ancestor scope, or `nil` otherwise.
```
stmontgomery, line 24:
```suggestion
  ///
  /// - Returns: A known issue context containing information about the known
  ///   issue, if the issue is considered "known" by this known issue scope or any
  ///   ancestor scope, or `nil` otherwise.
```

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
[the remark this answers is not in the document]
stmontgomery: I agree we should stop including the comment passed to `withKnownIssue` (if any) in the `comments` of the `Issue` constructed in `_matchError()`.

stmontgomery (commented):
A few nitpicky formatting comments I saw as I caught up with the latest changes

aroben:
@swift-ci please test

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,44 +8,84 @@
...
+  /// - Returns: A known issue context containing information about the known
+  ///   issue, if the issue is considered "known" by this known issue scope or any
+  ///   ancestor scope, or `nil` otherwise.
+  typealias Matcher = @Sendable (_ issue: Issue) -> Issue.KnownIssueContext?
+
+  /// Determine if an issue is known to this scope or any of its ancestor
+  /// scopes.
+  ///
+  /// Returns `nil` if the issue is not known.
+  var matcher: @Sendable (Issue) -> Issue.KnownIssueContext?
```
stmontgomery, line 32:
```suggestion
  /// The matcher function for this known issue scope.
  var matcher: Matcher
```

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,44 +8,84 @@
...
+  ///
+  /// - Parameters:
+  ///   - parent: The context that should be checked next if `issueMatcher`
+  ///     fails to match an issue.
+  ///   - issueMatcher: A function to invoke when an issue occurs that is used
+  ///     to determine if the issue is known to occur.
+  ///   - context: The context to be associated with issues matched by
+  ///     `issueMatcher`.
+  /// - Returns: A new instance of ``KnownIssueScope``.
+  init(parent: KnownIssueScope?, issueMatcher: @escaping KnownIssueMatcher, context: Issue.KnownIssueContext) {
```
stmontgomery, line 48: Could `parent:` have a default value of `= .current`? Then you wouldn't need to pass it in the places below where you initialize a `KnownIssueScope`

(If you do this, please update the DocC above to mention its default value)

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -8,44 +8,84 @@
...
 ///   - error: The error to test.
-///   - issueMatcher: A function to which `error` is passed (after boxing it in
-///     an instance of ``Issue``) to determine if it is known to occur.
+///   - scope: The known issue scope that is processing the error.
 ///   - comment: An optional comment to apply to any issues generated by this
 ///     function.
 ///   - sourceLocation: The source location to which the issue should be
 ///     attributed.
-private func _matchError(_ error: any Error, using issueMatcher: KnownIssueMatcher, comment: Comment?, sourceLocation: SourceLocation) throws {
+private func _matchError(_ error: any Error, using scope: KnownIssueScope, comment: Comment?, sourceLocation: SourceLocation) throws {
```
stmontgomery, line 83: Nit: it may be more fluent to rename the label `using:` to `in:` now. "Match error **_in_ scope**" reads better to me, at least

Tests/TestingTests/KnownIssueTests.swift
stmontgomery, line 1: Great tests!

stmontgomery (commented):
Just a few more suggestions and style cleanups, but I think the behavior and logic all make sense now!

Sources/Testing/Issues/KnownIssue.swift
aroben: Good idea!

aroben:
@swift-ci please test

stmontgomery (approved):

stmontgomery:
The macOS and Linux CI failures are expected to be resolved by https://github.com/swiftlang/swift/pull/80830, but the release of a new `main` snapshot toolchain is currently blocked for unrelated reasons.

PR #1010, opened 2025-03-10
Refine type-based test discovery mechanism to use test content records.

grynspan:
This PR changes how we discover tests in the type metadata section to more closely align with how the test content section works. This change will allow for a smoother transition to the test content section and our use of `@section` by using the same underlying structures (as much as is feasible.) Client code that uses the new `_TestDiscovery` module will need fewer changes to adapt.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1015, opened 2025-03-12
Fix crash when loading test content records from the type metadata section.

grynspan:
Looks like #880 and/or #1010 caused a regression: the compiler appears to be dead-code-stripping the classes we emit to contain test content records. This PR changes the design back to using a protocol so that the members we need are always covered by `@_alwaysEmitConformanceMetadata`. This makes `_TestDiscovery` a little harder to use with legacy lookup, but it's all experimental and eventually going to be removed anyway.

Resolves rdar://146809312.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Discovery+Legacy.swift
+++ b/Sources/Testing/Test+Discovery+Legacy.swift
@@ -11,35 +11,37 @@
 #if !SWT_NO_LEGACY_TEST_DISCOVERY
 @_spi(Experimental) @_spi(ForToolsIntegrationOnly) internal import _TestDiscovery
 
-/// A shadow declaration of `_TestDiscovery.TestContentRecordContainer` that
-/// allows us to add public conformances to it without causing the
-/// `_TestDiscovery` module to appear in `Testing.private.swiftinterface`.
-///
-/// This protocol is not part of the public interface of the testing library.
-@_alwaysEmitConformanceMetadata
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci test

grynspan:
Windows CI is down right now. Self-reviewing due to potential impact on the toolchain.

Sources/Testing/Test+Discovery+Legacy.swift
```diff
@@ -11,35 +11,37 @@
...
-/// A shadow declaration of `_TestDiscovery.TestContentRecordContainer` that
-/// allows us to add public conformances to it without causing the
-/// `_TestDiscovery` module to appear in `Testing.private.swiftinterface`.
-///
-/// This protocol is not part of the public interface of the testing library.
-@_alwaysEmitConformanceMetadata
-protocol TestContentRecordContainer: _TestDiscovery.TestContentRecordContainer {}
-
-/// An abstract base class describing a type that contains tests.
+/// A protocol base class describing a type that contains tests.
```
grynspan, line 14: Typo here fixed on main already.

Sources/Testing/Test+Discovery+Legacy.swift
```diff
@@ -11,35 +11,37 @@
...
 ///
 /// - Warning: This class is used to implement the `@Test` macro. Do not use it
```
grynspan, line 16: Typo here fixed on main already.

PR #1019, opened 2025-03-12
Add a dedicated `TestContentKind` type to `_TestDiscovery`.

grynspan:
This PR adds a `TestContentKind` type to `_TestDiscovery` that represents the `kind` field of a record and allows initialization from a string literal (i.e. a FourCC) or an integer literal:

```swift
extension MyType: DiscoverableAsTestContent {
  static var testContentKind: TestContentKind {
    "moof"
  }
}
```

At this time, it's not possible to use this type directly in test content records emitted by our macros because the type cannot be made public, but it allows for types conforming to `DiscoverableAsTestContent` to be a bit more expressive. Test library authors who are experimenting with `_TestDiscovery` may opt to export the symbols from `_TestDiscovery` and thus could use the type directly in their emitted records (but keep in mind this module is still experimental!)

The new type is `@frozen` and mostly `@inlinable` because its layout is set in stone by virtue of the test content record `kind` field having a fixed layout.

Resolves rdar://146855125.

### Checklist:


grynspan:
@swift-ci test

briancroom:
I like it. Thanks!

briancroom (approved):

Tests/TestingTests/DiscoveryTests.swift
stmontgomery, line 1: Glad to see these being moved into a dedicated test file

PR #1023, opened 2025-03-13
Continue encoding value for deprecated property in runner plan snapshot to avoid decoding error in clients

stmontgomery:
This fixes a regression in the encoding of the `Runner.Plan` snapshot types, which can manifest when using Xcode 16.

### Motivation:

The `isParallelizationEnabled` property was deprecated and changed from a stored property to a derived one in #901. That caused the value to no longer be encoded in `Runner.Plan.Action.RunOptions`, which can cause decoding errors in versions of Xcode which expect it to still be present.

### Modifications:

- Manually implement `Codable` conformance for the affected type to begin including a hardcoded value.

### Checklist:

Fixes rdar://146284519

### DIFF
--- a/Sources/Testing/Running/Runner.Plan.swift
+++ b/Sources/Testing/Running/Runner.Plan.swift
@@ -15,7 +15,7 @@ extension Runner {
     public enum Action: Sendable {
       /// A type describing options to apply to actions of case
       /// ``Runner/Plan/Action/run(options:)`` when they are run.
-      public struct RunOptions: Sendable, Codable {
+      public struct RunOptions: Sendable {
         /// Whether or not this step should be run in parallel with other tests.
         ///
         /// By default, all steps in a runner plan are run in parallel if the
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

briancroom (approved):

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -347,6 +347,29 @@ extension Runner.Plan {
   }
 }
 
+extension Runner.Plan.Action.RunOptions: Codable {
+  private enum CodingKeys: CodingKey {
+    case isParallelizationEnabled
+  }
+
+  public init(from decoder: any Decoder) throws {
```
grynspan, line 355: I _think_ the compiler can synthesize the `init` and `encode` members so long as the keys are defined. Compare the keys in `__CommandLineArguments_v0`. (I think you need to explicitly state a raw type of `String` too.)

grynspan: > 🛑 CodingKey case 'isParallelizationEnabled' does not match any stored properties

Dang it. Never mind.

grynspan (approved):

PR #1024, opened 2025-03-13
[Experimental] Implement JSON coding without using Foundation or `Codable`.

grynspan:
This PR replaces most of our uses of `JSONEncoder` with a home-grown implementation. Types conform to the internal `JSON.Serializable` protocol and emit strings, numbers, dictionaries, arrays, etc. Unlike with `Codable`, there is no synthesized implementation here.

Pros:

- Less reliance on Foundation (ideally we drop the dependency entirely at some point);
- No need to use existentials to encode a value (allowing us to eventually support Embedded Swift); and
- The potential for more inlining as all the encoding work is done in-module.

Cons:

- Don't Repeat Yourself;
- This implementation doesn't come with a decoder (which is harder to write);
- The implementation of `JSON.Serializable` cannot be supplied by the compiler;
- More code means more technical debt and a higher maintenance burden;
- The implementation is less configurable/customizable than Foundation's; and
- The implementation is not a full `Encoder` conformance, so it cannot be used to reimplement `CustomTestArgumentEncodable`.

Resolves rdar://146964016.

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_Foundation/Attachments/Attachable+Encodable+NSSecureCoding.swift
+++ b/Sources/Overlays/_Testing_Foundation/Attachments/Attachable+Encodable+NSSecureCoding.swift
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-#if canImport(Foundation)
+#if !SWT_NO_FOUNDATION && canImport(Foundation)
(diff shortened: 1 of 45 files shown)


grynspan:
I'm going to redo this as an `enum` now that I have a better idea of the constraints we need to solve for.

grynspan:
@swift-ci test

PR #1030, opened 2025-03-19
build: account for spaces in the toolchain

compnerd:
If the macro library path contains a space, the invocation will improperly truncate it. Account for this eventuality.


compnerd:
@swift-ci please test

grynspan (approved):
What if it contains quote marks? 🧌

PR #1032, opened 2025-03-20
Change `Attachment.attach()` to `Attachment.record()` and `Attachable.withUnsafeBufferPointer()` to `Attachable.withUnsafeBytes()`.

grynspan:
This PR updates the `attach()` function to make it static on `Attachment` and renames it `record()`, per the discussion [here](https://forums.swift.org/t/pitch-attachments/78072).

It also renames `Attachable.withUnsafeBufferPointer()` to `Attachable.withUnsafeBytes()` to more closely match the convention used by the standard library for raw buffers.

Finally, it removes the conditional dependence on the experimental language feature "SuppressedAssociatedTypes" in `AttachableContainer`.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -197,7 +197,6 @@ extension Array where Element == PackageDescription.SwiftSetting {
 
     result += [
       .enableUpcomingFeature("ExistentialAny"),
-      .enableExperimentalFeature("SuppressedAssociatedTypes"),
 
       .enableExperimentalFeature("AccessLevelOnImport"),
       .enableUpcomingFeature("InternalImportsByDefault"),
--- a/Sources/Testing/Attachments/AttachableContainer.swift
+++ b/Sources/Testing/Attachments/AttachableContainer.swift
@@ -12,23 +12,19 @@
 /// written to disk when a test is run and which contains another value that it
 /// stands in for.
 ///
-/// To attach an attachable value to a test report or test run output, use it to
-/// initialize a new instance of ``Attachment``, then call
-/// ``Attachment/attach(sourceLocation:)``. An attachment can only be attached
-/// once.
+/// To attach an attachable value to a test, pass it to ``Attachment/record(_:named:sourceLocation:)``.
+/// To further configure an attachable value before you attach it, use it to
+/// initialize an instance of ``Attachment`` and set its properties before
+/// passing it to ``Attachment/record(_:sourceLocation:)``. An attachable
+/// value can only be attached to a test once.
 ///
 /// A type can conform to this protocol if it represents another type that
 /// cannot directly conform to ``Attachable``, such as a non-final class or a
 /// type declared in a third-party module.
 @_spi(Experimental)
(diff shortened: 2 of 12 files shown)


grynspan:
@swift-ci test

Package.swift
```diff
@@ -197,7 +197,6 @@ extension Array where Element == PackageDescription.SwiftSetting {
 
     result += [
       .enableUpcomingFeature("ExistentialAny"),
-      .enableExperimentalFeature("SuppressedAssociatedTypes"),
```
younata, line 200: What was the SuppressedAssociatedTypes doing here anyway?

younata (approved):
Thanks for keeping this up to date!

Sources/Testing/Attachments/AttachableContainer.swift
```diff
@@ -12,23 +12,19 @@
...
+/// value can only be attached to a test once.
 ///
 /// A type can conform to this protocol if it represents another type that
 /// cannot directly conform to ``Attachable``, such as a non-final class or a
 /// type declared in a third-party module.
 @_spi(Experimental)
 public protocol AttachableContainer<AttachableValue>: Attachable, ~Copyable {
-#if hasFeature(SuppressedAssociatedTypes)
-  /// The type of the attachable value represented by this type.
-  associatedtype AttachableValue: ~Copyable
```
stmontgomery, line 27: Why don't we want or need this anymore?

Sources/Testing/Attachments/Attachment.swift
```diff
@@ -220,25 +220,61 @@ extension Attachment where AttachableValue: AttachableContainer & ~Copyable {
...
+  /// the testing library encodes it as data immediately. If the value cannot be
+  /// encoded and an error is thrown, that error is recorded as an issue in the
+  /// current test and the attachment is not written to the test report or to
+  /// disk.
+  ///
   /// An attachment can only be attached once.
   @_documentation(visibility: private)
-  public consuming func attach(sourceLocation: SourceLocation = #_sourceLocation) {
-    var attachmentCopy = Attachment<AnyAttachable>(self)
+  public static func record(_ attachment: consuming Self, sourceLocation: SourceLocation = #_sourceLocation) {
```
stmontgomery, line 239: In the associated pitch thread I suggested having both static and instance versions of this method (originally named "attach" but same idea). I'm curious what the thinking is with eliminating the instance methods and only offering static

Package.swift
grynspan: The idea was that you could define an `AttachableContainer` that wrapped a move-only type, but since it's still an experimental feature, I removed it to not bake it into the API before the feature is ready.

Sources/Testing/Attachments/Attachment.swift
grynspan: Overload overload—there are then six `record()` functions, three of which are overloads for `~Copyable` types. It's worth continued discussion though.

stmontgomery (approved):

PR #1035, opened 2025-03-21
Ensure `Locked+Platform.swift` is not stripped when statically linking.

grynspan:
All symbols in Locked+Platform.swift are referenced _indirectly_ by the Swift compiler due to `LockedWith` being generic, which can cause symbols from that file (including protocol conformance metadata) to be stripped at link time as if it were unused. This change (hopefully) works around that issue.

Works around rdar://147250336.

### Checklist:


grynspan:
@swift-ci test

mikeash (approved):

PR #1036, opened 2025-03-24
Remove unused "override comment" from ConditionTrait implementation

stmontgomery:
This removes a mechanism in the implementation of `ConditionTrait` for a condition closure to provide an "override" comment.

### Motivation:

While reviewing a draft evolution proposal for `ConditionTrait` (https://github.com/swiftlang/swift-evolution/pull/2740) I realized that we don't use this "comment override" mechanism anywhere in the testing library. I believe we did back when it was first added, but it's no longer used.

Removing this would allow the public API being proposed above to be simplified: the `evaluate()` method could return `Bool` instead of a tuple.

### Checklist:


stmontgomery:
@swift-ci please test

grynspan (approved):

PR #1044, opened 2025-03-30
Erase environment variables set by exit tests after reading them.

grynspan:
The environment variables we currently use to pass information from the parent process to the child process are implementation details of exit tests and may change or be removed in a future update. To minimize the risk of code relying on these environment variables, and to avoid accidentally trying to open inherited file descriptors twice, clear the variables after reading them.

### Checklist:


grynspan:
@swift-ci test

briancroom (approved):

PR #1047, opened 2025-04-01
Replace CRC-32 with SHA-256 in our macro target

grynspan:
This PR replaces CRC-32 with SHA-256 in our macro target.

We currently use CRC-32 to disambiguate test functions that would otherwise generate identical derived symbol names (using `context.makeUniqueName()`.) By replacing it with the first 64 bits of a SHA-256 hash, we reduce the odds of a collision.

We also currently use a 128-bit random number as a unique ID for exit tests during macro expansion (which is as unique as it gets, but unstable.) Replacing this random number with half of a SHA-256 hash allows us to generate _stable_ IDs (that are still statistically unique) which can help when debugging an exit test and may also improve cache quality for tools (e.g. an IDE's symbol cache.)

Because there is no SHA-256 implementation available in the Swift standard library or other components we can reliably link to in the macro target, I've borrowed the implementation of SHA-256 in swift-tools-support-core. The original version is [here](https://github.com/swiftlang/swift-tools-support-core/blob/main/Sources/TSCBasic/HashAlgorithms.swift).

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

briancroom (approved):

Sources/TestingMacros/Support/SHA256.swift
```diff
@@ -0,0 +1,184 @@
...
+// Copyright (c) 2014–2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// The contents of this file were copied more-or-less verbatim from
+/// [swift-tools-support-core](https://github.com/swiftlang/swift-tools-support-core/blob/main/Sources/TSCBasic/HashAlgorithms.swift).
+
```
pmattos, line 13: Consider also adding the commit hash to help future patches from upstream, etc.

grynspan: We do not expect SHA-256 to change in the future, and the maintainers of TSC aren't planning to make any further changes to the original copy of this code. :)

grynspan: Added.

grynspan:
@swift-ci test

PR #1049, opened 2025-04-04
Emit "barriers" into the stdout/stderr streams of an exit test.

grynspan:
This PR causes Swift Testing to write "barriers" (known sequences of bytes) to `stdout` and `stderr` in the child process created by an exit test. Then, in the parent, these values are used to splice off any leading or trailing output that wasn't generated by the exit test's body (such as content generated by the host process, XCTest/Xcode, etc.)

This reduces the amount of extraneous data reported back to the exit test's parent process.

Thanks to @briancroom for the suggestion.

### Checklist:


stmontgomery (approved):

grynspan:
@swift-ci test

PR #1055, opened 2025-04-07
Miscellaneous bookkeeping/cleanup of `_TestDiscovery`.

grynspan:
This PR has a few minor changes:

- The default implementation of `DiscoverableAsTestContent._testContentTypeNameHint` is not present if `SWT_NO_LEGACY_TEST_DISCOVERY` is defined;
- Fixed the precondition in `TestContentKind.init(stringLiteral:)` so it doesn't call `utf8CodeUnitCount` on single-character static strings;
- The implementation of `TestContentKind._fourCCValue` uses `isprint()` to check if the kind value looks like a FourCC rather than calling ICU-based functions;
- Fixed a type in a comment on `SectionBounds.Kind.segmentAndSectionName`; and
- Simplified the implementation of `TestContentKind.commentRepresentation` (in the macro target).

### Checklist:


grynspan:
@swift-ci test

briancroom (approved):

PR #1056, opened 2025-04-07
Add signal handler for `SIGABRT_COMPAT` on Windows.

grynspan:
The special-case signal handling implemented for Windows doesn't currently set a handler for `SIGABRT_COMPAT` (which is a synonym of `SIGABRT` with a different value.) This PR adds it to the list of signals we install handlers for.

### Checklist:


grynspan:
@swift-ci test

briancroom (approved):

stmontgomery (approved):

PR #1061, opened 2025-04-08
Bump the Swift version to 6.1.2

shahmishal:


grynspan (approved):

grynspan:
That remains an open question. Ours is not the only component with this problem.

stmontgomery:
I took a quick look and think there may be a viable way to plumb this info through the build-script / CMake build. I filed https://github.com/swiftlang/swift-testing/issues/1063.

PR #1069, opened 2025-04-10
Add `unsafe` keyword handling to macro expansions.

grynspan:
This PR changes the `@Test` and `#expect()` macros so they handle `unsafe` expressions the same way `try` and `await` expressions are handled. The propagation rules for `unsafe` aren't the same as for `try` and `await` (i.e. it doesn't colour the calling function), but the general way we handle the keyword is the same.

I haven't attempted to avoid inserting `unsafe` if a function is not marked `@unsafe` as it complicates the necessary logic but has no effects at runtime.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -96,7 +96,7 @@ let package = Package(
   }(),
 
   dependencies: [
-    .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "601.0.0-latest"),
+    .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "602.0.0-latest"),
   ],
 
   targets: [
--- a/Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
+++ b/Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
@@ -1,0 +1,159 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+import SwiftSyntax
+import SwiftSyntaxBuilder
+import SwiftSyntaxMacros
+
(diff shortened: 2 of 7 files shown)


grynspan:
@swift-ci test

Package.swift
```diff
@@ -96,7 +96,7 @@ let package = Package(
   }(),
 
   dependencies: [
-    .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "601.0.0-latest"),
+    .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "602.0.0-latest"),
```
stmontgomery, line 99: Can you update the hardcoded `GIT_TAG` commit hash in `CMakeRules.txt` for TestingMacros target to match this?

Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
```diff
@@ -0,0 +1,159 @@
...
+///   - context: The macro context in which the expression is being parsed.
+///
+/// - Returns: A set of effectful keywords such as `await` that are present in
+///   `node`.
+///
+/// This function does not descend into function declarations or closure
+/// expressions because they represent distinct lexical contexts and their
+/// effects are uninteresting in the context of `node` unless they are called.
+func findEffectKeywords(in node: some SyntaxProtocol, context: some MacroExpansionContext) -> Set<Keyword> {
+  // TODO: gather any effects from the lexical context once swift-syntax-#3037 and related PRs land
```
stmontgomery, line 66: Has this just not landed in the tag you're using? I see the PRs themselves have landed

Package.swift
grynspan: Will do Monday.

Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
grynspan: I need to wait a few more nightlies before I can start trying it out, looks like.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

suzannaratcliff (approved):

PR #1074, opened 2025-04-11
Add support for the `--attachments-path` CLI argument.

grynspan:
This PR adds support for the `--attachments-path` CLI argument on `swift test` as approved in [ST-0009](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0009-attachments.md). We will maintain support for the older `--experimental-attachments-path` version through Swift 6.2.

Resolves rdar://147753783.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
@@ -285,8 +285,8 @@ public struct __CommandLineArguments_v0: Sendable {
   /// The value of the `--repeat-until` argument.
   public var repeatUntil: String?
 
-  /// The value of the `--experimental-attachments-path` argument.
-  public var experimentalAttachmentsPath: String?
+  /// The value of the `--attachments-path` argument.
+  public var attachmentsPath: String?
 
   /// Whether or not the experimental warning issue severity feature should be
   /// enabled.
@@ -314,7 +314,7 @@ extension __CommandLineArguments_v0: Codable {
     case skip
     case repetitions
     case repeatUntil
-    case experimentalAttachmentsPath
+    case attachmentsPath
   }
 }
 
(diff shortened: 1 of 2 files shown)


grynspan:
@younata FYI.

grynspan:
@swift-ci test

grynspan:
See also https://github.com/swiftlang/swift-package-manager/pull/8492

briancroom (approved):

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -509,9 +510,9 @@ public func configurationForEntryPoint(from args: __CommandLineArguments_v0) thr
   }
 
   // Attachment output.
-  if let attachmentsPath = args.experimentalAttachmentsPath {
+  if let attachmentsPath = args.attachmentsPath {
     guard fileExists(atPath: attachmentsPath) else {
```
stmontgomery, line 514: Tangential this PR and can be fixed separately, but I notice this throws an error if the attachment directory doesn't exist yet. I find it helpful when tools create directories like this on the user's behalf if it doesn't exist yet.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -396,8 +396,9 @@ func parseCommandLineArguments(from args: [String]) throws -> __CommandLineArgum
   }
 
   // Attachment output
-  if let attachmentsPathIndex = args.firstIndex(of: "--experimental-attachments-path"), !isLastArgument(at: attachmentsPathIndex) {
-    result.experimentalAttachmentsPath = args[args.index(after: attachmentsPathIndex)]
+  if let attachmentsPathIndex = args.firstIndex(of: "--attachments-path") ?? args.firstIndex(of: "--experimental-attachments-path"),
+     !isLastArgument(at: attachmentsPathIndex) {
```
stmontgomery, line 400: Is this "last argument" check new in this PR? Could we report an error if the flag is passed but it lacks a path afterward? Or am I misreading and it will produce an error already? I just don't want it to be silently ignored

grynspan: It's not new and serves as a bounds check so we don't read past the end of the array if `--attachments-path` is somehow the last argument.

I want to do a cleanup pass on the logic here to make it more legible and also add support for `--x=y` but it isn't a top priority.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: I want to discuss the exact behaviour of the flag with the SwiftPM folks first; for now, this is the behaviour we already implemented.

PR #1075, opened 2025-04-14
Expose Issue severity and isFailure as API

suzannaratcliff:
Expose `Issue` `Severity` and `isFailure` as API

### Motivation:

In order for users to create issues on their tests but would not like those issues to fail their test it would be nice to be able to create an issue with a warning severity.

**Test Warnings:**
Severity:
```
  public enum Severity: Sendable {
    /// The severity level for an issue which should be noted but is not
    /// necessarily an error.
    ///
    /// An issue with warning severity does not cause the test it's associated
    /// with to be marked as a failure, but is noted in the results.
    case warning

    /// The severity level for an issue which represents an error in a test.
    ///
    /// An issue with error severity causes the test it's associated with to be
    /// marked as a failure.
    case error
  }
```

This will allow users to create issues but they will not have to fail the test

Example Usage:
```
Issue.record("My comment", severity: .warning)
```

Later on users can inspect the severity of their issue with the severity property.
```
/// The severity of the issue.
public var severity: Severity
```

```
/// If the issue is a failing issue.
public var isFailure: Bool
```

This API will be useful for teams who would like to inspect if their issue is a failing issue
https://developer.apple.com/documentation/xctest/xctestcase/record(_:)

### Modifications:

_[Describe the modifications you've done.]_
- Add isFailure variable
- Remove experimental spi from Severity 

### Checklist:


Sources/Testing/Issues/Issue.swift
```diff
@@ -101,8 +100,12 @@ public struct Issue: Sendable {
   }
 
   /// The severity of this issue.
-  @_spi(Experimental)
   public var severity: Severity
+  
+  /// If the issues is a failing issue.
+  public var isFailure: Bool {
```
grynspan, line 106: I don't think this member is necessary; whether or not a warning is a failure might depend on external configuration (some equivalent to `-Werror` perhaps.)

Sources/Testing/Issues/Issue.swift
```diff
@@ -101,8 +100,12 @@ public struct Issue: Sendable {
...
+    severity == .error
+  }
```
stmontgomery, line 108: I suggest adding this `isFailure` property via a separate PR as `@_spi` first, so that this PR can consist only of _removals_ of `@_spi` guards and similar affordances.

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -58,25 +58,6 @@ extension Issue {
...
-  ///   - comment: A comment describing the expectation.
-  ///   - sourceLocation: The source location to which the issue should be
-  ///     attributed.
-  ///
-  /// - Returns: The issue that was recorded.
-  ///
-  /// Use this function if, while running a test, an issue occurs that cannot be
-  /// represented as an expectation (using the ``expect(_:_:sourceLocation:)``
-  /// or ``require(_:_:sourceLocation:)-5l63q`` macros.)
-  @discardableResult public static func record(
```
grynspan, line 73: As with the other overload of `record()`, make sure that you provide a default argument for `severity` in the replacement to avoid source breakage.

grynspan:
@suzannaratcliff @stmontgomery I believe this PR is obsolete?

PR #1077, opened 2025-04-14
Add runtime issue kind and record issue with a kind

suzannaratcliff:
Add runtime issue kind and record issue with a kind

### Motivation:

Runtime issues are reports of programming errors at runtime.  These are different than test failures because those are pointing out issues in your project while runtime issues point out errors in how/what you code. 

This will allow anyone to report a runtime issue during a test.
### Modifications:

- Added support for specifying a kind when recording an issue
- Added a runtime issue type for users to report runtime issues

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -62,6 +62,7 @@ extension Issue {
   ///
   /// - Parameters:
   ///   - comment: A comment describing the expectation.
+  ///   - kind: The kind of issue.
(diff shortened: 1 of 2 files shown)


grynspan:
We should probably workshop the name of this case a bit and figure out what it can actually be used for. I wouldn't want it to be tightly coupled to Apple-specific implementation details, for instance. If it _is_ ultimately Apple-only, it may not make sense to expose it on other platforms.

grynspan (changes requested):
Please do not add `kind` as an argument to `record()`. It is intentionally omitted.

PR #1080, opened 2025-04-16
Introduce issue handling trait (as SPI)

stmontgomery:
This introduces a new trait type named `IssueHandlingTrait` as SPI. It allows observing, transforming, or filtering the issue(s) recorded during the test it's applied to. Here's a contrived example:

```swift
@Test(.transformIssues { issue in
  var issue = issue
  issue.comments.append("A comparison of two literals")
  return issue  // Or, return `nil` to suppress the issue
})
func example() {
  #expect(1 == 2)
}
```

### Motivation:

Sometimes it can be useful to customize an issue recorded during a test. For example, you might wish to add supplemental information to it, such by adding comments. Another example of this could be adding an attachment to an issue, which was a capability mentioned as a [future direction](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0009-attachments.md#future-directions) in [ST-0009 Attachments](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0009-attachments.md).

Other times, you might wish to suppress an issue which is later determined to be irrelevant or cannot be marked as a [known issue](https://swiftpackageindex.com/swiftlang/swift-testing/main/documentation/testing/known-issues). Or, you may simply want to be notified that an issue was recorded, to react to it in some other way while still recording the issue normally.

### Checklist:

Resolves rdar://140144041

### DIFF
--- a/Sources/Testing/Traits/IssueHandlingTrait.swift
+++ b/Sources/Testing/Traits/IssueHandlingTrait.swift
@@ -113,11 +113,11 @@ extension Trait where Self == IssueHandlingTrait {
   ///     this trait is applied to. It is passed a recorded issue, and returns
   ///     an optional issue to replace the passed-in one.
   ///
-  /// The `transformer` closure is called each time an issue is recorded by the
-  /// test this trait is applied to. The closure is passed the recorded issue,
-  /// and if it returns a non-`nil` value, that will be recorded instead of the
-  /// original. Otherwise, if the closure returns `nil`, the issue is suppressed
-  /// and will not be included in the results.
+  /// The `transformer` closure is called synchronously each time an issue is
+  /// recorded by the test this trait is applied to. The closure is passed the
+  /// recorded issue, and if it returns a non-`nil` value, that will be recorded
+  /// instead of the original. Otherwise, if the closure returns `nil`, the
+  /// issue is suppressed and will not be included in the results.
   ///
   /// The `transformer` closure may be called more than once if the test records
   /// multiple issues. If more than one instance of this trait is applied to a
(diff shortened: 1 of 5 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/Traits/IssueHandlingTrait.swift
```diff
@@ -0,0 +1,167 @@
...
+/// `nil` from the closure passed to ``Trait/transformIssues(_:)``.
+///
+/// When an instance of this trait is applied to a suite, it is recursively
+/// inherited by all child suites and tests.
+///
+/// To add this trait to a test, use one of the following functions:
+///
+/// - ``Trait/transformIssues(_:)``
+/// - ``Trait/filterIssues(_:)``
+@_spi(Experimental)
```
suzannaratcliff, line 27: Will you explain when this is run? Is it whenever an issue is recorded or always at the end of the test?

stmontgomery: Good clarification - I'll describe that in the docs. (The closure is called synchronously the moment each issue is recorded; it doesn't wait til the end of the test)

suzannaratcliff: Why is it called `transformIssues` then? I would expect `transformIssue`

stmontgomery: It's pluralized because more than one issue can be recorded by the test, and the transformer closure is called once for each recorded issue. (That's noted somewhere else in the docs)

stmontgomery: I'm totally open to bikeshedding the name though. I just prefer to find a name that conveys the fact that there can potentially be multiple issues

stmontgomery:
@swift-ci please test

stmontgomery:
The macOS CI failure is expected to be resolved by https://github.com/swiftlang/swift/pull/80830, but the release of a new main snapshot toolchain is currently blocked for unrelated reasons. Proceeding to merge.

PR #951, opened 2025-02-12
Enable Library Evolution in package builds for public library targets

stmontgomery:
This modifies `Package.swift` to enable Library Evolution for builds of the package.

### Motivation:

I recently landed a change (#931) which passed our project-level CI but later failed in Swift CI. The difference ended up being due to the latter building with Library Evolution (LE) enabled, whereas our project-level CI builds via SwiftPM and does not enable LE. The change was reverted (#950) but this revealed a gap in our testing strategy. We should always build these targets with LE enabled.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

Fixes rdar://144655439

PR #1082, opened 2025-04-17
Enable Library Evolution for package-based builds of the _TestDiscovery target

stmontgomery:
This enables Library Evolution (LE) for the `_TestDiscovery` target in Swift package-based builds. The CMake-based build is already configured to enable LE for this target, so this is just matching behavior.

(Note that we already enable LE for several other targets which vend public API as of #951.)

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -171,7 +171,7 @@ let package = Package(
       dependencies: ["_TestingInternals",],
       exclude: ["CMakeLists.txt"],
       cxxSettings: .packageSettings,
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
The macOS CI failure is expected to be resolved by https://github.com/swiftlang/swift/pull/80830. Proceeding to merge, given this is a known issue.

PR #1092, opened 2025-04-23
Include code comments before expectations which are preceded by try/await in recorded issues

stmontgomery:
This fixes an issue where code comments placed before an expectation like `#expect()` which has effect introducer keywords like `try` or `await` are ignored, and ensures they are included in the recorded issue if the expectation fails.

Consider this example of two failing expectations:

```swift
// Uh oh!
#expect(try x() == 2)

// Uh oh!
try #expect(x() == 2)
```

Prior to this PR, if `x()` returned a value other than `2`, there would be two issues recorded, but the second one would not have the comment `“Uh oh!”` because from the macro’s perspective, that code comment was on the `try` keyword and it could only see trivia associated with `#expect()`.

Now, with the recent swift-syntax fix from https://github.com/swiftlang/swift-syntax/pull/3037, the `try` keyword and its associated trivia can be included and this bug can be fixed. We recently adopted a new-enough swift-syntax in #1069, so the only fix needed is to adopt `lexicalContext` for this new purpose in our macro.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/ConditionMacro.swift
+++ b/Sources/TestingMacros/ConditionMacro.swift
@@ -163,7 +163,7 @@ extension ConditionMacro {
       let commentsArrayExpr = ArrayExprSyntax {
         // Lexical context is ordered innermost-to-outermost, so reverse it to
         // maintain the expected order.
-        for lexicalSyntaxNode in context.lexicalContext.effectExpressions.reversed() {
+        for lexicalSyntaxNode in context.lexicalContext.trailingEffectExpressions.reversed() {
           for commentTraitExpr in createCommentTraitExprs(for: lexicalSyntaxNode) {
             ArrayElementSyntax(expression: commentTraitExpr)
           }
--- a/Sources/TestingMacros/ConditionMacro.swift
+++ b/Sources/TestingMacros/ConditionMacro.swift
@@ -163,7 +163,7 @@ extension ConditionMacro {
       let commentsArrayExpr = ArrayExprSyntax {
         // Lexical context is ordered innermost-to-outermost, so reverse it to
         // maintain the expected order.
-        for lexicalSyntaxNode in context.lexicalContext.reversed() {
+        for lexicalSyntaxNode in context.lexicalContext.effectExpressions.reversed() {
           for commentTraitExpr in createCommentTraitExprs(for: lexicalSyntaxNode) {
             ArrayElementSyntax(expression: commentTraitExpr)
           }
--- a/Sources/TestingMacros/ConditionMacro.swift
+++ b/Sources/TestingMacros/ConditionMacro.swift
@@ -157,8 +157,17 @@ extension ConditionMacro {
         expandedFunctionName = conditionArgument.expandedFunctionName
       }
 
-      // Capture any comments as well (either in source or as a macro argument.)
+      // Capture any comments as well -- either in source, preceding the
+      // expression macro or one of its lexical context nodes, or as an argument
+      // to the macro.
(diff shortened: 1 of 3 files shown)


stmontgomery:
cc @hamishknight

briancroom (approved):

suzannaratcliff (approved):

Sources/TestingMacros/ConditionMacro.swift
```diff
@@ -157,8 +157,17 @@ extension ConditionMacro {
         expandedFunctionName = conditionArgument.expandedFunctionName
       }
 
-      // Capture any comments as well (either in source or as a macro argument.)
+      // Capture any comments as well -- either in source, preceding the
+      // expression macro or one of its lexical context nodes, or as an argument
+      // to the macro.
       let commentsArrayExpr = ArrayExprSyntax {
+        // Lexical context is ordered innermost-to-outermost, so reverse it to
```
grynspan, line 164: This will capture comments for distant nodes too, which seems incorrect?

grynspan (changes requested):

Sources/TestingMacros/ConditionMacro.swift
stmontgomery: Good callout — fixed in latest commit!

stmontgomery:
@swift-ci please test

Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
```diff
@@ -69,6 +69,19 @@ func findEffectKeywords(in node: some SyntaxProtocol, context: some MacroExpansi
   return effectFinder.effectKeywords
 }
 
+extension Sequence<Syntax> {
+  /// The syntax nodes in this sequence which are effectful expressions, such as
+  /// those for `try` or `await`.
+  var effectExpressions: some Sequence<Syntax> {
+    filter { node in
```
grynspan, line 76: I'm not certain what the compiler and swift-syntax are doing exactly, but this could still be incorrect:

```swift
// bad comment
try {
  #expect(...)
}()
```

This specific example is contrived, but do you see what I mean? I think you basically want to inspect the suffix of the lexical context sequence up to the first non-expr (and non-closure!) node.

grynspan (changes requested):

Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
stmontgomery: The part about only taking the _suffix_ of matching nodes absolutely makes sense. Implemented that in latest commit.

To err on the conservative/safe side right now I'm going to continue filtering to nodes of explicitly-specified types, rather than generalizing it to any `ExprSyntax`, and subtracting known-bad exprs like closures. I figure we can widen this over time, and it's primarily `try`/`await` that I care about at the moment.

stmontgomery:
@swift-ci please test

Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
grynspan: Makes sense!

grynspan: Also consider `UnsafeExprSyntax`!

grynspan (approved):

stmontgomery:
The macOS and Linux CI failures are expected to be resolved by https://github.com/swiftlang/swift/pull/80830, but the release of a new `main` snapshot toolchain is currently blocked for unrelated reasons. Proceeding to merge.

PR #1084, opened 2025-04-18
Enable experimental 'AllowUnsafeAttribute' feature to continue supporting 6.1 development snapshot toolchains

stmontgomery:
This fixes a build failure when attempting to build the `main` branch using a 6.1 development snapshot toolchain. This failure was introduced by #1069, which added usage of the new `@unsafe` attribute, and the failure was revealed when we set up the 6.1 CI jobs in #1083.

Here are some relevant related Swift PRs which give context around these changes:

- https://github.com/swiftlang/swift/pull/75413
- https://github.com/swiftlang/swift/pull/79645

See the code comment for more details.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1093, opened 2025-04-23
Fix build failure when building with pre-6.2 toolchain due to unrecognized `unsafe` keyword

stmontgomery:
This fixes another bit of fallout from #1069 when building this project's test targets with a 6.1 (or any pre-6.2) toolchain.

The `unsafe` keyword was introduced in 6.2 as part of [SE-0458: Opt-in Strict Memory Safety Checking](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md). Older toolchains are not aware of it, so the fix is to avoid emitting expressions involving that keyword when the macro plugin has been built using an older toolchain.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
+++ b/Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
@@ -111,7 +111,13 @@ func applyEffectfulKeywords(_ effectfulKeywords: Set<Keyword>, to expr: some Exp
 
   let needAwait = effectfulKeywords.contains(.await) && !expr.is(AwaitExprSyntax.self)
   let needTry = effectfulKeywords.contains(.try) && !expr.is(TryExprSyntax.self)
+
+  // The 'unsafe' keyword was introduced in 6.2 as part of SE-0458. Older
+  // toolchains are not aware of it, so avoid emitting expressions involving
+  // that keyword when the macro has been built using an older toolchain.
+#if compiler(>=6.2)
   let needUnsafe = effectfulKeywords.contains(.unsafe) && !expr.is(UnsafeExprSyntax.self)
+#endif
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift please test

stmontgomery:
@swift-ci please test 6.1

briancroom (approved):

grynspan (approved):

grynspan:
When building from source locally with the release 6.1 toolchain, I'm seeing the compiler continue to complain that "Attribute requires '-enable-experimental-feature AllowUnsafeAttribute'" on `__requiringUnsafe()`. So I think we need to guard the `@unsafe` attribute there with:

```swift
#if hasFeature(AllowUnsafeAttribute)
@unsafe
#endif
```

stmontgomery:
> When building from source locally with the release 6.1 toolchain, I'm seeing the compiler continue to complain that "Attribute requires '-enable-experimental-feature AllowUnsafeAttribute'" on `__requiringUnsafe()`

That should be fixed by #1084. Or, are you saying your local checkout has that fix but you're still seeing this error? With that fix plus this PR, I am able to build using the 6.1 release toolchain successfully.

grynspan:
I have `main` checked out and I see the diagnostic with:

> swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)
> Target: arm64-apple-macosx15.0

stmontgomery:
The macOS and Linux CI failures are expected to be resolved by https://github.com/swiftlang/swift/pull/80830, but the release of a new `main` snapshot toolchain is currently blocked for unrelated reasons. Proceeding to merge.

PR #1103, opened 2025-05-03
Restore DocC `@Comment` blocks to documentation line comments

stmontgomery:
This reverts a portion of the changes I made in #836 which converted triple-`///` style DocC `@Comment` blocks to double-`//` style code comments. That change was needed at the time because our documentation was still rendered using pre-Swift 6.1 toolchains, but now `@Comment` blocks are supported by DocC in documentation comments (as of https://github.com/swiftlang/swift-docc/pull/1107) so we can restore these.

I needed to move some of these comment blocks up before `## See Also` sections to avoid a DocC warning about items in lists not being links.

### Checklist:


stmontgomery:
@swift-ci please test

PR #1105, opened 2025-05-04
Synthesize display names for de facto suites with raw identifiers.

### LINKED ISSUE #1104: Raw identifiers for suite types aren't treated as display names
Run the following code:

```swift
import Testing

struct `My example tests` {
    @Test func `one little test`() {}
    @Test func `and/another/one`() {}
}
```

The test functions are correctly treated as having display names, but the suite is not. Because it doesn't have `@Suite`, its name is inferred at runtime instead of at compile time. Adding `@Suite` correctly infers the name.

This should be a mostly trivial fix: if the name of the type is a raw identifier and the suite is synthesized, treat the type name as the suite's display name.

~~There's an edge case here where backticks are used for names that are _not_ raw (e.g. struct \`notraw\`), but we would need to pull in swift-syntax at runtime to detect those. We can change the rules here and simply infer that any use of backticks around a type or function name produces a display name (but won't produce the diagnostic we get when you also specify a display name with a string literal, because it might be necessary in that case I guess.)~~ No there isn't, because such names won't have backticks when demangled. So we're fine.

grynspan:
This PR ensures that suite types that don't have the `@Suite` attribute but which _do_ have raw identifiers for names are correctly given display names the same way those with `@Suite` would be. This PR also ensures that we transform spaces in raw identifiers after they are demangled by the runtime--namely, the runtime replaces ASCII spaces (as typed by the user) with Unicode non-breaking spaces (which aren't otherwise valid in raw identifers) in order to avoid issues with existing uses of spaces in demangled names. We want to make sure that identifiers as presented to the user match what the user has typed, so we need to transform these spaces back. No changes in this area are needed for display names derived during macro expansion because we do the relevant work based on the source text which still has the original ASCII spaces.

This PR also deletes the "`raw$`" hack that I put in place when originally implementing raw identifier support as the entire toolchain supports them now.

Resolves #1104.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
9sswjj

grynspan:
Looks like the Windows toolchain doesn't support these identifiers yet @allevato @compnerd.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/Parameterization/TypeInfo.swift
```diff
@@ -142,6 +142,34 @@ func rawIdentifierAwareSplit<S>(_ string: S, separator: Character, maxSplits: In
...
+  /// normal ASCII spaces with non-breaking spaces to maintain compatibility
+  /// with historical usages of spaces in mangled name forms. Non-breaking
+  /// spaces are not otherwise valid in raw identifiers, so this transformation
+  /// is reversible.
+  private static func _rewriteNonBreakingSpacesAsASCIISpaces(in rawIdentifier: some StringProtocol) -> String? {
+    guard rawIdentifier.contains("\u{00A0}") else {
+      return nil
+    }
+
+    let result = rawIdentifier.lazy.map { c in
```
allevato, line 163: Not important: My experience has led me to operate on Unicode scalar views in situations like this where I don't need grapheme clustering so I don't pay the extra cost of it. It's probably not going to be a big bottleneck here, but that's just my habit 🤷🏻‍♂️

allevato:
> https://ci-external.swift.org/job/swift-testing-pull-request-windows/1474/consoleText
> 
> Looks like the Windows toolchain doesn't support these identifiers yet @allevato @compnerd.

I'm not sure what's going on there; I wouldn't have been able to merge the original support if it didn't work on Windows. Both the old C++ parser and the Swift parser handle them. Any ideas @compnerd ?

allevato (approved):

compnerd:
> > https://ci-external.swift.org/job/swift-testing-pull-request-windows/1474/consoleText
> > Looks like the Windows toolchain doesn't support these identifiers yet @allevato @compnerd.
> 
> I'm not sure what's going on there; I wouldn't have been able to merge the original support if it didn't work on Windows. Both the old C++ parser and the Swift parser handle them. Any ideas @compnerd ?

I'm not sure - could it be missing cherry-picks?

allevato:
> I'm not sure - could it be missing cherry-picks?

Hmm. The raw identifier changes were merged before 6.2 was branched, and the tag history [here](https://github.com/swiftlang/swift/pull/76636/commits/d94bd80c6263b7aad5b8cd9bc7308757f4fb1282) shows that it's present in the 6.2 dev snapshots so far: `swift-6.2-DEVELOPMENT-SNAPSHOT-2025-04-21-a`, `swift-6.2-DEVELOPMENT-SNAPSHOT-2025-04-23-a`, and `swift-6.2-DEVELOPMENT-SNAPSHOT-2025-04-24-a`.

But the Swift version info in that log says `Swift version 6.2-dev (LLVM d9449c924457b3a, Swift b57f8a9750f4b7b)`, and [commit `b57f8a9750f4b7b` is from March 9](https://github.com/swiftlang/swift/commit/b57f8a9750f4b7bd7c189a11a897a62be02d5e63).

Is something lagging behind?

grynspan:
@shahmishal Is our Windows job pinned to an old build?

Sources/Testing/Parameterization/TypeInfo.swift
grynspan: I can change it.

grynspan:
@swift-ci test

Tests/TestingTests/MiscellaneousTests.swift
```diff
@@ -297,15 +297,25 @@ struct MiscellaneousTests {
...
-    #expect(id.nameComponents == ["MiscellaneousTests", "`raw_identifier_provides_a_display_name`()"])
+    #expect(id.nameComponents == ["MiscellaneousTests", "`Test with raw identifier gets a display name`()"])
   }
 
+  @Test func `Suite type with raw identifier gets a display name`() throws {
+    struct `Suite With De Facto Display Name` {}
+    let typeInfo = TypeInfo(describing: `Suite With De Facto Display Name`.self)
+    let suite = Test(traits: [], sourceLocation: #_sourceLocation, containingTypeInfo: typeInfo, isSynthesized: true)
+    let displayName = try #require(suite.displayName)
+    #expect(displayName == "Suite With De Facto Display Name")
```
stmontgomery, line 316: Can we test (and formally define) what the `name` property of such a suite is? That'll be important to commit to in terms of tools integration

Sources/Testing/Test.swift
```diff
@@ -209,8 +209,13 @@ public struct Test: Sendable {
...
     isSynthesized: Bool = false
   ) {
-    self.name = containingTypeInfo.unqualifiedName
-    self.displayName = displayName
+    let name = containingTypeInfo.unqualifiedName
+    self.name = name
+    if let displayName {
+      self.displayName = displayName
+    } else if isSynthesized && name.count > 2 && name.first == "`" && name.last == "`" {
+      self.displayName = String(name.dropFirst().dropLast())
```
stmontgomery, line 217: Does this handle backticks in parameter labels?

Tests/TestingTests/MiscellaneousTests.swift
grynspan: It's the name of the type, same as it ever was.

Sources/Testing/Test.swift
grynspan: There are no parameters in suite names.

Tests/TestingTests/MiscellaneousTests.swift
stmontgomery: Can we add an `#expect` validating `name`, though?

grynspan:
@swift-ci test

grynspan:
macOS CI is waiting for resources. Tested at desk and it's passed CI previously.

PR #1115, opened 2025-05-15
Polling confirmations (under Experimental spi)

younata:
This implements polling confirmations, as described in [ST-NNNN](https://github.com/younata/swift-evolution/blob/younata/testing-polling-expectations/proposals/testing/NNNN-polling-confirmations.md).

### Motivation:

Being able to monitor changes in the background is something of immense value. Swift Testing already provides an API for this in the `confirmation` api. However, I've found the `confirmation` to be hard to work with at times - it requires you to set up a callback for when something changes, which is not always possible or even the right thing to do. Polling provides a very general approach to monitoring all kinds of changes.

### Modifications:

This adds a new set of macros for handling polling. A new public enum for the 2 separate polling behaviors, and new types to actually implement polling. All under the experimental spi.

### Checklist:


Sources/Testing/Polling/Polling.swift
```diff
@@ -0,0 +1,293 @@
...
+    return await withTaskGroup { taskGroup in
+      taskGroup.addTask {
+        do {
+          try await Task.sleep(for: timeout)
+        } catch {}
+        // Task.sleep will only throw if it's cancelled, at which point this
+        // taskgroup has already returned and we don't care about the value
+        // returned here.
+        return await pollingProcessor.didTimeout()
+      }
```
younata, line 225: Per [discussion in forum](https://forums.swift.org/t/pitch-polling-expectations/79866/15), this is unsafe & needs to be rethought.

Sources/Testing/Polling/Polling.swift
```diff
@@ -0,0 +1,293 @@
...
+  /// A type to record the last value returned by a closure returning an optional
+  /// This is only used in the `#require(until:)` macro returning an optional.
+  private actor Recorder<R: Sendable> {
+    var lastValue: R?
+
+    /// Record a new value to be returned
+    func record(value: R) {
+      self.lastValue = value
+    }
+  }
```
younata, line 183: Ugh. I really dislike this approach, but I wanted to get this PR up so that others could actually test this proposal out & I couldn't think of a better way to handle this. Hopefully one of the benefits of a non-timed-based approach to polling will be that returning values will be much simpler/not a tacked-on hack.

Sources/Testing/Polling/Polling.swift
```diff
@@ -0,0 +1,293 @@
...
+        isRequired: isRequired,
+        sourceLocation: sourceLocation
+      )
+    case .cancelled:
+      Issue(
+        kind: .system,
+        comments: comments,
+        sourceContext: sourceContext
+      ).record()
+    }
```
younata, line 85: a) I'm going to deduplicate this analysis with the nearly-identical section in the other version of `run`. But time constraints & a desire to let others try out this code were why this was left in.
b) I'd really like to be able to provide additional explanations behind why each issue is recorded, but it wasn't clear what additional arguments to Issue (if any, maybe I need to set a property and I missed that?) I should use to provide that additional context.

Tests/TestingTests/PollingTests.swift
```diff
@@ -0,0 +1,352 @@
...
+        var configuration = Configuration()
+        configuration.eventHandler = { event, _ in
+          if case .issueRecorded = event.kind {
+            failed()
+          }
+        }
+        await Test {
+          await #expect(until: .passesOnce) { false }
+        }.run(configuration: configuration)
+      }
```
younata, line 48: I'm going to write a test helper to make it much easier to verify that issues were reported, as `withKnownIssue` is the wrong semantics for what I want. I'm also tempted to make it available under `ForToolsIntegrationOnly`, but I also think that should be a separate pitch?

I really want something that I can use like:

```swift
let issues = await collectIssues {
    #expect(Bool(false))
}
```

Tests/TestingTests/PollingTests.swift
```diff
@@ -0,0 +1,352 @@
...
+}
+
+extension DurationProtocol {
+  fileprivate func isCloseTo(other: Self, within delta: Self) -> Bool {
+    var distance = self - other
+    if (distance < Self.zero) {
+      distance *= -1
+    }
+    return distance <= delta
+  }
```
younata, line 343: I want this to be in the swift standard library and apply to all numeric types. It would make comparing floating point types so much more doable.

In my ideal world, this would work like `1.000001 == 1 ± 1e-3` (or `1.00001 == 1 +- 1e-3`, to avoid using the hard-to-type ± symbol). But that would involve adding another ternary operator to the language, and would be much more involved that adding a method to a protocol in the standard library.

grynspan: swift-numerics has this, more or less, but we can't add it as a dependency of course. :)

Sources/Testing/Issues/Issue.swift
```diff
@@ -38,6 +38,15 @@ public struct Issue: Sendable {
...
     indirect case confirmationMiscounted(actual: Int, expected: any RangeExpression & Sendable)
 
+    /// An issue due to a polling confirmation having failed.
+    ///
+    /// This issue can occur when calling ``confirmation(_:until:within:pollingEvery:isolation:sourceLocation:_:)-455gr``
+    /// or
+    /// ``confirmation(_:until:within:pollingEvery:isolation:sourceLocation:_:)-5tnlk``
+    /// whenever the polling fails, as described in ``PollingStopCondition``.
+    @_spi(Experimental)
+    case pollingConfirmationFailed
```
grynspan: Should we include any associated values with this case? e.g. to inform the console output for this issue kind.

younata: That seems reasonable to do. Maybe for if an error was thrown, and if the stopping condition fails?

younata: Here's what I'm thinking right now:

```swift
/// A type defining why a Polling Confirmation failed
@_spi(Experimental)
public enum PollingConfirmationFailureReason: Sendable {
  /// The polling confirmation failed because the stop condition failed.
  ///
  /// If the stop condition is ``PollingStopCondition.firstPass``, then this
  /// this means that the polling confirmation exceeded the number of polling
  /// iterations without the `body` closure having returned true or a non-nil
  /// value.
  ///
  /// If the stop condition is ``PollingStopCondition.stopsPassing``, then this
  /// means that the `body` closure returned false or nil before the polling
  /// confirmation reached the number of polling iterations.
  case stopConditionFailed(PollingStopCondition)

  /// The polling confirmation failed because the `body` closure threw an error.
  indirect case errorThrown(_ error: any Error)
}
```

Thoughts?

grynspan: If an error is thrown, do we want to treat that as the polling confirmation failing _per se_, or just a separate issue of kind `.errorCaught()` and the polling confirmation just doesn't record any sort of result?

Tests/TestingTests/PollingTests.swift
```diff
@@ -0,0 +1,544 @@
...
+        try await confirmation(until: stop, pollingEvery: .milliseconds(1)) {
+          await incrementor.increment() != 0
+        }
+        let count = await incrementor.count
+        #expect(count == 100)
+      }
+
+      @available(_clockAPI, *)
+      @Test(
+        "When test configuration porvided, uses the test configuration",
```
plemarquand:
```suggestion
        "When test configuration provided, uses the test configuration",
```

Sources/Testing/Traits/PollingConfigurationTrait.swift
```diff
@@ -0,0 +1,74 @@
...
+/// To add this trait to a test, use the ``Trait/pollingConfirmationDefaults``
+/// function.
+@_spi(Experimental)
+@available(_clockAPI, *)
+public struct PollingConfirmationConfigurationTrait: TestTrait, SuiteTrait {
+  /// The stop condition to this configuration is valid for
+  public var stopCondition: PollingStopCondition
+
+  /// How long to continue polling for. If nil, this will fall back to the next
+  /// inner-most `PollingUntilStopsPassingConfigurationTrait.duration` value.
```
plemarquand: I think `PollingUntilStopsPassingConfigurationTrait` was renamed. Should this now be `PollingConfirmationConfigurationTrait`?

Sources/Testing/Traits/PollingConfigurationTrait.swift
```diff
@@ -0,0 +1,74 @@
...
+  /// If no non-nil values are found, then it will use 1 second.
+  public var duration: Duration?
+
+  /// The minimum amount of time to wait between polling attempts. If nil, this
+  /// will fall back to earlier `PollingUntilStopsPassingConfigurationTrait.interval`
```
plemarquand: And here

Sources/Testing/Traits/PollingConfigurationTrait.swift
```diff
@@ -0,0 +1,74 @@
...
+extension Trait where Self == PollingConfirmationConfigurationTrait {
+  /// Specifies defaults for polling confirmations in the test or suite.
+  ///
+  /// - Parameters:
+  ///   - stopCondition: The `PollingStopCondition` this trait applies to.
+  ///   - duration: The expected length of time to continue polling for.
+  ///     This value may not correspond to the wall-clock time that polling
+  ///     lasts for, especially on highly-loaded systems with a lot of tests
+  ///     running.
+  ///     if nil, polling will be attempted for approximately 1 second.
```
plemarquand:
```suggestion
  ///     If nil, polling will be attempted for approximately 1 second.
```

Tests/TestingTests/PollingTests.swift
```diff
@@ -0,0 +1,544 @@
...
+        if case .pollingConfirmationFailed = $0.kind {
+          return true
+        } else {
+          return false
+        }
+      })
+    }
+
+    @available(_clockAPI, *)
+    @Test("if the closures starts off as true, but becomes false")
```
plemarquand:
```suggestion
    @Test("if the closure starts off as true, but becomes false")
```

Sources/Testing/Polling/Polling.swift
```diff
@@ -0,0 +1,459 @@
...
+  ///   this will record an issue.
+  @discardableResult func evaluateOptional<R>(
+    isolation: isolated (any Actor)?,
+    _ body: @escaping () async -> sending R?
+  ) async throws -> R {
+    precondition(duration > Duration.zero)
+    precondition(interval > Duration.zero)
+    precondition(duration > interval)
+
+    let iterations = max(Int(duration.seconds() / interval.seconds()), 1)
```
plemarquand: While improbable, it is possible to provide inputs that will crash here:
```
let duration = Duration.seconds(Int.max)
let interval = Duration.milliseconds(1) 
```

plemarquand (commented):
Overall looking really great. Just one comment and a bunch of documentation nits.

Sources/Testing/Polling/Polling.swift
younata: Good catch! Thanks!

Sources/Testing/Polling/Polling.swift
```diff
@@ -0,0 +1,463 @@
...
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// Default values for polling confirmations.
+@available(_clockAPI, *)
+internal let defaultPollingConfiguration = (
```
stmontgomery: Can this be `private`?

```suggestion
private let _defaultPollingConfiguration = (
```

Sources/Testing/Confirmations/Polling.swift
stmontgomery, line 1: New files, such as this one and `PollingConfigurationTrait.swift`, need to be added to the CMake scripts. Can you be sure to add them there? See the file list in `Sources/Testing/CMakeLists.txt`.

stmontgomery, line 1: In terms of the placement of this new file, I'd like to suggest instead of creating a new `Polling/` subdirectory we create a new `Confirmations/` subdirectory, add move the existing `Confirmation.swift` to that new directory, and place this new file there too.

(And if you _do_ take this suggestion and move some files around, be sure to adjust the CMake files accordingly.)

Sources/Testing/Polling/Polling.swift
```diff
@@ -0,0 +1,463 @@
...
+  /// If the expression only returns true before the timeout is reached, then
+  /// no failure will be reported.
+  /// If the expression does not finish evaluating before the timeout is
+  /// reached, then a failure will be reported.
+  case stopsPassing
+}
+
+/// A type describing why polling failed
+@_spi(Experimental)
+public enum PollingFailureReason: Sendable, Codable {
```
stmontgomery: I think this `PollingFailureReason` enum would feel more natural nested within `PollingFailedError`, and then its name could be simplified to `Reason` (or fully-qualified, `PollingFailedError.Reason`).

grynspan: Or could `PollingFailedError` be the `enum` itself, then we don't need a nested type? (Either way we've got a `public enum` here.)

younata: Fair idea. The main reason for keeping `PollingFailedError` a struct is to make the conformance to `CustomIssueRepresentable` simple and obvious.

For what test authors consume, I expect `PollingFailedError.Reason` provides enough information for them to do anything custom - they can't really access `sourceContext` in the general case, and `comment` likely isn't that useful to handle?

I'm inclined to keep with @stmontgomery's suggestion right now, but I definitely see @grynspan's argument.

Sources/Testing/Confirmations/Polling.swift
```diff
@@ -0,0 +1,462 @@
...
+///   - stopCondition: When to stop polling.
+///   - duration: The expected length of time to continue polling for.
+///     This value may not correspond to the wall-clock time that polling lasts
+///     for, especially on highly-loaded systems with a lot of tests running.
+///     If nil, this uses whatever value is specified under the last
+///     ``PollingConfirmationConfigurationTrait`` added to the test or suite
+///     with a matching stopCondition.
+///     If no such trait has been added, then polling will be attempted for
+///     about 1 second before recording an issue.
+///     `duration` must be greater than 0.
```
jerryjrchen: Might also be worth mentioning that `duration > interval` based on the current implementation.

Although from a user perspective, should `duration == interval` be permissible? I would interpret that basically a single poll (so perhaps not very useful), but still a possible use case.

Sources/Testing/Confirmations/Polling.swift
```diff
@@ -0,0 +1,462 @@
...
+      return false
+    }
+  }
+}
+
+/// Confirm that some expression eventually returns a non-nil value
+///
+/// - Parameters:
+///   - comment: A user-specified comment describing this confirmation.
+///   - stopCondition: When to stop polling.
```
jerryjrchen, line 109: As obvious as it might sound, I think it would help to explicitly note that `nil` maps to the `false` stop condition. That could either go here or maybe in the docs for `PollingStopCondition`. Based on my reading of the docs for `firstPass`, I assumed it would only be used with `Bool` and not with optional return types as well:

```
  /// Evaluates the expression until the first time it returns true.
  /// If it does not pass once by the time the timeout is reached, then a
  /// failure will be reported.
  case firstPass
```

jerryjrchen:
The link in the description links to the old path for the proposal doc. I think it should be updated to point to https://github.com/younata/swift-evolution/blob/younata/testing-polling-expectations/proposals/testing/NNNN-polling-confirmations.md instead

(I think I have edit permissions to make this description change but wasn't sure if that would be polite to do 😅 )

Sources/Testing/Confirmations/Polling.swift
```diff
@@ -0,0 +1,462 @@
...
+    precondition(duration > interval)
+
+    let iterations = Int(exactly:
+        max(duration.seconds() / interval.seconds(), 1).rounded()
+    ) ?? Int.max
+    // if Int(exactly:) returns nil, then that generally means the value is too
+    // large. In which case, we should fall back to Int.max.
+
+    let failureReason: PollingFailedError.Reason
+    switch await poll(iterations: iterations, expression: body) {
```
jerryjrchen: Should the `isolation` parameter for this method also get forwarded to `poll`?

I'm not too familiar with actor isolated parameters so it's totally possible I'm missing something here.

Sources/Testing/Confirmations/Polling.swift
```diff
@@ -0,0 +1,462 @@
...
+  pollingDuration: Duration.seconds(1),
+  pollingInterval: Duration.milliseconds(1)
+)
+
+/// A type defining when to stop polling.
+/// This also determines what happens if the duration elapses during polling.
+@_spi(Experimental)
+public enum PollingStopCondition: Sendable, Equatable, Codable {
+  /// Evaluates the expression until the first time it returns true.
+  /// If it does not pass once by the time the timeout is reached, then a
```
jerryjrchen: Nit: would it make sense to say `duration` instead of `timeout` here? I don't see "timeout" appearing elsewhere in the polling confirmation API or documentation.

Sources/Testing/Confirmations/Polling.swift
```diff
@@ -0,0 +1,462 @@
...
+  fileprivate func process<R>(
+    expressionResult result: R?,
+    wasLastPollingAttempt: Bool
+  ) -> PollingProcessResult<R> {
+    switch self {
+    case .firstPass:
+      if let result {
+        return .succeeded(result)
+      } else {
+        return .continuePolling
```
jerryjrchen, line 229: I realised too late my note below is overly pedantic and I actually think this function is fairly clear as-is. I kinda already typed it out and figured I might as well share it though :)

I'd find it a bit clearer to check for the "timeout" scenario here compared to the call site in `poll`:

```suggestion
        return wasLastPollingAttempt ? .failed : .continuePolling
```

Or alternatively an early check on the last polling attempt which would simplify the switch:
```swift
    guard !wasLastPollingAttempt else {
      return result.map { .succeeded($0) } ?? .failed
    }

    switch self {
    // ...
```

Sources/Testing/Confirmations/Polling.swift
```diff
@@ -0,0 +1,462 @@
...
+    return issue
+  }
+}
+
+/// Poll expression within the duration based on the given stop condition
+///
+/// - Parameters:
+///   - comment: A user-specified comment describing this confirmation.
+///   - stopCondition: When to stop polling.
+///   - duration: The expected length of time to continue polling for.
```
jerryjrchen: Would it also be worth mentioning that time to run the body isn't accounted for in `duration`? I can see someone unfamiliar with the full details of polling confirmations writing something like:
```swift
      try await confirmation(until: .firstPass, within: .seconds(10)) {
        tenSecondLongOperation()
      }
```

And then being surprised it runs much longer than 10 seconds total.

Tests/TestingTests/PollingTests.swift
jerryjrchen, line 1: Thanks for the awesome comprehensive tests!

Tests/TestingTests/PollingTests.swift
```diff
@@ -0,0 +1,556 @@
...
+          within: .milliseconds(10)
+        )
+      )
+      func testUsesTestConfigurationOverSuiteConfiguration() async throws  {
+        let incrementor = Incrementor()
+        try await confirmation(until: stop, pollingEvery: .milliseconds(1)) {
+          await incrementor.increment() != 0
+        }
+        let count = await incrementor.count
+        #expect(await count == 10)
```
jerryjrchen: I think this `await` is unnecessary:
```suggestion
        #expect(count == 10)
```

Tests/TestingTests/PollingTests.swift
```diff
@@ -0,0 +1,556 @@
...
+          until: stop,
+          within: .milliseconds(50),
+          pollingEvery: .milliseconds(1)
+        ) {
+          await incrementor.increment() != 0
+        }
+        #expect(await incrementor.count == 50)
+      }
+
+#if !SWT_NO_EXIT_TESTS
```
jerryjrchen, line 439: Nice use of exit tests!

jerryjrchen (approved):
Overall this looks great to me, and I mostly had a bunch of nitpicks about the docs. I'll caveat by saying I've only read through the proposal recently and skimmed the pitch forum post, so I hope I didn't rehash a bunch of previous questions :)

Sources/Testing/Confirmations/Polling.swift
younata: Thanks for bringing this up!

I'll update the documentation & add some tests mentioning this.

I don't see any real harm in disallowing the case where `duration == interval`. It is not very useful - at least not at the moment - but to your point, it is a possible use case someone might have.

Sources/Testing/Confirmations/Polling.swift
younata: That's a good point. I'm going to update the docs for `PollingStopCondition` to be about pass/fail, and then have the documentation for both forms of `confirmation` specifically mention what is considered a pass/fail.

I'm also adding a test that explicitly documents the case where you return `false` when you pass in a closure returning `Optional<Bool>`.

Tests/TestingTests/PollingTests.swift
younata: I'm going to instead inline the `await incrementor.count` back into this and in the other places.

The reason this was separated out was to make debugging the test failure easier - the failure message when `incrementor.count` wasn't equal to 10 did not mention what the then-current value of `incrementor` was. But that's not necessary any longer.

Sources/Testing/Confirmations/Polling.swift
younata: I appreciate the feedback, and while pedantic, I do agree that makes it easier to read.

I think this is a good suggestion for clarity purposes. It does make the last line in `poll` redundant (`return .failed` after the for loop exits), but necessary to satisfy the compiler. I'll add a comment to the end of that method explaining this.

Sources/Testing/Issues/KnownIssue.swift
```diff
@@ -231,13 +231,13 @@ public func withKnownIssue(
   guard precondition() else {
     return try body()
   }
-  let scope = KnownIssueScope(issueMatcher: issueMatcher, context: Issue.KnownIssueContext(comment: comment))
+  let scope = IssueCapturingScope(captureSilently: false, issueMatcher: issueMatcher, context: Issue.KnownIssueContext(comment: comment))
```
younata, line 234: This is non-ideal. Ideally, `withKnownIssues` would call `captureIssues` and then operate on the number of issues returned, but I hit a few snags doing that which I haven't cleaned up yet.

Sources/Testing/Issues/IssueCapturing.swift
```diff
@@ -0,0 +1,165 @@
+//
+//  IssueCapturing.swift
+//  swift-testing
+//
+//  Created by Rachel Brindle on 8/24/26.
+//
+
```
younata, line 7: Replace this with the actual licensing header.

PR #1116, opened 2025-05-15
Include total number of suites in "run ended" console message

stmontgomery:
This enhances the console output message shown when a test run finishes by including the total number of suites which ran or skipped, after the total number of test functions. Example:

```
✔ Test run with 456 tests in 62 suites passed after 3.389 seconds.
```

The data was already being collected to support this, in a property named `suiteCount`, but it was not being used anywhere. So this PR adopts that property to augment the current "test run ended" message.

### Checklist:


stmontgomery:
@swift-ci please test

briancroom:
Nice, I like it!

briancroom (approved):

PR #1117, opened 2025-05-16
[CMake] Fix usage of lowercase or in condition

drodriguez:
CMake is a strange animal and some things are case sensitive, while others aren't. It seems that `if(<condition>)` logical operator `OR` is case sensitive, and `or` was not cutting it. At least for CMake 3.26.4.

If some CMake version allows `or`, it should also allow `OR`, so this change should not be a problem for those versions.

Most of the time these pieces are not actually hit, because the build system sets `SwiftTesting_MACROS` to `NO`, so this is skipped, which might explain why it has not been a problem.


drodriguez:
@michael-yuji can you check on your setup if this is a problem? I think a normal usage of `build-script` should not hit this line, which might explain why it was not a problem for you.

drodriguez:
@swift-ci test

grynspan (approved):

stmontgomery (approved):

stmontgomery (commented):
LGTM. Let me know if you need me to merge

michael-yuji (approved):
Lgtm too

PR #1120, opened 2025-05-20
Add a kind to Issue.record to record various kinds a of issues

suzannaratcliff:
Add a kind to Issue.record to record various kinds a of issues

### Motivation:

I would like to be able to specify what kind of issue I am recording when I record an issue.

### Modifications:

Add an optional parameter to `Issue.record` that takes in an issue `Kind`.

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -76,7 +76,6 @@ extension Issue {
   ///   - severity: The severity of the issue.
   ///   - sourceLocation: The source location to which the issue should be
   ///     attributed.
-  ///   - kind: The kind of the issue.
   ///
   /// - Returns: The issue that was recorded.
   ///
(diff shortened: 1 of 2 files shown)


Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -86,10 +87,11 @@ extension Issue {
   @discardableResult public static func record(
     _ comment: Comment? = nil,
     severity: Severity,
-    sourceLocation: SourceLocation = #_sourceLocation
+    sourceLocation: SourceLocation = #_sourceLocation,
+    kind: Kind = .unconditional
```
stmontgomery, line 91: I recall this is the overload of `record()` you added previously, and its primary difference from the existing, public API is that it includes a `severity:` parameter. This one is currently `@_spi(Experimental)`, and I think that SPI level makes sense given your active [pitch](https://forums.swift.org/t/pitch-test-issue-warnings/79285) which seeks to elevate this overload to non-SPI and replace the older one entirely. If that proposal advances and is accepted, I think we will want to expose this `record()` method as it's currently written and proposed, _without_ including a `kind:` parameter in its signature.

Given those plans, I think it may make more sense in this PR to expose an entirely new overload of the `record()` function which includes the `kind:` parameter, and leave the existing overload alone. And that overload can have **two** SPI groups—`@_spi(Experimental)` and `@_spi(ForToolsIntegrationOnly)`—to reflect that it's intended for use by integrated tools.

I would also suggest placing `kind:` as the first parameter in that new overload, and giving its comment parameter a label (instead of `_ comment: Comment?`).

stmontgomery:
@swift-ci please test

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -92,6 +92,34 @@ extension Issue {
...
+  ///
+  /// - Returns: The issue that was recorded.
+  ///
+  /// Use this function if, while running a test, an issue occurs that cannot be
+  /// represented as an expectation (using the ``expect(_:_:sourceLocation:)``
+  /// or ``require(_:_:sourceLocation:)-5l63q`` macros.)
+  @_spi(Experimental)
+  @_spi(ForToolsIntegrationOnly)
+  @discardableResult public static func record(
+    kind: Kind = .unconditional,
```
stmontgomery, line 113: Ok now that this is a distinct overload, I think you'll need to make `kind:` _not_ have a default value at all, because all parameters have a default value now and `Issue.record()` (zero args) may be ambiguous with the public API
```suggestion
    kind: Kind,
```

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -92,6 +92,34 @@ extension Issue {
...
   }
+  
+  /// Record an issue when a running test fails unexpectedly.
+  ///
+  /// - Parameters:
+  ///   - comment: A comment describing the expectation.
+  ///   - severity: The severity of the issue.
+  ///   - sourceLocation: The source location to which the issue should be
+  ///     attributed.
+  ///   - kind: The kind of the issue.
```
stmontgomery, line 103: `kind` can move up to the top, now

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -92,6 +92,34 @@ extension Issue {
     let issue = Issue(kind: .unconditional, severity: severity, comments: Array(comment), sourceContext: sourceContext)
     return issue.record()
   }
+  
+  /// Record an issue when a running test fails unexpectedly.
```
stmontgomery, line 96:
```suggestion
  /// Record an issue with a specific kind.
```

Sources/Testing/Issues/Issue+Recording.swift
grynspan: It should never be possible to create an issue with an arbitrary `kind` field via API or non-experimental SPI, so we need to nix that argument for any public interfaces.

Sources/Testing/Issues/Issue+Recording.swift
grynspan: It should be deleted.

grynspan (changes requested):
It is intentional that we don't provide API or SPI to create an `Issue` with an arbitrary `kind` field.

PR #1121, opened 2025-05-20
Refinements to IssueHandlingTrait SPI

stmontgomery:
A few refinements to `IssueHandlingTrait`, which is still SPI.

### Motivation:

Polish this SPI in anticipation of posting a pitch to promote it to public API soon.

### Modifications:

- Expose a `handleIssue(_:)` instance method to allow more easily composing multiple issue handling traits and calling their underlying handler closure. This is conceptually similar to what was done for [ST-0010: Public API to evaluate ConditionTrait](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0010-evaluate-condition.md).
- Refine the names of private decls.
- Add `- Returns:` in DocC for places it's missing.

### Checklist:

### DIFF
--- a/Sources/Testing/Traits/IssueHandlingTrait.swift
+++ b/Sources/Testing/Traits/IssueHandlingTrait.swift
@@ -26,20 +26,31 @@
 /// - ``Trait/filterIssues(_:)``
 @_spi(Experimental)
 public struct IssueHandlingTrait: TestTrait, SuiteTrait {
-  /// A function which transforms an issue and returns an optional replacement.
+  /// A function which handles an issue and returns an optional replacement.
   ///
   /// - Parameters:
-  ///   - issue: The issue to transform.
+  ///   - issue: The issue to handle.
   ///
   /// - Returns: An issue to replace `issue`, or else `nil` if the issue should
   ///   not be recorded.
-  fileprivate typealias Transformer = @Sendable (_ issue: Issue) -> Issue?
+  fileprivate typealias Handler = @Sendable (_ issue: Issue) -> Issue?
 
-  /// This trait's transformer function.
-  private var _transformer: Transformer
+  /// This trait's handler function.
+  private var _handler: Handler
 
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/Traits/IssueHandlingTrait.swift
```diff
@@ -131,7 +144,7 @@ extension Trait where Self == IssueHandlingTrait {
   /// record new issues, although they will only be handled by issue handling
   /// traits which precede this trait or were inherited from a containing suite.
   public static func transformIssues(_ transformer: @escaping @Sendable (Issue) -> Issue?) -> Self {
-    Self(transformer: transformer)
+    Self(handler: transformer)
```
suzannaratcliff, line 147: Do you want this to be handler: transformer?

suzannaratcliff: I guess that is the name of the parameter so it makes sense

Sources/Testing/Traits/IssueHandlingTrait.swift
```diff
@@ -26,20 +26,31 @@
...
+  fileprivate init(handler: @escaping Handler) {
+    _handler = handler
+  }
+
+  /// Handle a specified issue.
+  ///
+  /// - Parameters:
+  ///   - issue: The issue to handle.
+  ///
+  /// - Returns: An issue to replace `issue`, or else `nil` if the issue should
```
suzannaratcliff, line 50: Is issue mutable here? So that someone can just return the same issue or a slightly modified one?

stmontgomery: The return value is what ends up getting used, and it can be the unmodified issue which was passed in, a modified issue instance, or `nil` to completely suppress the issue.

suzannaratcliff (approved):

PR #1123, opened 2025-05-22
Introduce 'MarkerTrait' to unify the representation of boolean test attributes

stmontgomery:
This introduces a utility type named `MarkerTrait` which acts as a generic "marker" trait representing a boolean attribute of a test, and uses it to replace the `.hidden` trait and `Test.isSynthesized` property.

### Motivation:

Currently the testing library has an internal-only trait named `.hidden`, of type `HiddenTrait`, which is used in the project's own unit tests to represent fixture/example tests and suites. Separately, it also has a stored, boolean property on `Test` named `isSynthesized` which is meant for debugging purposes to keep track of suites which were synthesized during planning and didn't have a `@Suite` attribute.

Conceptually, both of these can be thought of as boolean attributes of a test/suite. A given `Test` instance is either hidden or not, and either synthesized or not. There's an opportunity to simplify the modeling by representing them both as a trait, which would make adding future boolean test attributes more straightforward. Plus, for `isSynthesized` specifically, doing this would eliminate a stored property on every `Test` instance.

### Modifications:

- Add `MarkerTrait` type and use it reimplement `.hidden`.
- Remove the `Test.isSynthesized` stored property and replace it with a new `.synthesized` marker trait.
- Only include the synthesized trait in debug builds (guarded by `#if DEBUG`) to avoid needless overhead in release builds where this information isn't used.
- Add a `Test.containsTrait()` utility function for checking whether a test contains some (equatable) trait.

### Checklist:


stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

grynspan:
I see where you're going with this, but it may make it harder to support Embedded Swift by relying even more on existentials.

There is also a performance cost to this change that we should be careful to measure.

PR #1130, opened 2025-05-25
Infer the types of function/closure arguments when captured by an exit test.

grynspan:
This PR adds the ability to infer the type of a parameter of a function or closure that encloses an exit test. For example, `x` here:

```swift
func f(x: Int) async {
  await #expect(processExitsWith: .failure) { [x] in
    ...
  }
}
```

This inference still fails if a parameter is shadowed by a variable with an incompatible type; we still need something like `decltype()` to solve for such cases. We emit a custom diagnostic of the form "🛑 Type of captured value 'x' is ambiguous" if the inferred type of the captured value doesn't match what the compiler thinks it is (see #1146). Still, being able to capture `@Test` function arguments with minimal ceremony is helpful:

```swift
@Test(arguments: 0 ..< 100) func f(i: Int) async {
  await #expect(exitsWith: .failure) { [i] in
    ...
  }
}
```

Also type inference for literals because "why not?"

> [!NOTE]
> Exit test value capture remains an experimental feature.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/TestingMacros/ConditionMacro.swift
```diff
@@ -611,6 +613,55 @@ extension ExitTestConditionMacro {
...
+       let captureClause = closureExpr.signature?.capture,
+       !captureClause.items.isEmpty {
+      // Disallow capture lists if the experimental feature is not enabled.
+      if !ExitTestExpectMacro.isValueCapturingEnabled {
+        diagnostics.append(.captureClauseUnsupported(captureClause, in: closureExpr, inExitTest: macro))
+      }
+    }
+
+    // Disallow exit tests in generic types and functions as they cannot be
+    // correctly expanded due to the use of a nested type with static members.
```
stmontgomery, line 643: Do we need to enforce this restriction if there is no capture list in the exit test body closure?

Sources/TestingMacros/Support/ClosureCaptureListParsing.swift
```diff
@@ -76,18 +77,71 @@ struct CapturedValueInfo {
...
+  /// - Returns: The Swift type of first parameter found whose name matches, or
+  ///   `nil` if none was found. The lexical context is searched in the order
+  ///   provided which, by default, starts with the innermost scope.
+  private static func _findTypeOfParameter(named parameterName: TokenSyntax, in lexicalContext: [Syntax]) -> TypeSyntax? {
+    for lexicalContext in lexicalContext {
+      var parameterType: TypeSyntax?
+      if let functionDecl = lexicalContext.as(FunctionDeclSyntax.self) {
+        parameterType = functionDecl.signature.parameterClause.parameters
+          .first { ($0.secondName ?? $0.firstName).tokenKind == parameterName.tokenKind }
+          .map(\.type)
```
stmontgomery, line 125: Does this need to contend with keywords like `borrowing` on the parameter type if present, or does it already? And could that be validated in a test either way?

Sources/TestingMacros/ConditionMacro.swift
grynspan: Yes. The emitted code doesn't work either way and this is a better diagnostic than we get by default.

Sources/TestingMacros/Support/ClosureCaptureListParsing.swift
grynspan: Potentially, but my hope is that we can get a `decltype()` equivalent before I need to care about that much nuance.

stmontgomery (approved):

Sources/TestingMacros/Support/ClosureCaptureListParsing.swift
grynspan: (Will revisit in a separate PR for the sake of clarity.)

PR #1134, opened 2025-06-02
Acknowledge unsafe API usages in code expanded from testing library macros

stmontgomery:
Acknowledge unsafe API usages from various testing library macros such as `@Test`, `@Suite`, and `#expect(processExitsWith:)` which are revealed in modules which enable the new opt-in strict memory safety feature in Swift 6.2.

### Motivation:

This fix allows clients of the testing library to enable [SE-0458: Opt-in Strict Memory Safety Checking](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md) if they wish and avoid diagnostics from the testing library macros in their modules. These warnings generally looked like this, before this fix:

```
⚠️ @__swiftmacro_22MemorySafeTestingTests19exampleTestFunction33_F2EA1AA3013574E5644E5A4339F05086LL0F0fMp_.swift:23:14: warning: expression uses unsafe constructs but is not marked with 'unsafe'
  Testing.Test.__store($s22MemorySafeTestingTests19exampleTestFunction33_F2EA1AA3013574E5644E5A4339F05086LL0F0fMp_25generator1e3470c498e8fe35fMu_, into: outValue, asTypeAt: type)
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### Modifications:

- Add test file to reproduce the new diagnostics. It's in a new module which opts-in to strict memory safety, and marked as a dependency of `TestingTests`.
- Add `unsafe` keyword in the appropriate places in our macros to acknowledge the existing unsafe usages.

### Checklist:

Fixes: rdar://151238560

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -127,8 +127,17 @@ let package = Package(
         "Testing",
         "_Testing_CoreGraphics",
         "_Testing_Foundation",
+        "MemorySafeTestingTests",
       ],
       swiftSettings: .packageSettings
+    ),
+    .target(
+      name: "MemorySafeTestingTests",
+      dependencies: [
+        "Testing",
+      ],
+      path: "Tests/_MemorySafeTestingTests",
+      swiftSettings: .packageSettings + .strictMemorySafety
     ),
 
     .macro(
@@ -354,6 +363,14 @@ extension Array where Element == PackageDescription.SwiftSetting {
     }
 
     return result
+  }
+
+  /// Settings necessary to enable Strict Memory Safety, introduced in
+  /// [SE-0458: Opt-in Strict Memory Safety Checking](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md#swiftpm-integration).
+  static var strictMemorySafety: Self {
+    // FIXME: Adopt official `.strictMemorySafety()` condition once the minimum
+    // supported toolchain is 6.2.
+    [.unsafeFlags(["-strict-memory-safety"])]
   }
 }
 
--- a/Sources/TestingMacros/ConditionMacro.swift
+++ b/Sources/TestingMacros/ConditionMacro.swift
@@ -499,7 +499,7 @@ extension ExitTestConditionMacro {
       recordDecl = """
       enum \(legacyEnumName): Testing.__TestContentRecordContainer {
         nonisolated static var __testContentRecord: Testing.__TestContentRecord {
-          \(enumName).testContentRecord
+          unsafe \(enumName).testContentRecord
         }
       }
       """
@@ -510,7 +510,7 @@ extension ExitTestConditionMacro {
         @available(*, deprecated, message: "This type is an implementation detail of the testing library. Do not use it directly.")
         enum \(enumName) {
           private nonisolated static let accessor: Testing.__TestContentRecordAccessor = { outValue, type, hint, _ in
-            Testing.ExitTest.__store(
+            unsafe Testing.ExitTest.__store(
               \(idExpr),
               \(bodyThunkName),
               into: outValue,
--- a/Sources/TestingMacros/TestDeclarationMacro.swift
+++ b/Sources/TestingMacros/TestDeclarationMacro.swift
@@ -475,7 +475,7 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
       """
       @available(*, deprecated, message: "This property is an implementation detail of the testing library. Do not use it directly.")
       private \(staticKeyword(for: typeName)) nonisolated let \(accessorName): Testing.__TestContentRecordAccessor = { outValue, type, _, _ in
-        Testing.Test.__store(\(generatorName), into: outValue, asTypeAt: type)
+        unsafe Testing.Test.__store(\(generatorName), into: outValue, asTypeAt: type)
       }
       """
     )
@@ -499,7 +499,7 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
       @available(*, deprecated, message: "This type is an implementation detail of the testing library. Do not use it directly.")
       enum \(enumName): Testing.__TestContentRecordContainer {
         nonisolated static var __testContentRecord: Testing.__TestContentRecord {
-          \(testContentRecordName)
+          unsafe \(testContentRecordName)
         }
       }
       """
--- a/Tests/_MemorySafeTestingTests/MemorySafeTestDecls.swift
+++ b/Tests/_MemorySafeTestingTests/MemorySafeTestDecls.swift
@@ -1,0 +1,27 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+@testable import Testing
+
+#if !hasFeature(StrictMemorySafety)
+#error("This file requires strict memory safety to be enabled")
+#endif
+
+@Test(.hidden)
+func exampleTestFunction() {}
+
+@Suite(.hidden)
+struct ExampleSuite {
+  @Test func example() {}
+}
+
+func exampleExitTest() async {
+  await #expect(processExitsWith: .success) {}
+}
--- a/Sources/TestingMacros/SuiteDeclarationMacro.swift
+++ b/Sources/TestingMacros/SuiteDeclarationMacro.swift
@@ -150,7 +150,7 @@ public struct SuiteDeclarationMacro: MemberMacro, PeerMacro, Sendable {
       """
       @available(*, deprecated, message: "This property is an implementation detail of the testing library. Do not use it directly.")
       private nonisolated static let \(accessorName): Testing.__TestContentRecordAccessor = { outValue, type, _, _ in
-        Testing.Test.__store(\(generatorName), into: outValue, asTypeAt: type)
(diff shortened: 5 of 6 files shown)


stmontgomery:
@swift-ci please test

stmontgomery:
This fails to build when using a pre-6.2 toolchain because those toolchains don't recognize the `unsafe` keyword. (Unlike the `@unsafe` _attribute_ which 6.1 toolchains are aware of.) I'll need to figure out some way to conditionalize this — hopefully without resorting to `#if` checks in the macro expansion code itself.

Package.swift
```diff
@@ -355,6 +364,14 @@ extension Array where Element == PackageDescription.SwiftSetting {
 
     return result
   }
+
+  /// Settings necessary to enable Strict Memory Safety, introduced in
+  /// [SE-0458: Opt-in Strict Memory Safety Checking](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md#swiftpm-integration).
+  static var strictMemorySafety: Self {
+    // FIXME: Adopt official `.strictMemorySafety()` condition once the minimum
+    // supported toolchain is 6.2.
+    [.unsafeFlags(["-strict-memory-safety"])]
```
grynspan, line 380: This flag prevents use of the package as a dependency. I am unsure if that's the case even if it's only applied to a test target.

Is there no upcoming feature flag you can enable? If not, consider adding a versioned package manifest file instead?

stmontgomery: Before posting this PR I locally confirmed that this flag is not problematic for clients, because it's limited to dependencies of test targets.

There is no upcoming feature because this is not "upcoming", it's intended to remain an opt-in language mode permanently. That's why there's a new SwiftPM package manifest API to enable it.

grynspan:
Can we conditionally mark the `__store` functions `@safe`?

stmontgomery:
> Can we conditionally mark the `__store` functions `@safe`?

Yes, I've gone ahead and done that, and removed the relevant `unsafe` keywords afterward.

stmontgomery:
> This fails to build when using a pre-6.2 toolchain because those toolchains don't recognize the `unsafe` keyword. (Unlike the `@unsafe` _attribute_ which 6.1 toolchains are aware of.) I'll need to figure out some way to conditionalize this — hopefully without resorting to `#if` checks in the macro expansion code itself.

OK I've adjusted the PR to only include the `unsafe` expression keyword when the macro plugin was built using a ≥ 6.2 compiler. This approach expands the scope of a similar fix I made earlier in #1093.

With this and some other adjustments, I am now able to build the package locally using a 6.1 toolchain successfully.

stmontgomery:
@swift-ci please test

Package.swift
```diff
@@ -127,9 +127,18 @@ let package = Package(
         "Testing",
         "_Testing_CoreGraphics",
         "_Testing_Foundation",
+        "MemorySafeTestingTests",
       ],
       swiftSettings: .packageSettings
     ),
+    .target(
```
grynspan, line 140:
```suggestion
    .testTarget(
```

Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
```diff
@@ -86,6 +86,17 @@ extension BidirectionalCollection<Syntax> {
 
 // MARK: - Inserting effect keywords/thunks
 
+/// Whether or not the `unsafe` expression keyword is supported.
+var isUnsafeKeywordSupported: Bool {
```
grynspan, line 90: Ah, good idea.

Sources/TestingMacros/Support/EffectfulExpressionHandling.swift
```diff
@@ -127,12 +138,7 @@ func applyEffectfulKeywords(_ effectfulKeywords: Set<Keyword>, to expr: some Exp
   let needAwait = effectfulKeywords.contains(.await) && !expr.is(AwaitExprSyntax.self)
   let needTry = effectfulKeywords.contains(.try) && !expr.is(TryExprSyntax.self)
 
-  // The 'unsafe' keyword was introduced in 6.2 as part of SE-0458. Older
-  // toolchains are not aware of it, so avoid emitting expressions involving
-  // that keyword when the macro has been built using an older toolchain.
-#if compiler(>=6.2)
-  let needUnsafe = effectfulKeywords.contains(.unsafe) && !expr.is(UnsafeExprSyntax.self)
-#endif
+  let needUnsafe = isUnsafeKeywordSupported && effectfulKeywords.contains(.unsafe) && !expr.is(UnsafeExprSyntax.self)
```
grynspan, line 141: _Perfetto!_

Sources/TestingMacros/ConditionMacro.swift
```diff
@@ -496,10 +496,11 @@ extension ExitTestConditionMacro {
       var recordDecl: DeclSyntax?
 #if !SWT_NO_LEGACY_TEST_DISCOVERY
       let legacyEnumName = context.makeUniqueName("__🟡$")
+      let unsafeKeyword = isUnsafeKeywordSupported ? TokenSyntax.keyword(.unsafe) : nil
       recordDecl = """
       enum \(legacyEnumName): Testing.__TestContentRecordContainer {
         nonisolated static var __testContentRecord: Testing.__TestContentRecord {
-          \(enumName).testContentRecord
+          \(unsafeKeyword) \(enumName).testContentRecord
```
grynspan: Minor nitpick, but this introduces a space before `\(enumName).testContentRecord` in the resulting expansion when `unsafe` is unsupported, which looks a bit weird. You could just make it the trailing trivia of the `TokenSyntax` instead and avoid it.

(Since this only affects 6.1, it's probably not worth worrying about, but thought I'd mention it.)

Tests/_MemorySafeTestingTests/MemorySafeTestDecls.swift
```diff
@@ -0,0 +1,31 @@
...
+#if !hasFeature(StrictMemorySafety)
+#error("This file requires strict memory safety to be enabled")
+#endif
+
+@Test(.hidden)
+func exampleTestFunction() {}
+
+@Suite(.hidden)
+struct ExampleSuite {
+  @Test func example() {}
```
grynspan, line 24:
```suggestion
  @Test(.hidden) func example() {}
```

stmontgomery: `.hidden` is inherited correctly nowadays and we don't need it redundantly on children.

Package.swift
```diff
@@ -355,6 +364,18 @@ extension Array where Element == PackageDescription.SwiftSetting {
 
     return result
   }
+
+  /// Settings necessary to enable Strict Memory Safety, introduced in
```
grynspan, line 374: Do we need a GitHub issue for follow-up here?

Package.swift
stmontgomery: This was intentional because I don't wish to incur the runtime overhead of a separate test target. I'll add an explanatory comment.

stmontgomery:
@swift-ci please test

Sources/TestingMacros/ConditionMacro.swift
stmontgomery: Fixed

Package.swift
stmontgomery: I don't think so, it's just a low priority clean up opportunity the next time we bump our minimum toolchain (to adopt the official `.strictMemorySafety()` API).

stmontgomery:
@swift-ci please test

Sources/TestingMacros/TestDeclarationMacro.swift
```diff
@@ -494,12 +494,13 @@ public struct TestDeclarationMacro: PeerMacro, Sendable {
 #if !SWT_NO_LEGACY_TEST_DISCOVERY
     // Emit a type that contains a reference to the test content record.
     let enumName = context.makeUniqueName(thunking: functionDecl, withPrefix: "__🟡$")
+    let unsafeKeyword: TokenSyntax? = isUnsafeKeywordSupported ? .keyword(.unsafe).with(\.trailingTrivia, .space) : nil
```
grynspan:
```suggestion
    let unsafeKeyword: TokenSyntax? = isUnsafeKeywordSupported ? .keyword(.unsafe, trailingTrivia: .space) : nil
```

This should be more efficient, and these tokens are on a hot path. See https://github.com/swiftlang/swift-syntax/blob/6de157b65d57948c8904c3a326cbccfa261215ee/Sources/SwiftSyntax/generated/Tokens.swift#L230

P4

grynspan (approved):

Sources/TestingMacros/TestDeclarationMacro.swift
stmontgomery: Oh, and we might as well use that approach over in [this existing code](https://github.com/swiftlang/swift-testing/blob/d6111205ce07f173dbc2519ddaf4eec800a76b94/Sources/TestingMacros/Support/EffectfulExpressionHandling.swift#L156), too.

stmontgomery:
@swift-ci please test

Sources/TestingMacros/TestDeclarationMacro.swift
grynspan: Sure thing.

PR #1146, opened 2025-06-10
Improve the diagnostics for a bad exit test capture.

grynspan:
This PR improves the diagnostics presented at compile time when an exit test captures an unsupported value. For example, given the following (bad) exit test:

```swift
struct NonCodableValue {}

let x = NonCodableValue()
await #expect(processExitsWith: .success) { [x = x as NonCodableValue] in
  _ = x
}
```

We currently get diagnostics of the form:

> 🛑 Global function '__checkClosureCall(identifiedBy:encodingCapturedValues:processExitsWith:observing:performing:expression:comments:isRequired:isolation:sourceLocation:)' requires that 'NonCodableValue' conform to 'Decodable'
> 🛑 Global function '__checkClosureCall(identifiedBy:encodingCapturedValues:processExitsWith:observing:performing:expression:comments:isRequired:isolation:sourceLocation:)' requires that 'NonCodableValue' conform to 'Encodable'
> ⚠️ No 'async' operations occur within 'await' expression

None of which actually tell the developer (clearly) what's wrong. With this PR, we instead get:

> 🛑 Type of captured value 'x' must conform to 'Sendable' and 'Codable' (from macro '__capturedValue')

Much better! The diagnostic is attributed to the temporary file containing the expansion of `#expect()` rather than to the original source file, but I've opened [an issue](https://github.com/swiftlang/swift-syntax/issues/3085) against swift-syntax with a fix in mind. Even with the misattribution, this diagnostic is still an improvement, yeah?

> [!NOTE]
> Exit test value capture remains an experimental feature.

### Checklist:

### DIFF
--- a/Sources/Testing/Expectations/Expectation+Macro.swift
+++ b/Sources/Testing/Expectations/Expectation+Macro.swift
@@ -572,3 +572,37 @@ public macro require(
   sourceLocation: SourceLocation = #_sourceLocation,
   performing expression: @escaping @Sendable @convention(thin) () async throws -> Void
 ) -> ExitTest.Result = #externalMacro(module: "TestingMacros", type: "ExitTestRequireMacro")
+
+/// Capture a sendable and codable value to pass to an exit test.
+///
+/// - Parameters:
+///   - value: The captured value.
+///   - name: The name of the capture list item corresponding to `value`.
+///
+/// - Returns: `value` verbatim.
+///
+/// - Warning: This macro is used to implement the `#expect(processExitsWith:)`
+///   macro. Do not use it directly.
+@freestanding(expression)
+public macro __capturedValue<T>(
+  _ value: T,
+  _ name: String
+) -> T = #externalMacro(module: "TestingMacros", type: "ExitTestCapturedValueMacro") where T: Sendable & Codable
+
+/// Emit a compile-time diagnostic when an unsupported value is captured by an
+/// exit test.
+///
+/// - Parameters:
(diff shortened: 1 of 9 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/Expectations/Expectation+Macro.swift
```diff
@@ -572,3 +572,37 @@ public macro require(
...
+/// - Returns: The result of a call to `fatalError()`. `value` is discarded at
+///   compile time.
+///
+/// - Warning: This macro is used to implement the `#expect(processExitsWith:)`
+///   macro. Do not use it directly.
+@freestanding(expression)
+public macro __capturedValue<T>(
+  _ value: borrowing T,
+  _ name: String
+) -> Never = #externalMacro(module: "TestingMacros", type: "ExitTestBadCapturedValueMacro") where T: ~Copyable & ~Escapable
```
briancroom, line 608: Making sure I remember how this works correctly.. in this context, the `where T: ~Copyable & ~Escapable` means "`T` is allowed to be non-copyable or non-escapable", right?

briancroom (approved):

Sources/Testing/Expectations/Expectation+Macro.swift
grynspan: Correct. Widest possible net.

PR #1150, opened 2025-06-11
Add an unreachable marker function.

grynspan:
This PR adds a wrapper around `__builtin_unreachable()` (`Builtin.unreachable()` when building the Swift standard library) that we can use in place of `fatalError()`. The benefit is that the generated code size for unreachable paths is significantly reduced. For example, given the following function compiled with `-O`:

```swift
@available(*, unavailable) func f() {
  fatalError("Unreachable")
}
```

The compiler currently produces:

```asm
sub    sp, sp, #0x20
stp    x29, x30, [sp, #0x10]
add    x29, sp, #0x10
mov    w8, #0x1                  ; =1
str    w8, [sp, #0x8]
mov    w8, #0xc                  ; =12
str    x8, [sp]
adrp   x0, 0
add    x0, x0, #0x6a8            ; "Fatal error"
adrp   x5, 0
add    x5, x5, #0x690            ; "UnreachableTest/S.swift"
mov    x3, #0x6e55               ; =28245
movk   x3, #0x6572, lsl #16
movk   x3, #0x6361, lsl #32
movk   x3, #0x6168, lsl #48
mov    x4, #0x6c62               ; =27746
movk   x4, #0x65, lsl #16
movk   x4, #0xeb00, lsl #48
mov    w1, #0xb                  ; =11
mov    w2, #0x2                  ; =2
mov    w6, #0x17                 ; =23
mov    w7, #0x2                  ; =2
bl     0x100000680               ; symbol stub for: Swift._assertionFailure(_: Swift.StaticString, _: Swift.String, file: Swift.StaticString, line: Swift.UInt, flags: Swift.UInt32) -> Swift.Never
brk    #0x1
```

But with this change:
```swift
@available(*, unavailable) func f() {
  swt_unreachable()
}
```

It instead compiles to simply:

```asm
brk    #0x1
```

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.CapturedValue.swift
+++ b/Sources/Testing/ExitTests/ExitTest.CapturedValue.swift
@@ -8,6 +8,8 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
+private import _TestingInternals
+
 @_spi(Experimental) @_spi(ForToolsIntegrationOnly)
 #if SWT_NO_EXIT_TESTS
 @available(*, unavailable, message: "Exit tests are not available on this platform.")
@@ -82,7 +84,7 @@ extension ExitTest {
         }
         return nil
 #else
-        fatalError("Unsupported")
+        swt_unreachable()
 #endif
       }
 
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci test

briancroom (approved):
Cool, makes sense!

PR #1152, opened 2025-06-13
Diagnose when we incorrectly infer the type of a capture list item in an exit test.

grynspan:
Follow-up to #1130, split out for clarity.

This PR adds a custom diagnostic at compile time if we incorrectly infer the type of a captured function argument or `self` in an exit test. For example:

```swift
func f(_ x: Int) async {
  let x = String(x) // local type of 'x' is String, not Int
  await #expect(processExitsWith: ...) { [x] in
    ...
  }
}
```

This improves our feedback to the developer when we encounter a pattern like that. The developer will now see:

> 🛑 Type of captured value 'x' is ambiguous

### Checklist:


Sources/TestingMacros/ExitTestCapturedValueMacro.swift
```diff
@@ -52,3 +53,26 @@ public struct ExitTestBadCapturedValueMacro: ExpressionMacro, Sendable {
...
+    of macro: some FreestandingMacroExpansionSyntax,
+    in context: some MacroExpansionContext
+  ) throws -> ExprSyntax {
+    let arguments = Array(macro.arguments)
+    let expr = arguments[0].expression
+    let nameExpr = arguments[1].expression.cast(StringLiteralExprSyntax.self)
+
+    // Diagnose that the type of 'expr' is invalid.
+    let name = nameExpr.representedLiteralValue ?? expr.trimmedDescription
+    let capture = ClosureCaptureSyntax(name: .identifier(name))
```
grynspan, line 73: This fake `ClosureCaptureSyntax` instance is created because we can't pass such a thing as an expression and we need to use a string instead. It's a temporary abstraction until https://github.com/swiftlang/swift-syntax/issues/3085 is resolved.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
Preemptively merging before all CI is complete as it passed previously and is passing now on Linux. Windows and Mac CI are stuck waiting.

PR #1153, opened 2025-06-13
Remove `@convention(thin)` from exit test macro declarations.

grynspan:
`@convention(thin)` is not fully plumbed through the compiler and should not be in our API surface, but also because of the experimental value capturing feature the closures may be thick anyway.

See also https://github.com/swiftlang/swift-evolution/pull/2884.

### Checklist:


grynspan:
@swift-ci test

Sources/Testing/Expectations/Expectation+Macro.swift
```diff
@@ -525,7 +525,7 @@ public macro expect(
   observing observedValues: [any PartialKeyPath<ExitTest.Result> & Sendable] = [],
   _ comment: @autoclosure () -> Comment? = nil,
   sourceLocation: SourceLocation = #_sourceLocation,
-  performing expression: @escaping @Sendable @convention(thin) () async throws -> Void
+  performing expression: @escaping @Sendable () async throws -> Void
```
stmontgomery, line 528: I see the check functions are changed from `@convention(thin)` to `@convention(c)`, is there a reason you chose not to match that in the macro decls?

grynspan: The macro declarations do not use `@convention(c)` intentionally. That would preclude `async throws` and would preclude capture lists. The closures in the expanded form aren't these closures, but are empty closures that contain the exit test thunk/gunk we generate.

PR #840, opened 2024-12-02
[WIP] Rethink how we capture expectation conditions and their subexpressions.

grynspan:
This PR completely rewrites how we capture expectation conditions.

### Explanation

For example, given the following expectation:

```swift
#expect(x.f() == 123)
```

We currently detect that there is a binary operation and emit code that calls the binary operator as a closure and passes the left-hand value and right-hand value, then checks that the result of the operation is `true`.

This is sufficient for simpler expressions like that one, but more complex ones (including any that involve `try` or `await` keywords) cannot be expanded correctly. With this PR, such expressions _can_ generally be expanded correctly.

The change involves rewriting the macro condition as a closure to which is passed a local, mutable "context" value. Subexpressions of the condition expression are then rewritten by walking the syntax tree of the expression (using typical swift-syntax API) and replacing them with calls into the context value that pass in the value and related state.

If the expectation ultimately fails, the collected data is transformed into an instance of the SPI type `Expression` that contains the source code of the expression and interesting subexpressions as well as the runtime values of those subexpressions.

Nodes in the syntax tree are identified by a unique ID which is composed of the swift-syntax ID for that node as well as all its parent nodes in a compact bitmask format. These IDs can be transformed into graph/trie key paths when expression/subexpression relationships need to be reconstructed on failure, meaning that a single rewritten node doesn't otherwise need to know its "place" in the overall expression.

### Examples

As an example, this expectation…

```swift
#expect(g() > 500)
```

… previously expanded to…

```swift
Testing.__checkBinaryOperation(
  g(),
  { $0 > $1() },
  500,
  expression: .__fromBinaryOperation(
    .__fromSyntaxNode("g()"),
    ">",
    .__fromSyntaxNode("500")
  ),
  comments: [],
  isRequired: false,
  sourceLocation: Testing.SourceLocation.__here()
).__expected()
```

… but will now expand to:

```swift
Testing.__checkCondition(
  { (__ec: inout Testing.__ExpectationContext) -> Swift.Bool in
    __ec(__ec(g(),0x2) > 500,0x0)
  },
  sourceCode: [
    0x0: "g() > 500",
    0x2: "g()"
  ],
  comments: [],
  isRequired: false,
  sourceLocation: Testing.SourceLocation.__here()
).__expected()
```

More interestingly, an expression with side effects or complex nested operations can also be translated. For example, this throwing expression…

```swift
#expect((try g() > 500) && true)
```

… was…

```swift
Testing.__checkValue(
  (try g() > 500) && true,
[...]

PR #1154, opened 2025-06-13
Support property accesses on ~Escapable types

natecook1000:
This adds support for having nonescapable types on the LHS of property accesses in the `#expect(...)` and `#require(...)` macros. This includes some additional changes that could lead to supporting a nonescapable result type in `#require(...)`, but that may need more lifetime support in the language.

### Motivation:

In the BinaryParsing library, I have a nonescapable `ParserSpan` type. In the tests, I need to write `#expect(span.count == 0)` instead of the more fluent `#expect(span.isEmpty)`.

### Modifications:

I've enabled lifetime annotations in the `Testing` and test targets, and then added overloads of `__checkPropertyAccess` and `__checkValue` that allow `~Escapable` parameters.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -116,7 +116,8 @@ let package = Package(
       ],
       exclude: ["CMakeLists.txt", "Testing.swiftcrossimport"],
       cxxSettings: .packageSettings,
-      swiftSettings: .packageSettings + .enableLibraryEvolution(),
+      swiftSettings: .packageSettings + .enableLibraryEvolution()
+        + [.enableExperimentalFeature("LifetimeDependence")],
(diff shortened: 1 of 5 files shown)


stmontgomery:
Quick note up front that currently we expect our `main` branch to remain buildable using a 6.1 toolchain. (No further back than that, luckily.) So you may need `#if compiler(>=6.2)` guards in some places.

grynspan:
I asked Nate to look at branching from #840 instead of main here as the implementations are different and we wouldn't want to accidentally introduce something on main that was unsupportable under the new implementation.

PR #1155, opened 2025-06-13
Add a test for property access on ~Escapable types

natecook1000:
This tests using a ~Escapable type as the LHS of a property access inside the `#expect` and `#require` macros. A ~Escapable return value from `#require` is unsupported.

### Checklist:


stmontgomery:
@swift-ci please test

grynspan:
I believe this test is dependent on #840, right?

natecook1000:
Right! I believe it's aimed at that branch, not main. Should I direct it somewhere else?

grynspan:
Nope, just confirming the dependency is why the test failed. You could redirect the PR to merge to that branch, and then it would pass?

grynspan:
@swift-ci test

PR #1142, opened 2025-06-07
Update outdated Suite macro documentation

Wouter01:
Remove the doc stating that the `@Suite` macro only works on `Copyable` types.

### Motivation:

When reading through the docs, I was quite confused seeing it stated that `~Copyable` types aren't supported by the `@Suite` macro, whereas it does work when doing so. I had a look at it looks like that restriction has been lifted by #619, but the docs weren't updated. 

### Modifications:

Updated the docs

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1159, opened 2025-06-16
Revert "Update outdated Suite macro documentation"

grynspan:
Reverts swiftlang/swift-testing#1142

This change isn't quite accurate. While we do support `~Copyable` suite types, we don't (and for now, can't) support `~Escapable` suite types and the documentation should reflect that.


grynspan:
@Wouter01 FYI. Sorry!

grynspan:
Self-reviewing and merging without CI. This change only affects documentation.

Wouter01:
Oops, I didn't think about those, thanks for notifying me and updating the docs!

PR #1161, opened 2025-06-17
Suppress the complex expansion of expectations when we see effects in the lexical context.

grynspan:
This PR changes the behaviour of the testing library for expressions such as:

```swift
try #expect(a == b)
```

Currently, we don't expand that expression correctly because we can't tell where the `try` keyword should be applied. It sometimes expands and sometimes doesn't. This PR detects the presence of those keywords (with a recent-enough toolchain) and, if found, disables the fancy expansion in favour of a simpler one that is less likely to fail to compile.

More thorough support for effectful expressions in expectations is tracked by #840 which involves fully refactoring the implementation of the `#expect()` macro. See also #162 for some more context.

### Checklist:


grynspan:
@swift-ci test

suzannaratcliff (approved):

PR #1163, opened 2025-06-17
Improve type inference for exit test value captures.

grynspan:
This PR refactors the (new) type inference logic for exit test capture lists to use a syntax visitor, which allows for the types of more complex expressions to be inferred.

For example, previously the type of this capture would not be inferred:

```swift
[x = try await f() as Int]
```

Even though the type (`Int`) is clearly present, because the `AsExprSyntax` is nested in an `AwaitExprSyntax` and then a `TryExprSyntax`.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/ClosureCaptureListParsing.swift
+++ b/Sources/TestingMacros/Support/ClosureCaptureListParsing.swift
@@ -50,60 +50,175 @@ struct CapturedValueInfo {
       return
     }
 
-    // Potentially get the name of the type comprising the current lexical
-    // context (i.e. whatever `Self` is.)
-    lazy var lexicalContext = context.lexicalContext
-    lazy var typeNameOfLexicalContext = {
-      let lexicalContext = lexicalContext.drop { !$0.isProtocol((any DeclGroupSyntax).self) }
-      return context.type(ofLexicalContext: lexicalContext)
-    }()
-
-    if let initializer = capture.initializer {
-      // Found an initializer clause. Extract the expression it captures.
-      self.expression = removeParentheses(from: initializer.value) ?? initializer.value
-
-      // Find the 'as' clause so we can determine the type of the captured value.
-      if let asExpr = self.expression.as(AsExprSyntax.self) {
-        self.type = if asExpr.questionOrExclamationMark?.tokenKind == .postfixQuestionMark {
-          // If the caller is using as?, make the type optional.
-          TypeSyntax(OptionalTypeSyntax(wrappedType: asExpr.type.trimmed))
-        } else {
-          asExpr.type
-        }
-      } else if let selfExpr = self.expression.as(DeclReferenceExprSyntax.self),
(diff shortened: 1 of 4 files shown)


grynspan:
@swift-ci test

Tests/TestingTests/ExitTestTests.swift
```diff
@@ -506,6 +507,41 @@ private import _TestingInternals
...
+
+#if false // intentionally fails to compile
+  @Test("Capturing a tuple")
+  func captureListWithTuple() async throws {
+    // A tuple whose elements conform to Codable does not itself conform to
+    // Codable, so we cannot actually express this capture list in a way that
+    // works with #expect().
+    await #expect(processExitsWith: .success) { [x = (0 as Int, 1 as Double, "2" as String)] in
+      #expect(x.0 == 0)
+      #expect(x.1 == 1)
```
suzannaratcliff, line 539: Interesting.  I would have assumed 1 was an Int.

Sources/TestingMacros/Support/ClosureCaptureListParsing.swift
```diff
@@ -50,62 +50,177 @@ struct CapturedValueInfo {
...
+          TypeSyntax(OptionalTypeSyntax(wrappedType: inferredType.trimmed))
+        } else {
+          inferredType
+        }
+      }
+    }
+
+    init(in context: C) {
+      self.context = context
+      super.init(viewMode: .sourceAccurate)
```
suzannaratcliff, line 117: Why is this `sourceAccurate` and not `all`?

Tests/TestingTests/ExitTestTests.swift
grynspan: `1` is an integer literal. `1 as Double` tells the compiler to use `1` as the value of an instance of `Double`. The default type of the integer literal `1` is `IntegerLiteralType`, which by default is `Int`.

Sources/TestingMacros/Support/ClosureCaptureListParsing.swift
grynspan: `.all` doesn't insert type information and has a non-zero cost to synthesize missing nodes. Since the code at this point is known to be syntactically correct (otherwise the macro expansion wouldn't have started) there's no need to try to insert additional nodes. `.sourceAccurate` just walks the code the test author typed.

Tests/TestingTests/ExitTestTests.swift
suzannaratcliff: So why can you compare an IntegerLiteralType to a Double?

suzannaratcliff (approved):

Tests/TestingTests/ExitTestTests.swift
grynspan: The type of `1` in the expression `x.1 == 1` is inferred to be `Double`.

PR #1174, opened 2025-06-25
Infer a display name for tests and suites which have a raw identifier name consisting of a single token

stmontgomery:
This tweaks the logic for inferring a display name for a test or suite which has a raw identifier name so that it will take effect even if the content of the raw identifier is a single, valid token.

### Motivation:

Typically, raw identifiers (introduced in [SE-0451](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0451-escaped-identifiers.md)) are intended for use when you want to include characters in the name of a test or suite which aren't valid in identifiers — such as whitespace or (some) punctuation. However, it can sometimes be useful to specify a raw identifier for a test or suite even when it isn't strictly needed, because of how the testing library implicitly uses the content of a raw identifier as the display name.

For example, consider a parameterized test function which currently has an _explicit_ display name consisting of a single word, and has a lengthy parameter label:

```swift
@Test("Authentication", arguments: [...])
func authentication(authenticationSessionCoordinator: MyAuthenticationSession) { ... }
```

This test will have the plain display name `Authentication` and that's what will be displayed in console output and in the UI of integrated tools/IDEs. Even though the display name is only a single word, it allows the test author to precisely control how the test is presented in results and omit the lengthy parameter label, which allows the output to be cleaner and easier to understand.

Now imagine the user wishes to modernize their tests by adopting the new raw identifiers feature. They do so by switching to the following:

```swift
@Test(arguments: [...])
func `Authentication`(authenticationSessionCoordinator: MyAuthenticationSession) { ... }
```

In this scenario, because of an edge case in the current handling logic of the testing library, _no display name will be inferred_ at all, and the console output and UI tools will show the full underlying name which is `Authentication(authenticationSessionCoordinator:)`. However this isn't ideal, since the user no longer has an idiomatic way to specify the succinct display name they want using a raw identifier because it happens to consist of a single token without any whitespace or special characters. Although this can be worked around by specifying two or more words inside the raw identifier, there isn't an obvious reason why doing so is necessary and users should be permitted to specify a single valid token as a display name if they want to.

### Modifications:

With the fix in this PR, the above test would have the expected display name of `Authentication` inferred. Specifically, this impacts the `displayName`, not the `name`, property of `Test`.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
@@ -11,7 +11,6 @@
 import SwiftSyntax
 import SwiftSyntaxBuilder
 import SwiftSyntaxMacros
-import SwiftParser
 
 extension FunctionDeclSyntax {
   /// Whether or not this function a `static` or `class` function.
@@ -39,7 +38,7 @@ extension FunctionDeclSyntax {
   /// colons.
   var completeName: DeclReferenceExprSyntax {
     func possiblyRaw(_ token: TokenSyntax) -> TokenSyntax {
-      if let rawIdentifier = token.rawIdentifier, !rawIdentifier.isValidSwiftIdentifier(for: .memberAccess) {
+      if let rawIdentifier = token.rawIdentifier {
         return .identifier("`\(rawIdentifier)`")
       }
       return .identifier(token.textWithoutBackticks)
--- a/Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
@@ -11,6 +11,7 @@
 import SwiftSyntax
 import SwiftSyntaxBuilder
 import SwiftSyntaxMacros
+import SwiftParser
 
 extension FunctionDeclSyntax {
   /// Whether or not this function a `static` or `class` function.
@@ -38,7 +39,7 @@ extension FunctionDeclSyntax {
   /// colons.
   var completeName: DeclReferenceExprSyntax {
     func possiblyRaw(_ token: TokenSyntax) -> TokenSyntax {
-      if let rawIdentifier = token.rawIdentifier {
+      if let rawIdentifier = token.rawIdentifier, !rawIdentifier.isValidSwiftIdentifier(for: .memberAccess) {
         return .identifier("`\(rawIdentifier)`")
       }
       return .identifier(token.textWithoutBackticks)
--- a/Sources/TestingMacros/Support/Additions/TokenSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/TokenSyntaxAdditions.swift
@@ -35,7 +35,7 @@ extension TokenSyntax {
   /// token, or `nil` if it does not represent one.
   var rawIdentifier: String? {
     let (textWithoutBackticks, backticksRemoved) = _textWithoutBackticks
-    if backticksRemoved {
+    if backticksRemoved, !textWithoutBackticks.isValidSwiftIdentifier(for: .memberAccess) {
       return textWithoutBackticks
     }
 
--- a/Sources/TestingMacros/Support/Additions/TokenSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/TokenSyntaxAdditions.swift
@@ -35,7 +35,7 @@ extension TokenSyntax {
   /// token, or `nil` if it does not represent one.
   var rawIdentifier: String? {
     let (textWithoutBackticks, backticksRemoved) = _textWithoutBackticks
-    if backticksRemoved, !textWithoutBackticks.isValidSwiftIdentifier(for: .memberAccess) {
+    if backticksRemoved {
       return textWithoutBackticks
     }
 
--- a/Sources/TestingMacros/Support/AttributeDiscovery.swift
+++ b/Sources/TestingMacros/Support/AttributeDiscovery.swift
@@ -11,7 +11,6 @@
 import SwiftSyntax
 import SwiftSyntaxBuilder
 import SwiftSyntaxMacros
-import SwiftParser
 
 /// A syntax rewriter that removes leading `Self.` tokens from member access
 /// expressions in a syntax tree.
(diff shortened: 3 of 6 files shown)


stmontgomery:
@swift-ci please test

grynspan:
This is not a bug but intentional behaviour, as we discussed previously. It's a change in behaviour for existing test functions/suites and so it needs to be reviewed by the community via the Swift Evolution process.

Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
```diff
@@ -38,7 +39,7 @@ extension FunctionDeclSyntax {
   /// colons.
   var completeName: DeclReferenceExprSyntax {
     func possiblyRaw(_ token: TokenSyntax) -> TokenSyntax {
-      if let rawIdentifier = token.rawIdentifier {
+      if let rawIdentifier = token.rawIdentifier, !rawIdentifier.isValidSwiftIdentifier(for: .memberAccess) {
```
grynspan: This is not the right bottleneck for this change if we do pursue it. Rather, we would change the implementation of `var rawIdentifier` so the change applies universally.

grynspan (changes requested):
.

Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
grynspan: Actually, in this position this is a **breaking change**.

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -138,11 +139,17 @@ struct AttributeInfo {
...
     }
 
-    // Disallow an explicit display name for tests and suites with raw
-    // identifier names as it's redundant and potentially confusing.
+    // If this declaration has a raw identifier name, implicitly use the content
+    // of that raw identifier as its display name.
     if let namedDecl = declaration.asProtocol((any NamedDeclSyntax).self),
        let rawIdentifier = namedDecl.name.rawIdentifier {
-      if let displayName, let displayNameArgument {
+      // Disallow having an explicit display name for tests and suites with raw
```
grynspan: The only actual change you ought to need here outside of unit tests is to change the line:

```swift
let rawIdentifier = namedDecl.name.rawIdentifier {
```

To, I think:

```swift
case let nameWithoutBackticks = namedDecl.name.textWithoutBackticks,
nameWithoutBackticks != namedDecl.name.text {
```

i.e. "if the name _had_ backticks in source, it's a display name regardless of whether the language considers it a raw ID."

grynspan:
```diff
diff --git a/Sources/TestingMacros/Support/AttributeDiscovery.swift b/Sources/TestingMacros/Support/AttributeDiscovery.swift
index 3d95df29..83953ff7 100644
--- a/Sources/TestingMacros/Support/AttributeDiscovery.swift
+++ b/Sources/TestingMacros/Support/AttributeDiscovery.swift
@@ -141,11 +141,12 @@ struct AttributeInfo {
     // Disallow an explicit display name for tests and suites with raw
     // identifier names as it's redundant and potentially confusing.
     if let namedDecl = declaration.asProtocol((any NamedDeclSyntax).self),
-       let rawIdentifier = namedDecl.name.rawIdentifier {
+       case let nameWithoutBackticks = namedDecl.name.textWithoutBackticks,
+       nameWithoutBackticks != namedDecl.name.text {
       if let displayName, let displayNameArgument {
         context.diagnose(.declaration(namedDecl, hasExtraneousDisplayName: displayName, fromArgument: displayNameArgument, using: attribute))
       } else {
-        displayName = StringLiteralExprSyntax(content: rawIdentifier)
+        displayName = StringLiteralExprSyntax(content: nameWithoutBackticks)
       }
     }
 
diff --git a/Tests/TestingTests/Support/GraphTests.swift b/Tests/TestingTests/Support/GraphTests.swift
index 5f956dd6..9de7e28f 100644
--- a/Tests/TestingTests/Support/GraphTests.swift
+++ b/Tests/TestingTests/Support/GraphTests.swift
@@ -53,8 +53,7 @@ struct GraphTests {
     #expect(graph.subgraph(at: "C2", "C3")?.value == 2468)
   }
 
-  @Test("subscript([K]) operator")
-  func `subscript`() {
+  @Test func `subscript([K]) operator`() {
     let graph = Graph<String, Int>(value: 123, children: [
       "C1": Graph(value: 456),
       "C2": Graph(value: 789, children: [
```

Sources/TestingMacros/Support/Additions/TokenSyntaxAdditions.swift
```diff
@@ -35,7 +35,7 @@ extension TokenSyntax {
   /// token, or `nil` if it does not represent one.
   var rawIdentifier: String? {
     let (textWithoutBackticks, backticksRemoved) = _textWithoutBackticks
-    if backticksRemoved, !textWithoutBackticks.isValidSwiftIdentifier(for: .memberAccess) {
+    if backticksRemoved {
```
grynspan: The distinction here is intentional. If you want `textWithoutBackticks`, use `textWithoutBackticks`.

stmontgomery:
@swift-ci please test

Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
```diff
@@ -263,19 +263,60 @@ struct TestDeclarationMacroTests {
...
+        default:
+          Issue.record("Change \(change) differs from expected change \(expectedChange)")
+        }
+      }
+    }
+  }
+
   @Test("Raw identifier is detected")
   func rawIdentifier() {
-    #expect(TokenSyntax.identifier("`hello`").rawIdentifier == nil)
```
grynspan, line 268: Can you re-add these cases please? They're still valid.

Tests/TestingTests/MiscellaneousTests.swift
```diff
@@ -28,6 +28,10 @@ private import Foundation
 
 @Sendable func freeSyncFunctionParameterized2(_ i: Int, _ j: String) {}
 
+#if compiler(>=6.2) && hasFeature(RawIdentifiers)
+@Test(.hidden, arguments: [0]) func `ValidSingleCapitalizedToken`(someLengthyParameterName: Int) {}
```
grynspan, line 32: Should we add a test that uses an escaped keyword name too?

grynspan (changes requested):

PR #1175, opened 2025-06-25
Lower the "ambiguous display name" diagnostic to a warning for some names.

grynspan:
This PR modifies the behaviour of this compile-time macro diagnostic:

> 🛑 "Attribute 'Test' specifies display name 'foo' for function with implicit display name 'bar'

If `bar` (in the above context) is _not_ considered a raw identifier by the language, we emit a warning instead of an error. This will allow us to adjust display-name-from-backticked-name inference (see #1174) without introducing a source-breaking change for a declaration such as:

```swift
@Test("subscript([K]) operator")
func `subscript`()
```

(The above is a real-world test function in our own package that would be impacted.)

Note that we don't actually have a code path that triggers this warning yet. #1174, if approved and merged, would introduce such a code path. Here's an example of what that would look like:

Screenshot showing the warning diagnostic presented for func subscript()

### Checklist:


grynspan:
@swift-ci test

Issue #863, opened 2024-12-16
"Migrating a test from XCTest" does not mention parallel-by-default


MahdiBM:
"Migrating a test from XCTest" does not mention parallel-by-default, which catches users off-guard when migrating from XCTest to swift-testing.

There should also be a sentence or two explaining what to do to go back to the old behavior of non-parallel tests.

I was in a meeting last night and some folks mentioned this issue.
Apparently Fastlane also does not support having the tests run in parallel mode? so they had to figure out the `--no-parallel` CLI command and use that to run the tests sequentially.

grynspan:
We do mention it in [Parallelization.md](https://github.com/swiftlang/swift-testing/blob/main/Sources/Testing/Testing.docc/Parallelization.md), but a nod here wouldn't be a bad idea.

iamleeg:
Is there anything left to do here now that #960 is merged?

stmontgomery:
> Is there anything left to do here now that [#960](https://github.com/swiftlang/swift-testing/pull/960) is merged?

I don't think so! This issue appears to have been addressed well by #960. Looks like we just forgot to close this issue afterward.

stmontgomery:
Fixed by #960

PR #1176, opened 2025-06-26
Highlight parallel-by-default test run in case of XCTest migration

petermolnar-dev:
Adding highlight and link to the XCTest Migration document to help understand and mitigate the parallel-by-default test execution.

### Motivation:

Handling https://github.com/swiftlang/swift-testing/issues/863

### Modifications:

Adding the Important aside, highlighting the parallel-by-default nature of Swift Testing test execution while migrating from XCTest.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/MigratingFromXCTest.md
+++ b/Sources/Testing/Testing.docc/MigratingFromXCTest.md
@@ -84,6 +84,8 @@ concurrency safety:
 For more information about suites and how to declare and customize them, see
 <doc:OrganizingTests>.
 
+> Important: While XCTest test cases running sequentially by default, the test execution happens in parallel by default in Swift Testing. See the [Run tests sequentially](#Run-tests-sequentially) section in this document for further information. 
+
 ### Convert setup and teardown functions
 
 In XCTest, code can be scheduled to run before and after a test using the


grynspan:
@iamleeg Would you please help us review?

grynspan:
@petermolnar-dev It looks like you've opened this PR and made changes using two separate GitHub accounts. Was this intentional?

Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -84,6 +84,8 @@ concurrency safety:
 For more information about suites and how to declare and customize them, see
 <doc:OrganizingTests>.
 
+> Important: While XCTest test cases running sequentially by default, the test execution happens in parallel by default in Swift Testing. See the [Run tests sequentially](#Run-tests-sequentially) section in this document for further information.
```
iamleeg, line 87: This sentence doesn't say anything that isn't already covered in "Run tests sequentially," below. Additionally, the aside's content isn't really relevant to the topic of this section, which is about creating test suites. I'd recommend thinking about the structure of this content differently.

Is there something that "Run tests sequentially" ought to tell developers, but doesn't? If so, add information there.

Is there something that "Convert test classes" ought to tell developers about concurrency, so they know to read "Run tests sequentially"? If so, work that into the prose in this section.

If your goal is to ensure that developers remember to think about concurrency, they should probably also remember to think about some of the other points in this doc. Perhaps you need to expand the Overview, to give developers a list of important points to consider when they migrate tests.

iamleeg (changes requested):
Thanks for making this PR!

stmontgomery:
Hi @petermolnar-dev, thank you for the PR. It looks like your motivation for submitting it was to resolve the issue https://github.com/swiftlang/swift-testing/issues/863, is that correct? We realized upon reviewing that issue that it was probably resolved already by @iamleeg's subsequent PR #960 — or at least, it's not clear what additional information beyond that should be added — we just forgot to _close_ #863 after #960 was merged.

I'm going to close this PR for now given that. However, if you feel there is additional guidance on the Migrating document which should still be added, please clarify and we can re-open and reconsider this PR.

PR #1178, opened 2025-06-26
Add (hidden) synchronous overloads of `#expect(throws:)`.

### LINKED ISSUE #1177: Spurious warnings about `noasync` calls in `#expect(throws:)`
Given the following test function:

```swift
@available(*, noasync)
func f() {}

@Test func g() {
  #expect(throws: (any Error).self) {
    f()
  }
}
```

In Swift 6.2 the compiler emits a warning due to improved typechecking during macro expansion:

> ⚠️ function 'f' is unavailable from asynchronous contexts; this will be an error in a future Swift language mode

Because the compiler is now inferring the type of the closure passed to `#expect(throws:)` as `async` _before_ macro expansion rather than inferring its type (as non-`async`) _after_.

grynspan:
This PR adds overloads of the `#expect(throws:)` and `#require(throws:)` macros that take synchronous closures. This is necessary due to a change in the compiler in Swift 6.2 that improves type checking on closures passed to macros. The change is a good thing, but it means that the compiler infers the type of closures passed to these macros as `async` even when they are synchronous and developers will now get warnings under some circumstances.

This PR does not constitute an API change. The new overloads are identical to their `async` peers and there is no change in the underlying macro expansion logic at compile time or runtime. The PR serves solely to suppress new spurious warnings that were not emitted in Swift 6.1 or earlier.

Resolves #1177.
Resolves rdar://149299786.

### Checklist:

### DIFF
--- a/Sources/Testing/Expectations/Expectation+Macro.swift
+++ b/Sources/Testing/Expectations/Expectation+Macro.swift
@@ -188,12 +188,128 @@ public macro require<T>(
 /// check that an error is _not_ thrown by it, do not use this macro. Instead,
 /// simply call the code in question and allow it to throw an error naturally.
 @discardableResult
+@freestanding(expression)
+@_documentation(visibility: private)
+public macro expect<E, R>(
+  throws errorType: E.Type,
+  _ comment: @autoclosure () -> Comment? = nil,
+  sourceLocation: SourceLocation = #_sourceLocation,
+  performing expression: () throws -> R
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

stmontgomery:
For tracking purposes, the relevant compiler diagnostic was downgraded to a warning in https://github.com/swiftlang/swift/pull/80853.

PR #1147, opened 2025-06-10
Use `posix_spawn_file_actions_adddup2()` to clear `FD_CLOEXEC`.

grynspan:
On [FreeBSD](https://man.freebsd.org/cgi/man.cgi?query=posix_spawn_file_actions_adddup2), [OpenBSD](https://github.com/openbsd/src/blob/master/lib/libc/gen/posix_spawn.c#L155), [Android](https://android.googlesource.com/platform/bionic/+/master/libc/bionic/spawn.cpp#103), and [Glibc ≥ 2.29](https://sourceware.org/bugzilla/show_bug.cgi?id=23640), `posix_spawn_file_actions_adddup2()` automatically clears `FD_CLOEXEC` if the file descriptors passed to it are equal. Relying on this behaviour eliminates a race condition when spawning child processes. This functionality is standardized in [POSIX.1-2024](https://pubs.opengroup.org/onlinepubs/9799919799/) thanks to [Austin Group Defect #411](https://www.austingroupbugs.net/view.php?id=411).

Some older Linuxes (Amazon Linux 2 in particular) don't have this functionality, so we do a runtime check of the Glibc version.

This PR is a follow-up to #1145.
Resolves #1140.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1183, opened 2025-06-27
FreeBSD: Gate GNU-only API

etcwilde:
The FreeBSD builds are currently using the GlibC modulemap to import the C runtimes. FreeBSD does not have `gnu_get_libc_version` resulting in build failures.

The use of this API was introduced in https://github.com/swiftlang/swift-testing/pull/1147

### DIFF
--- a/Sources/Testing/ExitTests/SpawnProcess.swift
+++ b/Sources/Testing/ExitTests/SpawnProcess.swift
@@ -136,7 +136,7 @@ func spawnExecutable(
             // standardized in POSIX.1-2024 (see https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_spawn_file_actions_adddup2.html
             // and https://www.austingroupbugs.net/view.php?id=411).
             _ = posix_spawn_file_actions_adddup2(fileActions, fd, fd)
-#if canImport(Glibc)
+#if canImport(Glibc) && !os(FreeBSD) && !os(OpenBSD)
             if _slowPath(glibcVersion.major < 2 || (glibcVersion.major == 2 && glibcVersion.minor < 29)) {
               // This system is using an older version of glibc that does not
               // implement FD_CLOEXEC clearing in posix_spawn_file_actions_adddup2(),
--- a/Sources/Testing/Support/Versions.swift
+++ b/Sources/Testing/Support/Versions.swift
@@ -153,7 +153,7 @@ let swiftStandardLibraryVersion: String = {
   return "unknown"
 }()
 
-#if canImport(Glibc)
+#if canImport(Glibc) && !os(FreeBSD) && !os(OpenBSD)
 /// The (runtime, not compile-time) version of glibc in use on this system.
 ///
 /// This value is not part of the public interface of the testing library.


grynspan (approved):
I thought the module was supposed to be named FreeBSD since they don't use Glibc? Ah well.

grynspan:
@swift-ci test

etcwilde:
> I thought the module was supposed to be named FreeBSD since they don't use Glibc? Ah well.

Yes, we should create a new module and module map for FreeBSD and call it something else, but since everything piggybacks on the Glibc module map, it was convenient to avoid re-writing the `#if canImport(<insert libc here>)` chains across the project.

CodingCarpincho (approved):

Sources/Testing/ExitTests/SpawnProcess.swift
```diff
@@ -136,7 +136,7 @@ func spawnExecutable(
             // standardized in POSIX.1-2024 (see https://pubs.opengroup.org/onlinepubs/9799919799/functions/posix_spawn_file_actions_adddup2.html
             // and https://www.austingroupbugs.net/view.php?id=411).
             _ = posix_spawn_file_actions_adddup2(fileActions, fd, fd)
-#if canImport(Glibc)
```
jakepetroules, line 139: Add !OpenBSD while you're add it? SwiftTesting also has a bunch of existing OpenBSD conditionals.

grynspan: I've attempted to support OpenBSD but it's entirely untested. It's unclear to me if it would have a module named Glibc (its stdlib is not GNU.)

grynspan:
Holding off on merging until we know if we need to make OpenBSD changes too.

Sources/Testing/ExitTests/SpawnProcess.swift
etcwilde: As of right now, Cygwin, FreeBSD, OpenBSD, Linux, and Haiku all use the Glibc module map to import the underlying C runtime. https://github.com/swiftlang/swift/blob/d3f86b4efdb8a3b1225e3ceab0cceddad97f9193/stdlib/public/Platform/CMakeLists.txt#L12

They aren't glibc, that is correct, but right now we are the ones controlling that file and it's the closest thing we have to `import libc` for, "Unix-y things that aren't Apple".

CC @3405691582, can you check if swift-testing is close to building and adding the `!os(OpenBSD)` would have a meaningful impact on your life, or if there are enough things that don't work that it wouldn't really have an impact? Thanks.

3405691582: swift-testing is building enough to build swiftpm for OpenBSD. We definitely need !os(OpenBSD) to gate off the missing `__GLIBC__` symbol -- and thus I imagine here as well. (This also needs to be backported to 6.2.)

grynspan: Okay, let's make that change, squashmerge, then we can cherry-pick for release/6.2. The nomination template is [here](https://github.com/swiftlang/.github/blob/main/PULL_REQUEST_TEMPLATE/release.md?plain=1).

Sources/Testing/ExitTests/SpawnProcess.swift
```diff
@@ -136,7 +136,7 @@ func spawnExecutable(
...
+#if canImport(Glibc) && !os(FreeBSD) && !os(OpenBSD)
```
compnerd, line 139: What do you think of some DeMorgan? `&& !(os(FreeBSD) || os(OpenBSD))`

etcwilde: It's equivalent. ¯\\\_(ツ)_/¯
I don't have a preference for or against it. I'm not sure it substantially helps readability here though.

grynspan: What's here is how I would have typed it. Although now it's annoying me and I'd probably rewrite the whole line as `#if SWT_USES_GLIBC` and define that in CMake and SPM as "the target is Linux". At least until we fix the name of the module elsewhere.

3405691582: Then you might have issues for musl platforms :)

In any event: given the 6.2 branch has this problem as well, I'd suggest improving the form of the conditional in a separate pr may be more pragmatic, however.

grynspan: I can't win!

grynspan:
@swift-ci test

Sources/Testing/ExitTests/SpawnProcess.swift
etcwilde: With CMake, we can ask it to tell us if the symbol is available and define the macro that way. https://cmake.org/cmake/help/v4.1/module/CheckSymbolExists.html

grynspan: It's probably not worth the effort—this is mostly me griping about the module name.

grynspan:
@etcwilde Thus is ready to squashmerge, yes?

etcwilde:
Merged, now to pick some cherries.

PR #1184, opened 2025-06-28
order limits.h before stdlib.h to workaround for glibc _FORTIFY_SOURCE

lhoward:
`limits.h` must be included before `stdlib.h` when building with glibc and having `_FORTIFY_SOURCE` set to a non-zero value.

When building with `_FORTIFY_SOURCE`, `realpath()` is inlined, and its definition depends on whether `limits.h` has been included or not (clearly, this is a terrible idea in terms of interacting with Clang modules and should probably be fixed upstream). If the definition differs from the one in SwiftGlibc, then _TestingInternals will not build.

### DIFF
--- a/Sources/_TestingInternals/include/Includes.h
+++ b/Sources/_TestingInternals/include/Includes.h
@@ -29,6 +29,12 @@
 #include <ctype.h>
 #include <errno.h>
 #include <inttypes.h>
+/// limits.h must be included before stdlib.h with glibc, otherwise the
+/// fortified realpath() in this module will differ from the one in SwiftGlibc.
+/// glibc bug: https://sourceware.org/bugzilla/show_bug.cgi?id=30516
+#if __has_include(<limits.h>)
+#include <limits.h>
+#endif
 /// Guard against including `signal.h` on WASI. The `signal.h` header file
 /// itself is available in wasi-libc, but it's just a stub that doesn't actually
 /// do anything. And also including it requires a special macro definition
@@ -95,10 +101,6 @@
 
 #if __has_include(<pwd.h>)
 #include <pwd.h>
-#endif
-
-#if __has_include(<limits.h>)
-#include <limits.h>
 #endif
 
 #if __has_include(<spawn.h>)


grynspan:
@swift-ci test

lhoward:
Obviously we could also conditionalise this based on platform but I presume it's generally safe to include `<limits.h>` first.

lhoward:
> Can we put it in alphabetical position? Also, can you include a URL to whatever issue tracks the upstream fix?

Yes, will do.

grynspan:
Is there a URL we can point to tracking the issue in glibc?

lhoward:
> Is there a URL we can point to tracking the issue in glibc?

I haven't filed a bug yet, will do (sorry, it's Sunday, negotiating bug fixing around family life!).

grynspan:
Understood. This can certainly wait until Monday!

lhoward:
> Understood. This can certainly wait until Monday!

It appears applying for a glibc [bug tracker](https://sourceware.org/bugzilla/) account is a manual process so, I'll have to wait until I hear back...

grynspan:
Okay, fair enough. When you get the issue filed please comment here so we can refer back to it in the future.

It's not a blocking issue, so we can merge this change without it.

grynspan:
@swift-ci test

lhoward:
Updated.

glibc bug here: https://sourceware.org/bugzilla/show_bug.cgi?id=33120

grynspan:
@swift-ci test

grynspan:
@swift-ci test

lhoward:
FYI upstream bug is a duplicate, first report was in 2023:

https://sourceware.org/bugzilla/show_bug.cgi?id=30516

grynspan:
That's okay, we just want a paper trail. Want to update the link? We can then merge the PR I think.

grynspan:
@lhoward If you need this in our 6.2 release, please create a new PR targetting release/6.2 including this PR's commit and following [this](https://github.com/swiftlang/.github/blob/main/PULL_REQUEST_TEMPLATE/release.md?plain=1) template. Thanks!

PR #1188, opened 2025-06-30
Ensure that when .serialized is applied to a parameterized @Test func, its test cases are serialized

stmontgomery:
This fixes a regression introduced in the changes for #901. When `.serialized` is applied directly to a parameterized `@Test` function, not to a containing suite, its test cases are no longer serialized. This PR resolves and restores the original behavior by ensuring there is a non-`nil` scope provider returned for the test function.

### Checklist:

Fixes rdar://154529146

### DIFF
--- a/Tests/TestingTests/TestSupport/TestingAdditions.swift
+++ b/Tests/TestingTests/TestSupport/TestingAdditions.swift
@@ -108,6 +108,25 @@ extension Runner {
     inModuleOf fileID: String = #fileID,
     configuration: Configuration = .init()
   ) async {
+    let plan = await Runner.Plan(selecting: testName, inModuleOf: fileID, configuration: configuration)
+    self.init(plan: plan, configuration: configuration)
+  }
+}
+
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

Tests/TestingTests/TestSupport/TestingAdditions.swift
```diff
@@ -103,6 +103,25 @@ extension Runner {
   ///   - fileID: The `#fileID` string whose module should be used to locate
   ///     the test function to run.
   ///   - configuration: The configuration to use for running.
+  init(
```
grynspan, line 106: ???

stmontgomery: This is used [here](https://github.com/swiftlang/swift-testing/blob/9f0c83b8e7ffad569691c99cc895f891701fa666/Tests/TestingTests/Traits/ParallelizationTraitTests.swift#L17)

PR #1190, opened 2025-06-30
[6.2] Ensure that when .serialized is applied to a parameterized @Test func, its test cases are serialized

stmontgomery:
- **Explanation**: This fixes a regression when `.serialized` is applied directly to a parameterized `@Test` function, not to a containing suite, its test cases are no longer serialized.
- **Scope**: Affects the `.serialized` trait when applied to `@Test` functions, specifically, not to suite types.
- **Issues**: rdar://154529146
- **Original PRs**: https://github.com/swiftlang/swift-testing/pull/1188
- **Risk**:
  
- **Testing**: New regression test added
- **Reviewers**: @suzannaratcliff, @briancroom, @grynspan


stmontgomery:
@swift-ci please test

PR #1191, opened 2025-07-01
Adopt Source Location-Aware Unique Name Generation for `@Test` Macro

twhitt14:
This PR updates the `swift-testing` library to leverage the new source location-aware `makeUniqueName` API from `swift-syntax`. This ensures that `@Test` functions with the same name at different locations generate unique identifiers, preventing compiler crashes and redeclaration errors.

#### Motivation
With the previous approach, duplicate `@Test` function names could cause symbol conflicts and compiler crashes. By incorporating source location into the unique name, we ensure robust test discovery and execution.

#### Changes
- Updated the `makeUniqueName(thunking:)` helper to use the new source location-aware API.
- Added a test to verify that duplicate function names generate different identifiers.
- (For development) Temporarily updated `Package.swift` to use the local `swift-syntax` package.

#### Impact
- Fixes crashes and redeclaration errors caused by duplicate `@Test` function names.
- Ensures reliable test macro expansion and execution.

Resolves swiftlang/swift#82631

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -103,7 +103,7 @@ let package = Package(
   ],
 
   dependencies: [
-    .package(url: "https://github.com/swiftlang/swift-syntax.git", from: "602.0.0-latest"),
+    .package(path: "../swift-syntax"),
   ],
 
   targets: [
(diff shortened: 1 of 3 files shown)


grynspan:
It is intentional that we don't include source location information in the unique names we generate. If we did, then any time you inserted or removed a line in one test, the stack traces for all subsequent tests in the same file would change.

Please allow the compiler team to address the compiler crash tracked by this issue. Thanks!

PR #1192, opened 2025-07-02
Require `UTType` for image attachments.

grynspan:
This PR simplifies our image attachment code so that it always requires `UTType` instead of providing implementations for older Apple platforms. For more information about the `UTType` API, watch [this](https://developer.apple.com/videos/play/tech-talks/10696/) video. There is a kitty.

Image attachments depend on `CGImage` and are Apple-specific at this time. Non-Apple image attachment support is a future direction.

> [!NOTE]
> Image attachments are an experimental feature.

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageWrapper.swift
+++ b/Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageWrapper.swift
@@ -73,7 +73,7 @@ public struct _AttachableImageWrapper<Image>: Sendable where Image: AttachableAs
   /// the result is undefined.
   var contentType: UTType {
     get {
-      _contentType ?? (encodingQuality < 1.0 ? .jpeg : .png)
+      _contentType ?? .image
     }
     set {
       precondition(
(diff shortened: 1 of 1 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

Package.swift
```diff
@@ -217,6 +217,7 @@ let package = Package(
       name: "SymbolShowcase",
       dependencies: [
         "Testing",
+        "_Testing_CoreGraphics",
```
stmontgomery: What's this needed for?

Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
```diff
@@ -23,46 +25,9 @@ extension Attachment {
   ///     to a test report or to disk. If `nil`, the testing library attempts
   ///     to derive a reasonable filename for the attached value.
   ///   - contentType: The image format with which to encode `attachableValue`.
```
stmontgomery, line 27: Consider documenting the default `nil` value, and saying "if any" etc.

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageWrapper.swift
```diff
@@ -61,7 +62,7 @@ public struct _AttachableImageWrapper<Image>: Sendable where Image: AttachableAs
   var encodingQuality: Float
 
   /// Storage for ``contentType``.
-  private var _contentType: (any Sendable)?
+  private var _contentType: UTType?
```
stmontgomery, line 65: Could we consolidate this property with `contentType` now?

Package.swift
grynspan: Bad merge, I think. Will remove.

Sources/Overlays/_Testing_CoreGraphics/Attachments/Attachment+AttachableAsCGImage.swift
grynspan: `nil` is described in the prose below. "if any" doesn't make sense here since there will be an image format one way or another.

Sources/Overlays/_Testing_CoreGraphics/Attachments/_AttachableImageWrapper.swift
grynspan: No, see the additional logic in `contentType`. It doesn't simply forward `_contentType`.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

PR #1197, opened 2025-07-02
Add a temporary package product that builds our cross-import overlays.

grynspan:
This PR adds a **temporary** product to our package that builds the cross-import overlay targets needed for image attachment support. This allows developers to try out this new code.

This product will be removed from the package in a future update.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -84,6 +84,16 @@ let package = Package(
     )
 #endif
 
+    result += [
+      .library(
+        name: "_Testing_ExperimentalImageAttachments",
+        targets: [
+          "_Testing_AppKit",
+          "_Testing_CoreGraphics",
+        ]
+      )
+    ]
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -84,6 +84,16 @@ let package = Package(
     )
 #endif
 
+    result += [
```
stmontgomery, line 87: How about guarding this with `if buildingForDevelopment { ...`

grynspan: ~~Can do!~~ That'll break once we merge and it gets into a daily tag.

grynspan:
@swift-ci test

stmontgomery (approved):

grynspan:
macOS CI is jammed up. Merging before it completes. Tested at desk.

PR #1199, opened 2025-07-03
Apply the 'traitRelated' tag to more tests and suites

stmontgomery:
This is a small, test-only enhancement which applies the `.traitRelated` tag to more tests and suites which are related to specific traits or the trait subsystem.

### Checklist:


stmontgomery:
@swift-ci please test

PR #1202, opened 2025-07-04
Adopt `SuspendingClock.systemEpoch`.

grynspan:
This PR adopts the new `systemEpoch` API added with [SE-0473](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0473-clock-epochs.md).

This API is back-deployed, so we don't need availability annotations on Darwin. It is not available with the 6.1 toolchain though, so the old `unsafeBitCast()` calls remain if building from source with a 6.1 toolchain.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

PR #1203, opened 2025-07-07
Use a custom `AttachableImageFormat` type instead of directly relying on `UTType`.

grynspan:
This PR creates a platform-agnostic type to represent image formats for image attachments instead of relying directly on `UTType`. The implementation still requires `UTType` on Apple platforms, but on non-Apple platforms we can use the same type to represent those platforms' platform-specific image format enums (e.g. on Windows, it can box `CLSID`.) This reduces the platform-specific API surface area for image attachments.

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableImageFormat+UTType.swift
+++ b/Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableImageFormat+UTType.swift
@@ -1,0 +1,100 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024–2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if SWT_TARGET_OS_APPLE && canImport(CoreGraphics)
+@_spi(Experimental) public import Testing
+
+public import UniformTypeIdentifiers
+
+@available(_uttypesAPI, *)
+extension AttachableImageFormat {
+  /// Get the content type to use when encoding the image, substituting a
+  /// concrete type for `UTType.image` in particular.
+  ///
+  /// - Parameters:
+  ///   - imageFormat: The image format to use, or `nil` if the developer did
+  ///     not specify one.
+  ///   - preferredName: The preferred name of the image for which a type is
+  ///     needed.
+  ///
+  /// - Returns: An instance of `UTType` referring to a concrete image type.
+  ///
+  /// This function is not part of the public interface of the testing library.
+  static func computeContentType(for imageFormat: Self?, withPreferredName preferredName: String) -> UTType {
+    guard let imageFormat else {
+      // The developer didn't specify a type. Substitute the generic `.image`
+      // and solve for that instead.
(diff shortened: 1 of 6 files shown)


grynspan:
@swift-ci test

Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableImageFormat+UTType.swift
```diff
@@ -0,0 +1,100 @@
...
+      if contentType != .image {
+        // The developer explicitly specified a type.
+        return contentType
+      }
+
+      // The developer didn't specify a concrete type, so try to derive one from
+      // the preferred name's path extension.
+      let pathExtension = (preferredName as NSString).pathExtension
+      if !pathExtension.isEmpty,
+         let contentType = UTType(filenameExtension: pathExtension, conformingTo: .image),
```
grynspan, line 53: Note to self: on Windows, `ImageCodecInfo::FilenameExtension` should help us here. Windows does not have hierarchical type declarations (does it?)

grynspan: > Pointer to a null-terminated string that contains all file-name extensions associated with the codec. The extensions are separated by semicolons.

Sources/Testing/Attachments/AttachableImageFormat.swift
```diff
@@ -0,0 +1,93 @@
...
+/// on the attachment's preferred name.
+///
+/// The PNG and JPEG image formats are always supported. The set of additional
+/// supported image formats is platform-specific:
+///
+/// - On Apple platforms, you can use [`CGImageDestinationCopyTypeIdentifiers()`](https://developer.apple.com/documentation/imageio/cgimagedestinationcopytypeidentifiers())
+///   from the [Image I/O framework](https://developer.apple.com/documentation/imageio)
+///   to determine which formats are supported.
+@_spi(Experimental)
+@available(_uttypesAPI, *)
```
briancroom, line 27: What's the thinking behind including this on the new, generic type?

Sources/Testing/Attachments/AttachableImageFormat.swift
```diff
@@ -0,0 +1,93 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// A type describing image formats supported by the system that can be used
+/// when attaching an image to a test.
+///
+/// When you attach an image to a test, you can pass an instance of this type to
```
briancroom, line 14: It might be helpful to elaborate a little bit on what "image" means here - specifically, that it's some structured representation of an image that's something other than a generic `Data` / bag of bytes, and that `AttachableImageFormat` is involved in configuring the process of transforming it into the `Data`-style representation.

Sources/Testing/Attachments/AttachableImageFormat.swift
grynspan: The rest of the interface is unusable without `UTType`, so it would be odd to say "you can use this type, but not do anything with it."

Sources/Testing/Attachments/AttachableImageFormat.swift
grynspan: Will work with @iamleeg to finesse these docs when we move them to public.

PR #1205, opened 2025-07-07
Disallow the `@Test` attribute on operator declarations.

### LINKED ISSUE #1204: `@Test` doesn't work consistently on operator functions
This works:

```swift
prefix operator ^

@Test(arguments: [(0)])
prefix func ^(_ lhs: Int) {
  #expect(1 == 3)
}
```

But if you remove the `_` argument label or change it to an `infix` or `postfix` operator, you get a compiler error in the macro expansion. Either `@Test` should properly handle these cases or the macro should detect and reject attempts to use it on an operator.

grynspan:
This PR prevents `@Test` from being applied to an operator declaration such as:

```swift
@Test(arguments: ...)
static func +(lhs: A, rhs: B) { ... }
```

Now, the following error will be emitted by the compiler:

> 🛑 Attribute 'Test' cannot be applied to an operator

Previously, applying `@Test` to an operator produced undefined/unstable effects.

Resolves #1204.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
+++ b/Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
@@ -34,6 +34,16 @@ extension FunctionDeclSyntax {
       .contains(.keyword(.nonisolated))
   }
 
+  /// Whether or not this function declares an operator.
+  var isOperator: Bool {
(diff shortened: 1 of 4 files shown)


grynspan:
@swift-ci test

stmontgomery (approved):

PR #1210, opened 2025-07-08
Clarify the number of supported argument collections in article which discusses combinatoric parameterized testing

stmontgomery:
This is a small adjustment to the documentation article [Implementing parameterized tests](https://swiftpackageindex.com/swiftlang/swift-testing/main/documentation/testing/parameterizedtesting), in the section [Test with more than one collection](https://swiftpackageindex.com/swiftlang/swift-testing/main/documentation/testing/parameterizedtesting#Test-with-more-than-one-collection). It currently has a sentence which ends:

> …, elements from the second collection are passed as the second argument, **and so forth.**

The "and so forth" has caused some confusion since it implies that you can pass more than two argument collections, but in reality only two are supported currently. Eventually this could potentially be expanded with further improvements to the testing library, but that is tracked separately and the docs should reflect its current capabilities.

Fixes rdar://154647425

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/ParameterizedTesting.md
+++ b/Sources/Testing/Testing.docc/ParameterizedTesting.md
@@ -146,11 +146,11 @@ func makeLargeOrder(of food: Food, count: Int) async throws {
 ```
 
 Elements from the first collection are passed as the first argument to the test
-function, elements from the second collection are passed as the second argument,
-and so forth.
+function, and elements from the second collection are passed as the second
+argument.
 
 Assuming there are five cases in the `Food` enumeration, this test function
-will, when run, be invoked 500 times (5 x 100) with every possible combination
+will, when run, be invoked 500 times (5 × 100) with every possible combination
 of food and order size. These combinations are referred to as the collections'
 Cartesian product.
 


stmontgomery:
@swift-ci please test

Sources/Testing/Testing.docc/ParameterizedTesting.md
```diff
@@ -146,11 +146,11 @@ func makeLargeOrder(of food: Food, count: Int) async throws {
...
 
 Elements from the first collection are passed as the first argument to the test
-function, elements from the second collection are passed as the second argument,
-and so forth.
+function, and elements from the second collection are passed as the second
+argument.
 
 Assuming there are five cases in the `Food` enumeration, this test function
-will, when run, be invoked 500 times (5 x 100) with every possible combination
+will, when run, be invoked 500 times (5 × 100) with every possible combination
```
stmontgomery: Drive-by improvement to use the multiplication sign character instead of the letter 'x' since this is representing multiplication.

grynspan: This was actually intentional per @iamleeg and the DevPubs style guide.

Sources/Testing/Testing.docc/ParameterizedTesting.md
```diff
@@ -146,11 +146,11 @@ func makeLargeOrder(of food: Food, count: Int) async throws {
 ```
 
 Elements from the first collection are passed as the first argument to the test
-function, elements from the second collection are passed as the second argument,
-and so forth.
+function, and elements from the second collection are passed as the second
```
grynspan, line 149: This strictly isn't true (wasn't before either) due to tuple desugaring. Do we want to adjust the phrasing to mention that somehow?

stmontgomery: This sentence refers to the code example which precedes it, and I think if you zoom out and reread the section in its entity it is clear in context.

Then, the subsequent paragraphs contrast it with the tuple desugaring behavior:

> To avoid the combinatoric semantics shown above, use [zip()](https://developer.apple.com/documentation/swift/zip(_:_:)):
> […]
> The zipped sequence will be “destructured” into two arguments automatically, then passed to the test function for evaluation.

Sources/Testing/Testing.docc/ParameterizedTesting.md
stmontgomery: Removed

PR #1212, opened 2025-07-08
Reset the working directory of child processes we spawn on Windows.

### LINKED ISSUE #1209: Exit tests crash main process on Windows (testing fails)
### Description

When enabling exit tests on Windows, the tests fail indicating that `swift.exe` unexpectedly exited:

```
Error: Process completed with exit code 1.
```

### Reproduction

We're seeing this with any exit test we enable on Windows, to reproduce simply create a test that uses `#expect(processExitsWith: .failure)` and testing will fail with this message

### Expected behavior

I'd expect tests to succeed and not crash (except for the expected crashing code) on Windows like the other platforms

### Environment

https://github.com/swiftlang/swift-foundation/actions/runs/16126481664/job/45504492458?pr=1406

```
Swift version 6.2-dev (LLVM 3c4a54b02dd62b7, Swift a40a7be6945a895)
Target: x86_64-unknown-windows-msvc
Build config: +assertions
```

```
→ Testing Library Version: 6.3-dev (25b61ef4667a6d1)
```

### Additional information

Note that the same exit tests in https://github.com/swiftlang/swift-foundation/actions/runs/16126481664?pr=1406 pass on Linux and macOS but hit this issue on Windows

grynspan:
This PR modifies the Windows implementation of `spawnProcess()` so that it sets the working directory of the new process to "C:\" (or thereabouts). This prevents a race condition on Windows because that system won't let you delete a directory if it's the working directory of any process. See [The Old New Thing](https://devblogs.microsoft.com/oldnewthing/20101109-00/?p=12323) for a very on-the-nose blog post.

Note that we do not specify the value of the working directory in an exit test body. A test should generally not rely on it anyway because it is global state and any thread could change its value at any time.

I haven't written a unit test for this change because it's unclear what I could write that would be easily verifiable, and because I don't know what state I might perturb outside such a test by calling `SetCurrentDirectory()`.

Resolves #1209.
(This is a speculative fix.)

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/SpawnProcess.swift
+++ b/Sources/Testing/ExitTests/SpawnProcess.swift
@@ -297,9 +297,9 @@ func spawnExecutable(
       //
       // SEE: https://devblogs.microsoft.com/oldnewthing/20101109-00/?p=12323
       let workingDirectoryPath: UnsafeMutablePointer<wchar_t>? = {
-        let systemDrive = Environment.variable(named: "SYSTEMDRIVE") ?? "C:"
+        var systemDrive = Environment.variable(named: "SYSTEMDRIVE") ?? "C:"
         if systemDrive.last == ":" {
-          return #"\#(systemDrive)\"#
+          systemDrive = #"\#(systemDrive)\"#
         }
         return systemDrive.withCString(encodedAs: UTF16.self) { _wcsdup($0) }
       }()
--- a/Sources/Testing/ExitTests/SpawnProcess.swift
+++ b/Sources/Testing/ExitTests/SpawnProcess.swift
@@ -284,6 +284,29 @@ func spawnExecutable(
       let commandLine = _escapeCommandLine(CollectionOfOne(executablePath) + arguments)
       let environ = environment.map { "\($0.key)=\($0.value)" }.joined(separator: "\0") + "\0\0"
 
+      // On Windows, a process holds a reference to its current working
+      // directory, which prevents other processes from deleting it. This causes
+      // code to fail if it tries to set the working directory to a temporary
+      // path. SEE: https://github.com/swiftlang/swift-testing/issues/1209
+      //
+      // This problem manifests for us when we spawn a child process without
+      // setting its working directory, which causes it to default to that of
+      // the parent process. To avoid this problem, we set the working directory
+      // of the new process to the root directory of the boot volume (which is
+      // unlikely to be deleted, one hopes).
+      //
+      // SEE: https://devblogs.microsoft.com/oldnewthing/20101109-00/?p=12323
+      let workingDirectoryPath: UnsafeMutablePointer<wchar_t>? = {
+        let systemDrive = Environment.variable(named: "SYSTEMDRIVE") ?? "C:"
+        if systemDrive.last == ":" {
+          return #"\#(systemDrive)\"#
+        }
+        return systemDrive.withCString(encodedAs: UTF16.self) { _wcsdup($0) }
+      }()
+      defer {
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/ExitTests/SpawnProcess.swift
```diff
@@ -229,6 +229,7 @@ func spawnExecutable(
       return pid
     }
   }
+
```
grynspan, line 232: 🤷‍♂️

grynspan:
@swift-ci test Linux

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
I'm not planning to apply the same change to the `posix_spawn()` path (using `posix_spawn_file_actions_addchdir_np()`), but if we decide we need it: [posix_spawn_file_actions_addchdir.txt](https://github.com/user-attachments/files/21125104/posix_spawn_file_actions_addchdir.txt)

Sources/Testing/Support/FileHandle.swift
```diff
@@ -719,4 +719,37 @@ func setFD_CLOEXEC(_ flag: Bool, onFileDescriptor fd: CInt) throws {
...
 }
 #endif
+
+/// The path to the root directory of the boot volume.
+///
+/// On Windows, this string is usually of the form `"C:\"`. On UNIX-like
+/// platforms, it is always equal to `"/"`.
+let rootDirectoryPath: String = {
+#if os(Windows)
+  if let systemDrive = Environment.variable(named: "SYSTEMDRIVE")?.utf16 {
```
compnerd: It would be safer to use the combination of `GetWindowsDirectoryW` and `PathCchStripToRoot` to get to `C:\` over the environment which the user could modify.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

briancroom (approved):

Sources/Testing/Support/FileHandle.swift
```diff
@@ -719,4 +719,35 @@ func setFD_CLOEXEC(_ flag: Bool, onFileDescriptor fd: CInt) throws {
...
+      _ = GetSystemWindowsDirectoryW(buffer.baseAddress!, UINT(buffer.count))
+      let rStrip = PathCchStripToRoot(buffer.baseAddress!, buffer.count)
+      if rStrip == S_OK || rStrip == S_FALSE {
+        result = String.decodeCString(buffer.baseAddress!, as: UTF16.self)?.result
+      }
+    }
+  }
+
+  // If we weren't able to get a path, fall back to "C:\" on the assumption that
+  // it's the common case and most likely correct.
```
jakepetroules, line 747: Why not just throw an error instead of falling back to C:\? Or even fatalError. If the above fails certainly something catastrophic has occurred.

grynspan: It's dead code with no way to test it either way, and isn't practically useful since in almost all scenarios `"C:\"` is the correct value.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

grynspan:
@swift-ci test

grynspan:
@swift-ci test Linux

grynspan:
@swift-ci test macOS

PR #1213, opened 2025-07-08
Bump the required compiler version for `systemEpoch` to 6.3.

grynspan:
There are some workflows including Xcode 26 betas where the compiler version is 6.2 but `systemEpoch` isn't available yet, so bump the requirement for it to 6.3. This is temporary and we can delete the check outright after 6.2 ships.

### Checklist:


grynspan:
@swift-ci test

grynspan:
Small/safe change, self-reviewing.

PR #1217, opened 2025-07-10
Simplify `NSImageRep`-checking logic to use `Bundle`.

grynspan:
This PR simplifies the logic we use to check if subclasses of `NSImageRep` are likely to be thread-safe by using Foundation's `Bundle` instead of querying dyld. AppKit always links Foundation, so there's no need for us to roll our own logic for this check.

### Checklist:


grynspan:
@swift-ci test

stmontgomery (approved):

PR #1218, opened 2025-07-10
Make `ExitTest._current` immutable.

grynspan:
This PR makes the private `ExitTest._current` stored property a `let` instead of a `var`. This has no practical effect but is more correct.

### Checklist:


grynspan:
@swift-ci test

grynspan:
Will self-review and merge after CI finishes.

PR #1219, opened 2025-07-10
[6.2] Reset the working directory of child processes we spawn on Windows.

grynspan:
- **Explanation**: Works around a behavioural difference on Windows where a process prevents the deletion of its CWD which means exit tests can cause Foundation unit tests to fail.
- **Scope**: Windows exit tests.
- **Issues**: #1209
- **Original PRs**: #1212
- **Risk**: Low, but it is a Windows-specific behavioural change so non-zero?
- **Testing**: Existing CI jobs and a new unit test.
- **Reviewers**: @stmontgomery @jmschonfeld @briancroom


grynspan:
@swift-ci test

briancroom (approved):

PR #1225, opened 2025-07-15
rdar://140293091 Clarify the outcome of applying a timeLimit trait to a suite.

iamleeg:
Make it clearer that the time limit applies to each test or case in the suite, not to the overall runtime of the test suite.

### Motivation:

I received feedback that a developer couldn't work out whether `.timeLimit(_:)` on a suite limits each test, or the whole suite.

### Modifications:

- Add documentation that explains that the time limit applies to each test or case in a parameterized test, when you apply it to a suite.
- Remove some passive-voice sentences in the same documents so that it's clearer what a developer does, and what the testing library does for them.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/LimitingExecutionTime.md
+++ b/Sources/Testing/Testing.docc/LimitingExecutionTime.md
@@ -40,8 +40,8 @@ hour (60 x 60 seconds) to execute, the task in which it's running is
 and the test fails with an issue of kind
 ``Issue/Kind-swift.enum/timeLimitExceeded(timeLimitComponents:)``.
 
-- Note: If multiple time limit traits apply to a test, the shortest time limit
-  is used.
+- Note: If multiple time limit traits apply to a test, the testing library uses
+  the shortest time limit.
 
 The testing library may adjust the specified time limit for performance reasons
 or to ensure tests have enough time to run. In particular, a granularity of (by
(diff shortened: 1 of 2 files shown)


Sources/Testing/Traits/TimeLimitTrait.swift
```diff
@@ -83,6 +83,12 @@ extension Trait where Self == TimeLimitTrait {
   /// test cases individually. If a test has more than one time limit associated
   /// with it, the shortest one is used. A test run may also be configured with
   /// a maximum time limit per test case.
+  ///
+  /// If you apply this trait to a test suite, then it sets the time limit for
+  /// each test in the suite, or each test case in parameterized tests in a suite.
```
stmontgomery: Should "…in **a** suite." be "…in _**the**_ suite." ?
```suggestion
  /// each test in the suite, or each test case in parameterized tests in the suite.
```

Sources/Testing/Testing.docc/LimitingExecutionTime.md
```diff
@@ -40,22 +40,25 @@ hour (60 x 60 seconds) to execute, the task in which it's running is
...
 
-### Time limits applied to test suites
+### Apply time limits to test suites
 
-When a time limit is applied to a test suite, it's recursively applied to all
-test functions and child test suites within that suite.
+When you apply a time limit to a test suite, the testing library recursively
+applies it to all test functions and child test suites within that suite.
+The time limit applies to each test in the test suite and any child test suites,
+or each test case for parameterized tests.
```
stmontgomery, line 57: I liked the _"For example, if a suite contains five…"_ example you gave in the symbol documentation, and feel it could be valuable here in the article too IMO.

PR #1226, opened 2025-07-16
Only clear exit test ID env var upon successful lookup

stmontgomery:
This changes the logic in `ExitTest.findInEnvironmentForEntryPoint()` to only clear the environment variable containing the ID of the exit test to run if an exit test is successfully located.

### Motivation:

This ensures that if a tool integrates with the testing library and has both its own built-in copy of the library but also may call into the ABI entry point, both usage scenarios can successfully look up the exit test. Without this fix, if an earlier attempt to look up an exit test fails, the environment variable would be cleared which prevents a subsequent lookup attempt from succeeding.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -235,7 +235,7 @@ extension ExitTest {
 
   /// The name of the environment variable used to identify the exit test to
   /// call in a spawned exit test process.
-  static let idEnvironmentVariableName = "SWT_EXIT_TEST_ID"
+  private static let _idEnvironmentVariableName = "SWT_EXIT_TEST_ID"
 
   /// Call the exit test in the current process.
   ///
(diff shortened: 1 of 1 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -233,6 +233,10 @@ extension ExitTest {
 #endif
   }
 
+  /// The name of the environment variable used to identify the exit test to
+  /// call in a spawned exit test process.
+  static let idEnvironmentVariableName = "SWT_EXIT_TEST_ID"
```
grynspan: Let's leave this out for the moment—it's probably better if we come back and define a "keys we care about" type inside the `Environment` type instead and redefine _all_ such keys to values of that type.

stmontgomery:
@swift-ci please test

PR #1231, opened 2025-07-17
Validate format of event stream version CLI argument when passed

stmontgomery:
This adds validation of the user-specified `--event-stream-version` (or `--experimental-event-stream-version`) command-line arguments when they are passed, to ensure they have a valid format.

### Motivation:

Currently, these flags only accept integers like `0`, `1`, etc. But I am moving towards making the entry point and event stream versions align with the Swift version, rather than having an independent versioning scheme which users must look up or keep track of. Once this effort is finished, in order to use the event stream format included in (for example) Swift 6.3, a user would pass `--event-stream-version 6.3`.

In https://github.com/swiftlang/swift-package-manager/pull/8944, I recently landed a SwiftPM change which will permit any arbitrary string for these arguments, and this means we need to begin validating the format of supported versions within the testing library. In a subsequent PR, I plan to introduce support for non-integer versions, at which point having some existing validation logic will be even more valuable.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
@@ -255,15 +255,16 @@ public struct __CommandLineArguments_v0: Sendable {
   /// whichever occurs first.
   public var eventStreamOutputPath: String?
 
-  /// The version of the event stream schema to use when writing events to
-  /// ``eventStreamOutput``.
+  /// The value of the `--event-stream-version` or `experimental-event-stream-version`
+  /// argument, representing the version of the event stream schema to use when
+  /// writing events to ``eventStreamOutput``.
   ///
   /// The corresponding stable schema is used to encode events to the event
   /// stream. ``ABI/Record`` is used if the value of this property is `0` or
   /// higher.
   ///
   /// If the value of this property is `nil`, the testing library assumes that
-  /// the newest available schema should be used.
+  /// the current supported (non-experimental) version should be used.
   public var eventStreamVersion: Int?
 
   /// The value(s) of the `--filter` argument.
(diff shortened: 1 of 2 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -255,15 +255,16 @@ public struct __CommandLineArguments_v0: Sendable {
   /// whichever occurs first.
   public var eventStreamOutputPath: String?
 
-  /// The version of the event stream schema to use when writing events to
-  /// ``eventStreamOutput``.
+  /// The value of the `--event-stream-version` or `experimental-event-stream-version`
```
suzannaratcliff: Nit: Should this be `--experimental-event-stream-version`?

suzannaratcliff (approved):

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: No, this is not an experimental flag.

grynspan (approved):

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
stmontgomery: Oops yes, thank you!

stmontgomery:
@swift-ci please test

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
stmontgomery: The experimental variant of `--event-stream-version` is still relevant; it indicates a user wishes to use an experimental _version_, and that flag serves as explicit acknowledgement.

grynspan: Oh, the dashes? Okay.

PR #1237, opened 2025-07-24
Don't use `_getSuperclass()` from the Swift runtime.

grynspan:
To determine if a class is a subclass of another, we can use `T.self is U.Type` in an implicitly-opened existential context. Prior to the Swift 6 language mode, type existentials weren't openable, so we had to recursively call a runtime function to check this. That's no longer a concern for us.

### Checklist:


grynspan:
@swift-ci test

briancroom:
Cool!

PR #1239, opened 2025-07-24
Introduce a new `ABI.VersionNumber` type.

grynspan:
This PR introduces and plumbs through an internal `ABI.VersionNumber` type that represents the JSON event stream version that should be used during a test run. This type supports integers (such as the current version 0) and semantic versions (e.g. 6.3 or 6.3.0). Care is taken to ensure backwards compatibility for ABI versions -1 and 0 which do not accept a string in this position.

The new type necessarily conforms to `Comparable` and `Codable`. I've adjusted `__CommandLineArguments_v0` to support this new type while maintaining backwards-compatibility with Xcode 16.

For the moment, these are all implementation details that have no impact on our supported interface as there is no newer JSON schema than 0. We intend to make changes to the schema in the next Swift release (assumed to be 6.3) that will require a Swift Evolution review.

### Checklist:


grynspan:
Related:

https://github.com/swiftlang/swift-package-manager/pull/8944
#1231

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/ABI/ABI.swift
```diff
@@ -18,7 +18,7 @@ extension ABI {
   /// A protocol describing the types that represent different ABI versions.
   protocol Version: Sendable {
     /// The numeric representation of this ABI version.
```
stmontgomery, line 20: nit: Can we generalize this description now since it's not a trivial number anymore?

Sources/Testing/ABI/ABI.VersionNumber.swift
```diff
@@ -0,0 +1,124 @@
...
+
+    /// The patch, revision, or bug fix version.
+    var patchComponent: Int8 = 0
+  }
+}
+
+extension ABI.VersionNumber {
+  init(_ majorComponent: Int8, _ minorComponent: Int8, _ patchComponent: Int8 = 0) {
+    self.init(majorComponent: majorComponent, minorComponent: minorComponent, patchComponent: patchComponent)
+  }
```
stmontgomery: Stylistic but I'd prefer labels e.g. `major:`. If we did, I don't think we'd need the word "component" on the labels despite the property names having that suffix

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -259,13 +259,31 @@ public struct __CommandLineArguments_v0: Sendable {
...
   ///
   /// If the value of this property is `nil`, the testing library assumes that
   /// the current supported (non-experimental) version should be used.
-  public var eventStreamVersion: Int?
+  public var eventStreamSchemaVersion: String? {
+    get {
+      eventStreamVersionNumber.map { String(describing: $0) }
+    }
+    set {
+      eventStreamVersionNumber = newValue.flatMap { ABI.VersionNumber($0) }
```
stmontgomery: Could we `preconditionFailure` or something if the passed-in string could not be successfully parsed despite being non-`nil`?

Tests/TestingTests/SwiftPMTests.swift
```diff
@@ -304,7 +337,8 @@ struct SwiftPMTests {
   @Test("Experimental ABI version requires --experimental-event-stream-version argument")
   func experimentalABIVersionNeedsExperimentalFlag() {
     #expect(throws: (any Error).self) {
-      let experimentalVersion = ABI.CurrentVersion.versionNumber + 1
+      var experimentalVersion = ABI.CurrentVersion.versionNumber
+      experimentalVersion.minorComponent += 1
```
stmontgomery, line 347: Does incrementing `patchComponent` also require the `--experimental-...` flag?

Tests/TestingTests/SwiftPMTests.swift
```diff
@@ -230,21 +230,54 @@ struct SwiftPMTests {
...
+    #expect(ABI.version(forVersionNumber: versionNumber) == nil)
+  }
+
   @Test("--event-stream-output-path argument (writes to a stream and can be read back)",
         arguments: [
           ("--event-stream-output-path", "--event-stream-version", ABI.v0.versionNumber),
           ("--experimental-event-stream-output", "--experimental-event-stream-version", ABI.v0.versionNumber),
-          ("--experimental-event-stream-output", "--experimental-event-stream-version", ABI.v1.versionNumber),
+          ("--experimental-event-stream-output", "--experimental-event-stream-version", ABI.v6_3.versionNumber),
+          ("--experimental-event-stream-output", "--experimental-event-stream-version", ABI.v6_3.versionNumber),
```
stmontgomery: Accidental duplicate?

Sources/Testing/ABI/ABI.VersionNumber.swift
```diff
@@ -0,0 +1,124 @@
...
+        )
+      }
+      self = result
+    }
+  }
+
+  func encode(to encoder: any Encoder) throws {
+    var container = encoder.singleValueContainer()
+    if majorComponent <= 0 && minorComponent == 0 && patchComponent == 0 {
+      try container.encode(majorComponent)
```
stmontgomery, line 125: Encoding and decoding pre-6.3 versions as plain integer instead of string is important to maintain backwards compatibility. Can you include a comment noting that rationale in both methods for future reference?

Sources/Testing/ABI/ABI.VersionNumber.swift
```diff
@@ -0,0 +1,124 @@
...
+  }
+}
+
+// MARK: - Codable
+
+extension ABI.VersionNumber: Codable {
+  init(from decoder: any Decoder) throws {
+    let container = try decoder.singleValueContainer()
+    if let number = try? container.decode(Int8.self) {
+      self.init(majorComponent: number)
```
stmontgomery: I wonder if we should also enforce that `number == 0` here, else throw an error, so that someone does not use (say) `7` as an integer expecting it to be interpreted the same as `"7.0.0"`? In other words, put a ceiling on the versions for which this backwards-compatibility shim is accepted.

Sources/Testing/ABI/ABI.VersionNumber.swift
grynspan: We have that initializer automatically. This is a shorthand initializer.

Sources/Testing/ABI/ABI.swift
grynspan: What do you want to change it to?

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: Sure.

Tests/TestingTests/SwiftPMTests.swift
grynspan: Yes.

Tests/TestingTests/SwiftPMTests.swift
grynspan: Yes, thanks.

Sources/Testing/ABI/ABI.VersionNumber.swift
grynspan: Sure.

Sources/Testing/ABI/ABI.VersionNumber.swift
grynspan: I don't think we should given that the _string_ `"7"` would be a valid version number, equivalent to `"7.0.0"`. Adding logic to block `7` and `"7"` is extra code for a case that we're able to handle, and we need to be able to parse this as an integer (`0`) forever anyway.

grynspan:
@swift-ci test

Sources/Testing/ABI/ABI.VersionNumber.swift
stmontgomery: Writing an explicit `init` supersedes the automatic memberwise one, no?

grynspan: Not if it's in an extension, no.

stmontgomery: Ah right, in an extension. OK!

Sources/Testing/ABI/ABI.swift
```diff
@@ -54,28 +81,28 @@ extension ABI {
...
-  ///   @Available("Swift Testing ABI", introduced: 1)
+  ///   @Available(Swift, introduced: 6.3)
   /// }
   @_spi(Experimental)
-  public enum v1: Sendable, Version {
-    static var versionNumber: Int {
-      1
+  public enum v6_3: Sendable, Version {
+    static var versionNumber: VersionNumber {
+      VersionNumber(6, 3)
```
stmontgomery, line 105: These were the `init` usages which made me suggest having explicit `major: ...` labels, to clarify what the numbers meant.

stmontgomery (approved):
LGTM, thank you for the iteration! My final comment was that I thought explicit labels e.g. `majorComponent:...` for the initializer usages would be helpful, but it's stylistic and I leave it up to you

Sources/Testing/ABI/ABI.swift
grynspan: This is, again, just for convenience. If it really bothers you I can remove the convenience initializer.

grynspan: Made the convenience initializer take `_const` values so that you can only use it with integer literals so it's always clear at the call site.

grynspan:
@swift-ci test

grynspan:
macOS CI is backlogged. Tested at desk.

PR #545, opened 2024-07-17
Remove the legacy Xcode 16 entry point function on non-Darwin platforms.

grynspan:
This PR removes `swt_copyABIEntryPoint_v0()`, which is an older version of the approved `swt_abiv0_getEntryPoint()` entry point function used by Xcode 16 Beta 1 but not used on non-Darwin platforms.

The symbol remains available on Darwin for the benefit of Xcode 16 beta users. It will be removed from the Darwin builds in a future update.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1240, opened 2025-07-24
Remove `swt_copyABIEntryPoint_v0()`.

grynspan:
`swt_copyABIEntryPoint_v0()` is an older alias of `swt_abiv0_getEntryPoint()` with a slightly different signature that was used by Xcode 16.0. Xcode 16.3 and later use the newer function, so we can remove the older one now.

`swt_copyABIEntryPoint_v0()` is not exported on non-Apple platforms.

### Checklist:


grynspan:
@swift-ci test

grynspan:
See also: #545

PR #1241, opened 2025-07-25
Silence a warning on FreeBSD/OpenBSD when calling `pthread_set_name_np()`.

grynspan:
On FreeBSD and OpenBSD, `pthread_set_name_np()` returns `void`, so we don't need to suppress the result (assigning it to `_`) and the Swift compiler complains that we do.

### Checklist:


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

PR #1165, opened 2025-06-19
Enable exit test value capturing

grynspan:
This PR enables the value capture subfeature of exit tests and adds documentation for it.

Resolves #1157.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1243, opened 2025-07-29
Avoid passive voice and clarify macro in exit test docs.

iamleeg:
Review recent changes in the exit test documentation.

### Motivation:

@grynspan asked me to review #1165 but it was already merged by the time I got there, so here are my review suggestions as a fresh PR.

### Modifications:

- Specified which macro we mean by "this macro" in the exit tests docs.
- Removed instances of the passive voice.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/exit-testing.md
+++ b/Sources/Testing/Testing.docc/exit-testing.md
@@ -128,13 +128,14 @@ value using the `as` operator:
 
 Every value you capture in an exit test must conform to [`Sendable`](https://developer.apple.com/documentation/swift/sendable)
 and [`Codable`](https://developer.apple.com/documentation/swift/codable). The
-testing library passes each value to the exit test body by following this process:
+testing library passes each value to the exit test body by performing the
+following steps:
 
 1. It encodes each value using [`encode(to:)`](https://developer.apple.com/documentation/swift/encodable/encode(to:))
-in the parent process
+   in the parent process
 2. It passes the encoded value to the child process
 3. It decodes each value using [`init(from:)`](https://developer.apple.com/documentation/swift/decodable/init(from:))
-in the child process
+   in the child process
 
 If a captured value's type does not conform to both `Sendable` and `Codable`, or
 if the value is not explicitly specified in the exit test body's capture list,


Sources/Testing/Testing.docc/exit-testing.md
```diff
@@ -107,11 +107,12 @@ func `Customer won't eat food unless it's nutritious`(_ food: Food) async {
 }
 ```
 
-- Note: If you use this macro with a Swift compiler version lower than 6.3, it
-  doesn't support capturing state.
+- Note: If you use the ``expect(processExitsWith:observing:_:sourceLocation:performing:)``
```
grynspan: `require` too.

Sources/Testing/Testing.docc/exit-testing.md
```diff
@@ -126,10 +127,10 @@ value using the `as` operator:
...
 
 Every value you capture in an exit test must conform to [`Sendable`](https://developer.apple.com/documentation/swift/sendable)
-and [`Codable`](https://developer.apple.com/documentation/swift/codable). Each
-value is encoded by the parent process using [`encode(to:)`](https://developer.apple.com/documentation/swift/encodable/encode(to:))
-and is decoded by the child process [`init(from:)`](https://developer.apple.com/documentation/swift/decodable/init(from:))
-before being passed to the exit test body.
+and [`Codable`](https://developer.apple.com/documentation/swift/codable). The
+testing library encodes each value using [`encode(to:)`](https://developer.apple.com/documentation/swift/encodable/encode(to:))
+in the parent process, and decodes it using [`init(from:)`](https://developer.apple.com/documentation/swift/decodable/init(from:))
+in the child process, before passing it to the exit test body.
```
grynspan: The new commas are incorrect, at least the way I was taught (circa 1854.)

grynspan (changes requested):

Sources/Testing/Testing.docc/exit-testing.md
iamleeg: The words "and", and "before", are acting as coordinating conjunctions between independent clauses in this sentence. While you can always find a grammar "rule" that says you should use a comma in the same place that another "rule" says you shouldn't use one, I think the commas separate the independent clauses here.

Sources/Testing/Testing.docc/exit-testing.md
iamleeg: Ah I missed that because the original text had "this macro" and there were only two to choose from, of which `require` wasn't one.

Sources/Testing/Testing.docc/exit-testing.md
grynspan: We could split the difference and make this a list of steps instead?

grynspan (approved):

PR #1246, opened 2025-07-30
Various `VersionNumber` changes of merit.

grynspan:
This PR does things to `ABI.VersionNumber`:

- Renames it to `VersionNumber` as we do have some use cases that aren't related to JSON schema versioning. I initially didn't want to make this type a general version number type, but it's just too useful not to do so. Alas.
- Changes the type of the `swiftStandardLibraryVersion` global variable to `VersionNumber?`.
- Changes the type of the `glibcVersion` global variable to `VersionNumber`.
- Adds `swiftCompilerVersion` representing the version of the Swift compiler used to compile Swift Testing. We need this value when computing the JSON schema version (see next bullet.)
- Clamps the range of supported JSON schema versions to the Swift compiler version _unless_ we've explicitly defined a schema version higher than it:

  | Compiler | Highest Defined Schema | Requested | Result |
  |-|-|-|-|
  | 1.0 | 1.0 | 1.0 | 1.0 | | 2.0 | 1.0 | 1.0 | 1.0 | | 1.0 | 2.0 | 1.0 | 1.0 |
  | 1.0 | 1.0 | 2.0 | `nil` |
  | 2.0 | 2.0 | 1.0 | 1.0 |
  | 2.0 | 1.0 | 2.0 | 1.0 |
  | 1.0 | 2.0 | 2.0 | 2.0 |
  | 2.0 | 2.0 | 2.0 | 2.0 |

  The reasoning here is that, when we're built with a given compiler version, we presumably know about all JSON schema versions up to and including the one aligned with that compiler, so if you ask for the schema version aligned with the compiler, it's equivalent to whatever we support that's less than or equal to the compiler version. But if you ask for something greater than the compiler version, and we haven't defined it, we don't know anything about it and can't provide it.

  This reasoning breaks down somewhat if you build an old version of the Swift Testing package with a new compiler, but in general we don't support that sort of configuration for very long (and we can't predict the future anyway.)

### Checklist:

### DIFF
--- a/Sources/_TestingInternals/include/Versions.h
+++ b/Sources/_TestingInternals/include/Versions.h
@@ -16,6 +16,12 @@
 
 SWT_ASSUME_NONNULL_BEGIN
 
+/// Get the version of the compiler used to build the testing library.
+///
+/// - Returns: An integer containing the packed major, minor, and patch
+///   components of the compiler version. For more information, see
+///   [ClangImporter.cpp](https://github.com/swiftlang/swift/blob/36246a2c8e9501cd29a75f34c9631a8f4e2e1e9b/lib/ClangImporter/ClangImporter.cpp#L647)
+///   in the Swift repository.
 static inline uint64_t swt_getSwiftCompilerVersion(void) {
 #if defined(__SWIFT_COMPILER_VERSION)
   return __SWIFT_COMPILER_VERSION;
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
@@ -78,7 +78,7 @@ extension ABI {
     /// - Warning: Tags are not yet part of the JSON schema.
     ///
     /// @Metadata {
-    ///   @Available("Swift Testing ABI", introduced: 1)
+    ///   @Available("Swift Testing ABI", introduced: 6.3)
     /// }
     var _tags: [String]?
 


grynspan:
@swift-ci test

grynspan:
@swift-ci test Linux

grynspan:
@swift-ci test

grynspan:
@swift-ci test macOS

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/_TestingInternals/include/Versions.h
```diff
@@ -16,6 +16,14 @@
 
 SWT_ASSUME_NONNULL_BEGIN
 
+static inline uint64_t swt_getSwiftCompilerVersion(void) {
```
stmontgomery, line 25: Could this be documented, for example to show what format the number is expected to be in?

Sources/Testing/Support/Versions.swift
```diff
@@ -141,23 +142,50 @@ var targetTriple: String? {
...
-/// machine-readable. It is added to the output of a test run when using
-/// an event writer.
+/// This value is unavailable on some earlier Apple runtime targets. On those
+/// targets, this property has a value of `5.0.0`.
 ///
 /// This value is not part of the public interface of the testing library.
-let swiftStandardLibraryVersion: String = {
-  if #available(_swiftVersionAPI, *) {
-    return String(describing: _SwiftStdlibVersion.current)
+let swiftStandardLibraryVersion: VersionNumber? = {
```
stmontgomery, line 149: It's not clear to me under what conditions this would be `nil` despite reading the documentation, so I'm curious why it's Optional?

Sources/Testing/ABI/ABI.swift
```diff
@@ -55,7 +58,24 @@ extension ABI {
...
+      //
+      // Note that it is possible for the Swift compiler version to be lower
+      // than the highest defined ABI version (e.g. if you use a 6.2 toolchain
+      // to build this package's release/6.3 branch with a 6.3 ABI defined.)
+      //
+      // Note also that building an old version of Swift Testing with a newer
+      // compiler may produce incorrect results here. We don't generally support
+      // that configuration though.
+      if versionNumber > swiftCompilerVersion {
+        return nil
```
stmontgomery, line 74: Thanks!

Sources/Testing/Support/Versions.swift
grynspan: The stdlib function we call has limited availability.

Sources/_TestingInternals/include/Versions.h
grynspan: Oops, yes.

grynspan: Done.

grynspan:
@swift-ci test

PR #1247, opened 2025-07-30
Remove experimental spi from issue severity

suzannaratcliff:
Remove experimental spi from issue severity

### Motivation:

The Issue Severity proposal was approved.

https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0013-issue-severity-warning.md

### Modifications:

Removed experimental spi from issue severity and updated comments.

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -65,7 +65,7 @@ extension Issue {
   /// or ``require(_:_:sourceLocation:)-5l63q`` macros.)
   @discardableResult public static func record(
     _ comment: Comment? = nil,
-    severity: Severity,
+    severity: Severity = .error,
     sourceLocation: SourceLocation = #_sourceLocation
   ) -> Self {
     let sourceContext = SourceContext(backtrace: .current(), sourceLocation: sourceLocation)
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -50,7 +50,7 @@ extension Issue {
     return self
   }
 
-  /// Record an issue when a running test fails unexpectedly.
+  /// Records an issue that a test encounters while it's running.
   ///
   /// - Parameters:
   ///   - comment: A comment describing the expectation.
@@ -63,6 +63,8 @@ extension Issue {
   /// represented as an expectation (using the ``expect(_:_:sourceLocation:)``
   /// or ``require(_:_:sourceLocation:)-5l63q`` macros.)
   @_disfavoredOverload
+  @_documentation(visibility: private)
+  @available(*, deprecated, message: "Use record(_:severity:sourceLocation:) instead.")
   @discardableResult public static func record(
     _ comment: Comment? = nil,
     sourceLocation: SourceLocation = #_sourceLocation
@@ -70,11 +72,13 @@ extension Issue {
     record(comment, severity: .error, sourceLocation: sourceLocation)
   }
 
-  /// Record an issue when a running test and an issue occurs.
+  /// Records an issue that a test encounters while it's running.
   ///
   /// - Parameters:
   ///   - comment: A comment describing the expectation.
-  ///   - severity: The severity level of the issue.  The testing library marks the test as failed if the severity is greater than ``Issue/Severity/warning``.  The default is ``Issue/Severity/error``.
+  ///   - severity: The severity level of the issue.  The testing library marks the
+  ///   test as failed if the severity is greater than ``Issue/Severity/warning``.
+  ///   The default is ``Issue/Severity/error``.
   ///   - sourceLocation: The source location to which the issue should be
   ///     attributed.
   ///
@@ -83,6 +87,7 @@ extension Issue {
   /// Use this function if, while running a test, an issue occurs that cannot be
   /// represented as an expectation (using the ``expect(_:_:sourceLocation:)``
   /// or ``require(_:_:sourceLocation:)-5l63q`` macros.)
+  /// 
   /// @Metadata {
   ///   @Available(Swift, introduced: 6.3)
   /// }
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -50,10 +50,11 @@ extension Issue {
     return self
   }
 
-  /// Record an issue when a running test fails unexpectedly.
+  /// Record an issue when a running test and an issue occurs.
   ///
   /// - Parameters:
   ///   - comment: A comment describing the expectation.
+  ///   - severity: The severity level of the issue.  This factor impacts whether the issue constitutes a failure.
   ///   - sourceLocation: The source location to which the issue should be
   ///     attributed.
   ///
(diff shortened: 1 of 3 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -62,27 +63,6 @@ extension Issue {
...
-  ///
-  /// - Returns: The issue that was recorded.
-  ///
-  /// Use this function if, while running a test, an issue occurs that cannot be
-  /// represented as an expectation (using the ``expect(_:_:sourceLocation:)``
-  /// or ``require(_:_:sourceLocation:)-5l63q`` macros.)
-  @_spi(Experimental)
   @discardableResult public static func record(
     _ comment: Comment? = nil,
     severity: Severity,
```
stmontgomery: Now, this should gain a default value of `.error`. That couldn't be added until now, when the API is promoted, because it would've been ambiguous with the prior overload
```suggestion
    severity: Severity = .error,
```

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -50,10 +50,11 @@ extension Issue {
     return self
   }
 
-  /// Record an issue when a running test fails unexpectedly.
+  /// Record an issue when a running test and an issue occurs.
   ///
   /// - Parameters:
   ///   - comment: A comment describing the expectation.
+  ///   - severity: The severity level of the issue.  This factor impacts whether the issue constitutes a failure.
```
stmontgomery: We should note the default value (error)

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -50,10 +50,11 @@ extension Issue {
     return self
   }
 
-  /// Record an issue when a running test fails unexpectedly.
+  /// Record an issue when a running test and an issue occurs.
```
grynspan: Please have @iamleeg review these documentation comments. :)

stmontgomery:
To enable the posting of issues with warning severity to event handlers, we need to also change `Configuration.EventHandlingOptions.isWarningIssueRecordedEventEnabled` to be `true` by default

Sources/Testing/Issues/Issue+Recording.swift
grynspan: I believe we need to remove the old overload if we do that (which was part of the plan).

grynspan:
> To enable the posting of issues with warning severity to event handlers, we need to also change Configuration.EventHandlingOptions.isWarningIssueRecordedEventEnabled to be true by default

But only for `v6_3` onward and presumably the Xcode 16 legacy path, right? `v0` can't handle them.

Sources/Testing/Issues/Issue.swift
```diff
@@ -84,7 +84,6 @@ public struct Issue: Sendable {
   ///
   /// - ``warning``
   /// - ``error``
-  @_spi(Experimental)
   public enum Severity: Sendable {
```
stmontgomery, line 91: Ah, I forgot about **availability**: We'll want to annotate each of the newly added, non-`@_spi` APIs with Swift 6.3 availability in DocC, such as:
```swift
/// @Metadata {
///   @Available(Swift, introduced: 6.3)
/// }
```
We have several examples of this throughout the codebase.

stmontgomery:
> > To enable the posting of issues with warning severity to event handlers, we need to also change `Configuration.EventHandlingOptions.isWarningIssueRecordedEventEnabled` to be `true` by default.
> 
> But only for `v6_3` onward and presumably the Xcode 16 legacy path, right? `v0` can't handle them.

Compatibility with older event stream versions is handled by this entry point logic:

https://github.com/swiftlang/swift-testing/blob/f0047e6fb754fa52b306d642ec8c1257c16d5eae/Sources/Testing/ABI/EntryPoints/EntryPoint.swift#L625-L628

suzannaratcliff:
@stmontgomery I think that `Configuration.EventHandlingOptions.isWarningIssueRecordedEventEnabled` is already set to true

stmontgomery:
> @stmontgomery I think that `Configuration.EventHandlingOptions.isWarningIssueRecordedEventEnabled` is already set to true

I don't think it is; maybe we're talking about different things. I'm referring to this line in `Configuration.swift`:

https://github.com/swiftlang/swift-testing/blob/f0047e6fb754fa52b306d642ec8c1257c16d5eae/Sources/Testing/Running/Configuration.swift#L193

Sources/Testing/Issues/Issue+Recording.swift
iamleeg: There's definitely something weird here…"when a running test occurs"? I think this should be something like:
```suggestion
  /// Records an issue that a test encounters while it's running.
```

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -62,18 +62,20 @@ extension Issue {
...
   ///
   /// - Parameters:
   ///   - comment: A comment describing the expectation.
-  ///   - severity: The severity of the issue.
+  ///   - severity: The severity level of the issue.  This factor impacts whether the
+  ///     issue constitutes a failure.  The default is .error.
```
iamleeg: Better to tell developers what level constitutes a failure:
```suggestion
  ///   - severity: The severity level of the issue.  The testing library marks the test as failed if the severity is greater than `.mildInconvenience`.  The default is `.error`.
```

iamleeg: I couldn't get the line wraps correct in the GitHub UI, sorry.

iamleeg (changes requested):
Small but important issues in the documentation.

Sources/Testing/Issues/Issue+Recording.swift
grynspan: These should use symbolic links (double backticks, produces links in DocC to the relevant enum cases.)

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -62,18 +62,19 @@ extension Issue {
...
     record(comment, severity: .error, sourceLocation: sourceLocation)
   }
 
-  /// Record an issue when a running test fails unexpectedly.
+  /// Record an issue when a running test and an issue occurs.
   ///
   /// - Parameters:
   ///   - comment: A comment describing the expectation.
-  ///   - severity: The severity of the issue.
+  ///   - severity: The severity level of the issue.  The testing library marks the test as failed if the severity is greater than ``Issue/Severity/warning``.  The default is ``Issue/Severity/error``.
```
stmontgomery: Thanks for switching to DocC symbol references for the enum cases. Last thing here is a style nitpick: can you wrap the lines to 80 characters, similar to what we do elsewhere? (And indent to match)

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -62,18 +62,19 @@ extension Issue {
   /// Use this function if, while running a test, an issue occurs that cannot be
   /// represented as an expectation (using the ``expect(_:_:sourceLocation:)``
   /// or ``require(_:_:sourceLocation:)-5l63q`` macros.)
+  @_disfavoredOverload
   @discardableResult public static func record(
```
stmontgomery, line 68: The reason we need to keep this overload is to avoid a source breakage in case any clients are taking a reference to this function. The `@_disfavoredOverload` you added is good — that will help ensure that despite this overload having fewer parameters, it will be disfavored.

I think we should consider adding two other attributes to this older overload:

- `@_documentation(visibility: private)`, to prevent this overload from appearing in rendered API documentation.
- Deprecated availability such as `@available(*, deprecated, message: "Use record(_:severity:sourceLocation:) instead.")`. This is so that anyone who is currently obtaining a function reference will get a deprecation warning and be prompted to switch to the newer replacement overload.

Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -82,10 +83,12 @@ extension Issue {
   /// Use this function if, while running a test, an issue occurs that cannot be
   /// represented as an expectation (using the ``expect(_:_:sourceLocation:)``
   /// or ``require(_:_:sourceLocation:)-5l63q`` macros.)
-  @_spi(Experimental)
+  /// @Metadata {
```
stmontgomery, line 91: For each of these `@Metadata` blocks, we typically include a blank line above them to separate them from ordinary documentation

grynspan: (Makes it easier to read when reviewing the source code or Swift textual interface. Makes no difference in DocC AFAIK.)

Sources/Testing/Running/Configuration.swift
```diff
@@ -187,10 +187,7 @@ public struct Configuration: Sendable {
     ///
     /// By default, events matching this criteria are not delivered to event
     /// handlers since this is an experimental feature.
-    ///
-    /// - Warning: Warning issues are not yet an approved feature.
-    @_spi(Experimental)
-    public var isWarningIssueRecordedEventEnabled: Bool = false
+    public var isWarningIssueRecordedEventEnabled: Bool = true
```
grynspan, line 187: The last sentence of the documentation comment is no longer true.

PR #1256, opened 2025-08-08
Emit a diagnostic if a display name string is empty

stmontgomery:
This introduces an error diagnostic emitted from the testing library's `@Test` and `@Suite` macros if the display name string literal is empty. For example:

```swift
@Test("") // 🛑 Attribute 'Test' specifies an empty display name for this function (from macro 'Test')
func example() {
  ...
}
```

Fixes rdar://157603034

### Motivation:

It's not recommended for the display name string to be empty since it can cause confusing test results.

### Checklist:


stmontgomery:
@swift-ci please test

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -643,6 +643,22 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+  /// but should not be.
+  ///
+  /// - Parameters:
+  ///   - displayNameExpr: The display name string literal expression.
+  ///
+  /// - Returns: A diagnostic message.
+  static func emptyDisplayName(_ displayNameExpr: StringLiteralExprSyntax) -> Self {
+    Self(
+      syntax: Syntax(displayNameExpr),
+      message: "Display name string should not be empty",
```
stmontgomery: Open to feedback on the phrasing

stmontgomery:
@swift-ci please test Linux

stmontgomery:
@swift-ci please test Linux

grynspan:
Should we give it a fix-it to remove the display name? Also, have you read the compiler diagnostic style guide? It might help you find the right voice for the message.

Sources/TestingMacros/Support/Additions/StringLiteralExprSyntaxAdditions.swift
```diff
@@ -0,0 +1,30 @@
...
+import SwiftSyntax
+
+extension StringLiteralExprSyntax {
+  /// Whether or not this string literal expression is completely empty.
+  ///
+  /// The value of this property is `false` even if the literal contains one or
+  /// more interpolation segments whose evaluated expressions are an empty
+  /// string. The value is only `true` if this string literal contains only one
+  /// empty string segment.
+  var isEmptyLiteral: Bool {
```
grynspan: This isn't necessary as you can just check if the represented string is empty instead.

grynspan (changes requested):

Sources/TestingMacros/Support/Additions/StringLiteralExprSyntaxAdditions.swift
stmontgomery: Great, thanks—I couldn't find anything, but I guess that's because the convenience getter I needed was in `SwiftParser` which wasn't imported in this file (yet)

Sources/TestingMacros/Support/DiagnosticMessage.swift
grynspan: I'll mull it over and give you feedback Monday.

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test Linux

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -643,6 +643,22 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+      severity: .warning
+    )
```
grynspan, line 678: Should be easy enough to provide a fix-it to remove the empty display name. Bonus points for offering a second fix-it to turn it into a placeholder and force the test author to write something.

Sources/TestingMacros/Support/DiagnosticMessage.swift
grynspan: How about:

> Attribute \(_macroName(attribute)) specifies an empty display name for \(_kindString(for: decl))

Which is consistent with this diagnostic we already have:

> Attribute \(_macroName(attribute)) specifies display name '\(displayNameFromAttribute.representedLiteralValue!)' for \(_kindString(for: decl)) with implicit display name '\(decl.name.textWithoutBackticks)'

grynspan: See:

```swift
  static func declaration(
    _ decl: some NamedDeclSyntax,
    hasExtraneousDisplayName displayNameFromAttribute: StringLiteralExprSyntax,
    fromArgument argumentContainingDisplayName: LabeledExprListSyntax.Element,
    using attribute: AttributeSyntax
  ) -> Self {
```

stmontgomery:
> Should we give it a fix-it to remove the display name?

Added a fix-it to remove the display name argument!

> Also, have you read the compiler diagnostic style guide? It might help you find the right voice for the message.

I've now read the [guide](https://github.com/swiftlang/swift/blob/main/docs/Diagnostics.md). Among its recommendations I saw:

> …a particular diagnostic should be a warning if the intent of the code is clear and it won't immediately result in a crash.

Based on that, I feel like the severity of this should be a warning and not an error. In terms of the phrasing of the diagnostic message I'm still open to suggestions. I didn't see much in the style guide that would contradict my current phrasing but I could also imagine saying _"Display name string is empty"_ rather than saying it _"should not"_ be empty.

stmontgomery:
@swift-ci test

Sources/TestingMacros/Support/DiagnosticMessage.swift
stmontgomery: Missed this reply until now. Sounds good

stmontgomery:
@swift-ci test

stmontgomery:
@swift-ci test Windows

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -643,6 +643,37 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+    fromArgument argumentContainingDisplayName: LabeledExprListSyntax.Element,
+    using attribute: AttributeSyntax
+  ) -> Self {
+    Self(
+      syntax: Syntax(displayNameExpr),
+      message: "Attribute \(_macroName(attribute)) specifies an empty display name for \(_kindString(for: decl))",
+      severity: .warning,
+      fixIts: [
+        FixIt(
+          message: MacroExpansionFixItMessage("Remove '\(displayNameExpr.representedLiteralValue!)'"),
```
grynspan: We know the literal value is `""` so no need to include it in the format string.

Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -643,6 +643,37 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+  /// - Returns: A diagnostic message.
+  static func declaration(
+    _ decl: some NamedDeclSyntax,
+    hasEmptyDisplayName displayNameExpr: StringLiteralExprSyntax,
+    fromArgument argumentContainingDisplayName: LabeledExprListSyntax.Element,
+    using attribute: AttributeSyntax
+  ) -> Self {
+    Self(
+      syntax: Syntax(displayNameExpr),
+      message: "Attribute \(_macroName(attribute)) specifies an empty display name for \(_kindString(for: decl))",
```
grynspan: I think we're missing the name of `decl` here, so it doesn't read quite right. Sorry. Try:

```suggestion
      message: "Attribute \(_macroName(attribute)) specifies an empty display name for this \(_kindString(for: decl))",
```

stmontgomery:
@swift-ci test

Sources/TestingMacros/Support/DiagnosticMessage.swift
stmontgomery: Now added a second fix-it to fill in the display name with an editor placeholder

stmontgomery:
@swift-ci test

PR #1258, opened 2025-08-09
Remove optionality from `_AttachableImageWrapper.wrappedValue` on Windows.

grynspan:
This PR makes `_AttachableImageWrapper.wrappedValue` non-optional on Windows (as it is on Darwin.) As currently implemented, it allows for failures when calling `CopyImage()` and `CopyIcon()`, but in practice the only way these can fail is due to heap exhaustion[^msDocs]. Swift treats allocation failures as almost universally fatal, so we should do the same.

[^msDocs]: Microsoft's documentation for these functions does not list any other failure modes.

### Checklist:


grynspan:
@swift-ci test

PR #1262, opened 2025-08-11
Enhance documentation for URL attachments.

### LINKED ISSUE #1255: Enhance documentation for URL attachments
Enhance documentation for [`Attachment.init(contentsOf:named:sourceLocation:)`](https://developer.apple.com/documentation/testing/attachment/init(contentsof:named:sourcelocation:)). In particular, provide info on what kind of URLs can be passed in, what the initializer awaits on, and how to use this (since there's intentionally no equivalent `record()` overload.)

grynspan:
Enhance documentation for [`Attachment.init(contentsOf:named:sourceLocation:)`](https://developer.apple.com/documentation/testing/attachment/init(contentsof:named:sourcelocation:)). In particular, provide info on what kind of URLs can be passed in, what the initializer awaits on, and how to use this (since there's intentionally no equivalent `record()` overload.)

Resolves #1255.

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
+++ b/Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
@@ -51,7 +51,7 @@ extension Attachment where AttachableValue == _AttachableURLWrapper {
   ///
   /// - Throws: Any error that occurs attempting to read from `url`.
   ///
-  /// Use this initialize to create an instance of ``Attachment`` that
+  /// Use this initializer to create an instance of ``Attachment`` that
   /// represents a local file or directory:
   ///
   /// ```swift
--- a/Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
+++ b/Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
@@ -62,14 +62,14 @@ extension Attachment where AttachableValue == _AttachableURLWrapper {
   ///
   /// When you call this initializer and pass it the URL of a file, it reads or
   /// maps the contents of that file into memory. When you call this initializer
-  /// and pass it the URL of a directory, it creates a temporary ZIP file of the
+  /// and pass it the URL of a directory, it creates a temporary zip file of the
   /// directory before reading or mapping it into memory. These operations may
   /// take some time, so this initializer suspends the calling task until they
   /// are complete.
   ///
   /// - Important: This initializer supports creating attachments from file URLs
-  ///   only. It does not support creating attachments from resources on the
-  ///   Internet or the local network.
+  ///   only. If you pass it a URL other than a file URL, such as an HTTPS URL,
+  ///   the testing library throws an error.
   ///
   /// @Metadata {
   ///   @Available(Swift, introduced: 6.2)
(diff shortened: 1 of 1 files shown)


Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
```diff
@@ -51,6 +51,26 @@ extension Attachment where AttachableValue == _AttachableURLWrapper {
   ///
   /// - Throws: Any error that occurs attempting to read from `url`.
   ///
+  /// Use this initialize to create an instance of ``Attachment`` that
```
iamleeg:
```suggestion
  /// Use this initializer to create an instance of ``Attachment`` that
```

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
```diff
@@ -51,6 +51,26 @@ extension Attachment where AttachableValue == _AttachableURLWrapper {
...
+  ///
+  /// ```swift
+  /// let url = try await FoodTruck.saveMenu(as: .pdf)
+  /// let attachment = try await Attachment(contentsOf: url)
+  /// Attachment.record(attachment)
+  /// ```
+  ///
+  /// When you call this initializer and pass it the URL of a file, it reads or
+  /// maps the contents of that file into memory. When you call this initializer
+  /// and pass it the URL of a directory, it creates a temporary ZIP file of the
```
iamleeg:
```suggestion
  /// and pass it the URL of a directory, it creates a temporary `zip` file of the
```

I don't think ZIP is an initialism in this context.

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
```diff
@@ -51,6 +51,26 @@ extension Attachment where AttachableValue == _AttachableURLWrapper {
...
+  /// directory before reading or mapping it into memory. These operations may
+  /// take some time, so this initializer suspends the calling task until they
+  /// are complete.
+  ///
+  /// - Important: This initializer supports creating attachments from file URLs
+  ///   only. It does not support creating attachments from resources on the
+  ///   Internet or the local network.
```
iamleeg: Does it throw an error if someone tries that? Might be worth calling that out if so, and otherwise explaining what errors it encounters.

iamleeg (approved):
Looks good, just a few suggestions.

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
grynspan: Code voice also isn't correct as it's just the name of the filetype. 🤔

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
grynspan: Yes, we throw an error in that case.

Sources/Overlays/_Testing_Foundation/Attachments/Attachment+URL.swift
grynspan: https://en.wikipedia.org/wiki/ZIP_(file_format)

Historically "ZIP" is written uppercased even though it's not an initialism. Is there a style guide rule here?

iamleeg: Not that I can find, no. An example, though: over where Apple document the zip file type, they use lower case. https://developer.apple.com/documentation/uniformtypeidentifiers/uttype-swift.struct/zip

grynspan: Lowercase it is.

grynspan:
@swift-ci test

PR #1263, opened 2025-08-12
Remove a workaround for a `URL.path` bug on Windows.

grynspan:
This PR removes a workaround for a bug in Foundation's `URL` on Windows where the `path` property would not be formatted correctly. See https://github.com/swiftlang/swift-foundation/pull/1038.

### Checklist:


grynspan:
@swift-ci test

PR #1266, opened 2025-08-13
Allow attaching an `IWICBitmapSource` instance directly.

grynspan:
This PR adjusts the experimental WIC bitmap attachment support I recently added so that a test author can attach an instance of `IWICBitmapSource` (which is the parent type of the already-supported `IWICBitmap`.) Protocols and conformances are adjusted to match.

This PR then explicitly adds conformances to every COM class in WIC that subclasses `IWICBitmapSource` since we can't see COM class inheritance in Swift (yet?)

I could have left the protocols as-is, but then you'd have to convert an `IWICBitmapSource` to an `IWICBitmap` (by allocating a new COM object) then cast it back to an `IWICBitmapSource` (by calling `QueryInterface()` and juggling the refcount) in order to actually attach it.

### Checklist:


grynspan:
@swift-ci test

Sources/Overlays/_Testing_WinSDK/Attachments/IWICBitmapSource+AttachableAsIWICBitmapSource.swift
```diff
@@ -0,0 +1,119 @@
...
+  ///
+  /// - Throws: Any error that occurs while calling `QueryInterface()`. In
+  ///   practice, this function is not expected to throw an error as it should
+  ///   always be possible to cast a valid instance of `IWICBitmap` to
+  ///   `IWICBitmapSource`.
+  ///
+  /// - Important: This function consumes a reference to `self` even if the cast
+  ///   fails.
+  consuming func cast(to _: IWICBitmapSource.Type) throws -> UnsafeMutablePointer<IWICBitmapSource> {
+    try self.withMemoryRebound(to: IUnknown.self, capacity: 1) { `self` in
```
grynspan, line 81: The casts in this file to `IUnknown` are safe insofar as we use them to access fixed members of the COM vtable. The casts would become unsafe if we allowed the resulting pointers to escape _and_ if any of the types we use them on have multiple non-virtual inheritance to `IUnknown`.

grynspan:
@swift-ci test

Sources/Overlays/_Testing_WinSDK/Attachments/IWICBitmapSource+AttachableAsIWICBitmapSource.swift
stmontgomery: Can we capture this in a code comment for future reference?

stmontgomery (approved):

grynspan:
@swift-ci test

PR #1267, opened 2025-08-14
Test that the fixups are generated if display name is empty

jerryjrchen:
No functional change: just adds new test cases for the change in #1256

### Motivation:

Slightly improves effective test coverage for Stuart's earlier change.

### Modifications:

* Moved the tests previously added to `apiMisuseErrors` -> `apiMisuseErrorsIncludingFixIts`, and include the expected fixits

  It would probably be ok to leave them in `apiMisuseErrors` as well, but I think it would be kinda redundant since we also test the expected message along with the fixits already.

### Checklist:

### DIFF
--- a/Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
+++ b/Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
@@ -273,8 +273,7 @@ struct TestDeclarationMacroTests {
          ),
        #"@Suite("") struct S {}"#:
        (
-         message:
-         "Attribute 'Suite' specifies an empty display name for this structure",
+         message: "Attribute 'Suite' specifies an empty display name for this structure",
          fixIts: [
            ExpectedFixIt(
              message: "Remove display name argument",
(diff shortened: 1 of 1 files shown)


stmontgomery (approved):

Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
```diff
@@ -241,6 +233,63 @@ struct TestDeclarationMacroTests {
...
+                 .replace(
+                   oldSourceCode: ##"#""#"##,
+                   newSourceCode: #""\#(EditorPlaceholderExprSyntax("display name"))""#)
+               ])
+           ]
+         ),
+       #"@Suite("") struct S {}"#:
+       (
+         message:
+         "Attribute 'Suite' specifies an empty display name for this structure",
```
stmontgomery: nit: I would typically place these on the same line

jerryjrchen:
@swift-ci test

PR #1279, opened 2025-08-20
Include severity and isFailure for Issues in the published JSON event stream

jerryjrchen:
### Motivation:

- Proposed and accepted in ST-0013 [^1], so we can move ahead and include this in the event stream.

- This will _not_ be available in event stream version 0 as this is considered a breaking change [^2]

### Modifications:

- Promote existing _severity/_isFailure fields -> severity/isFailure without the underscores

  This effectively makes them part of the official JSON ABI. The underscored names will no longer be accessible, but they were always experimental so this is not considered a blocker.

- Only emit severity/isFailure when using event stream version >= 6.3

  This removes access to the underscored names from earlier versions, but again, these were experimental.

- Remove hidden `__CommandLineArguments_v0` argument `isWarningIssueRecordedEventEnabled`

  This was only for internal testing use, but is obsolete because we test issue severity by specifying event stream version >= 6.3.

### Checklist:

[^1]: https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0013-issue-severity-warning.md
[^2]: https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0013-issue-severity-warning.md#integration-with-supporting-tools

Fixes rdar://158621033

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
@@ -26,13 +26,21 @@ extension ABI {
 
     /// The severity of this issue.
     ///
-    /// - Warning: Severity is not yet part of the JSON schema.
-    var _severity: Severity
-    
+    /// Prior to 6.3, this is nil.
+    ///
+    /// @Metadata {
+    ///   @Available(Swift, introduced: 6.3)
+    /// }
+    var severity: Severity?
+
     /// If the issue is a failing issue.
     ///
-    /// - Warning: Non-failing issues are not yet part of the JSON schema.
-    var _isFailure: Bool
+    /// Prior to 6.3, this is nil.
+    ///
+    /// @Metadata {
+    ///   @Available(Swift, introduced: 6.3)
+    /// }
+    var isFailure: Bool?
 
     /// Whether or not this issue is known to occur.
     var isKnown: Bool
@@ -51,13 +59,20 @@ extension ABI {
     var _error: EncodedError<V>?
 
     init(encoding issue: borrowing Issue, in eventContext: borrowing Event.Context) {
-      _severity = switch issue.severity {
-      case .warning: .warning
-      case .error: .error
-      }
-      _isFailure = issue.isFailure
+      // >= v0
       isKnown = issue.isKnown
       sourceLocation = issue.sourceLocation
+
+      // >= v6.3
+      if V.versionNumber >= ABI.v6_3.versionNumber {
+        severity = switch issue.severity {
+        case .warning: .warning
+        case .error: .error
+        }
+        isFailure = issue.isFailure
+      }
+
+      // Experimental
       if let backtrace = issue.sourceContext.backtrace {
         _backtrace = EncodedBacktrace(encoding: backtrace, in: eventContext)
       }
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -1046,10 +1046,11 @@ extension ExitTest {
         // TODO: improve fidelity of issue kind reporting (especially those without associated values)
         .unconditional
       }
-      let severity: Issue.Severity = switch issue._severity {
+      let severity: Issue.Severity = switch issue.severity {
       case .warning:
         .warning
-      case .error:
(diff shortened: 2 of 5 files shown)


jerryjrchen:
@swift-ci test

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -1046,11 +1046,14 @@ extension ExitTest {
         // TODO: improve fidelity of issue kind reporting (especially those without associated values)
         .unconditional
       }
-      let severity: Issue.Severity = switch issue._severity {
+      let severity: Issue.Severity = switch issue.severity {
       case .warning:
         .warning
       case .error:
         .error
+      case .none:
```
grynspan:
```suggestion
      case nil, .error:
```

Tests/TestingTests/SwiftPMTests.swift
```diff
@@ -286,6 +286,47 @@ struct SwiftPMTests {
...
+      }
+
+      let configuration = try configurationForEntryPoint(withArguments:
+          ["PATH", "--event-stream-output-path", temporaryFilePath, "--experimental-event-stream-version", "6.3"]
+      )
+
+      await test.run(configuration: configuration)
+    }
+
+    let decodedRecords = try FileHandle(forReadingAtPath: temporaryFilePath).readToEnd()
```
grynspan: I assume this was copied from another test; should we factor it into a helper?

grynspan (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -26,13 +26,17 @@ extension ABI {
 
     /// The severity of this issue.
```
stmontgomery, line 27: It might be helpful to note under what circumstances this will be `nil` (same for `isFailure`)

Tests/TestingTests/EntryPointTests.swift
```diff
@@ -30,11 +30,12 @@ struct EntryPointTests {
     }
   }
 
-  @Test("Entry point with WarningIssues feature enabled exits with success if all issues have severity < .error")
-  func warningIssues() async throws {
+  @Test("Entry point without WarningIssues feature enabled exits with success if all issues have severity < .error")
```
stmontgomery: When I wrote the display name of this test (and the one below), I had considered adding a "feature" concept and that's why I wrote `WarningIssues feature`. But that's not really relevant, and now the tests are both more concerned with the _event stream version_ rather than whether or not a feature is enabled. So maybe we should rephrase the display names? E.g.
```suggestion
  @Test("Entry point using event stream version 0 exits with success if all issues have severity < .error")
```

And below
```swift
@Test("Entry point using event stream version 6.3 propagates warning issues and exits with success if all issues have severity < .error")
```

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
jerryjrchen: Is it ok to say "Prior to 6.3, this is nil.", or do we need to explain in more depth?

stmontgomery: Yeah that's fine. Something to explain that the reason the value might be missing is because of our desire to maintain compatibility with older versions. 

Having that written down might allow us to more confidently change the property to non-`Optional` in the future if/when we drop support for pre-6.3 versions.

Tests/TestingTests/SwiftPMTests.swift
jerryjrchen: Added ~helpers~ a helper to wrap the common behaviour for reading and deserialising ABI.Record JSON from a file. Take a look and let me know what you think.

If we're looking to do more testing on the event stream longer term, it might make sense to create some sort of `EventStreamFile` object to make it slightly easier to work with arrays of `ABI.Record`. In particular, we do a lot of filtering on `Record.kind` which is a little awkward since it's an enum with an associated value.

I'm not doing that in this change to keep it more lightweight, but wanted to bring that up as a possible future direction.

jerryjrchen:
@swift-ci test

stmontgomery (approved):

PR #1282, opened 2025-08-21
Ensure attachments created in exit tests are forwarded to the parent.

grynspan:
This PR ensures that when an exit test records an attachment, it is forwarded to the parent process.

It introduces a local/private dependency on `Data` in `EncodedAttachment` so that we can use its base64 encoding/decoding and its ability to map files from disk. If Foundation is not available, it falls back to encoding/decoding `[UInt8]` and reading files into memory with `fread()` (the old-fashioned way).

Resolves rdar://149242118.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedAttachment.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedAttachment.swift
@@ -8,6 +8,10 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
+#if canImport(Foundation)
+private import Foundation
+#endif
+
 extension ABI {
   /// A type implementing the JSON encoding of ``Attachment`` for the ABI entry
   /// point and event stream output.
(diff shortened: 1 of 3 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test Windows

Sources/Testing/ABI/Encoded/ABI.EncodedAttachment.swift
```diff
@@ -8,25 +8,144 @@
...
 }
 
 // MARK: - Codable
 
 extension ABI.EncodedAttachment: Codable {}
+
+extension ABI.EncodedAttachment.Bytes: Codable {
+  func encode(to encoder: any Encoder) throws {
+#if canImport(Foundation)
+    // If possible, encode this structure as Base64 data.
```
stmontgomery, line 77: It makes me a little nervous to be encoding and decoding differently based on the _build-time_ availability of `Foundation`. I know this probably is not a concern right now, but what if we began transmitting these encoded payloads between different copies of the testing library, such as between platforms where one platform doesn't have Foundation and the other does?

I see below how the decoding logic gracefully falls back to raw bytes, but what about the following scenario: The sender doesn't have Foundation, so it encodes raw bytes. Then the recipient does have Foundation so it attempts to decode as `Data`. The decoding attempt could _succeed_ if it happens to be valid Base64-encoded data, but the decoded content could be invalid since it wasn't actually Base64-encoded. Is that a plausible failure mode?

I feel like that could be alleviated by including within the encoded representation a value indicating what format the bytes are in, so the recipient can confidently know which decoding approach to take.

grynspan: It won't be valid Base64-encoded data. `[UInt8]` will encode as `[1, 2, 3, ...]` while `Data` will encode as a Base64 _string_.

stmontgomery (approved):

PR #1284, opened 2025-08-21
Add experimental SPI to cancel a running test.

### LINKED ISSUE #120: XCSkip equivalents in swift-testing
### Description

The old XCTest based test case:
```swift
func testDemo() throws {
    let value: Int? = ...
    guard let value else {
        throw XCTSkip("Some skip message here")

    }
    XCTAssertEqual(value, 3)
}
```

The closest migrating to swift-testing I could think of. 
```swift
func testDemo() throws {
    let value: Int? = ...
    guard let value else {
        withKnownIssue {
            throw "Some skip message here"
        }
        return
    }
    #expect(BloomFilter(type: type).value == expectedValue)
}
extension String: Error {}
```

### Expected behavior

`XCSkip` have an equivalent in swift-testing

### Actual behavior

`XCSkip` have no equivalent in swift-testing

### Steps to reproduce

_No response_

_No response_

_No response_

### LINKED ISSUE #1289: Task cancellation within a test is handled inconsistently
Today, the task cancellation APIs behave inconsistently in Swift Testing:

- `UnsafeCurrentTask.cancel()` within a test body cancels the current test case but has no other observable effects
- `UnsafeCurrentTask.cancel()` during trait evaluation will cancel the entire test run (!)
-  `try Task.checkCancellation()` after task cancellation will record an error even though the test's task has been cancelled

`Task.cancel()` and `UnsafeCurrentTask.cancel()` don't throw, so they don't implicitly provide for early termination, but they can be paired with `throw CancellationError()` or even just `return`.

The only reasonable behaviour for task cancellation in a test target is that, if you cancel the task associated with a test, it cancels the test.

grynspan:
This PR introduces `Test.cancel()` and `Test.Case.cancel()` which cancel the current test/suite and the current test case, respectively. For example:

```swift
@Test(arguments: [Food.burger, .fries, .iceCream])
func `Food truck is well-stocked`(_ food: Food) throws {
  if food == .iceCream && Season.current == .winter {
    try Test.Case.cancel("It's too cold for ice cream.")
  }
  // ...
}
```

These functions work by cancelling the child task associated with the current test or test case, then throwing an error to end local execution early. Compare `XCTSkip()` which, in Swift, is just a thrown error that the XCTest harness special-cases, or `XCTSkip()` in Objective-C which actually throws an exception to force the caller to exit early.

Resolves #120.
Resolves #1289.
Resolves rdar://159150449.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
@@ -68,10 +68,31 @@ extension ABI {
 
     /// The comments the test author provided for this event, if any.
     ///
+    /// The value of this property contains the comments related to the primary
+    /// user action that caused this event to be generated.
+    ///
+    /// Some kinds of events have additional associated comments. For example,
+    /// when using ``withKnownIssue(_:isIntermittent:sourceLocation:_:)``, there
+    /// can be separate comments for the "underlying" issue versus the known
+    /// issue matcher, and either can be `nil`. In such cases, the secondary
+    /// comment(s) are represented via a distinct property depending on the kind
+    /// of that event.
+    ///
     /// - Warning: Comments at this level are not yet part of the JSON schema.
     var _comments: [String]?
 
     /// A source location associated with this event, if any.
+    ///
+    /// The value of this property represents the source location most closely
+    /// related to the primary user action that caused this event to be
+    /// generated.
+    ///
+    /// Some kinds of events have additional associated source locations. For
+    /// example, when using ``withKnownIssue(_:isIntermittent:sourceLocation:_:)``,
+    /// there can be separate source locations for the "underlying" issue versus
+    /// the known issue matcher. In such cases, the secondary source location(s)
+    /// are represented via a distinct property depending on the kind of that
+    /// event.
     ///
     /// - Warning: Source locations at this level of the JSON schema are not yet
     ///   part of said JSON schema.


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
Todo:

- [ ] Remove filtering in legacy event stream (Xcode will ignore)
- [ ] Add experimental JSON output
- [ ] Plumb through exit tests

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/ABI/Encoded/ABI.EncodedAttachment.swift
```diff
@@ -38,12 +38,6 @@ extension ABI {
     /// - Warning: Inline attachment content is not yet part of the JSON schema.
     var _bytes: Bytes?
 
-    /// The source location where this attachment was created.
-    ///
-    /// - Warning: Attachment source locations are not yet part of the JSON
-    ///   schema.
-    var _sourceLocation: SourceLocation?
```
stmontgomery, line 45: What's the reason for this removal? Just not used anywhere (yet)?

Sources/Testing/Events/Event.swift
```diff
@@ -121,6 +133,18 @@ public struct Event: Sendable {
...
+    ///
+    /// - Parameters:
+    ///   - skipInfo: A ``SkipInfo`` with details about the cancelled test.
+    ///
+    /// This event is generated by a call to ``Test/cancel(_:sourceLocation:)``.
+    ///
+    /// The test that was cancelled is contained in the ``Event/Context``
+    /// instance that was passed to the event handler along with this event.
+    @_spi(Experimental)
+    indirect case testCancelled(_ skipInfo: SkipInfo)
```
stmontgomery, line 146: We have `case testSkipped` which has an associated `SkipInfo` already. Instead of needing this (and test**Case**Cancelled) could we distinguish skip from a cancellation on the receiving side based on whether or not the `testSkipped` event was preceded by a `testStarted` event?

Sources/Testing/Running/Runner.swift
```diff
@@ -355,9 +355,6 @@ extension Runner {
   /// This function sets ``Test/Case/current``, then invokes the test case's
   /// body closure.
   private static func _runTestCase(_ testCase: Test.Case, within step: Plan.Step) async throws {
-    // Exit early if the task has already been cancelled.
-    try Task.checkCancellation()
```
stmontgomery, line 359: It looks like this change will allow the enclosing function to remove `throws`, is that right? I recall there's a `try?` in `_run(_:)` below, and if this PR allows us to remove that that'd be nice

Sources/Testing/Events/Event.swift
grynspan: They are different events and are handled differently in the console output recorder. A tool that doesn't care about the distinction can just treat them as equivalent.

Sources/Testing/ABI/Encoded/ABI.EncodedAttachment.swift
grynspan: Yep, I just landed it and now I've moved it up a level.

Sources/Testing/Running/Runner.Plan.swift
```diff
@@ -193,6 +193,52 @@ extension Runner.Plan {
...
+  /// - Parameters:
+  ///   - test: The test whose action will be determined.
+  ///
+  /// - Returns: A tuple containing the action to take for `test` as well as any
+  ///   error that was thrown during trait evaluation. If more than one error
+  ///   was thrown, the first-caught error is returned.
+  private static func _determineAction(for test: Test) async -> (Action, (any Error)?) {
+    // FIXME: Parallelize this work. Calling `prepare(...)` on all traits and
+    // evaluating all test arguments should be safely parallelizable.
+    await withTaskGroup(returning: (Action, (any Error)?).self) { taskGroup in
```
stmontgomery, line 215: It seems like the rationale for hoisting this out into a standalone function is that you want to enclose it in a task group. Is that right? Could we add a comment to that effect to the doc block?

Sources/Testing/Running/Runner.swift
grynspan: Resolved as much as possible.

stmontgomery (approved):

Sources/Testing/Running/Runner.Plan.swift
grynspan: Will add a comment.

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -64,6 +66,17 @@ extension ABI {
...
+    /// The comments the test author provided for this event, if any.
+    ///
+    /// - Warning: Comments at this level are not yet part of the JSON schema.
+    var _comments: [String]?
+
+    /// A source location associated with this event, if any.
+    ///
+    /// - Warning: Source locations at this level of the JSON schema are not yet
+    ///   part of said JSON schema.
+    var _sourceLocation: SourceLocation?
```
stmontgomery, line 99: This `_sourceLocation`, and (to a lesser extent) the `_comments` property above make me nervous to be on this `EncodedEvent` type directly. (I realize they're underscored/experimental, but nevertheless.) I like the way `(Encoded)Issue` is modeled more: it has an optional property related to source location, and that allows distinguishing the location of the original issue from (say) the location of a `withKnownIssue` which _matched_ that issue, if there was one. 

In other words, there are some situations where an event may have more than one source location associated with it. And it can be useful to clearly distinguish the purpose of each location, and allow them to be `nil` independently.

Instead of placing these properties on this type directly, could we introduce a type such as `ABI.EncodedCancellation` and an accompanying underscored/experimental property like `_cancellation`? (Actual names TBD.) Then, these properties could be confined to that type.

grynspan: Sure.

grynspan: Thinking it over, since this type just provides a model for JSON generation, this isn't necessary as these fields just don't need to be present when they don't apply (hence why they're optional).

grynspan: Added comments per our offline discussion explaining the use case for these new fields.

grynspan:
@swift-ci test

PR #1253, opened 2025-08-06
Add experimental AdvancedConsoleOutputRecorder skeleton framework

tienquocbui:
Add experimental AdvancedConsoleOutputRecorder skeleton framework

### Motivation:

The current console output for `swift test` is a static log, which presents challenges for developers in understanding test progress and diagnosing failures, especially in large, parallel test suites. This PR introduces the foundational "skeleton" for a new, advanced console reporter to address these issues.

The recorder is marked as experimental and must be explicitly enabled via the `SWT_ENABLE_EXPERIMENTAL_CONSOLE_OUTPUT` environment variable, ensuring it doesn't affect existing functionality.

### Modifications:

- **Added `Event.AdvancedConsoleOutputRecorder.swift`**: New experimental console output recorder skeleton with:
  - Basic structure and minimal configuration options
  - Currently delegates to `Event.ConsoleOutputRecorder` for actual output
  - Foundation framework ready for future development of advanced features

The implementation provides the foundation structure following established patterns in the codebase, ready for future development of advanced console features.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #956, opened 2025-02-12
Plumb the schema version through the JSON ABI layers.

grynspan:
This PR plumbs the schema version (0 or 1) through the Swift types and functions that produce our JSON output, allowing them to change behaviour based on which version is in use.

As a proof-of-concept as much as anything else, this PR also adds an (unsupported) `_tags` field to the JSON output for a test record, but only with ABI version 1.

I've split this PR into two commits: the first plumbs the version through as an integer argument, while the second uses the type system to represent different ABI versions. The latter uses a protocol that we can extend so that different ABI versions have differing behaviours (although that's too coarse-grained for things like `_tags`.)

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1287, opened 2025-08-26
Define an experimental event stream version and use it to conditionalize relevant content

stmontgomery:
This PR formally defines an "experimental" event stream version (named `ABI.ExperimentalVersion`) which represents content  that is considered experimental. It then adopts the new experimental version in several places to conditionalize the inclusion of fields on event stream models which are not yet officially included in any defined, supported version. Finally, it uses this new version everywhere that intentionally always uses the _highest experimental_ version, such as the exit test back channel and the [recently-added](#1253) experimental console output recorder.

### Motivation:

The event stream now has an established versioning system (as of #956) and can easily conditionalize content based on version. As a general rule, we prefer to exclude content which has not gone through Swift Evolution review when delivering data to event stream consumers which expect to receive content from a supported version. The fields this PR conditionalizes have underscore prefixes and have code-level documentation indicating their unofficial-ness, but the fact that they are included at all in supported/non-experimental version streams could lead to misuse or unintentional breakages in the future if the names or semantics of these fields change.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/ABI.swift
+++ b/Sources/Testing/ABI/ABI.swift
@@ -45,7 +45,8 @@ extension ABI {
   /// The current supported ABI version (ignoring any experimental versions.)
   typealias CurrentVersion = v0
 
-  /// The highest supported ABI version (including any experimental versions.)
+  /// The highest defined and supported ABI version (including any experimental
+  /// versions.)
   typealias HighestVersion = v6_3
 
 #if !hasFeature(Embedded)
@@ -123,6 +124,14 @@ extension ABI {
   public enum v6_3: Sendable, Version {
     static var versionNumber: VersionNumber {
       VersionNumber(6, 3)
+    }
+  }
+
+  /// A namespace and type representing the ABI version whose symbols are
+  /// considered experimental.
+  enum ExperimentalVersion: Sendable, Version {
+    static var versionNumber: VersionNumber {
+      VersionNumber(99, 99)
     }
   }
 }
--- a/Sources/Testing/ABI/ABI.swift
+++ b/Sources/Testing/ABI/ABI.swift
@@ -115,9 +115,15 @@ extension ABI.Version {
   ///
   /// Otherwise, the value of this property is `false`.
   static var includesExperimentalFields: Bool {
-    versionNumber < ABI.v6_3.versionNumber
-      || (versionNumber >= ABI.v6_3.versionNumber && _shouldIncludeExperimentalFlags == true)
-      || versionNumber >= ABI.ExperimentalVersion.versionNumber
+    switch versionNumber {
+    case ABI.ExperimentalVersion.versionNumber...:
+      true
+    case ABI.v6_3.versionNumber...:
+      _shouldIncludeExperimentalFlags == true
+    default:
+      // Maintain behavior for pre-6.3 versions.
+      true
+    }
   }
 }
 
--- a/Sources/Testing/ABI/ABI.swift
+++ b/Sources/Testing/ABI/ABI.swift
@@ -131,7 +131,7 @@ extension ABI {
   /// considered experimental.
   enum ExperimentalVersion: Sendable, Version {
     static var versionNumber: VersionNumber {
-      VersionNumber(99, 99)
+      VersionNumber(99, 0)
     }
   }
 }
--- a/Sources/Testing/ABI/ABI.swift
+++ b/Sources/Testing/ABI/ABI.swift
@@ -94,6 +94,33 @@ extension ABI {
 #endif
 }
 
+/// The value of the environment variable flag which enables experimental event
+/// stream fields, if any.
+private let _shouldIncludeExperimentalFlags = Environment.flag(named: "SWT_EXPERIMENTAL_EVENT_STREAM_FIELDS_ENABLED")
+
+extension ABI.Version {
+  /// Whether or not experimental fields should be included when using this
+  /// ABI version.
+  ///
+  /// The value of this property is `true` if any of the following conditions
+  /// are satisfied:
+  ///
+  /// - The version number is less than 6.3. This is to preserve compatibility
+  ///   with existing clients before the inclusion of experimental fields became
+  ///   opt-in starting in 6.3.
+  /// - The version number is greater than or equal to 6.3 and the environment
+  ///   variable flag `SWT_EXPERIMENTAL_EVENT_STREAM_FIELDS_ENABLED` is set to a
+  ///   true value.
+  /// - The version number is greater than or equal to that of ``ABI/ExperimentalVersion``.
+  ///
+  /// Otherwise, the value of this property is `false`.
+  static var includesExperimentalFields: Bool {
+    versionNumber < ABI.v6_3.versionNumber
+      || (versionNumber >= ABI.v6_3.versionNumber && _shouldIncludeExperimentalFlags == true)
+      || versionNumber >= ABI.ExperimentalVersion.versionNumber
+  }
+}
+
 // MARK: - Concrete ABI versions
 
 extension ABI {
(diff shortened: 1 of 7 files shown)


stmontgomery:
@swift-ci please test

Sources/Testing/ABI/ABI.swift
```diff
@@ -125,6 +126,14 @@ extension ABI {
       VersionNumber(6, 3)
     }
   }
+
+  /// A namespace and type representing the ABI version whose symbols are
+  /// considered experimental.
+  enum ExperimentalVersion: Sendable, Version {
+    static var versionNumber: VersionNumber {
+      VersionNumber(99, 99)
```
grynspan: Nitpick: the stdlib would write this as just 99.0, not 99.99, right?

grynspan (commented):
We should see what the VS Code plugin might be relying on before blocking stuff off.

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -76,29 +76,26 @@ extension ABI {
...
       name = test.name
       displayName = test.displayName
       sourceLocation = test.sourceLocation
       id = ID(encoding: test.id)
 
-      if V.versionNumber >= ABI.v6_3.versionNumber {
+      // Experimental
+      if V.versionNumber >= ABI.ExperimentalVersion.versionNumber {
+        if isParameterized == true {
+          _testCases = test.uncheckedTestCases?.map(EncodedTestCase.init(encoding:))
```
plemarquand, line 96: The VS Code Swift extension relies on this (and it has since it added Swift Testing support): https://github.com/swiftlang/vscode-swift/blob/main/src/TestExplorer/TestParsers/SwiftTestingOutputParser.ts#L324

stmontgomery: Great, thanks for confirming. Do you know if there's anything else I'm conditionalizing in this PR that would be affected?

One way I could handle this is by modifying my PR to preserve this property in all current versions (including v0) as a special case, then file a new issue to formally propose inclusion of this property in a future event stream version, and work with you to adopt that new version.

Another idea is to keep this change as I've written it, but introduce a new, undocumented environment variable which you could set from the VS Code plugin which would override this check and preserve this field in the current event stream v0. (And still pursue the formal proposal step I mentioned in the previous paragraph, which would allow us to eventually remove that env var workaround.)

I like the latter approach more, because it avoids new clients from having "free" access to this field and potentially misusing it or getting broken before it's been formally proposed. How does that sound?

Sources/Testing/ABI/ABI.swift
stmontgomery: I wasn't sure, but I'm fine with that

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
stmontgomery: @plemarquand I've implemented an even more permissive workaround: keep it in pre-6.3 versions (including v0), and in versions ≥ 6.3, don't include it by default but allow opting-in via an env var (until the new field can be proposed formally and is accepted). See my [code comment](https://github.com/swiftlang/swift-testing/pull/1287/files#diff-81e9e4399a13a86a9186fadc620ee532da6139bb1d059c0a77c9430de6003978R93) for more details. How does that sound?

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -76,29 +76,39 @@ extension ABI {
     /// The tags associated with the test.
     ///
     /// - Warning: Tags are not yet part of the JSON schema.
-    ///
```
grynspan, line 79: I actually do want to promote this in 6.3!

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -98,8 +98,12 @@ extension ABI {
       instant = EncodedInstant(encoding: event.instant)
       self.messages = messages.map(EncodedMessage.init)
       testID = event.testID.map(EncodedTest.ID.init)
-      if eventContext.test?.isParameterized == true {
-        _testCase = eventContext.testCase.map(EncodedTestCase.init)
+
+      // Experimental
```
grynspan: Does VS Code need this?

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -76,29 +76,39 @@ extension ABI {
...
+      // including this field in versions earlier than 6.3 (including v0). In
+      // 6.3 and later, don't include it by default but allow opting-in to it
+      // via an environment variable. (This is to maintain compatibility until
+      // the field has been formally proposed and accepted, and discourage new
+      // clients from becoming dependent on it in the mean time.) Finally,
+      // in the special experimental version always include this field.
+      if isParameterized == true,
+         (V.versionNumber < ABI.v6_3.versionNumber
+          || V.versionNumber >= ABI.ExperimentalVersion.versionNumber
+          || Environment.flag(named: "SWT_ENABLE_EXPERIMENTAL_EVENT_STREAM_TEST_CASE_ENCODING") == true) {
```
grynspan: Checking an environment flag on each call is expensive. You should cache this.

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
plemarquand: @stmontgomery that works. VS Code will always provide the environment var and once it is up for formal proposal we'll add code that favours using the new, undescoreless variable if it exists.

plemarquand: I put up https://github.com/swiftlang/vscode-swift/pull/1804 to set the env var.

grynspan: I'd like to find a more scalable solution as we're liable to find ourselves needing similar workarounds in the future.

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
stmontgomery: The new approach covers all existing fields

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
stmontgomery: Implemented a generalized approach in latest commit

stmontgomery:
@swift-ci please test

stmontgomery:
@swift-ci please test

Sources/Testing/ABI/ABI.swift
```diff
@@ -93,6 +94,33 @@ extension ABI {
...
+  ///   with existing clients before the inclusion of experimental fields became
+  ///   opt-in starting in 6.3.
+  /// - The version number is greater than or equal to 6.3 and the environment
+  ///   variable flag `SWT_EXPERIMENTAL_EVENT_STREAM_FIELDS_ENABLED` is set to a
+  ///   true value.
+  /// - The version number is greater than or equal to that of ``ABI/ExperimentalVersion``.
+  ///
+  /// Otherwise, the value of this property is `false`.
+  static var includesExperimentalFields: Bool {
+    versionNumber < ABI.v6_3.versionNumber
```
grynspan: This might be more readable as a range switch.

Tests/TestingTests/SwiftPMTests.swift
```diff
@@ -376,7 +376,7 @@ struct SwiftPMTests {
     }
     #expect(testRecords.count == 1)
     for testRecord in testRecords {
-      if version.versionNumber >= ABI.v6_3.versionNumber {
+      if version.versionNumber < ABI.v6_3.versionNumber || version.versionNumber >= ABI.ExperimentalVersion.versionNumber {
```
grynspan:
```suggestion
      if version.includesExperimentalFields {
```

grynspan (approved):

stmontgomery:
@swift-ci please test

Sources/Testing/ABI/ABI.swift
stmontgomery: Implemented this suggestion now

PR #1290, opened 2025-08-27
Implement hierarchical console output with comprehensive test result display

tienquocbui:
Implement hierarchical console output with comprehensive test result display

### Modifications:
- Implemented a tree structure using Unicode box-drawing characters (┌─, ├─, ╰─, │) that properly displays the relationship between modules, test suites, and individual tests
- **Test Result Summary**: 
  - Right-aligned duration formatting in consistent `x.xxs` format
  - Updated summary format: "X tests completed in Y.ZZs (pass: A, fail: B, skip: C)"
- **Detailed Failure Section**: 
  - Complete hierarchical path display for failed tests
  - Source location information with file and line numbers
- **Unit Tests**: Added 8 automated tests following established patterns to validate output formatting, hierarchy generation, and failure handling

### Checklist:


Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -31,7 +31,95 @@ extension Event {
       /// This is needed because ABI.EncodedEvent doesn't contain full test context.
       var testStorage: [String: ABI.EncodedTest<V>] = [:]
       
-      // Future storage for result data and other event information can be added here
+      /// Hierarchical test node structure for building the test tree.
+      /// Maps test ID string to its corresponding hierarchy node.
+      var nodes: [String: _HierarchyNode] = [:]
```
grynspan: We have a type for a tree/graph structure called `Graph`. Could you use that?

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -65,70 +153,627 @@ extension Event {
...
+                spacingPrefix = childPrefix
+              }
+              output += "\(spacingPrefix)\n"  // Add the vertical line continuation
+            }
+          }
+        }
+      }
+    } else {
+      // Test case line
+      let treePrefix = isLast ? "╰─ " : "├─ "
```
grynspan: Anywhere you're using Unicode characters, we need fallback paths that use ASCII characters instead.

stmontgomery:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
tienquocbui: I've added fallback for ASCII characters.

```
// MARK: - ASCII Fallback Support

extension Event.AdvancedConsoleOutputRecorder {
  /// Get the appropriate tree drawing character with ASCII fallback.
  ///
  /// - Parameters:
  ///   - unicode: The Unicode box-drawing character to use.
  ///   - ascii: The ASCII fallback character(s) to use.
  ///
  /// - Returns: The appropriate character based on terminal capabilities.
  private func _treeCharacter(unicode: String, ascii: String) -> String {
    // Use ASCII fallback on Windows or when ANSI escape codes are disabled
    // This follows the same pattern as Event.Symbol
#if os(Windows)
    return ascii
#else
    if options.base.useANSIEscapeCodes {
      return unicode
    } else {
      return ascii
    }
#endif
  }
  
  /// Get the tree branch character (├─).
  private var _treeBranch: String {
    _treeCharacter(unicode: "├─ ", ascii: "|- ")
  }
  
  /// Get the tree last branch character (╰─).
  private var _treeLastBranch: String {
    _treeCharacter(unicode: "╰─ ", ascii: "`- ")
  }
  
  /// Get the tree first branch character (┌─).
  private var _treeFirstBranch: String {
    _treeCharacter(unicode: "┌─ ", ascii: ".- ")
  }
  
  /// Get the tree vertical line character (│).
  private var _treeVertical: String {
    _treeCharacter(unicode: "│", ascii: "|")
  }
}
```

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
tienquocbui: I've replaced the custom hierarchy structures with the existing `Graph` as requested

stmontgomery:
@swift-ci please test

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -64,71 +142,720 @@ extension Event {
...
+  ///   - testID: The test ID to parse.
+  /// - Returns: An array of key path components.
+  private func _parseTestIDToKeyPath(_ testID: String) -> [String] {
+    // Swift Testing test IDs include source location information
+    // We need to extract the logical hierarchy path without source locations
+    // Examples:
+    // Suite: "TestingTests.HierarchyDemoTests/NestedSuite"
+    // Test: "TestingTests.HierarchyDemoTests/failingTest()/HierarchyDemoTests.swift:21:4"
+    
+    let components = testID.split(separator: "/").map(String.init)
```
stmontgomery: We should probably use our backtick-aware split utility

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -64,71 +142,720 @@ extension Event {
...
+  }
+  
+  /// Get the tree last branch character (╰─).
+  private var _treeLastBranch: String {
+    _treeCharacter(unicode: "╰─ ", ascii: "`- ")
+  }
+  
+  /// Get the tree first branch character (┌─).
+  private var _treeFirstBranch: String {
+    _treeCharacter(unicode: "┌─ ", ascii: ".- ")
```
stmontgomery: What about two dashes `--` ?

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -58,8 +58,8 @@ func entryPoint(passing args: __CommandLineArguments_v0?, eventHandler: Event.Ha
       if Environment.flag(named: "SWT_ENABLE_EXPERIMENTAL_CONSOLE_OUTPUT") == true {
         // Use experimental AdvancedConsoleOutputRecorder
         var advancedOptions = Event.AdvancedConsoleOutputRecorder<ABI.HighestVersion>.Options()
-        advancedOptions.base = .for(.stderr)
-
+        advancedOptions.base = Event.ConsoleOutputRecorder.Options.for(.stderr)
+
```
stmontgomery: nit: spurious whitespace here

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
grynspan: That makes sense. If you want to go the extra mile, Windows _does_ have box-drawing characters in its default Terminal font, although they're not the same as the Unicode characters you're using. :)

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
grynspan: I'm like 90% sure we have infrastructure to parse a test ID string outright.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -58,12 +58,11 @@ func entryPoint(passing args: __CommandLineArguments_v0?, eventHandler: Event.Ha
...
         // Use experimental AdvancedConsoleOutputRecorder
         var advancedOptions = Event.AdvancedConsoleOutputRecorder<ABI.HighestVersion>.Options()
-        advancedOptions.base = .for(.stderr)
-
+        advancedOptions.base = Event.ConsoleOutputRecorder.Options.for(.stderr)
         let eventRecorder = Event.AdvancedConsoleOutputRecorder<ABI.HighestVersion>(options: advancedOptions) { string in
           try? FileHandle.stderr.write(string)
         }
-
+
```
stmontgomery: nit: whitespace changes in this file look unintentional

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -628,15 +627,15 @@ public func configurationForEntryPoint(from args: __CommandLineArguments_v0) thr
...
-    break
+    switch args.eventStreamVersionNumber {
+    case .some(..<ABI.v6_3.versionNumber):
+      // If the event stream version was explicitly specified to a value < 6.3,
+      // disable the warning issue event to maintain legacy behavior.
+      configuration.eventHandlingOptions.isWarningIssueRecordedEventEnabled = false
+    default:
+      // Otherwise the requested event stream version is ≥ 6.3, so don't change
+      // the warning issue event setting.
+      break
```
stmontgomery: nit: I think the indentation here was unintentional, probably the result of a merge conflict resolution where these lines were recently modified. You can "outdent" them one level.

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -64,71 +146,863 @@ extension Event {
...
+    _treeCharacter(unicode: "├─ ", ascii: "|- ")
+  }
+  
+  /// Get the tree last branch character (╰─).
+  private var _treeLastBranch: String {
+    _treeCharacter(unicode: "╰─ ", ascii: "`- ")
+  }
+  
+  /// Get the tree first branch character (┌─).
+  private var _treeFirstBranch: String {
```
stmontgomery: Per offline discussion, I think you said you were planning to remove this

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -64,71 +146,863 @@ extension Event {
...
+              }
+            }
+            
+            // 2. Location and Source Code Context
+            if let sourceLocation = issue.sourceLocation {
+              output += "\n"
+              output += "  Location: \(sourceLocation.fileName):\(sourceLocation.line):\(sourceLocation.column)\n"
+              
+              // 3. Source Code Context (2-3 lines before/after)
+              let codeContext = _getSourceCodeContext(for: sourceLocation)
```
stmontgomery: This is a cool idea in theory but as we discussed offline, it isn't possible to obtain the context of surrounding lines of code in all cases (for example, when running on a remotely-connected device) and so I'm hesitant to include this. However, I do think it would be very useful to show the unmodified source code of the `#expect` (or similar expectation) and use that as a place to denote the subexpressions and their values on subsequent lines. But for this PR, I would suggest excluding this since it won't have a consistent experience in all usage contexts.

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -64,71 +146,863 @@ extension Event {
...
+    if node.isSuite {
+      // Suite header
+      let treePrefix: String
+      if prefix.isEmpty {
+        if isSingleRoot {
+          // Single root module: no tree prefix, flush left
+          treePrefix = ""
+        } else if isFirstRoot {
+                  // Multiple roots: first root uses first branch character
+        treePrefix = _treeFirstBranch
```
stmontgomery: nit: indentation

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
tienquocbui: @grynspan after doing some research, I think we could implement a 3-tiered fallback mechanism like this:
- Default (macOS/Linux): Use the Unicode characters.
- Windows fallback: render the specific box-drawing characters from [code page 437](https://www.ascii-codes.com)
- Final fallback (ASCII): we'll use the simple ASCII set.

What do you think of this approach? It seems like a good balance between providing the best experience on each platform while still ensuring absolute compatibility.

grynspan: That makes sense to me. We can also make it configurable (with appropriate default state of course) if somebody really wants ASCII, for instance.

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -64,71 +142,716 @@ extension Event {
...
+    
+    return keyPath.isEmpty ? [testID] : keyPath
+  }
+  
+  /// Get all root nodes (module-level nodes) from the Graph.
+  ///
+  /// - Parameters:
+  ///   - testTree: The Graph to extract root nodes from.
+  /// - Returns: Array of key paths for root nodes (modules).
+  private func _getRootNodes(from testTree: Graph<String, _HierarchyNode?>) -> [[String]] {
```
grynspan: FYI it's not idiomatic to use the word `get` in Swift function names.

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
```diff
@@ -64,71 +142,716 @@ extension Event {
...
+    }
+  }
+  
+  /// Check if a string contains Swift file pattern without using Foundation.
+  ///
+  /// - Parameters:
+  ///   - string: The string to check.
+  /// - Returns: True if the string contains ".swift:" pattern.
+  private func _containsSwiftFile(_ string: String) -> Bool {
+    let target = ".swift:"
```
grynspan: This string can be present in any component of a test ID.

The last component of a test ID string should always be the filename, so we don't really need to check each component. If you're doing this because of some limitation in the JSON data, can you clarify so we can improve it? Thanks!

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -58,8 +58,7 @@ func entryPoint(passing args: __CommandLineArguments_v0?, eventHandler: Event.Ha
       if Environment.flag(named: "SWT_ENABLE_EXPERIMENTAL_CONSOLE_OUTPUT") == true {
         // Use experimental AdvancedConsoleOutputRecorder
         var advancedOptions = Event.AdvancedConsoleOutputRecorder<ABI.HighestVersion>.Options()
-        advancedOptions.base = .for(.stderr)
-
+        advancedOptions.base = Event.ConsoleOutputRecorder.Options.for(.stderr)
```
grynspan: What's the change here for?

Sources/Testing/Events/Recorder/Event.AdvancedConsoleOutputRecorder.swift
tienquocbui: I've simplified it to just use `component.hasSuffix(".swift:")` since the filename should indeed always be the last component. The original complex approach was unnecessary - I was being overly cautious about where ".swift:" might appear

PR #1291, opened 2025-08-28
Clarify note about supported target types in "Defining test functions" article

stmontgomery:
This adjusts the wording of a `- Note:` in our [Defining test functions](https://developer.apple.com/documentation/testing/definingtests) documentation article which advises that users only import the testing library in test targets. It expands the kinds of supported targets to include "library targets meant for test targets", which is a supported use case.

### Motivation:

The wording should not imply that it's _never_ supported to vend a library which extends the testing library. There is an important distinction between libraries which are distributed to end users (e.g. those bundles within apps or alongside executables) vs. libraries intended solely to support test targets or other testing libraries.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/DefiningTests.md
+++ b/Sources/Testing/Testing.docc/DefiningTests.md
@@ -25,8 +25,9 @@ contains the test:
 import Testing
 ```
 
-- Note: Only import the testing library into a test target. Importing the
-  testing library into an application, library, or binary target isn't
+- Note: Only import the testing library into a test target or library meant for
+  test targets. Importing the testing library into a target intended for
+  distribution such as an application, app library, or executable target isn't
   supported or recommended. Test functions aren't stripped from binaries when
   building for release, so logic and fixtures of a test may be visible to anyone
   who inspects a build product that contains a test function.


Sources/Testing/Testing.docc/DefiningTests.md
```diff
@@ -25,8 +25,9 @@ contains the test:
 import Testing
 ```
 
-- Note: Only import the testing library into a test target. Importing the
-  testing library into an application, library, or binary target isn't
+- Note: Only import the testing library into a test target or library meant for
+  test targets. Importing the testing library into a target intended for
+  distribution such as an application, app library, or executable target isn't
```
stmontgomery, line 30: Open to suggestions for better phrasing @iamleeg, but I hope my intent is clear!

Sources/Testing/Testing.docc/DefiningTests.md
```diff
@@ -25,8 +25,9 @@ contains the test:
 import Testing
 ```
 
-- Note: Only import the testing library into a test target. Importing the
-  testing library into an application, library, or binary target isn't
+- Note: Only import the testing library into a test target or library meant for
```
jerryjrchen, line 28: Very nit: would it be clearer to say "library used *only* in test targets"? I can see the original wording as being interpreted overly broadly (e.g. some sort of utils used in both an executable and test library)

stmontgomery: I see what you're going for there. I worded it this way to keep it slightly more open because it's also supported (though less common) to have one test-only library which is used exclusively by another test-only library which in turn is used by exclusively by test targets. So the guidance is really "keep the transitive closure of these libraries confined to testing contexts". It's just that I'd like this doc phrasing to read naturally. So I felt "meant for test targets" captured that well enough, without excluding that use case

grynspan: This for example is valid in a library if you are building from source:

```swift
#if canImport(Testing)
public func foo(_ test: Test) {
  ...
}
#endif
```

PR #1297, opened 2025-09-02
Extend test/task cancellation support to test case evaluation.

grynspan:
This PR extends the special-casing of `SkipInfo` and `CancellationError` added in #1284 to also cover test case evaluation during test planning. If either error is thrown from `evaluateTestCases()`, we treat it as test cancellation the same way we do for trait evaluation.

Example:

```swift
func websites() async throws -> [Website] {
  guard let www = Web() else {
    try Test.cancel("The Web doesn't exist, is this 1992?")
  }
  return try await www.downloadEverything()
}

@Test(arguments: try await websites())
func browseWeb(at website: Website) { ... }
```

### Checklist:


grynspan:
@swift-ci test

PR #1298, opened 2025-09-03
Propagate `SkipInfo` to child tasks when a test is cancelled.

grynspan:
This PR ensures that, when a test is cancelled, its test cases report the same `SkipInfo` structure as they are recursively cancelled. For instance:

```swift
@Test(arguments: [a ... z]) func foo(arg: T) throws {
  try Test.cancel("Message")
}
```

Prior to this change, the `.testCaseCancelled` events would be posted without the user-supplied comment `"Message"` and each one would need to construct a source context by capturing a backtrace from deep within Swift Testing and the stdlib.

With this change in place, said events post with the same comment and source context as is associated with the initial `.testCancelled` event.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Cancellation.swift
+++ b/Sources/Testing/Test+Cancellation.swift
@@ -45,6 +45,16 @@ private struct _TaskReference: Sendable {
   private nonisolated(unsafe) var _unsafeCurrentTask = Locked<UnsafeCurrentTask?>()
 
   init() {
+    // WARNING! Normally, allowing an instance of `UnsafeCurrentTask` to escape
+    // its scope is dangerous because it could be used unsafely after the task
+    // ends. However, because we take care not to allow the task object to
+    // escape the task (by only storing it in a task-local value), we can ensure
+    // these unsafe scenarios won't occur.
+    //
+    // TODO: when our deployment targets allow, we should switch to calling the
+    // `async` overload of `withUnsafeCurrentTask()` from the body of
+    // `withCancellationHandling(_:)`. That will allow us to use the task object
+    // in a safely scoped fashion.
     _unsafeCurrentTask = withUnsafeCurrentTask { Locked(rawValue: $0) }
   }
 


grynspan:
@swift-ci test

grynspan:
@swift-ci test Linux

Sources/Testing/Running/SkipInfo.swift
```diff
@@ -69,7 +69,7 @@ extension SkipInfo {
       self = skipInfo
     } else if error is CancellationError, Task.isCancelled {
       // Synthesize skip info for this cancellation error.
-      let backtrace = Backtrace(forFirstThrowOf: error) ?? .current()
+      let backtrace = Backtrace(forFirstThrowOf: error)
```
briancroom, line 72: Good, good! That fallback had caught my eye in the previous PR and I almost asked about it; I'm happy to see this follow-up improvement.

Sources/Testing/Test+Cancellation.swift
```diff
@@ -47,8 +45,7 @@ private struct _TaskReference: Sendable {
   private nonisolated(unsafe) var _unsafeCurrentTask = Locked<UnsafeCurrentTask?>()
 
   init() {
-    let unsafeCurrentTask = withUnsafeCurrentTask { $0 }
-    _unsafeCurrentTask = Locked(rawValue: unsafeCurrentTask)
+    _unsafeCurrentTask = withUnsafeCurrentTask { Locked(rawValue: $0) }
```
briancroom, line 58: Aside: Given that this is doing something [explicitly documented as unsafe](https://developer.apple.com/documentation/swift/withunsafecurrenttask(body:)-6gvhl#discussion), might it be worth expanding the comments on `_TaskReference` to provide more context on why it's necessary and how we're avoiding bugs from undefined behavior?

Sources/Testing/Test+Cancellation.swift
```diff
@@ -208,15 +212,13 @@ extension Test: TestCancellable {
   /// test alone, call ``Test/Case/cancel(_:sourceLocation:)`` instead.
   @_spi(Experimental)
   public static func cancel(_ comment: Comment? = nil, sourceLocation: SourceLocation = #_sourceLocation) throws -> Never {
-    try Self.cancel(
-      comments: Array(comment),
-      sourceContext: SourceContext(backtrace: .current(), sourceLocation: sourceLocation)
-    )
+    let skipInfo = SkipInfo(comment: comment, sourceContext: SourceContext(backtrace: nil, sourceLocation: sourceLocation))
```
briancroom, line 225: Why is the `nil` backtrace better than the previous `.current()` in this API context?

grynspan: Consistency with test skipping. Per the comment there:

```
      // We don't need to consider including a backtrace here because it will
      // primarily contain frames in the testing library, not user code.
```

Sources/Testing/Test+Cancellation.swift
grynspan: I confirmed that this specific pattern is safe because we don't extend the lifetime of the task object past the task itself. In the future we should use the `async` variant of `withUnsafeCurrentTask()` to make this clearer, but it's not available on our minimum deployment targets.

I will add a comment explaining the theory here.

grynspan:
@swift-ci test

briancroom (approved):

grynspan:
@swift-ci test

PR #1302, opened 2025-09-08
Improve `ExitStatus.description` for signals.

grynspan:
This PR uses the `sys_signame` array present on many UNIX-like systems to derive a better description for values of type `ExitStatus` and `ExitTest.Condition`. (On Linux, the equivalent `sigabbrev_np()` is used. Windows and WASI don't have an equivalent API.)

Before:

```swift
let s = ExitStatus.signal(SIGABRT)
print(s) // ".signal(12345)"
```

After:

```swift
let s = ExitStatus.signal(SIGABRT)
print(s) // ".signal(SIGABRT)"
```

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitStatus.swift
+++ b/Sources/Testing/ExitTests/ExitStatus.swift
@@ -97,7 +97,7 @@ extension ExitStatus: Equatable {}
 
 // MARK: - CustomStringConvertible
 
-#if os(Linux)
+#if os(Linux) && !SWT_NO_DYNAMIC_LINKING
 /// Get the short name of a signal constant.
 ///
 /// This symbol is provided because the underlying function was added to glibc
@@ -126,13 +126,14 @@ extension ExitStatus: CustomStringConvertible {
       withUnsafeBytes(of: sys_signame) { sys_signame in
         sys_signame.withMemoryRebound(to: UnsafePointer<CChar>.self) { sys_signame in
           if signal > 0 && signal < sys_signame.count {
-            signalName = String(validatingCString: sys_signame[Int(signal)])
+            signalName = String(validatingCString: sys_signame[Int(signal)])?.uppercased()
           }
         }
       }
 #elseif os(Linux)
-      signalName = _sigabbrev_np?(signal)
-        .flatMap(String.init(validatingCString:))
+#if !SWT_NO_DYNAMIC_LINKING
+      signalName = _sigabbrev_np?(signal).flatMap(String.init(validatingCString:))
+#endif
 #elseif os(Windows) || os(WASI)
       // These platforms do not have API to get the programmatic name of a
       // signal constant.
@@ -141,7 +142,7 @@ extension ExitStatus: CustomStringConvertible {
 #endif
 
       if let signalName {
-        return ".signal(SIG\(signalName.uppercased()) → \(signal))"
+        return ".signal(SIG\(signalName) → \(signal))"
       }
       return ".signal(\(signal))"
     }


grynspan:
@swift-ci test

grynspan:
@finagolfin and @3405691582 Could you smoke-test this on Android/OpenBSD please? Thanks!

grynspan:
@swift-ci test

Sources/Testing/ExitTests/ExitStatus.swift
```diff
@@ -96,17 +96,56 @@ public enum ExitStatus: Sendable {
...
 #endif
 extension ExitStatus: CustomStringConvertible {
   public var description: String {
     switch self {
     case let .exitCode(exitCode):
-      ".exitCode(\(exitCode))"
+      return ".exitCode(\(exitCode))"
     case let .signal(signal):
-      ".signal(\(signal))"
+      var signalName: String?
```
jerryjrchen, line 121: Style nit: I think I'd find this easier to read as a helper, so the `description` body becomes:
```swift
    case let .signal(signal):
      if let signalName = _lookupSignalName(signal) {
        return ".signal(SIG\(signalName.uppercased()))"
      }
      return ".signal(\(signal))"
...

private static function _lookupSignalName(signal: CInt) { /* ... */ }
```

Sources/Testing/ExitTests/ExitStatus.swift
```diff
@@ -96,17 +96,56 @@ public enum ExitStatus: Sendable {
...
+        .flatMap(String.init(validatingCString:))
+#elseif os(Windows) || os(WASI)
+      // These platforms do not have API to get the programmatic name of a
+      // signal constant.
+#else
+#warning("Platform-specific implementation missing: signal names unavailable")
+#endif
+
+      if let signalName {
+        return ".signal(SIG\(signalName.uppercased()))"
```
stmontgomery: Could we continue including the numeric value in addition to the symbolic name for convenience? I imagine that could be convenient when comparing against another data source which only prints the number

Tests/TestingTests/ExitTestTests.swift
```diff
@@ -13,6 +13,15 @@ private import _TestingInternals
 
 #if !SWT_NO_EXIT_TESTS
 @Suite("Exit test tests") struct ExitTestTests {
+  @Test("Signal names are reported (where supported)") func signalName() {
+    let exitStatus = ExitStatus.signal(SIGABRT)
+#if SWT_TARGET_OS_APPLE || os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android)
+    #expect(String(describing: exitStatus) == ".signal(SIGABRT)")
+#else
+    #expect(String(describing: exitStatus) == ".signal(\(SIGABRT))")
```
jerryjrchen, line 21: Nit: in favour of keeping the test more explicit, could we just use the actual value of the signal here (I admittedly have no idea if it is consistent on the platforms that this would apply to)?

```suggestion
    #expect(String(describing: exitStatus) == ".signal(6)")
```

grynspan: No, because the value of `SIGABRT` is not always `6`.

grynspan:
@swift-ci test

Sources/Testing/ExitTests/ExitStatus.swift
grynspan: I would agree if it got much longer but the overall body of `description` is mostly just this code anyway and we don't have anything else calling this function.

grynspan:
@swift-ci test

3405691582:
(FWIW, the toolchain successfully built after merge on OpenBSD.)

finagolfin:
Building this Testing repo natively on Android after this pull worked fine, but I could not build the tests with the following error:
```
/data/data/com.termux/files/home/swift-testing/Tests/TestingTests/AttachmentTests.swift:58:9: error: deinit of non-copyable type not visible in the current module
 56 |   @Test func moveOnlyDescription() {
 57 |     let attachableValue = MyAttachable(string: "<!doctype html>")
 58 |     let attachment = Attachment(attachableValue, named: "AttachmentTests.saveValue.html")
    |         `- error: deinit of non-copyable type not visible in the current module
 59 |     #expect(attachment.description.contains(#""\#(attachment.preferredName)""#))
 60 |     #expect(attachment.description.contains("'MyAttachable'"))

macro expansion @Test:40:32: error: global variable must be a compile-time constant to use @_section attribute
`- /data/data/com.termux/files/home/swift-testing/Tests/TestingTests/AttachmentTests.swift:47:4: note: expanded code originates here
 41 | @Suite("Attachment Tests")
 42 | struct AttachmentTests {
 43 |   @Test func saveValue() {
    |   `- note: in expansion of macro 'Test' on instance method 'saveValue()' here
 44 |     let attachableValue = MyAttachable(string: "<!doctype html>")
 45 |     let attachment = Attachment(attachableValue, named: "AttachmentTests.saveValue.html")
 46 |     Attachment.record(attachment)
 47 |   }
    +--- macro expansion @Test -----------------------------------------
    |38 | #endif
    |39 | @available(*, deprecated, message: "This property is an implementation detail of the testing library. Do not use it directly.")
    |40 | private nonisolated static let $s12TestingTests010AttachmentB0V9saveValue4TestfMp_33testContentRecord6e232f07b7901dbafMu_: Testing.__TestContentRecord = (
    |   |                                `- error: global variable must be a compile-time constant to use @_section attribute
    |41 |   0x74657374, /* 'test' */
    |42 |   0,
    +-------------------------------------------------------------------
 48 | 
 49 |   @Test func description() {
```
Does compiling Testing trunk currently require a recent trunk snapshot toolchain? I'm building it with an older June 17 trunk snapshot toolchain that I natively built on Android too, so I'll need to build one of the latest trunk snapshot toolchains first if so.

Anyway, the exit tests are disabled on Android, so this test won't be run there anyway.

grynspan:
@finagolfin That sounds like https://github.com/swiftlang/swift/issues/84184. Looks like a bug in 6.2 with move-only types, but I don't have more info.

PR #1304, opened 2025-09-09
Move the package version to VERSION.txt.

grynspan:
This change moves the definition of Swift Testing's version to a new file, VERSION.txt. This file doesn't represent a standard, _per se_, but it is common for open-source projects to list their version info in such a file. And by putting our version info in this file, we can share it between package builds and CMake builds.

Currently using C23's `#embed` which is supported in clang as a non-standard extension. C++26 will add support for `#embed`.

### Checklist:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,13 @@
+##
+## This source file is part of the Swift.org open source project
+##
+## Copyright (c) 2024-2025 Apple Inc. and the Swift project authors
+## Licensed under Apache License v2.0 with Runtime Library Exception
+##
+## See https://swift.org/LICENSE.txt for license information
+## See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+##
+
 # EditorConfig documentation: https://editorconfig.org
 
 root = true
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -23,7 +23,6 @@ jobs:
     uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
     with:
       license_header_check_project_name: "Swift"
-      license_header_check_enabled: false
       docs_check_enabled: false
       unacceptable_language_check_enabled: false
       format_check_enabled: false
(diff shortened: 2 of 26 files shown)


grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

grynspan:
@swift-ci test

Sources/_TestingInternals/Versions.cpp
```diff
@@ -10,11 +10,37 @@
...
 #if defined(SWT_TESTING_LIBRARY_VERSION)
+  // The current environment explicitly specifies a version string to return.
   return SWT_TESTING_LIBRARY_VERSION;
+#elif __has_embed("../../version.txt")
+  static constinit auto version = [] () constexpr {
+    // Read the version from version.txt at the root of the package's repo.
+    char version[] = {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wc23-extensions"
+#embed "../../version.txt"
```
stmontgomery: Looks like you upper-cased the `VERSION.txt` filename, so can we fix here and the other (2-3) spots which reference it in this PR?

stmontgomery:
Also, looks like there's a conflict with my PR #1312 since I touched many CMake files

grynspan:
Will fix the conflict.

stmontgomery (approved):

3405691582:
FYI: for OpenBSD, the native platform compiler doesn't support C23 yet, so this causes an error (`function-like macro '__has_embed' is not defined`). I'll send a pr soon to add an extra `__STDC_VERSION__` check around this after some testing.

grynspan:
We should always be building with (a recent) clang, but if there's an OpenBSD-specific issue, can it be solved by setting the `SWT` constant at build time?

3405691582:
I don't think so, but instead, we could probably also do something like a platform specific `-D__has_embed(x)=0` to avoid the issue too, which might be a little neater.

grynspan:
Or just `#if defined(__has_embed) && __has_embed(...) ...` no?

3405691582:
That won't work on older clang versions (see e.g. https://godbolt.org/z/edGb78fbd).

PR #1305, opened 2025-09-09
Add task names to structured tasks we create.

### LINKED ISSUE #1303: Provide useful names for Tasks where possible
### Motivation

In a test process running lots of complex asynchronous code, when things go wrong, it can be difficult to debug the process and understand what each of the Tasks in the process are supposed to be doing.

### Proposed solution

To help make this a little easier, whenever the testing library creates a Task for running asynchronous work, it should attempt to specify a `name` reflecting the task's purpose to aid with debugging.

Potentially interesting info:
* Is it a task spun up for producing the runner plan / evaluating arguments, or for running a test?
* What is the name/identifier of the test that the task for?

grynspan:
This PR sets a task name for every child task created by Swift Testing. These names are useful for debugging.

Example result:

Screenshot 2025-09-10 at 11 49 42 AM

Resolves #1303.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -8,6 +8,23 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
+/// Make a (decorated) task name from the given undecorated task name.
+///
+/// - Parameters:
+///   - taskName: The undecorated task name to modify.
+///
+/// - Returns: A copy of `taskName` with a common prefix applied, or `nil` if
+///   `taskName` was `nil`.
+func makeTaskName(_ taskName: String?) -> String? {
+  let prefix = "[Swift Testing]"
+  return taskName.map { taskName in
+#if DEBUG
+    precondition(!taskName.hasPrefix(prefix), "Applied prefix '\(prefix)' to task name '\(taskName)' twice. Please file a bug report at https://github.com/swiftlang/swift-testing/issues/new")
+#endif
+    return "\(prefix) \(taskName)"
+  }
+}
+
 /// A type that runs tests according to a given configuration.
 @_spi(ForToolsIntegrationOnly)
 public struct Runner: Sendable {
@@ -162,7 +179,8 @@ extension Runner {
     try await withThrowingTaskGroup { taskGroup in
       for element in sequence {
         // Each element gets its own subtask to run in.
-        taskGroup.addTask(name: taskNamer(element)) {
+        let taskName = taskNamer(element)
+        taskGroup.addTask(name: makeTaskName(taskName)) {
           try await body(element)
         }
 
@@ -442,12 +460,12 @@ extension Runner {
         }
 
         await withTaskGroup { [runner] taskGroup in
-          let taskName = if iterationCount > 1 {
+          var taskName = if iterationCount > 1 {
             "test run (iteration #\(iterationIndex + 1))"
           } else {
             "test run"
           }
-          _ = taskGroup.addTaskUnlessCancelled(name: taskName) {
+          _ = taskGroup.addTaskUnlessCancelled(name: makeTaskName(taskName)) {
             try? await _runStep(atRootOf: runner.plan.stepGraph)
           }
           await taskGroup.waitForAll()
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -320,7 +320,13 @@ extension Runner {
 
     // Figure out how to name child tasks.
     func taskNamer(_ childGraph: Graph<String, Plan.Step?>) -> String? {
-      childGraph.value.map { $0.test.humanReadableName() }
+      childGraph.value.map { step in
+        let testName = step.test.humanReadableName()
+        if step.test.isSuite {
+          return "suite \(testName)"
+        }
+        return "test \(testName)"
+      }
     }
 
     // Run the child nodes.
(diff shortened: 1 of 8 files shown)


grynspan:
@swift-ci test

Sources/Testing/Running/Runner.swift
```diff
@@ -150,16 +150,19 @@ extension Runner {
   /// - Parameters:
   ///   - sequence: The sequence to enumerate.
   ///   - body: The function to invoke.
+  ///   - elementNamer: A function to invoke for each element in `sequence`. The
```
stmontgomery:
```suggestion
  ///   - taskNamer: A function to invoke for each element in `sequence`. The
```

Sources/Testing/Traits/TimeLimitTrait.swift
```diff
@@ -263,6 +263,7 @@ struct TimeoutError: Error, CustomStringConvertible {
 /// Invoke a function with a timeout.
 ///
 /// - Parameters:
+///   - taskName: The name of the child task that runs `body`.
```
stmontgomery:
```suggestion
///   - taskName: The name of the child task that runs `body`, if any.
```

grynspan:
@swift-ci test

Sources/Testing/Test+Discovery.swift
```diff
@@ -100,8 +101,9 @@ extension Test {
       if useLegacyMode && result.isEmpty {
         let generators = Generator.allTypeMetadataBasedTestContentRecords().lazy.compactMap { $0.load() }
         await withTaskGroup { taskGroup in
-          for generator in generators {
-            taskGroup.addTask { await generator.rawValue() }
+          for (i, generator) in generators.enumerated() {
+            let taskName = "legacy test discovery (#\(i))"
```
stmontgomery: Although our implementation refers to this as "legacy", it's still the current supported discovery mechanism in production toolchains. It might be better for the user-visible task name to omit the word "legacy" until we enable using the new mechanism in production. (We could leave a "FIXME" comment to remind us to update this when that happens.)

Sources/Testing/Running/Runner.swift
```diff
@@ -314,8 +318,13 @@ extension Runner {
       }
     }
 
+    // Figure out how to name child tasks.
+    func taskNamer(_ childGraph: Graph<String, Plan.Step?>) -> String? {
+      childGraph.value.map { $0.test.humanReadableName() }
```
stmontgomery: With the new prefix, this will end up being something like this, right?

> [Swift Testing] "Foo bar"

if the test has a display name `"Foo bar"`. Display names in particular might be somewhat hard to parse out of context and understand that it represents an individual test. It's more clear when it's a function name, but I worry the display name case will be unclear. Maybe refine the prefix further and have

> [Swift Testing] Test: "Foo bar"
…or…
> [Swift Testing] Test: anotherExample()

grynspan: Changed.

grynspan:
@swift-ci test

Sources/Testing/Test+Discovery.swift
grynspan: "type-based"?

grynspan:
@swift-ci test

Sources/Testing/Support/Additions/TaskAdditions.swift
```diff
@@ -0,0 +1,26 @@
...
+//
+
+/// Make a (decorated) task name from the given undecorated task name.
+///
+/// - Parameters:
+///   - taskName: The undecorated task name to modify.
+///
+/// - Returns: A copy of `taskName` with a common prefix applied, or `nil` if
+///   `taskName` was `nil`.
+func makeTaskName(_ taskName: String?) -> String? {
```
jerryjrchen: Could we consider calling this `makeDecoratedTaskName`? It's used a lot in contexts where we already have a "task name", e.g.:
```swift
        let taskName = taskNamer(element)
        taskGroup.addTask(name: makeTaskName(taskName)) {
```

Which would be a little clearer as:
```swift
        let taskName = taskNamer(element)
        taskGroup.addTask(name: makeDecoratedTaskName(taskName)) {
```

Sources/Testing/Test+Discovery.swift
stmontgomery: Sure

Sources/Testing/Support/Additions/TaskAdditions.swift
grynspan: Given the usage pattern (ensconcing the string being passed to `addTask(name:)`) I think that would end up adding verbosity without a clear benefit.

We could define our own `Task.Name` type:

```swift
extension Task<Never, Never> {
  struct Name: Sendable, ExpressibleByStringLiteral, ExpressibleByStringInterpolation {
    ...

    func decorate() -> String {
      "[Swift Testing] \(self.whatever)"
    }
  }
}

...

let taskName: Task.Name = "blah blah"
taskGroup.addTask(name: taskName.decorate()) {
}
```

But that's probably ultimately overkill.

grynspan: Renamed the function `decorateTaskName()` either way.

grynspan:
@swift-ci test

Sources/Testing/Running/Runner.swift
```diff
@@ -150,16 +150,20 @@ extension Runner {
   /// - Parameters:
   ///   - sequence: The sequence to enumerate.
   ///   - body: The function to invoke.
+  ///   - taskNamer: A function to invoke for each element in `sequence`. The
+  ///     result of this function is used to name each child task.
```
stmontgomery, line 153: nit: I think the parameter is before `body`

grynspan:
@swift-ci test

PR #1312, opened 2025-09-11
Enable license header soundness check

stmontgomery:
Enable the (currently-disabled) license header soundness check via GitHub Actions. I expect this will fail, but I'm trying to first understand what the failures are so we can hopefully resolve them and enable this.

### Checklist:


stmontgomery:
See also: https://github.com/swiftlang/github-workflows/pull/158

PR #1313, opened 2025-09-12
Enable the unacceptable language soundness check and resolve identified issues

stmontgomery:
This enables the (currently-disabled) unacceptable language soundness check via GitHub Actions and resolves all the issues that running the script identified.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/LimitingExecutionTime.md
+++ b/Sources/Testing/Testing.docc/LimitingExecutionTime.md
@@ -18,7 +18,7 @@ Some tests may naturally run slowly: they may require significant system
 resources to complete, may rely on downloaded data from a server, or may
 otherwise be dependent on external factors.
 
-If a test may stall indefinitely or may consume too many system resources to
+If a test might stall indefinitely or might consume too many system resources to
 complete effectively, consider setting a time limit for it so that it's marked
 as failing if it runs for an excessive amount of time. Use the
 ``Trait/timeLimit(_:)-4kzjp`` trait as an upper bound:


Sources/Testing/Testing.docc/LimitingExecutionTime.md
```diff
@@ -18,7 +18,7 @@ Some tests may naturally run slowly: they may require significant system
 resources to complete, may rely on downloaded data from a server, or may
 otherwise be dependent on external factors.
 
-If a test may hang indefinitely or may consume too many system resources to
+If a test may stall indefinitely or may consume too many system resources to
```
grynspan: This is publicly visible. @iamleeg is "stall" an acceptable substitute here?

stmontgomery: I also considered "suspend", although that has some existing connotations with Swift concurrency. "Pause" or "stop" are also options, although stall is probably still my first choice. Interested in others' suggestions

iamleeg: I think "stall" is a good choice or "wait", I've been using "freeze" in the context of UI performance but that doesn't apply here. By the way, I might replace "may" with "might", because "may" has a second meaning of "has permission to".

stmontgomery: Ok, thanks everyone! Since there's general support for the word 'stall' and no strong feelings for any of the alternatives, I'll stick with that.

> By the way, I might replace "may" with "might", because "may" has a second meaning of "has permission to".

Sure, I'll swap in the word "might"

PR #1314, opened 2025-09-12
Enable the Docs soundness check

stmontgomery:
Enable the (currently-disabled) Docs soundness check via GitHub Actions.

### Checklist:

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -23,6 +23,6 @@ jobs:
     uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
     with:
       license_header_check_project_name: "Swift"
-      docs_check_container_image: "swiftlang/swift:nightly-6.2-noble"
+      docs_check_container_image: "swift:6.2-noble"
       format_check_enabled: false
       api_breakage_check_enabled: false


Sources/Testing/Testing.docc/Documentation.md
```diff
@@ -35,10 +35,8 @@ their problems.
 
 #### Related videos
 
-@Links(visualStyle: compactGrid) {
-  - <doc://com.apple.documentation/videos/play/wwdc2024/10179>
-  - <doc://com.apple.documentation/videos/play/wwdc2024/10195>
-}
+- [Meet Swift Testing](https://developer.apple.com/videos/play/wwdc2024/10179)
```
grynspan, line 38: Can you explain why this particular change was necessary?

Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -556,9 +556,11 @@ test function with an instance of this trait type to control whether it runs:
   }
 }
 
-<!-- TODO: document Test.cancel() and Test.Case.cancel() here
+<!-- TODO: document Test.cancel() and Test.Case.cancel() here, and update
+     relevant links to use proper DocC symbol references.
+
 If a test has already started running and you determine it cannot complete and
-should end early without failing, use ``Test/cancel(_:sourceLocation:)`` instead
+should end early without failing, use `Test/cancel(_:sourceLocation:)` instead
```
grynspan, line 563: I assume it complained here because the symbol isn't valid yet?

stmontgomery: Right

Sources/Testing/Testing.docc/Documentation.md
stmontgomery: This is because the swift-docc-plugin does not support links of this style. It’s off topic for this repo, but platform vendors which distribute Swift Testing can choose to revert this specific change if they wish.

stmontgomery: Based on that, I ended up removing the `@Links` directive so that it would render to a reasonable format when using swift-docc-plugin. This has a nice side effect of allowing these links to be visible on our rendered docs on SwiftPackageIndex for the first time; currently they are missing.

grynspan (approved):

PR #1318, opened 2025-09-16
OpenBSD needs execinfo as well.

3405691582:
Add OpenBSD to the CMake clause that adds libexecinfo.

### Motivation:

`swift test` fails on some projects with `ld: error: undefined reference due to --no-allow-shlib-undefined: backtrace`; `backtrace` is provided by `libexecinfo` on OpenBSD.

### Modifications:

Add OpenBSD to the CMake clause that adds libexecinfo.

### Checklist:


stmontgomery (approved):

stmontgomery (commented):
Thanks!

PR #1319, opened 2025-09-16
Attachment lifetimes

grynspan:
This PR introduces a new experimental trait, `.savingAttachments(if:)`, that can be used to control whether a test's attachments are saved or not.

XCTest has API around the [`XCTAttachment.Lifetime`](https://developer.apple.com/documentation/xctest/xctattachment/lifetime-swift.enum) enumeration that developers can use to control whether attachments are saved to a test report in Xcode. This enumeration has two cases:

```objc
/*
 * Attachment will be kept regardless of the outcome of the test.
 */
XCTAttachmentLifetimeKeepAlways = 0,

/*
 * Attachment will only be kept when the test fails, and deleted otherwise.
 */
XCTAttachmentLifetimeDeleteOnSuccess = 1
```

I've opted to implement something a bit more granular. A developer can specify `.savingAttachments(if: .testFails)` and `.savingAttachments(if: .testPasses)` or can call some custom function of their own design like `runningInCI` or `hasPlentyOfFloppyDiskSpace`. The default behaviour if this trait is not used is to always save attachments, which is equivalent to `XCTAttachmentLifetimeKeepAlways`. `XCTAttachmentLifetimeDeleteOnSuccess` is, in effect, equivalent to `.savingAttachments(if: .testFails)`, but I hope reads a bit more clearly in context.

Here's a usage example:

```swift
@Test(.savingAttachments(if: .testFails))
func `best test ever`() {
  Attachment.record("...") // only saves to the test report or to disk if the
                           // next line is uncommented.
  // Issue.record("sadness")
}
```

I've taken the opportunity to update existing documentation for `Attachment` and `Attachable` to try to use more consistent language: a test records an attachment and then the testing library saves it (somewhere). I'm sure I've missed some spots, so please point them out if you see them.

Resolves rdar://138921461.

### Checklist:


Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -47,9 +47,19 @@ extension Runner {
       return
     }
 
-    configuration.eventHandler = { [eventHandler = configuration.eventHandler] event, context in
+    configuration.eventHandler = { [oldEventHandler = configuration.eventHandler] event, context in
```
grynspan, line 50: This code isn't new, but needed to move from `handleEvent` as it imposed the wrong layering (saving the attachment before any event handlers actually ran.)

Sources/Testing/Traits/AttachmentSavingTrait.swift
```diff
@@ -0,0 +1,300 @@
...
+  public struct Condition: Sendable {
+    /// The testing library saves attachments if the test passes.
+    public static var testPasses: Self {
+      Self { !$0.hasFailed }
+    }
+
+    /// The testing library saves attachments if the test fails.
+    public static var testFails: Self {
+      Self { $0.hasFailed }
+    }
```
stmontgomery, line 49: I think we should include a built-in condition representing a test which did not _fail_ but recorded one or more issues with **warning** severity. If a test recorded a warning issue, preserving the attachments might be useful to help diagnose the problem.

grynspan: Sure, gimme a symbol name and I can add it.

stmontgomery: We brainstormed this together, and came up with a more general purpose idea:

```swift
public static func testRecordsIssue(
  _ comment: Comment? = nil,
  matching predicate: @Sendable @escaping (_ issue: Issue) -> Bool
) -> Self
```

Example usage:

```swift
@Test(.savingAttachments(if: .testRecordsIssue { $0.severity >= .warning }))
func example() {
  ...
}
```

Sources/Testing/Attachments/AttachableWrapper.swift
```diff
@@ -8,9 +8,8 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-/// A protocol describing a type that can be attached to a test report or
-/// written to disk when a test is run and which contains another value that it
-/// stands in for.
+/// A protocol describing a type whose instances can be recorded and saved as
+/// part of a test run and which contains another value that it stands in for.
```
stmontgomery, line 12: Nice little refinement to this description

Sources/Testing/Traits/AttachmentSavingTrait.swift
```diff
@@ -0,0 +1,337 @@
...
+
+    /// Whether or not this condition needs to inspect individual issues (which
+    /// implies a slower path.)
+    fileprivate var inspectsIssues = false
+
+    /// The condition function.
+    ///
+    /// - Parameters:
+    ///   - condition: The function to call. The result of this function
+    ///     determines if the condition is satisfied or not.
```
stmontgomery: Do the docs for this closure property need to be updated?

Sources/Testing/Traits/AttachmentSavingTrait.swift
```diff
@@ -0,0 +1,337 @@
...
+
+  public func scopeProvider(for test: Test, testCase: Test.Case?) -> Self? {
+    // This function should apply directly to test cases only. It doesn't make
+    // sense to apply it to suites or test functions since they don't run their
+    // own code.
+    //
+    // NOTE: this trait can't reliably affect attachments recorded when other
+    // traits are evaluated (we may need a new scope in the TestScoping protocol
+    // for that.)
+    testCase != nil ? self : nil
```
stmontgomery, line 120: I think in this case, since `isRecursive` is `true`, you can just omit the `scopeProvider()` implementation entirely and it will have the equivalent behavior you have here.

Sources/Testing/Traits/AttachmentSavingTrait.swift
```diff
@@ -0,0 +1,337 @@
...
+  ///   the `--attachments-path` option. The testing library saves attachments
+  ///   to the specified directory.
+  ///
+  /// If you add this trait to a test, any attachments that test records are
+  /// stored in memory until the test finishes running. The testing library then
+  /// evaluates `condition` and, if the condition is met, saves the attachments.
+  public static func savingAttachments(
+    if condition: Self.Condition,
+    sourceLocation: SourceLocation = #_sourceLocation
+  ) -> Self {
```
stmontgomery, line 249: How about we add a `Comment` field to these? Could be useful to describe the reason why some attachments are being kept or not kept.

(I'm not sure exactly where it would be used initially, but I'm also unsure where the proposed `sourceLocation:` parameter would be shown for these traits, so it seems like we might as well include both of them or neither.)

Sources/Testing/Traits/AttachmentSavingTrait.swift
```diff
@@ -0,0 +1,337 @@
...
+/// test records are stored in memory until the test finishes running. The
+/// testing library then evaluates the instance's condition and, if the
+/// condition is met, saves the attachments.
+@_spi(Experimental)
+public struct AttachmentSavingTrait: TestTrait, SuiteTrait {
+  /// A type that describes the conditions under which the testing library
+  /// will save attachments.
+  ///
+  /// You can pass instances of this type to ``Trait/savingAttachments(if:)``.
+  public struct Condition: Sendable {
```
stmontgomery, line 40: My biggest feedback on this PR and associated proposal is that I think there should be a streamlined way to express "save attachments _unconditionally_", whether the test passed or failed. That can be expressed in the current implementation using `.savingAttachments(if: true)` but it feels pretty indirect.

Maybe zero-arg `.savingAttachments()` (or `.savesAttachments()`)?

Sources/Testing/Events/Event.swift
```diff
@@ -191,6 +191,14 @@ public struct Event: Sendable {
   /// The instant at which the event occurred.
   public var instant: Test.Clock.Instant
 
+  /// Whether or not this event was deferred.
+  ///
+  /// A deferred event is handled significantly later than when was posted.
+  ///
+  /// We currently use this property in our tests, but do not expose it as API
+  /// or SPI. We can expose it in the future if tools need it.
+  var wasDeferred: Bool = false
```
stmontgomery, line 201: Could this be guarded by `#if DEBUG`? The tests which use it already rely on debug builds by virtue of `@testable`

Sources/Testing/Traits/AttachmentSavingTrait.swift
grynspan: Clarify?

Sources/Testing/Traits/AttachmentSavingTrait.swift
grynspan: I'd prefer it be explicit so I have a place to put the comment, but we can always delete this code later if we find it's redundant.

Sources/Testing/Events/Event.swift
grynspan: No because our tests may build in release mode with testability enabled too.

Sources/Testing/Traits/AttachmentSavingTrait.swift
grynspan: That's the default behaviour, so the streamlined spelling is "".

Sources/Testing/Traits/AttachmentSavingTrait.swift
grynspan: We can add one later if we have a use for it, but at this time we do not.

Sources/Testing/Events/Event.swift
stmontgomery: We can guard the small section in the test that checks it, too

Sources/Testing/Traits/AttachmentSavingTrait.swift
stmontgomery: It specifies a `- Parameter` named `condition` but I don't see such a param

grynspan: It's `borrowing Context`, I must have dropped the label. Will re-add it.

grynspan: Eh, removed the comment.

Sources/Testing/Events/Event.swift
grynspan: Done.

Sources/Testing/Traits/AttachmentSavingTrait.swift
grynspan: Discussed with Stuart and we'll defer this question to a separate PR.

PR #1320, opened 2025-09-17
Work around lack of embed in OpenBSD's clang.

3405691582:
Work around lack of embed in OpenBSD's clang.

### Motivation:

For various reasons, the Swift toolchain for OpenBSD relies on using the platform's native clang, which is 16. clang 19 is the most recent version that will not emit an error with the new __has_embed features in C23.

Since swift-testing is experimentally supported by OpenBSD and thus to make swift-testing build again on the platform, work around the issue with a platform-specific command-line specified macro override in swiftpm and in cmake, though this does not seem to block the cmake build.

Furthermore, we can use cmake trickery to subsitute the version file contents instead of using embed. This may not be possible to do with swiftpm, but I don't know for sure.

### Modifications:

* Add -D__has_embed=0 and use cmake's file(STRINGS) to the cmake build.
* Add -D__has_embed=0 to Package.swift when the platform is OpenBSD.

### Checklist:


3405691582:
@swift-ci please test.

3405691582:
cc @grynspan from our conversation in #1304.

Package.swift
```diff
@@ -200,7 +200,9 @@ let package = Package(
     .target(
       name: "_TestingInternals",
       exclude: ["CMakeLists.txt"],
-      cxxSettings: .packageSettings
+      cxxSettings: .packageSettings + [
+          .unsafeFlags(["-D__has_embed(x)=0"], .when(platforms: [.openbsd]))
```
grynspan: There's a proper way to define macros in C++ in SwiftPM. See waaaay below (and maybe add this macro there.)

Although if you define `SWT_TESTING_LIBRARY_VERSION` do we still need this?

Sources/_TestingInternals/CMakeLists.txt
```diff
@@ -23,6 +23,12 @@ if("${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}" STREQUAL "MSVC" OR
 elseif(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
   target_compile_options(_TestingInternals PRIVATE
     -fno-exceptions -fPIC)
+elseif(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
```
grynspan: It should not be necessary to define `__has_embed` if `SWT_TESTING_LIBRARY_VERSION` is defined since the preprocessor will (should?) ignore the `#elif` when it's defined.

Can we define `SWT_TESTING_LIBRARY_VERSION` in CompilerSettings.cmake instead of this file?

Sources/_TestingInternals/CMakeLists.txt
```diff
@@ -23,6 +23,12 @@ if("${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}" STREQUAL "MSVC" OR
...
+  file(STRINGS "../../VERSION.txt" SWT_VERSION_FILE_CONTENTS)
```
grynspan: I think you want `LIMIT_COUNT 1`?

Package.swift
3405691582: You're right -- it might be simpler just to set SWT_TESTING_LIBRARY_VERSION when OpenBSD to 0.0, given that I don't know how to do something like `file(STRINGS)` in spm and thus version information wouldn't be available anyway.

Sources/_TestingInternals/CMakeLists.txt
3405691582: Good point. Done.

Sources/_TestingInternals/CMakeLists.txt
3405691582: Done.

Package.swift
grynspan: If you set it to `0` (not `"0"`) we'll interpret it as `nullptr` and fall back gracefully in Swift. :)

Package.swift
```diff
@@ -466,6 +468,7 @@ extension Array where Element == PackageDescription.CXXSetting {
 
       .define("SWT_NO_LEGACY_TEST_DISCOVERY", .whenEmbedded()),
       .define("SWT_NO_LIBDISPATCH", .whenEmbedded()),
+      .define("SWT_TESTING_LIBRARY_VERSION", to: "\"0.0\"", .when(platforms: [.openbsd]))
```
grynspan:
```suggestion

      // OpenBSD's version of clang doesn't support __has_embed.
      .define("SWT_TESTING_LIBRARY_VERSION", to: "0", .when(platforms: [.openbsd]))
```

Sources/_TestingInternals/CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ if("${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}" STREQUAL "MSVC" OR
 elseif(CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
   target_compile_options(_TestingInternals PRIVATE
     -fno-exceptions -fPIC)
+elseif(CMAKE_SYSTEM_NAME STREQUAL "OpenBSD")
+  target_compile_options(_TestingInternals PRIVATE
+    -fno-exceptions
+    -DSWT_TESTING_LIBRARY_VERSION="${SWT_VERSION_FILE_CONTENTS}")
```
grynspan: Looks like you moved the definition of `${SWT_VERSION_FILE_CONTENTS}` to CompilerSettings.cmake, but not that of `SWT_TESTING_LIBRARY_VERSION`.

What' I'd ask you do here is:

- Remove this `-D` line from this file, as it's inconsistent with how we define other C++ macros
- Add an equivalent `add_compile_definitions()` call in CompilerSettings.cmake after the call to `file(STRINGS)`
- And this last one is nitpicky but call the intermediate CMake value `${SWT_TESTING_LIBRARY_VERSION}` instead of `${SWT_VERSION_FILE_CONTENTS}` for consistency and so it's clear they refer to the same value.

Thanks much! :)

Package.swift
3405691582: Done.

Sources/_TestingInternals/CMakeLists.txt
3405691582: Ah, got it. I thought to keep it more local to where it is used but got this to apply neatly with some generator expression magic.

Package.swift
3405691582: Done manually (while testing).

grynspan:
I assume you've been able to test this change, yes? Ideally both the package and CMake avenues, but since OpenBSD support is experimental I'll be okay with "whatever @3405691582 thinks is sufficient." 😬

3405691582:
Yes, I have been running `swift build` and the CMake build as part of the toolchain to test.

grynspan:
I've updated the PR description to cover the Amazon Linux 2 failure. See here: https://ci.swift.org/job/oss-swift-package-amazon-linux-2-aarch64/4545

The buildbot over yonder has clang 13:

> [2025-09-15T04:17:46.060Z] -- The C compiler identification is Clang 13.0.0
> [2025-09-15T04:17:46.060Z] -- The CXX compiler identification is Clang 13.0.0

grynspan:
Sorry, looks like an earlier comment I made didn't land? We noticed the same issue on Amazon Linux 2 which has clang **13** 😱 hence me adjusting the PR. Thanks!

3405691582:
Out of curiosity -- do you need to adjust the Package.swift for the Linux variants?

grynspan:
We can't reliably distinguish Amazon Linux 2 in the package, so no. The effect is that there will be a build warning when building from a package on AL2, but that's unavoidable. AL2 is quite old and an outlier in various other ways, so this doesn't bother me too too much.

grynspan:
Looks like Android was impacted as well. https://ci-external.swift.org/job/swift-main-windows-toolchain/1597

grynspan:
Sorry for the noise, @3405691582! Just working through CI problems here. We'll merge this change shortly.

3405691582:
No problems, it's interesting to see the ancillary changes :D

cmake/modules/shared/CompilerSettings.cmake
```diff
@@ -42,3 +42,9 @@ if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
   add_compile_definitions("SWT_NO_DYNAMIC_LINKING")
   add_compile_definitions("SWT_NO_PIPES")
 endif()
+
+file(STRINGS "../VERSION.txt" SWT_TESTING_LIBRARY_VERSION LIMIT_COUNT 1)
```
mhegazy, line 46: @3405691582 and @grynspan this change broke building testing macros that include this but do not have the same relative path see https://github.com/swiftlang/swift-testing/blob/05b46397beaf2dfad30fb853faa725fc4089cdc3/Sources/TestingMacros/CMakeLists.txt#L19.

grynspan: D'oh. I knew that looked fishy. `${PROJECT_SOURCE_DIR}/VERSION.txt` should be the right path I think.

grynspan: #1327.

grynspan:
@3405691582 I put together a branch [here](https://github.com/swiftlang/swift-testing/commit/cb1d0c2be03ce07baa1a042d0d8f2dbf8dd113b9) that uses `.incbin` instead of `#embed`. However, given that OpenBSD's maintainers merged LLVM 19 [in June](https://github.com/openbsd/src/commit/6e236bcae9e00c4e5ae6b71137e522854c930ce8), it's only a matter of time before a new release includes it and this all becomes moot. So I haven't opened a PR, but it's in our back pocket if it becomes critical.

PR #1327, opened 2025-09-17
Fix the path to VERSION.txt used in CMake.

grynspan:
Fixes a bug introduced in #1320.

### Checklist:


grynspan:
Merging before CI finishes as none of it uses CMake anyway so this won't affect it.

mhegazy:
thanks!

PR #1328, opened 2025-09-18
Drop remaining Swift 6.1 compiler support.

grynspan:
This PR drops the remaining bits in the package that build with the Swift 6.1 compiler. Swift 6.2 has been released and our main branch no longer supports Swift 6.1 toolchains (and hasn't for a while now.)

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -164,7 +164,7 @@ let package = Package(
         "Testing",
       ],
       path: "Tests/_MemorySafeTestingTests",
-      swiftSettings: .packageSettings + .strictMemorySafety()
+      swiftSettings: .packageSettings + [.strictMemorySafety()]
     ),
 
     .macro(


Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -349,10 +349,7 @@ extension ExitTest {
   ///
   /// - Warning: This function is used to implement the
   ///   `#expect(processExitsWith:)` macro. Do not use it directly.
-#if compiler(>=6.2)
-  @safe
-#endif
-  public static func __store<each T>(
+  @safe public static func __store<each T>(
```
jerryjrchen, line 352: What does `@safe` do? I'm having a hard time finding it in the docs somehow

jerryjrchen (approved):
LGTM. Just had one question about a feature I'm personally not familiar with.

Sources/Testing/ExitTests/ExitTest.swift
grynspan: It marks these functions as safe to use despite taking unsafe arguments. See [Strict Memory Safety](https://docs.swift.org/compiler/documentation/diagnostics/strict-memory-safety/).

stmontgomery: > What does `@safe` do? I'm having a hard time finding it in the docs somehow

That might be a good GitHub documentation enhancement request to file for the [Swift Book](https://github.com/swiftlang/swift-book/issues/new/choose)

jerryjrchen: Added here: https://github.com/swiftlang/swift-book/issues/394

PR #1329, opened 2025-09-19
Merge more of the image attachments codebase duplicated between Darwin and Windows.

grynspan:
This PR eliminates a bunch of duplicated code between Darwin and Windows that's used to support image attachments. We're able to do this because `_AttachableImageWrapper` is now a class and can be included in conforms-to generic constraints, so we can write
`where AttachableValue: _AttachableImageWrapper<T> & AttachableWrapper`. It was previously a structure and structures can't be used in generic constraints this way.

There's a new public, underscored protocol introduced here, `_AttachableAsImage`, that serves as a "base" protocol for `AttachableAsCGImage` and `AttachableAsIWICBitmapSource`. I intend to promote this protocol to API, but it will need a Swift Evolution proposal first. I'm going to include it in a future "image attachments refinement" proposal that will cover a few other things.

### Checklist:


Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
```diff
@@ -39,7 +39,7 @@ private import ImageIO
 ///   @Available(Swift, introduced: 6.3)
 /// }
 @available(_uttypesAPI, *)
-public protocol AttachableAsCGImage: SendableMetatype {
+public protocol AttachableAsCGImage: _AttachableAsImage, SendableMetatype {
```
jerryjrchen, line 42: Nit: Is the `SendableMetatype` conformance declaration needed here if `_AttachableAsImage` already declares `SendableMetatype`?

stmontgomery, line 42: Since `SendableMetatype` is now stated in the parent protocol, does it still make sense to reiterate here?

Sources/Overlays/_Testing_WinSDK/Attachments/AttachableAsIWICBitmapSource.swift
```diff
@@ -113,7 +113,7 @@ public protocol _AttachableByAddressAsIWICBitmapSource {
 /// you have an image in another format that needs to be attached to a test,
 /// first convert it to an instance of one of the types above.
 @_spi(Experimental)
-public protocol AttachableAsIWICBitmapSource: SendableMetatype {
+public protocol AttachableAsIWICBitmapSource: _AttachableAsImage, SendableMetatype {
```
stmontgomery, line 116: Since `SendableMetatype` is now stated in the parent protocol, does it still make sense to reiterate here?

Sources/Testing/Attachments/Images/_AttachableAsImage.swift
```diff
@@ -0,0 +1,54 @@
...
+  ///
+  /// This function is not responsible for releasing the image returned from
+  /// `_copyAttachableIWICBitmapSource(using:)`.
+  ///
+  /// The default implementation of this function when `Self` conforms to
+  /// `Sendable` does nothing.
+  ///
+  /// This function is not part of the public interface of the testing library.
+  /// It may be removed in a future update.
+  func _deinitializeAttachableValue()
```
stmontgomery, line 53: Would it make sense for the default, no-op implementations of these methods to be in an `extension` here in this file? Or is doing so in the child protocols needed to maintain compatibility?

Sources/Testing/Attachments/Images/_AttachableAsImage.swift
```diff
@@ -0,0 +1,54 @@
...
+  /// Manually deinitialize any resources associated with this image.
+  ///
+  /// The implementation of this function cleans up any resources (such as
+  /// handles or COM objects) associated with this image. The testing library
+  /// automatically invokes this function as needed.
+  ///
+  /// This function is not responsible for releasing the image returned from
+  /// `_copyAttachableIWICBitmapSource(using:)`.
+  ///
+  /// The default implementation of this function when `Self` conforms to
```
jerryjrchen, line 48: Would it make sense to codify this below? e.g.

```swift
@available(_uttypesAPI, *)
extension _AttachableAsImage where Self: Sendable {
  func _deinitializeAttachableValue() {}
}
```

Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
grynspan: Yes because `_AttachableAsImage` won't appear in documentation until it's promoted to API, so the required conformance to `SendableMetatype` won't be visible either.

Sources/Overlays/_Testing_WinSDK/Attachments/AttachableAsIWICBitmapSource.swift
grynspan: Yes because `_AttachableAsImage` won't appear in documentation until it's promoted to API, so the required conformance to `SendableMetatype` won't be visible either.

Sources/Testing/Attachments/Images/_AttachableAsImage.swift
grynspan: The default implementations are platform-specific. On Darwin, it's a no-op because images are refcounted by Swift. On Windows, it needs to be implemented on a type-by-type basis because the types we support do not participate in Swift ARC.

Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
```diff
@@ -95,6 +79,8 @@ extension AttachableAsCGImage {
   public var _attachmentScaleFactor: CGFloat {
     1.0
   }
+
+  public func _deinitializeAttachableValue() {}
```
jerryjrchen, line 83: I know this isn't a functional change, but I'm curious why this safe as a default no-op even when the conforming type to `AttachableAsCGImage` is non-`Sendable`.

Wouldn't it be possible to construct some `FooImage: AttachableAsCGImage` that copies manually managed memory during `_copyAttachableValue`, and then the default `_deinitializeAttachableValue` would then end up leaking it?

I'm thinking the alternative would be adding this implementation manually for the ObjC types that aren't `Sendable` but won't leak with a no-op of deinit.

Sources/Testing/Attachments/Images/Attachment+_AttachableAsImage.swift
```diff
@@ -8,9 +8,6 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-#if SWT_TARGET_OS_APPLE && canImport(CoreGraphics)
```
jerryjrchen, line 11: 🆒 This is effectively deduping the code in `Sources/Overlays/_Testing_WinSDK/Attachments/Attachment+AttachableAsIWICBitmapSource.swift` then?

Sources/Testing/Attachments/Images/Attachment+_AttachableAsImage.swift
```diff
@@ -52,12 +49,8 @@ extension Attachment {
     named preferredName: String? = nil,
     as imageFormat: AttachableImageFormat? = nil,
     sourceLocation: SourceLocation = #_sourceLocation
-  ) where T: AttachableAsCGImage, AttachableValue == _AttachableImageWrapper<T> {
-    let imageWrapper = _AttachableImageWrapper(
-      image: image._copyAttachableValue(),
-      imageFormat: imageFormat,
-      deinitializingWith: { _ in }
-    )
+  ) where AttachableValue: _AttachableImageWrapper<T> & AttachableWrapper {
```
jerryjrchen, line 52: I'm not really getting why this is using `AttachableWrapper` rather than `Attachable`. It compiles ok with just `Attachable` and the `Attachment` type only requires `AttachableValue: Attachable & ~Copyable` right?

Sources/Testing/Attachments/Images/_AttachableAsImage.swift
jerryjrchen: Kinda like https://github.com/swiftlang/swift-testing/pull/1329#discussion_r2370047409 I guess, but I would say this should only be added specifically for `Sendable` types

Sources/Testing/Attachments/Images/Attachment+_AttachableAsImage.swift
grynspan: Yes. So the diff gets a bit confused.

Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
grynspan: These underscored members aren't part of our API surface, and the only way you leak here is if you implement them wrong with a type that doesn't fully participate in Swift ARC (such as an unsafe pointer.) _Caveat emptor_.

Sources/Testing/Attachments/Images/_AttachableAsImage.swift
grynspan: This is already implemented in the places we need it to be implemented.

Sources/Testing/Attachments/Images/Attachment+_AttachableAsImage.swift
grynspan: Yes, it can compile either way, so it doesn't really matter which we pick, but `_AttachableImageWrapper` is a wrapper type so you could argue it's more semantically correct this way. (In other words, flip a coin.)

PR #1331, opened 2025-09-22
Exclude Xcode 16.4 from GitHub Actions.

grynspan:
We don't support building with Xcode 16.4 which only has Swift 6.1.

### Checklist:


jerryjrchen (approved):

grynspan:
Xcode 26 failure is unrelated.

grynspan:
Merging before CI completes since this change is removing a specific job and should not affect others.

PR #1332, opened 2025-09-22
Change the semantics of `Test.cancel()`.

grynspan:
This PR changes the semantics of `Test.cancel()` such that it only cancels the current test case if the current test is parameterized. This PR also removes `Test.Case.cancel()`, but we may opt to add `Test.Case.cancelAll()` or some such in the future to re-add an interface that cancels an entire parameterized test.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -762,7 +762,7 @@ extension ExitTest {
     }
     configuration.eventHandler = { event, eventContext in
       switch event.kind {
-      case .issueRecorded, .valueAttached, .testCancelled, .testCaseCancelled:
+      case .issueRecorded, .valueAttached, .testCancelled:
         eventHandler(event, eventContext)
       default:
         // Don't forward other kinds of event.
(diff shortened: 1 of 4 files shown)


Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -762,7 +762,7 @@ extension ExitTest {
     }
     configuration.eventHandler = { event, eventContext in
       switch event.kind {
-      case .issueRecorded, .valueAttached, .testCancelled, .testCaseCancelled:
+      case .issueRecorded, .valueAttached, .testCancelled:
```
grynspan, line 765: Exit tests no longer need to be concerned with `.testCaseCancelled` events.

Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -206,6 +206,7 @@ extension Test {
   static func withCurrent<R>(_ test: Self, perform body: () async throws -> R) async rethrows -> R {
     var runtimeState = Runner.RuntimeState.current ?? .init()
     runtimeState.test = test
+    runtimeState.testCase = nil
```
grynspan, line 209: This change looks out-of-place, but it ensures that when we set the current test, we clear the current test _case_. This has no effect on third-party tests but ensures when we run nested tests of our own that they don't get their states intermingled (which breaks `Test.cancel()`'s new semantics.)

stmontgomery: Good catch. This is probably something we'd probably want even if we'd never sought to add test cancellation

PR #1334, opened 2025-09-23
Promote `SourceLocation._filePath` to proper, non-underscored SPI.

grynspan:
This PR introduces an `@_spi var filePath` property on `SourceLocation` with the intent of replacing the existing underscored/unsupported `_filePath` property.

Because Xcode 16 through 26 directly encodes and decodes instances of `SourceLocation`, we can't just stop encoding/decoding the `_filePath` key and need to keep it there for at least a Swift release or two, until Apple's fork of Swift Testing has been updated to include the new encoding key. At that point we can remove the old one. (We'll need to introduce an `EncodedSourceLocation` structure at that point and keep including the underscored key for the v0 JSON stream, but that's not a "today" problem.)

### Checklist:

### DIFF
--- a/Tests/TestingTests/SourceLocationTests.swift
+++ b/Tests/TestingTests/SourceLocationTests.swift
@@ -132,7 +132,7 @@ struct SourceLocationTests {
 
   @available(swift, deprecated: 100000.0)
   @Test("SourceLocation._filePath property")
-  func sourceLocationFilePath() {
+  func sourceLocation_FilePath() {
     var sourceLocation = #_sourceLocation
     #expect(sourceLocation._filePath == #filePath)
 


Sources/Testing/SourceAttribution/SourceLocation.swift
```diff
@@ -167,4 +164,54 @@ extension SourceLocation: CustomStringConvertible, CustomDebugStringConvertible
 
 // MARK: - Codable
 
-extension SourceLocation: Codable {}
+extension SourceLocation: Codable {
+  private enum _CodingKeys: String, CodingKey {
+    case fileID
+    case filePath
+    case _filePath
```
stmontgomery: Can we include some documentation about the purpose of this legacy key?

Sources/Testing/SourceAttribution/SourceLocation.swift
```diff
@@ -167,4 +164,54 @@ extension SourceLocation: CustomStringConvertible, CustomDebugStringConvertible
...
+  public init(from decoder: any Decoder) throws {
+    let container = try decoder.container(keyedBy: _CodingKeys.self)
+    fileID = try container.decode(String.self, forKey: .fileID)
+    line = try container.decode(Int.self, forKey: .line)
+    column = try container.decode(Int.self, forKey: .column)
+
+    // For simplicity's sake, we won't be picky about which key contains the
+    // file path.
+    filePath = try container.decodeIfPresent(String.self, forKey: ._filePath)
+      ?? try container.decode(String.self, forKey: .filePath)
```
stmontgomery: I see the comment, but I think the newer key ought to be preferred if it's present. (That would allow us to eventually stop encoding the old key, at least in theory. It also feels more appropriate.) How about this instead:

```swift
filePath = try container.decodeIfPresent(String.self, forKey: .filePath)
  ?? try container.decodeIfPresent(String.self, forKey: ._filePath)
  ?? try container.decode(String.self, forKey: .filePath)
```

The first two are optional (decode**IfPresent**) intentionally, while the final one is non-optional.

grynspan: That's redundant and can just be spelled:

```swift
filePath = try container.decodeIfPresent(String.self, forKey: .filePath)
  ?? try container.decode(String.self, forKey: ._filePath)
```

The only difference being the thrown error's message will mention `_filePath` instead of `filePath`, but… I'm okay with that because the decode path is only used in our own unit tests.

PR #1336, opened 2025-09-24
Adopt typed throws in the implementation of `#expect()`/`#require()`.

grynspan:
This PR adopts typed throws (`throws(E)` instead of `throws`) in the `#expect()` and `#require()` macros' declarations and in their implementations. In particular, the return type of the implementation functions changes from `Result<R, any Error>` to `Result<R, ExpectationFailedError>` and all implementation functions that take a closure have been updated with an additional generic `E` parameter.

`#expect(throws:)` and `#require(throws:)` take two generic error types, `EExpected` and `EActual`. The former is the error type specified by the test author, while the latter is the actual error type inferred from the body closure. In common usage, both will resolve to `any Error`.

The deprecated `#expect {} throws: {}` and `#require {} throws: {}` variants are marked `@_unavailableInEmbedded` and continue to use `any Error`.

If you see `throws(any Error)` in the diff, take it as meaning that section of the code was audited, but for now we don't have something better to use than `any Error`.

### Checklist:


Sources/Testing/Support/Additions/ResultAdditions.swift
```diff
@@ -48,7 +48,7 @@ extension Result {
   ///
   /// - Warning: This function is used to implement the `#expect()` and
   ///   `#require()` macros. Do not call it directly.
-  @discardableResult public func __required<T>() throws -> T where Success == T? {
+  @discardableResult public func __required<T>() throws(any Error) -> T where Success == T? {
```
grynspan, line 51: Can't use `throws(some Error)` unfortunately.

(We'll likely need to rewrite this function for Embedded Swift to throw some public-but-opaque error type AKA `AnyTestingLibraryError`.)

PR #1338, opened 2025-09-25
Use generics to declare suite types in macro expansions.

grynspan:
This PR adopts generics instead of existentials when declaring suite types in the expansions of `@Suite` and `@Test` and in the interface of the internal `TypeInfo` helper type.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Macro.swift
+++ b/Sources/Testing/Test+Macro.swift
@@ -157,16 +157,16 @@ extension Test {
   ///
   /// - Warning: This function is used to implement the `@Test` macro. Do not
   ///   call it directly.
-  public static func __function(
+  public static func __function<S>(
     named testFunctionName: String,
-    in containingType: (any ~Copyable.Type)?,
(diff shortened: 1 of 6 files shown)


Sources/Testing/Test+Macro.swift
```diff
@@ -157,16 +157,16 @@ extension Test {
   ///
   /// - Warning: This function is used to implement the `@Test` macro. Do not
   ///   call it directly.
-  public static func __function(
+  public static func __function<S>(
     named testFunctionName: String,
-    in containingType: (any ~Copyable.Type)?,
+    in containingType: S.Type?,
```
jerryjrchen, line 162: `S` here is a shorthand for Suite?

grynspan: Yeah. As a bad habit from C++, I tend to spell my generic types with one letter unless it's ambiguous.

briancroom (approved):

PR #1342, opened 2025-09-25
Treat `@_unavailableInEmbedded` like other availability attributes during macro expansion.

grynspan:
This PR adds support for mapping `@_unavailableInEmbedded` to a condition trait like we do with `@available()` etc.

Example:

```swift
@_unavailableInEmbedded
@Test func `stuff that doesn't work in Embedded Swift`() {
  ...
}
```

Which expands to include a trait of the form `.__unavailableInEmbedded(sourceLocation: ...)` of type `ConditionTrait`.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/AvailabilityGuards.swift
+++ b/Sources/TestingMacros/Support/AvailabilityGuards.swift
@@ -140,24 +140,6 @@ private func _createAvailabilityTraitExpr(
   }
 }
 
-/// Create an expression that contains a test trait for symbols that are
-/// unavailable in Embedded Swift.
-///
-/// - Parameters:
-///   - attribute: The `@_unavailableInEmbedded` attribute.
-///   - context: The macro context in which the expression is being parsed.
-///
-/// - Returns: An instance of `ExprSyntax` representing an instance of
-///   ``Trait`` that can be used to prevent a test from running in Embedded
-///   Swift.
-private func _createNoEmbeddedAvailabilityTraitExpr(
-  from attribute: AttributeSyntax,
-  in context: some MacroExpansionContext
-) -> ExprSyntax {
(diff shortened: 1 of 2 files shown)


Sources/TestingMacros/Support/AvailabilityGuards.swift
```diff
@@ -290,5 +295,16 @@ func createSyntaxNode(
     }
   }
 
+  // Handle Embedded Swift.
+  if decl.unavailableInEmbeddedAttribute != nil {
```
jerryjrchen, line 299: Is this basically rewriting the test body to wrap with an availability check of `!hasFeature(Embedded)`? So that a test author doesn't need to manually wrap it themselves?

Tests/TestingTests/RunnerTests.swift
```diff
@@ -819,6 +819,29 @@ final class RunnerTests: XCTestCase {
     await fulfillment(of: [testStarted], timeout: 0.0)
   }
 
+  @Suite(.hidden) struct UnavailableInEmbeddedTests {
+    @Test(.hidden)
+    @_unavailableInEmbedded
+    func embedded() {}
```
jerryjrchen, line 825: Would it help to add some code that doesn't compile for embedded to the test body, just to ensure the availability guards we add work properly?

Sources/TestingMacros/Support/AvailabilityGuards.swift
grynspan: Not the test body, but our thunk function body.

Tests/TestingTests/RunnerTests.swift
grynspan: We have no way to actually run our tests in Embedded Swift, so that's moot for now.

PR #1346, opened 2025-10-01
AdvancedConsoleOutputRecorderTests is missing an import of Foundation

stmontgomery:
Import `Foundation` in `AdvancedConsoleOutputRecorderTests`.

This is required due to `MemberImportVisibility` although in practice this problem only shows up when building for non-macOS Apple platforms, such as iOS, due to the differing deployment target versions we use for non-macOS platforms.

### Checklist:


stmontgomery:
For more background context about this kind of build error, see the older PR https://github.com/swiftlang/swift-testing/pull/1108.

PR #1349, opened 2025-10-02
Normalize the order in which we specify our Linux/FreeBSD/OpenBSD checks.

grynspan:
This PR adjusts some of our `#if os(...)` checks to always specify Linux, FreeBSD, OpenBSD, and Android in a consistent order. This makes it easier to find such code in the repo. I've also found a couple of spots where OpenBSD was missing and have fixed them.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/WaitFor.swift
+++ b/Sources/Testing/ExitTests/WaitFor.swift
@@ -85,11 +85,7 @@ private let _childProcessContinuations = LockedWith<pthread_mutex_t, [pid_t: Che
 /// A condition variable used to suspend the waiter thread created by
 /// `_createWaitThread()` when there are no child processes to await.
 private nonisolated(unsafe) let _waitThreadNoChildrenCondition = {
-#if os(FreeBSD) || os(OpenBSD)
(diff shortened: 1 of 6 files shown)


Sources/Testing/ExitTests/WaitFor.swift
```diff
@@ -85,11 +85,7 @@ private let _childProcessContinuations = LockedWith<pthread_mutex_t, [pid_t: Che
 /// A condition variable used to suspend the waiter thread created by
 /// `_createWaitThread()` when there are no child processes to await.
 private nonisolated(unsafe) let _waitThreadNoChildrenCondition = {
-#if os(FreeBSD) || os(OpenBSD)
-  let result = UnsafeMutablePointer<pthread_cond_t?>.allocate(capacity: 1)
-#else
   let result = UnsafeMutablePointer<pthread_cond_t>.allocate(capacity: 1)
```
jerryjrchen, line 88: This is picking up the `typealias pthread_cond_t = _TestingInternals.pthread_cond_t?` on the BSDs right?

grynspan: Yep

PR #1350, opened 2025-10-03
Add a property to `Test.Case` to get the values of its arguments.

grynspan:
This PR adds an experimental property to `Test.Case` to allow callers to get the test case's arguments as a type-erased array. For example:

```swift
if let testCase = Test.Case.current {
  for value in testCase.argumentValues {
    print("\(value): \(type(of: value))")
  }
}
```

We already have an `arguments` property of a type that we do not intend to make API, so the new property is named `argumentValues` pending any SPI renaming.

### Checklist:

### DIFF
--- a/Sources/Testing/Parameterization/Test.Case.swift
+++ b/Sources/Testing/Parameterization/Test.Case.swift
@@ -100,6 +100,46 @@ extension Test {
       }
     }
 
+    /// The arguments passed to this test case, if any.
+    ///
+    /// The values in this array correspond to the arguments to the test
+    /// function of which this instance is a test case.
+    ///
+    /// ### Passing tuples to test functions
+    ///
+    /// If your test function takes a tuple (including a key-value pair from a
+    /// dictionary) as its input and maps it to more than one argument, the
+    /// value of this property represents each member of the tuple as a separate
+    /// argument. For example, given the following test function:
+    ///
+    /// ```swift
+    /// @Test(arguments: [(Food.burger, Temperature.hot), (.iceCream, .cold),
+    ///   (.burrito, .hot), (.popsicle, .cold)])
+    /// func `Serving temperature for food`(food: Food, temp: Temperature) {
+    ///   // ...
+    /// }
+    /// ```
+    ///
+    /// The value of this property will be an array with two elements. The first
+    /// element in the array will be an instance of `Food` and the second
+    /// element will be an instance of `Temperature`.
+    ///
+    /// ### Test functions with no arguments
+    ///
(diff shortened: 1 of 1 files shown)


Sources/Testing/Parameterization/Test.Case.swift
```diff
@@ -100,6 +100,46 @@ extension Test {
...
+    /// element will be an instance of `Temperature`.
+    ///
+    /// ### Test functions with no arguments
+    ///
+    /// If your test function is not parameterized, the testing library assigns
+    /// it a single test case and the value of this property for that test case
+    /// is the empty array.
+    @_spi(Experimental)
+    @_unavailableInEmbedded
+    public var argumentValues: [any Sendable] {
```
jerryjrchen, line 134: What's the motivation for this being non-nil for the `.nonParameterized` case? This came to mind when comparing to the existing property `arguments`:

```
public var arguments: [Argument]?
```

jerryjrchen (approved):

Sources/Testing/Parameterization/Test.Case.swift
grynspan: If there is no practical difference between an empty collection and `nil`, best practice is typically to not make a value optional. That means there's one less bit of complexity you need to account for.

stmontgomery, line 134: Can you add a unit test for this new property?

stmontgomery:
I know this is just an SPI at the moment but assuming the intention is to advance it to a public API, I'm interested in discussing the merits of exposing just this one property vs. exposing the backing `Test.Case.Argument` type and an array of argument instances. Having a richer type to represent an argument could have some advantages worth weighing.

Of course, nothing prevents us from pursuing both; they could complement each other. Said a different way, I'm not certain yet whether I agree with this statement from the PR description (**emphasis** mine):

> We already have an `arguments` property of a type **that we do not intend to make API**, so the new property is named `argumentValues` pending any SPI renaming.

I think we should consider both ideas, and that could take place during an API pitch thread.

grynspan:
> I think we should consider both ideas, and that could take place during an API pitch thread.

The comment is more to the point that I couldn't go and name the new property `arguments` because it'd break against the existing property, and any renaming would need discussion. So I think you're just agreeing with me. :)

grynspan:
In any event I'm also fine leaving this specific PR on the table. I think there's value in allowing these values to be accessible but I don't feel too strongly about the precise shape of the API we would add.

Fortunately the values are all constrained to `Sendable & Copyable` so we can do this without hurting ourselves too much.

PR #1351, opened 2025-10-03
Adopt `Mutex`.

### LINKED ISSUE #538: Adopt `Mutex`
Swift Testing should adopt `Mutex` from [SE-0433](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0433-mutex.md).

We won't be able to adopt on Darwin due to back-deployment requirements, but we can adopt for Linux/Windows. Linux needs one `pthread_mutex_t` for the condition lock used in exit tests, but it can be hand-rolled.

grynspan:
This PR adopts `Mutex` on all platforms except Darwin (where we still need to back-deploy further than `Mutex` is available.)

Resolves #538.

### Checklist:


Sources/Testing/Support/Locked.swift
```diff
@@ -52,30 +22,48 @@ protocol Lockable {
...
   init(rawValue: T) {
-    _storage = _Storage.create(minimumCapacity: 1, makingHeaderWith: { _ in rawValue })
+#if SWT_TARGET_OS_APPLE && canImport(os)
+    _storage = .create(minimumCapacity: 1, makingHeaderWith: { _ in rawValue })
     _storage.withUnsafeMutablePointerToElements { lock in
-      L.initializeLock(at: lock)
+      lock.initialize(to: .init())
     }
+#else
+    nonisolated(unsafe) let rawValue = rawValue
```
grynspan, line 53: It's necessary for us to write `nonisolated(unsafe)` here and a few other places because `T` isn't universally `Sendable`. We have some types (pointers) that we need to guard with locks which are not sendable, and the compiler is concerned there may be data races. We know what we're doing (famous last words) but the compiler can't reason about it.

Sources/Testing/Support/Locked.swift
```diff
@@ -52,30 +22,48 @@ protocol Lockable {
...
-  /// A type providing heap-allocated storage for an instance of ``Locked``.
-  private final class _Storage: ManagedBuffer<T, L> {
-    deinit {
-      withUnsafeMutablePointerToElements { lock in
-        L.deinitializeLock(at: lock)
-      }
+struct Locked<T> {
+  /// A type providing storage for the underlying lock and wrapped value.
+#if SWT_TARGET_OS_APPLE && canImport(os)
+  private typealias _Storage = ManagedBuffer<T, os_unfair_lock_s>
```
jerryjrchen, line 28: > This PR adopts Mutex on all platforms except Darwin (where we still need to back-deploy further than Mutex is available.)


Would we be able to unify the Darwin implementation to use `Mutex` in the future? Might be worth a comment here if so

Sources/Testing/Support/Locked.swift
```diff
@@ -52,30 +22,48 @@ protocol Lockable {
...
+#else
+  private final class _Storage {
```
jerryjrchen, line 30: Are we using the wrapper `_Storage` type here to ensure we share a single reference to the lock?

Sources/Testing/Support/Locked.swift
jerryjrchen, line 53: This seems to me like it would be unsafe if `T` isn't `Sendable`. Are we working around a limitation of `RawRepresentable` conformance where we can't restrict to only `T: Sendable`?

Sources/Testing/Support/Locked.swift
```diff
@@ -52,30 +22,48 @@ protocol Lockable {
...
+#else
+    nonisolated(unsafe) let rawValue = rawValue
+    _storage = _Storage(rawValue)
+#endif
   }
 
   var rawValue: T {
-    withLock { $0 }
+    withLock { rawValue in
+      nonisolated(unsafe) let rawValue = rawValue
```
jerryjrchen, line 60: Similar q here, why is this safe to do with a potentially non-Sendable value?

Sources/Testing/ExitTests/WaitFor.swift
```diff
@@ -80,7 +80,42 @@ func wait(for pid: consuming pid_t) async throws -> ExitStatus {
 }
 #elseif SWT_TARGET_OS_APPLE || os(Linux) || os(FreeBSD) || os(OpenBSD)
 /// A mapping of awaited child PIDs to their corresponding Swift continuations.
-private let _childProcessContinuations = LockedWith<pthread_mutex_t, [pid_t: CheckedContinuation<ExitStatus, any Error>]>()
+private nonisolated(unsafe) let _childProcessContinuations = {
```
jerryjrchen, line 83: Hmm, isn't this explicitly not adopting `Mutex` on the non-Darwin platforms `os(Linux) || os(FreeBSD) || os(OpenBSD)`?

Sources/Testing/Support/Locked.swift
grynspan: "We know what we're doing." :)

Sources/Testing/Support/Locked.swift
grynspan: It's not safe in the general case, but we know what we're doing. See https://github.com/swiftlang/swift-testing/pull/1351#discussion_r2402889210

Sources/Testing/ExitTests/WaitFor.swift
grynspan: This _specific_ case needs to be a `pthread_mutex_t` so we can use it with a `pthread_cond_t`. All other uses of `Locked` can be done with `Mutex`.

Sources/Testing/Support/Locked.swift
grynspan: In a hypothetical future where our minimum deployment target on Darwin platforms includes `Mutex`, yes.

Sources/Testing/Support/Locked.swift
grynspan: `Mutex` is a move-only type meaning it needs to be owned by a reference-counted object to be used here. (We could theoretically make `Locked` a class, but it complicates things by forcing us to have `convenience` on some initializers, but not on Darwin.)

Tests/TestingTests/Support/LockTests.swift
```diff
@@ -13,50 +13,13 @@ private import _TestingInternals
 
 @Suite("Locked Tests")
 struct LockTests {
```
stmontgomery, line 15: This test suite is looking pretty minimal now. Could we include a test that involves concurrency — e.g. use a task group to repeatedly increment a `Locked` counter and ensure it's correct at the end?

grynspan: We certainly can in the technical sense, but we'd just be testing `Mutex` and/or `os_unfair_lock` which already have plenty of test coverage. Do you feel strongly about it?

grynspan: Eh, why not. Added!

stmontgomery: I know the heavy-lifting is on the underlying lock type, but we still have some involvement — for example with os_unfair_lock we still need to safely obtain the lock instance and lock/unlock it. So it seems like we still have some logic in our layer which could theoretically have bugs.

stmontgomery (approved):
Thanks!! This is a huge simplification

PR #1352, opened 2025-10-04
Get exit tests functioning on OpenBSD.

grynspan:
This PR adjusts the implementation of exit tests on OpenBSD so that they correctly spawn child processes there.

Without this change, exit tests all abnormally terminate with exit code 127. `ktrace` and `kdump` report:

```
63170 swift-testingPackageTes CALL  sigaction(SIGHUP,0x932dd7e5ca0,0)
63170 swift-testingPackageTes STRU  struct sigaction { handler=SIG_DFL, mask=0<>, flags=0<> }
63170 swift-testingPackageTes RET   sigaction 0
63170 swift-testingPackageTes CALL  sigaction(SIGINT,0x932dd7e5ca0,0)
63170 swift-testingPackageTes STRU  struct sigaction { handler=SIG_DFL, mask=0<>, flags=0<> }
63170 swift-testingPackageTes RET   sigaction 0
[...]
63170 swift-testingPackageTes CALL  sigaction(SIGKILL,0x932dd7e5ca0,0)
63170 swift-testingPackageTes RET   sigaction -1 errno 22 Invalid argument
63170 swift-testingPackageTes CALL  exit(127)
```

OpenBSD is more strict in its handling of `posix_spawnattr_setsigdefault()` than our other target platforms. Specifically, the child process will self-terminate after forking but before execing if either `SIGKILL` or `SIGSTOP` is in the signal mask passed to that function. Other platforms silently ignore these signals (since their handlers cannot be set per POSIX anyway.)

With this change, exit tests function correctly and Swift Testing's test suite passes with zero issues on OpenBSD. 🥳

openbsd

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/SpawnProcess.swift
+++ b/Sources/Testing/ExitTests/SpawnProcess.swift
@@ -113,6 +113,13 @@ func spawnExecutable(
       withUnsafeTemporaryAllocation(of: sigset_t.self, capacity: 1) { allSignals in
         let allSignals = allSignals.baseAddress!
         sigfillset(allSignals)
+#if os(OpenBSD)
+        // On OpenBSD, attempting to set the signal handler for SIGKILL or
+        // SIGSTOP will cause the child process of a call to posix_spawn() to
+        // exit abnormally with exit code 127. See https://man.openbsd.org/sigaction.2#ERRORS
+        sigdelset(allSignals, SIGKILL)
+        sigdelset(allSignals, SIGSTOP)
+#endif
         posix_spawnattr_setsigdefault(attrs, allSignals);
         flags |= CShort(POSIX_SPAWN_SETSIGDEF)
       }


grynspan:
@3405691582 please review!

3405691582 (approved):

grynspan:
For future reference, here are a few other projects on GitHub that spawn processes the way we do and need to do something like this:

https://github.com/mys-lang/mys/blob/f3cb604e95eadd9833466dbd43ac9f5913f31d60/mys/uv/src/unix/process.c#L801
https://github.com/notinaboat/UnixIO.jl/blob/169676f22654e0d41224452181cd12237f9836c2/src/UnixIO.jl#L1294
https://github.com/tcltk/tcl/blob/7f29ced6d2e866cd66483b9c6b725ff281960e12/unix/tclUnixPipe.c#L526

grynspan:
Oh and hey our very own swift-llbuild: https://github.com/swiftlang/swift-llbuild/blob/9cd06072bc98869d6058ecf0cbb62533ca10c02f/src/llbuild3/LocalExecutor.cpp#L1084

PR #1356, opened 2025-10-05
Dynamically look up `backtrace(3)` on Android.

### LINKED ISSUE #1135: Failing to build for Android in SPI
Looks like Swift Testing is failing to build for Android on Swift Package Index's site: https://swiftpackageindex.com/builds/875FEA40-0F76-47AF-A7E3-54A39616BA9C

```
/host/spi-builder-workspace/Sources/Testing/SourceAttribution/Backtrace.swift:70:56: error: cannot convert value of type 'UnsafeMutablePointer<UnsafeMutableRawPointer>' to expected argument type 'UnsafeMutablePointer<UnsafeMutableRawPointer?>'
 68 | #elseif os(Android)
 69 |       initializedCount = addresses.withMemoryRebound(to: UnsafeMutableRawPointer.self) { addresses in
 70 |         .init(clamping: backtrace(addresses.baseAddress!, .init(clamping: addresses.count)))
    |                                                        |- error: cannot convert value of type 'UnsafeMutablePointer<UnsafeMutableRawPointer>' to expected argument type 'UnsafeMutablePointer<UnsafeMutableRawPointer?>'
    |                                                        `- note: arguments to generic parameter 'Pointee' ('UnsafeMutableRawPointer' and 'UnsafeMutableRawPointer?') are expected to be equal
 71 |       }
 72 | #elseif os(Linux) || os(FreeBSD)
```

We have special-casing to convert the pointers here to remove optionality because, at one point, the signature of `backtrace()` was annotated such that the pointers were non-optional. This has either changed or is inconsistent across different Android versions.

grynspan:
The `backtrace()` function on Android was added with API level 33, but at least some external Android builds target earlier Android NDKs, so we'll dynamically look up the function. If it's not present or dynamic loading isn't available, we just produce an empty backtrace.

See: https://cs.android.com/android/_/android/platform/bionic/+/731631f300090436d7f5df80d50b6275c8c60a93:libc/include/execinfo.h;l=52

Resolves #1135.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/Backtrace.swift
+++ b/Sources/Testing/SourceAttribution/Backtrace.swift
@@ -40,6 +40,16 @@ public struct Backtrace: Sendable {
     self.addresses = addresses.map { Address(UInt(bitPattern: $0)) }
   }
 
+#if os(Android) && !SWT_NO_DYNAMIC_LINKING
+  /// The `backtrace()` function.
+  ///
+  /// This function was added to Android with API level 33, which is higher than
+  /// our minimum deployment target, so we look it up dynamically at runtime.
+  private static let _backtrace = symbol(named: "backtrace").map {
+    castCFunction(at: $0, to: (@convention(c) (UnsafeMutablePointer<UnsafeMutableRawPointer?>, CInt) -> CInt).self)
+  }
+#endif
+
   /// Get the current backtrace.
   ///
   /// - Parameters:
@@ -66,9 +76,9 @@ public struct Backtrace: Sendable {
         initializedCount = .init(clamping: backtrace(addresses.baseAddress!, .init(clamping: addresses.count)))
       }
 #elseif os(Android)
-      initializedCount = addresses.withMemoryRebound(to: UnsafeMutableRawPointer.self) { addresses in
-        .init(clamping: backtrace(addresses.baseAddress!, .init(clamping: addresses.count)))
-      }
+#if !SWT_NO_DYNAMIC_LINKING
+      initializedCount = .init(clamping: _backtrace?(addresses.baseAddress!, .init(clamping: addresses.count)))
+#endif
 #elseif os(Linux) || os(FreeBSD) || os(OpenBSD)
       initializedCount = .init(clamping: backtrace(addresses.baseAddress!, .init(clamping: addresses.count)))
 #elseif os(Windows)


finagolfin:
Should we just wait for `#availability` checking to be added for Android, swiftlang/swift#84574? Seems like Mads is close to getting that working; we simply disable this backtrace code in our Android SDK bundles for now.

grynspan:
No, because the signature is not consistent (optionality has changed) between the different toolchains floating around out there. So even with `#available`, I wouldn't be able to type the function call correctly.

finagolfin:
The signature issue is irrelevant: the only one that matters is the one currently in Bionic. The other signature comes from my Android SDK bundle, which uses a Termux library that was backported to API 24, but I'm sunsetting my bundle in favor of the official SDK bundle that we're working on, which will simply delegate to the Bionic headers in the NDK.

grynspan:
You're the domain expert, so if you think we should wait, I defer to you. However this change will also work even without `#available()` (in case that doesn't land) and should work with older compilers.

finagolfin:
I'm not too worried about backwards compatibility, particularly considering I only [got this library building for Android earlier this year](https://github.com/finagolfin/swift-android-sdk/releases/tag/6.1).

I'm certainly no expert on `dlsym()` or `#available`, neither of which I've ever used, but I thought that if we are on the verge of a Swift-native solution, better to use that: what do you think?

Thanks for working on this, but that was my only concern.

grynspan:
I think it really depends on how long until `#available` is functional on Android. If it's going to be a long time, this solution is ready now and will be fully functional. If it's ready to ship next week, then yeah we should wait.

finagolfin:
Mads is working on it and while he hit some stumbling blocks in that pull, hopefully we can add it soon, let's see.

grynspan:
I'll leave it to @finagolfin to decide.

finagolfin:
I'd rather avoid these old C methods if `#available` is around the bend, let's see what Mads comes up with this month.

grynspan:
As we don't have a clear timeline for `#available` support on Android, I'm going to go ahead and merge this PR. As it's self-contained, we can revert it easily enough in the future.

finagolfin:
`#available` is working well on Android in our testing, 😄 running more tests now, including switching this feature over to using it. His frontend pull should be ready for review tomorrow, but who knows how long it'll take to then get in.

As you say, this is trivial to update once his feature is available, so we can always do that later. 👍

grynspan:
Either way, a gift for you: https://github.com/swiftlang/swift/issues/84927.

PR #1357, opened 2025-10-05
Support command-line arguments specified as `--foo=bar`.

grynspan:
SwiftPM and `swift test` use Swift Argument Parser which allows developers to specify arguments of the form `--foo=bar`. Our bare-bones argument parser doesn't currently recognize that pattern, which means that the developer could write `--foo=bar` but get the wrong behavior.

This PR adds support for that pattern by changing how we parse command-line arguments to allow for both `--foo bar` and `--foo=bar`.

### Checklist:

### DIFF
--- a/Tests/TestingTests/SwiftPMTests.swift
+++ b/Tests/TestingTests/SwiftPMTests.swift
@@ -492,4 +492,25 @@ struct SwiftPMTests {
     let args = try parseCommandLineArguments(from: ["PATH", "--verbosity", "12345"])
     #expect(args.verbosity == 12345)
   }
+
+  @Test("--foo=bar form")
+  func equalsSignForm() throws {
+    // We can split the string and parse the result correctly.
+    do {
+      let args = try parseCommandLineArguments(from: ["PATH", "--verbosity=12345"])
+      #expect(args.verbosity == 12345)
+    }
+
+    // We don't overrun the string and correctly handle empty values.
+    do {
+      let args = try parseCommandLineArguments(from: ["PATH", "--xunit-output="])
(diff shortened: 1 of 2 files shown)


Tests/TestingTests/SwiftPMTests.swift
```diff
@@ -90,6 +90,7 @@ struct SwiftPMTests {
   @available(_regexAPI, *)
   func filter() async throws {
     let configuration = try configurationForEntryPoint(withArguments: ["PATH", "--filter", "hello"])
+    print(configuration)
```
plemarquand: errant `print`?

grynspan: Yep, thanks.

Tests/TestingTests/SwiftPMTests.swift
```diff
@@ -492,4 +500,25 @@ struct SwiftPMTests {
...
+      #expect(args.verbosity == 12345)
+    }
+
+    // We don't overrun the string and correctly handle empty values.
+    do {
+      let args = try parseCommandLineArguments(from: ["PATH", "--xunit-output="])
+      #expect(args.xunitOutput == "")
+    }
+
+    // We split at the first equals-sign.
```
stmontgomery, line 517: If you moved these comments down to the line above each `#expect`, they'd be expectation comments in the results! 😅

Makes me wonder if we ought to look for code comments above `do { ... }` blocks in lexical scope (probably stopping at the nearest-enclosing function scope)

grynspan: We can't, we don't have that lexical context.

PR #1362, opened 2025-10-08
Add support for experimental "custom availability domains" feature.

grynspan:
This PR updates our parsing of the `@available` attribute during expansion of the `@Test` and `@Suite` macros to support custom availability domains of the form:

```swift
@available(Foo) @Test func f() {}
```

This change means the macro expansion will include `if #available(Foo)` instead of `if #available(Foo, *)` in this case.

Resolves rdar://160323829.

### Checklist:

### DIFF
--- a/Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
+++ b/Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
@@ -415,7 +415,12 @@ struct TestDeclarationMacroTests {
         [
           #"#if os(moofOS)"#,
           #".__available("moofOS", obsoleted: nil, message: "Moof!", "#,
-        ]
+        ],
+      #"@available(customAvailabilityDomain) @Test func f() {}"#:
+        [
+          #".__available("customAvailabilityDomain", introduced: nil, "#,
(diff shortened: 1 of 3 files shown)


Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
```diff
@@ -415,7 +415,12 @@ struct TestDeclarationMacroTests {
         [
           #"#if os(moofOS)"#,
           #".__available("moofOS", obsoleted: nil, message: "Moof!", "#,
-        ]
+        ],
+      #"@available(customAvailabilityDomain) @Test func f() {}"#:
+        [
+          #".__available("customAvailabilityDomain", introduced: nil, "#,
+          #"guard #available (customAvailabilityDomain) else"#,
```
stmontgomery, line 422: Is the whitespace intentional?
```suggestion
          #"guard #available(customAvailabilityDomain) else"#,
```

grynspan: Yes, see previous cases.

grynspan:
Ignoring failed Windows CI as it's a known issue with the toolchain.

PR #1363, opened 2025-10-08
Adopt `Mutex` (take 2)

### LINKED ISSUE #538: Adopt `Mutex`

grynspan:
This PR adopts `Mutex` on all platforms except Darwin (where we still need to back-deploy further than `Mutex` is available.)

Blocked by https://github.com/swiftlang/swift/pull/84771.
Reapplies #1351.
Resolves #538.
Resolves rdar://131832797.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Locked.swift
+++ b/Sources/Testing/Support/Locked.swift
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-private import _TestingInternals
+internal import _TestingInternals
 private import Synchronization
 
 /// A type that wraps a value requiring access from a synchronous caller during


grynspan:
Should be identical.

grynspan:
Note to self: BSD commit?

jerryjrchen:
I can't see the CI results in #1353 (https://ci.swift.org/job/oss-swift-pr-test-ubuntu-22_04/9662/). Was the reason this was backed out because of a compilation failure due to https://github.com/swiftlang/swift/pull/84771?

grynspan:
That PR fixes the issue we ran into which our own CI jobs won't catch. See that issue for more info.

grynspan:
The bug fixed in https://github.com/swiftlang/swift/pull/84771 may still be present in the 6.2.x toolchain, but we don't expect to build this repo's main branch in release mode there, so I'm going to merge the change now. If we find this causes problems with the 6.2 toolchain, we can always revert again.

PR #1367, opened 2025-10-16
Make `Attachment` always conform to `CustomStringConvertible`.

grynspan:
This PR ensures that `Attachment` conforms to `CustomStringConvertible` even if the value it wraps is non-copyable. We accomplish this by creating a new helper function that is able to box an arbitrary `~Copyable` value in `Any` if its type is actually `Copyable`. (This requires a newer runtime on Apple platforms, so we fall back to a generic (`~Copyable`) implementation on older macOS/iOS/etc.) We then call this helper function in `Attachment.description` and if it returns a value, we can stringify it.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/CopyableAdditions.swift
+++ b/Sources/Testing/Support/Additions/CopyableAdditions.swift
@@ -41,7 +41,7 @@ func boxCopyableValue(_ value: borrowing some ~Copyable) -> Any? {
   }
 }
 #else
-func boxCopyableValue(_ value: borrowing some ~Copyable) -> Never? {
+func boxCopyableValue(_ value: borrowing some ~Copyable) -> Void? {
   nil
 }
 #endif


cmake/modules/shared/AvailabilityDefinitions.cmake
```diff
@@ -17,4 +17,5 @@ add_compile_options(
   "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -define-availability -Xfrontend \"_regexAPI:macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0\">"
   "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -define-availability -Xfrontend \"_swiftVersionAPI:macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0\">"
   "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -define-availability -Xfrontend \"_typedThrowsAPI:macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0\">"
+  "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -define-availability -Xfrontend \"_castingWithNonCopyableGenerics:macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0\">"
```
jerryjrchen, line 20: Do you know where I could find more docs on `_castingWithNonCopyableGenerics`? Search isn't turning up anything 😅

Sources/Testing/Attachments/Attachment.swift
```diff
@@ -193,21 +193,17 @@ public struct AnyAttachable: AttachableWrapper, Sendable, Copyable {
...
-  }
-}
-
-extension Attachment: CustomStringConvertible {
+extension Attachment: CustomStringConvertible where AttachableValue: ~Copyable {
   /// @Metadata {
   ///   @Available(Swift, introduced: 6.2)
   ///   @Available(Xcode, introduced: 26.0)
   /// }
   public var description: String {
```
jerryjrchen, line 201: I see this lets us implement a single `description` that kinda handles both `Copyable` and `~Copyable` types. Does this result in us being able to provide `String(describingForTest:)` on any types we couldn't before?

Sources/Testing/Attachments/Attachment.swift
```diff
@@ -193,21 +193,17 @@ public struct AnyAttachable: AttachableWrapper, Sendable, Copyable {
...
 
-extension Attachment where AttachableValue: ~Copyable {
-  @_documentation(visibility: private)
-  public var description: String {
-    let typeInfo = TypeInfo(describing: AttachableValue.self)
-    return #""\#(preferredName)": instance of '\#(typeInfo.unqualifiedName)'"#
-  }
-}
-
-extension Attachment: CustomStringConvertible {
```
jerryjrchen, line 204: Based on the PR description:

> (This requires a newer runtime on Apple platforms, so we fall back to a generic (~Copyable) implementation on older macOS/iOS/etc.)

Doesn't this regress the ability to provide a description based on `String(describingForTest:)` on older runtimes without `_castingWithNonCopyableGenerics` available? We'd always hit the fallback case using `TypeInfo`.

Sources/Testing/Support/Additions/CopyableAdditions.swift
```diff
@@ -0,0 +1,47 @@
...
+///
+/// - Returns: A copy of `value` in an existential box, or `nil` if the type of
+///   `value` does not conform to `Copyable`.
+///
+/// When using Embedded Swift, this function always returns `nil`.
+#if !hasFeature(Embedded)
+@available(_castingWithNonCopyableGenerics, *)
+func boxCopyableValue(_ value: borrowing some ~Copyable) -> Any? {
+  withUnsafePointer(to: value) { address in
+    return (address as? any _CopyablePointer)?.load()
```
jerryjrchen, line 40: Oh, this is super neat if I'm understanding correctly:
- `address as? any _CopyablePointer` casts to a `_CopyablePointer` which gives access for loading the pointee
- The only conforming type to this protocol is `UnsafePointer` where its `Pointee: Copyable`
- Therefore this cast and subsequent load only succeeds if the pointer has a `Copyable` pointee

Did I get that right?

jerryjrchen (approved):
LGTM. Are there already existing test cases that test the Attachment description behaviours? If not, would it be feasible to include as part of this PR?

cmake/modules/shared/AvailabilityDefinitions.cmake
grynspan: I invented it. :) We can discuss the mechanism I'm using here privately.

grynspan:
We already have test coverage, yes.

Sources/Testing/Support/Additions/CopyableAdditions.swift
grynspan: Correct! That `UnsafePointer` is the only conforming type isn't salient, but we only need the one type to make the function work so there's no sense in making other types conform.

Sources/Testing/Attachments/Attachment.swift
grynspan: Yes, but the tradeoff is we get unconditional conformance which is better when printing/recording an `Attachment` in the general case.

Sources/Testing/Attachments/Attachment.swift
grynspan: No, `String(describingForTest:)` can't handle non-copyable types because they still can't be conditionally cast to an existential, so there's no way to take a non-copyable value and conditionally turn it into `some CustomTestStringConvertible` while also falling back to a generic string in cases where the type does not conform.

grynspan:
Holding for @stmontgomery to look at.

stmontgomery (approved):

PR #1373, opened 2025-10-21
Use the new `#available(Android API, *)` instead to look for `backtrace()`

finagolfin:
Mads just added this compiler feature for Android in swiftlang/swift#84574

I tested this locally on an Android API 35 device in the Termux app, no problem, before transferring the Testing test runner to an API 28 device. It would not link there, since it was an executable and `libc.so` at API 28 didn't have `backtrace()`, so I couldn't run the tests that way.

I then ran `patchelf --add-needed libandroid-execinfo.so swift-testingPackageTests.xctest` to have the test runner use the backported libexecinfo from the Termux app, which got all the tests to run. Three backtrace tests correctly failed because the runtime `#available` checking disabled this call, showing that runtime version checking worked. 😄

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -25,6 +25,7 @@ jobs:
       ios_host_exclude_xcode_versions: '[{"xcode_version": "16.2"}, {"xcode_version": "16.3"}, {"xcode_version": "16.4"}]'
       enable_wasm_sdk_build: true
       enable_android_sdk_build: true
+      enable_android_sdk_checks: true
       android_ndk_versions: '["r28c"]'
       android_sdk_triples: '["aarch64-unknown-linux-android33", "x86_64-unknown-linux-android28"]'
   soundness:


grynspan:
We need to continue to build with the Swift 6.2 toolchain on other platforms. Do we need to preserve that functionality for Android given there is no _official_ Android 6.2 toolchain? (Also pinging @compnerd.)

If so, the right way forward is to hold off on merging this PR until Swift 6.3 is released, at which point we'll drop 6.2 support and can proceed.

Sources/Testing/SourceAttribution/Backtrace.swift
```diff
@@ -77,8 +67,10 @@ public struct Backtrace: Sendable {
       }
 #elseif os(Android)
 #if !SWT_NO_DYNAMIC_LINKING
```
grynspan: You can get rid of this `#if`.

grynspan:
> It would not link there, since it was an executable and libc.so at API 28 didn't have backtrace(), so I couldn't run the tests that way.

Does this mean that compiling against API 33+ and then _running_ on an older device will fail? If so, that's not what we want: it indicates that weak linking isn't working, which means you can't use `if #available(Android)` to conditionally enable new NDK API use.

Sources/Testing/SourceAttribution/Backtrace.swift
finagolfin: I thought so too initially, but the alternative is Android with static linking, in which case it will hit the final `#else` clause below and error.

grynspan: That's not what we use `SWT_NO_DYNAMIC_LINKING` for. We use it to indicate if `dlsym()` or equivalent is available. (I named it bad.) So you can just remove it. All Android codepaths will go through here. Even statically-linked Android binaries should be able to use weakly-linked NDK symbols.

finagolfin:
> Do we need to preserve that functionality for Android given there is no official Android 6.2 toolchain?

Not sure what you mean, is that because trunk in this repo is currently auto-merged to release/6.2? If so, we could also check `#if compiler(>=6.3)` to make sure this is not parsed by the 6.2 toolchain.

> Does this mean that compiling against API 33+ and then running on an older device will fail? If so, that's not what we want: it indicates that weak linking isn't working, which means you can't use if #available(Android) to conditionally enable new NDK API use.

Yep, _if_ it's an executable, which seem to be required on Android to resolve all symbols at startup. Since the SwiftPM-produced test runner executable for this repo, that I built against a target, ie minimum, of API 24, doesn't invoke these methods from the toolchain's `libTesting.so`, but keeps them in the test runner and invokes them from there, I believe that's why it failed.

I don't think it will be an issue in Android apks, which almost always ship with native code in shared libraries instead, but haven't looked into it in detail.

stmontgomery:
> If so, we could also check `#if compiler(>=6.3)` to make sure this is not parsed by the 6.2 toolchain.

That was going to be my suggestion as well.

grynspan:
> Not sure what you mean, is that because trunk in this repo is currently auto-merged to release/6.2? If so, we could also check `#if compiler(>=6.3)` to make sure this is not parsed by the 6.2 toolchain.

It is possible and supported to build Swift Testing as a package using the current released toolchain (6.2). So we need to keep that working.

> Yep, _if_ it's an executable, which seem to be required on Android to resolve all symbols at startup. Since the SwiftPM-produced test runner executable for this repo, that I built against a target, ie minimum, of API 24, doesn't invoke these methods from the toolchain's `libTesting.so`, but keeps them in the test runner and invokes them from there, I believe that's why it failed.

I would expect weak linking to NDK symbols to work even for executable targets. If it doesn't, then we can't use `#available` here and must continue using dynamic lookup indefinitely.

Sources/Testing/SourceAttribution/Backtrace.swift
finagolfin: I see, I thought you meant remove just the `#if`, meaning adding an `&&`. I kept the check based on your adding it before and misreading [this doc](https://github.com/swiftlang/swift-testing/blob/main/Documentation/Porting.md#runtime-test-discovery-with-static-linkage), will remove it.

Unsure how this will interact with static linking executables, given the dynamic linking issue I mentioned, but considering most will static link this into shared libraries eventually, may not matter.

finagolfin:
> It is possible and supported to build Swift Testing as a package using the current released toolchain (6.2). So we need to keep that working.

Ah, I see, changing the dynamic linking check to `#if compiler(>=6.3)` instead okay then?

> I would expect weak linking to NDK symbols to work even for executable targets. If it doesn't, then we can't use #available here and must continue using dynamic lookup indefinitely.

I don't know whether it should work differently for executables or not, was just presenting my hypothesis, as most won't use executables for testing their apps anyway. However, I'll look into it with Mads and let you know.

grynspan:
> I don't know whether it should work differently for executables or not, was just presenting my hypothesis, as most won't use executables for testing their apps anyway. However, I'll look into it with Mads and let you know.

Let's see if we can solve the weak-linking mystery and then revisit. If it turns out it's fine, or it's some narrow edge case we really don't need to care about, great. If there's a bug in how `#available` functions on Android, I'd prefer we resolve that bug before proceeding. Thanks!

grynspan:
When you tried to run on API 28, had you set a minimum deployment target at compile time? Do we have a mechanism for doing so when building for Android? If it built against API 35 and you didn't set a minimum deployment target, I'd expect it to not bother weak-linking (same as on Darwin.)

finagolfin:
Yep, target, ie minimum, was API 24, [mentioned it above](https://github.com/swiftlang/swift-testing/pull/1373#issuecomment-3426513988). Mads and I are looking into it, could be something specific to the toolchain I natively built in the Termux app on Android, will let you know in a day or two.

grynspan:
D'oh, misread. Okay.

finagolfin:
It was ninja edited in, so maybe you read the first version. 😉

Sources/Testing/SourceAttribution/Backtrace.swift
compnerd: @grynspan's suggestion makes sense - testing requires dynamic linking (i.e. you cannot statically link the testing framework). We know that the platform supports dynamic symbol resolution and because this is within the testing framework, you will be dynamically linking. As a result, just the new handling for the lookup is needed.

grynspan: You can in fact statically link Swift Testing!

finagolfin: In any case, we agreed on removing this check, which he had added, not me.

grynspan: Yarp.

finagolfin:
We looked into the weak linking issue: it only shows up in the natively-built toolchain in the Termux app, _not_ when cross-compiling. I'll look into the Termux issue later, but no reason to keep this waiting on that.

Rebased and made the changes discussed, ready to go ahead.

grynspan:
Alright. We can merge once CI is passing.

finagolfin:
Been discussing this pull with @marcprux and @madsodgaard: turns out this pull won't build with the next trunk Android SDK snapshot tag either, because this new `#available(Android <API>, *)` feature requires Android NDK 28 or later, but we currently build the Swift SDK for Android and use it with LTS NDK 27 only.

I just [passed in NDK 28 on github to the official Docker scripts that build the SDK snapshots](https://github.com/swift-android-sdk/swift-docker/actions/runs/19188601646), and other than four compiler validation suite tests failing, everything worked, including locally cross-compiling this pull.

@compnerd, what do you think about switching the trunk 6.3 SDK snapshots over to NDK 28 now, both in Docker and in the Windows toolchain? We can still keep the current 6.2 release branch and [the official 6.2 CI for Android](https://ci.swift.org/job/oss-swift-6.2-package-swift-sdk-for-android/) on LTS NDK 27, thus only offering LTS NDK support for actual releases.

This will allow us to start preparing for the final 6.3 release in 4-5 months, so it will then use what will likely be the next LTS NDK 30 around that time.

compnerd:
There's certainly a strong reason to do that - the availability would simplify the Android build maintenance. I'm not totally opposed to it, but the LTS NDK is certainly preferable. We should verify that there's no other fallout from the migration before we commit to it I think.

finagolfin:
I tried to change the NDK version used here, which appears to have disabled running the tests again. It won't work till the next trunk tag of the Swift SDK for Android anyway.

Sources/Testing/SourceAttribution/Backtrace.swift
```diff
@@ -40,7 +40,7 @@ public struct Backtrace: Sendable {
     self.addresses = addresses.map { Address(UInt(bitPattern: $0)) }
   }
 
-#if os(Android) && !SWT_NO_DYNAMIC_LINKING
+#if compiler(<6.3) && os(Android)
```
aluco100, line 43: why have you changed the order of conditionals here?

Sources/Testing/SourceAttribution/Backtrace.swift
```diff
@@ -76,7 +76,13 @@ public struct Backtrace: Sendable {
         initializedCount = .init(clamping: backtrace(addresses.baseAddress!, .init(clamping: addresses.count)))
       }
 #elseif os(Android)
-#if !SWT_NO_DYNAMIC_LINKING
+#if compiler(>=6.3)
+      if #available(Android 33, *) {
+        initializedCount = addresses.withMemoryRebound(to: UnsafeMutableRawPointer.self) { addresses in
+          .init(clamping: backtrace(addresses.baseAddress!, .init(clamping: addresses.count)))
```
aluco100, line 82: maybe an if let for baseAddress here?

Sources/Testing/SourceAttribution/Backtrace.swift
finagolfin: No particular reason, just the perception that such compiler versioning is more important.

aluco100: which conditional has more complexity, the compiler version check or the os check?

Sources/Testing/SourceAttribution/Backtrace.swift
finagolfin: I believe this almost always succeeds- unless the allocation up top fails?- so the unwrapping is a mere formality, but I'm just bringing back Jonathan's earlier code here: I didn't write this.

aluco100: I see, for some reason (correct me I'm wrong) the closure gives you a buggy object with nil property references (such as a de-reference issue) maybe you will lead this into a SIGTRAP I think

Sources/Testing/SourceAttribution/Backtrace.swift
finagolfin: No idea what you mean by complexity in this context, I just figured that the compiler check is more generally known by Swift devs.

Sources/Testing/SourceAttribution/Backtrace.swift
finagolfin: In my experience, this always works and is a common idiom in the Swift corelibs.

grynspan: This unwrap cannot fail. Hence the forced unwrap. There is no point in an `if let` because we know the `else` case will never be executed.

`!` is considered acceptable style within the Swift toolchain under these circumstances.

Sources/Testing/SourceAttribution/Backtrace.swift
grynspan: Actually, `compiler()` has special powers in the compiler (heh). If it evaluates to `false`, subsequent conditions are not even parsed. This means that you can write something like `#if compiler(>=6.3) && objectFormat(ELF)` and an older compiler won't complain that it doesn't know what `objectFormat()` is.

It's good to get into the habit of putting `compiler()` first for this reason.

grynspan:
@finagolfin I'm approving now because obviously the code is shaped the way we want. Please don't merge it until we know for certain that the Android build will pass and is correctly emitting the requisite weak symbol reference.

finagolfin:
Yep, as referenced above, this would break both the Android and Windows CI unless we switched both to use NDK 28 or later, so we are in _no_ hurry. 😉

.github/workflows/pull_request.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       ios_host_exclude_xcode_versions: '[{"xcode_version": "16.2"}, {"xcode_version": "16.3"}, {"xcode_version": "16.4"}]'
       enable_wasm_sdk_build: true
       enable_android_sdk_build: true
+      android_ndk_version: '["r28c"]'
```
jakepetroules: This needs to be `android_ndk_versions` (plural) now; I changed it.

jakepetroules: Also, r27d is the LTS NDK release, so maybe we should keep that _and_ include r28?

finagolfin: > r27d is the LTS NDK release, so maybe we should keep that and include r28?

We can't, as this new `#available(Android` feature in trunk and 6.3 _only_ compiles with NDK 28 or later- the Bionic libc changed how it versioned its APIs by API level starting with NDK 28 and this feature requires that change- hence the drive to get all trunk/6.3 CI on NDK 28 or 29 now, in preparation for the likely next LTS NDK 30.

finagolfin:
Tested this locally and it worked fine, now it can be tested with the official Android workflow too, since the first trunk Android SDK snapshot that supports this new feature just shipped.

I think someone will need to trigger the CI though, as some permission system likely doesn't like me modifying the CI config for this repo, and appears to have disabled running the tests on this pull.

Sources/Testing/SourceAttribution/Backtrace.swift
```diff
@@ -76,7 +76,16 @@ public struct Backtrace: Sendable {
...
+        }
+      }
+      else {
+        print("Backtraces are unavailable on Android API 32 and older.")
```
finagolfin: @grynspan, what do you think of this warning?

.github/workflows/pull_request.yml
finagolfin: Heh, turns out I forgot to make this change you mentioned, tests are running again now that I added it. 👍

finagolfin:
Good news, this pull built fine with Android NDK 28 on the GitHub CI, including an additional triple that I added to check building both before and after this `backtrace()` API was included. Of course, the official workflows for Android do not actually run the tests on Android, so I build and run those natively on Android instead, and we cannot get this in till [the official Android CI](https://ci.swift.org/job/oss-swift-package-swift-sdk-for-android/) switch to NDK 28 or later, which is still being discussed.

Sources/Testing/SourceAttribution/Backtrace.swift
grynspan: It's not necessary (and we wouldn't use `print()` for this anyway.) You can leave it out.

grynspan:
> Of course, the official workflows for Android do not actually run the tests on Android, so I build and run those natively on Android instead, and we cannot get this in till [the official Android CI](https://ci.swift.org/job/oss-swift-package-swift-sdk-for-android/) switch to NDK 28 or later, which is still being discussed.

I think we can just sit on this PR until that's resolved. Once it is, we can merge it (I don't think any further noodling is required at this point.)

grynspan:
@finagolfin Bumped your branch. Any news on NDK 28?

finagolfin:
Not yet, will let you know. 👍

finagolfin:
Tried out the new Android emulator testing that was recently added to the official workflow and hit the long-standing SwiftPM cross-compilation bug with macros again, as expected since I've been seeing this for years now, swiftlang/swift-package-manager#8094:
```
error: Failed opening '/home/runner/work/swift-testing/swift-testing/.build/aarch64-unknown-linux-android33/debug/index/store/v5/units/ConditionMacroTests.swift.o-14BKXGGRUDYQH': No such file or directory
```

grynspan:
Should be able to stub out those test files _à la_ Windows.

finagolfin:
Deleted this branch by mistake and it's not allowing me to reopen this, so I have newly reopened it at #1546.

PR #1377, opened 2025-10-23
Adopt `@c`.

grynspan:
This PR adopts the new `@c` attribute from [SE-0495](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0495-cdecl.md). Older compilers will continue to use `@_cdecl`.

### Checklist:

### DIFF
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -18,7 +18,7 @@ on:
 jobs:
   create_merge_pr:
     name: Create PR to merge main into release branch
-    uses: swiftlang/github-workflows/.github/workflows/create_automerge_pr.yml@0.0.3
+    uses: swiftlang/github-workflows/.github/workflows/create_automerge_pr.yml@0.0.4
     with:
       head_branch: main
       base_branch: release/6.3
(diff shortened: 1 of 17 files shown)


Sources/_TestingInterop/FallbackEventHandler.swift
```diff
@@ -55,7 +55,11 @@ package typealias FallbackEventHandler = @Sendable @convention(c) (
 /// Get the current fallback event handler.
 ///
 /// - Returns: The currently-set handler function, if any.
+#if compiler(>=6.3.1)
+@c
```
jerryjrchen, line 68: What about using `@objc` instead? This will only be called from Swift or ObjC, so my understanding is that we could add ObjC types to this interface later if needed. [My reading of the proposal](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0495-cdecl.md#objc-global-functions) is also that `@objc` is a closer match for `@_cdecl` today:

> Note: The attribute @objc can be used on a global function to replace @_cdecl as it preserves the behavior of the unofficial attribute.

grynspan: `@objc` is not available on non-Apple targets.

stmontgomery (approved):

grynspan:
Holding onto this until `@c` is in a build.

grynspan:
Windows failure is unrelated.

PR #1379, opened 2025-10-24
Add an SPI `Attachment.uncheckedPreferredName` property.

grynspan:
This PR adds a for-tools SPI property `uncheckedPreferredName` to `Attachment`. We can then use this property to distinguish when the developer provided or did not provide a name for an attachment.

Tools can then use this property and apply their own naming rules when consuming attachments from the event stream.

Naming is hard; I'm open to other/better names.

### Checklist:


grynspan:
Actually never mind.

PR #1381, opened 2025-10-27
[WIP, DNM] Enable exit tests on Android API level 28 and newer.

grynspan:
This PR enables exit tests on Android.

### Checklist:


Sources/Testing/ExitTests/SpawnProcess.swift
```diff
@@ -194,6 +203,9 @@ func spawnExecutable(
       // spawned child process if we control its execution.
       var environment = environment
       environment["SWT_CLOSEFROM"] = String(describing: highestFD + 1)
+#elseif os(Android)
+      // Android does not have posix_spawn_file_actions_addclosefrom_np() nor
+      // closefrom(2), so we don't attempt this operation there.
```
jakepetroules: Android has close_range (which closefrom is typically trivially implemented in terms of). See swiftlang/swift-subprocess#172

grynspan: `close_range()` requires an upper bound, which we don't know. (Also setting `CLOSE_RANGE_CLOEXEC` is a good way to break code that opened a higher fd and _doesn't_ want `FD_CLOEXEC` set.)

finagolfin:
Thanks, Jonathan, will try it and let you know. 👍

finagolfin:
Seeing a couple build errors because of `@available` attributes:
```
/data/data/com.termux/files/home/swift-testing/Sources/Testing/ExitTests/WaitFor.swift:84:58: [1;31merror: [1;39m'ExitStatus' is only available in Android 28 or newer[0;0m
 [0;36m81 |[0;0m #elseif SWT_TARGET_OS_APPLE || os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android)
 [0;36m82 |[0;0m /// A mapping of awaited child PIDs to their corresponding Swift continuations.
 [0;36m83 |[0;0m private nonisolated(unsafe) let _childProcessContinuations = {
    [0;36m|[0;0m                                 `- [1;39mnote: [1;39madd '@available' attribute to enclosing let[0;0m
 [0;36m84 |[0;0m   let result = ManagedBuffer<[pid_t: CheckedContinuation<ExitStatus, any Error>], pthread_mutex_t>.create(
    [0;36m|[0;0m                                                          |- [1;31merror: [1;39m'ExitStatus' is only available in Android 28 or newer[0;0m
    [0;36m|[0;0m                                                          `- [1;39mnote: [1;39madd 'if #available' version check[0;0m
 [0;36m85 |[0;0m     minimumCapacity: 1,
 [0;36m86 |[0;0m     makingHeaderWith: { _ in [:] }

/data/data/com.termux/files/home/swift-testing/Sources/Testing/ExitTests/WaitFor.swift:106:68: [1;31merror: [1;39m'ExitStatus' is only available in Android 28 or newer[0;0m
[0;36m102 |[0;0m ///
[0;36m103 |[0;0m /// - Throws: Whatever is thrown by `body`.
[0;36m104 |[0;0m private func _withLockedChildProcessContinuations<R>(
    [0;36m|[0;0m              `- [1;39mnote: [1;39madd '@available' attribute to enclosing global function[0;0m
[0;36m105 |[0;0m   _ body: (
[0;36m106 |[0;0m     _ childProcessContinuations: inout [pid_t: CheckedContinuation<ExitStatus, any Error>],
    [0;36m|[0;0m                                                                    `- [1;31merror: [1;39m'ExitStatus' is only available in Android 28 or newer[0;0m
[0;36m107 |[0;0m     _ lock: UnsafeMutablePointer<pthread_mutex_t>
[0;36m108 |[0;0m   ) throws -> R
```
Not going to spend time fixing this given the bigger SwiftPM issue, but if you want to keep iterating, happy to build it locally for you.

grynspan:
Thanks, I'll address those and ping you when it's ready to try again.

Sources/Testing/ExitTests/SpawnProcess.swift
jakepetroules: The idea is that you pass `~0U` (max int) as the upper bound, e.g. when reading SWT_CLOSEFROM, call `close_range(x, ~0U)` instead of `closefrom(x)`. That is how closefrom is implemented internally on the platforms where it exists anyways -- close_range is the syscall, where the kernel can clamp the upper bound down to the actual highest open fd, while closefrom is only in libc.

- https://codebrowser.dev/glibc/glibc/io/closefrom.c.html#29
- https://github.com/freebsd/freebsd-src/blob/9349214a28152a3cb2424459976f571a9c8fc5df/lib/libc/sys/closefrom.c#L36

grynspan:
`sig_t` is not a portable spelling, need a custom `typedef`.

Sources/Testing/ExitTests/SpawnProcess.swift
grynspan: Oh, that works? Neat, okay.

grynspan:
@finagolfin I believe I've got all the availability annotations in place now.

Package.swift
```diff
@@ -398,8 +411,8 @@ extension Array where Element == PackageDescription.SwiftSetting {
 
       .define("SWT_TARGET_OS_APPLE", .whenApple()),
 
-      .define("SWT_NO_EXIT_TESTS", .whenEmbedded(or: .when(platforms: [.iOS, .watchOS, .tvOS, .visionOS, .wasi, .android]))),
-      .define("SWT_NO_PROCESS_SPAWNING", .whenEmbedded(or: .when(platforms: [.iOS, .watchOS, .tvOS, .visionOS, .wasi, .android]))),
+      .define("SWT_NO_EXIT_TESTS", .whenEmbedded(or: .when(platforms: [.iOS, .watchOS, .tvOS, .visionOS, .wasi] + androidIfCompiler6_3))),
```
finagolfin: Compiles fine natively on Android, but no extra tests were run, so I examined this logic and it is wrong: you want these two defines set to disable these tests when _not_ running Swift 6.3 on Android, not when it is.

grynspan: I'm good at Swift! 🫠

grynspan: We're past the branch point for Swift 6.3, so this is moot.

PR #1383, opened 2025-10-27
Ensure `NonisolatedNonsendingByDefault` doesn't break exit tests.

### LINKED ISSUE #1375: Exit tests fail with SIGTRAP on macOS when the 'NonisolatedNonsendingByDefault' upcoming feature is enabled
### Description

In the following snippet, the exit test fails unexpectedly with a SIGTRAP when the upcoming Swift feature 'NonisolatedNonsendingByDefault' is enabled.

[Swift package demonstrating the issue](https://github.com/user-attachments/files/23069306/NonisolatedNonsendingByDefaultExitTestFailure.zip)

### Reproduction

```swift
import Testing

@Test func bug() async throws {
    // When the upcoming feature 'NonisolatedNonsendingByDefault' is enabled, exit tests can't exit successfully:
    // Expectation failed: .success → .signal(SIGTRAP → 5)
    await #expect(processExitsWith: .success) { }
}
```

### Expected behavior

The process should exit normally and the exit test should pass.

### Environment

I can reproduce the issue with Swift 6.2 and a recent main snapshot.

macOS 26.1 Beta (25B5072a)

Swift 6.2:

    Apple Swift version 6.2 (swift-6.2-RELEASE)
    Testing Library Version: 6.2 (3fdabe5392108d8)

main snapshot:

    Apple Swift version 6.3-dev (LLVM 3a427639f2a06cf, Swift 5417f00a58578de)
    Testing Library Version: 6.3-dev (f817ce5863fbd6b)

### Additional information

_No response_

grynspan:
When `NonisolatedNonsendingByDefault` is enabled, overload resolution of `ExitTest.__store()` picks the overload that takes any old `T` instead of taking a function. This overload exists only to suppress certain unhelpful compiler diagnostics and its implementation immediately aborts, which causes the described failure.

Adding `nonisolated(nonsending)` or `@concurrent` to the "good" overload doesn't appear to satisfy the type checker, so mark the "bad" overload as explicitly disfavoured instead.

I am unable to add a unit test for this case due to https://github.com/swiftlang/swift-package-manager/issues/9293.

Resolves #1375.

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -392,6 +392,7 @@ extension ExitTest {
   ///
   /// - Warning: This function is used to implement the
   ///   `#expect(processExitsWith:)` macro. Do not use it directly.
+  @_disfavoredOverload
   @safe public static func __store<T>(
     _ id: (UInt64, UInt64, UInt64, UInt64),
     _ body: T,


grynspan:
I suspect there's an underlying bug in how textual Swift interfaces are used at compile time, but at this point I don't have a repro case I can hand off to the compiler team.

PR #1384, opened 2025-10-28
Avoid `unsafeBitCast()` in the new interop target.

grynspan:
This PR uses `Unmanaged` to avoid calling `unsafeBitCast()` in the new `_TestingInterop` target on platforms that can use `Atomic`. This improves type safety (yay!)

### Checklist:

### DIFF
--- a/Sources/_TestingInterop/FallbackEventHandler.swift
+++ b/Sources/_TestingInterop/FallbackEventHandler.swift
@@ -76,6 +76,12 @@ package func _swift_testing_getFallbackEventHandler() -> FallbackEventHandler? {
     return fallbackEventHandler.pointee
   }
 #else
+  // If we had a setter, this load would present a race condition because
+  // another thread could store a new value in between the load and the call to
+  // `takeUnretainedValue()`, resulting in a use-after-free on this thread. We
+  // would need a full lock in order to avoid that problem. However, because we
+  // instead have a one-time installation function, we can be sure that the
+  // loaded value (if non-nil) will never be replaced with another value.
   return _fallbackEventHandler.load(ordering: .sequentiallyConsistent).map { fallbackEventHandler in
     fallbackEventHandler.takeUnretainedValue().rawValue
   }


Sources/_TestingInterop/FallbackEventHandler.swift
```diff
@@ -99,8 +101,15 @@ package func _swift_testing_installFallbackEventHandler(_ handler: FallbackEvent
     return true
   }
 #else
-  let handler = unsafeBitCast(handler, to: UnsafeRawPointer.self)
-  return _fallbackEventHandler.compareExchange(expected: nil, desired: handler, ordering: .sequentiallyConsistent).exchanged
+  let handler = Unmanaged.passRetained(handler as AnyObject)
```
jerryjrchen: What happening in the `handler as AnyObject` here? Is the closure getting copied onto the heap with type information for us to retrieve later?

jerryjrchen (approved):

Sources/_TestingInterop/FallbackEventHandler.swift
grynspan: Yeah it's just getting boxed in an existential object which is then refcounted and type-safe.

stmontgomery (approved):

Sources/_TestingInterop/FallbackEventHandler.swift
```diff
@@ -99,8 +101,15 @@ package func _swift_testing_installFallbackEventHandler(_ handler: FallbackEvent
     return true
   }
 #else
-  let handler = unsafeBitCast(handler, to: UnsafeRawPointer.self)
```
jerryjrchen, line 102: This new version also improves upon the memory lifetime safety right? It occurs to me now that the previous version required the caller of `_swift_testing_installFallbackEventHandler` to ensure it stays boxed on the heap and allocated forever after being installed.

grynspan:
I'm going to add a path for Embedded Swift that just sets a global (since there's no concurrency support in Embedded Swift at this time.)

Sources/_TestingInterop/FallbackEventHandler.swift
grynspan: `@convention( c)` functions are just function pointers and don't carry any state. They aren't refcounted in Swift and bitcasting one to a pointer doesn't box it.

jerryjrchen: To clarify, I was thinking of something like this being ok:
```swift
static let _fallbackEventHandler: FallbackEventHandler = { ... }
_swift_testing_installFallbackEventHandler(_fallbackEventHandler)
```

But this being not:
```swift
func performInstall() {
  let localhandler: FallbackEventHandler = { ... }
  _swift_testing_installFallbackEventHandler(localHandler)
}
```

Am I understanding correctly that option 2 would have compiled but have been unsafe?

grynspan: Perfectly safe because `FallbackEventHandler` is `@convention(c)` and is not allocated.

Sources/_TestingInterop/FallbackEventHandler.swift
```diff
@@ -26,8 +26,17 @@ private nonisolated(unsafe) let _fallbackEventHandler = {
...
+  let rawValue: FallbackEventHandler
+
+  init(rawValue: FallbackEventHandler) {
+    self.rawValue = rawValue
+  }
+}
+
 /// The installed event handler.
-private nonisolated(unsafe) let _fallbackEventHandler = Atomic<UnsafeRawPointer?>(nil)
+private let _fallbackEventHandler = Atomic<Unmanaged<_FallbackEventHandlerStorage>?>(nil)
```
jerryjrchen, line 39: Is this the change you're referring to here?
> I'm going to add a path for Embedded Swift that just sets a global (since there's no concurrency support in Embedded Swift at this time.)

Package.swift
```diff
@@ -105,6 +105,17 @@ let package = Package(
       )
     )
 
+#if DEBUG
+    // Build _TestingInterop for debugging/testing purposes only. It is
+    // important that clients do not link to this product/target.
+    result += [
+      .library(
+        name: "_TestingInterop_DO_NOT_USE",
+        targets: ["_TestingInterop_DO_NOT_USE"]
```
jerryjrchen, line 114: For testing purposes, would it matter to have this explicitly defined as `type: .dynamic,` here?

grynspan: Nope, we aren't testing it, just compiling it.

Sources/_TestingInterop/FallbackEventHandler.swift
grynspan: That's covered here, yeah. `AnyObject` is unavailable in Embedded Swift.

PR #1385, opened 2025-10-29
Enable `Test.cancel()` as API.

### LINKED ISSUE #120: XCSkip equivalents in swift-testing

grynspan:
Pending approval of [ST-0016](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0016-test-cancellation.md).

Resolves #120.
Resolves rdar://159150449.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Cancellation.swift
+++ b/Sources/Testing/Test+Cancellation.swift
@@ -223,16 +223,13 @@ extension Test: TestCancellable {
   /// current test has been cancelled, but does not attempt to cancel the test a
   /// second time.
   ///
-  /// @Comment {
-  ///   TODO: Document the interaction between an exit test and test
-  ///   cancellation. In particular, the error thrown by this function isn't
-  ///   thrown into the parent process and task cancellation doesn't propagate
-  ///   (because the exit test _de facto_ runs in a detached task.)
-  /// }
-  ///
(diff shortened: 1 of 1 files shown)


grynspan:
@iamleeg Would you please have a look at the documentation changes here? Thanks!

iamleeg (commented):
I think present tense is better throughout to make the prose more engaging, for example:

> If a test is running and you determine it cannot complete and
needs to end early without failing, use ``Test/cancel(_:sourceLocation:)`` to
cancel the test:

Other than that, this all looks great, thanks!

PR #1388, opened 2025-11-03
Capture the current working directory early on OpenBSD.

### LINKED ISSUE #1348: Improve `CommandLine.executablePath` implementation on OpenBSD
Currently on OpenBSD, our implementation of `CommandLine.executablePath` just takes `argv[0]` verbatim. This works poorly if a relative path is used to start the process (e.g. `./foo`). But we can improve it somewhat by calling `realpath()` as early as possible during execution. Something like:

```c++
#include <link.h>
#include <atomic>

static std::atomic<char *> executablePath = nullptr;

__attribute__((__constructor__(101)))
static void tryResolvingExecutablePath(void) {
  (void)dl_iterate_phdr([] (struct dl_phdr_info *info, size_t size, void *cookie) -> int {
    char *path = nullptr;

    if (info->dlpi_name) {
      if (strchr(info->dlpi_name, '/')) {
        path = realpath(info->dlpi_name, nullptr);
      }
      if (!path) {
        path = strdup(info->dlpi_name);
      }
      // NOTE: we do not attempt to resolve $PATH-relative paths here because we don't expect test executables to be installed into e.g. /usr/local/bin.
    }
    executablePath.store(path);

    return 1; // we only care about the first image
  }, nullptr);
}

extern "C" char *swt_getExecutablePath(void);
char *swt_getExecutablePath(void) {
  return executablePath.load();
}
```

We would need to use a constructor function to ensure this code runs as early as possible to minimize the risk of somebody changing the current working directory and breaking the relative path.

grynspan:
This PR adds some OpenBSD-specific code to capture the current working directory as early as possible. We then use that directory when argv[0] appears to be a relative path so that the following incantation works correctly:

```sh
./.build/debug/myPackageTests.xctest --testing-library swift-testing
```

This logic is not necessary on other platforms because they all provide a way to get the path to the current executable. OpenBSD has no such API.

Resolves #1348.

### Checklist:

### DIFF
--- a/Sources/_TestingInternals/include/ExecutablePath.h
+++ b/Sources/_TestingInternals/include/ExecutablePath.h
@@ -17,12 +17,13 @@
 SWT_ASSUME_NONNULL_BEGIN
 
 #if defined(__OpenBSD__)
-/// Get the executable path of the current process.
+/// Get the current working directory as it was set shortly after the process
+/// started and before `main()` has been called.
 ///
 /// This function is necessary on OpenBSD so that we can (as correctly as
 /// possible) resolve the executable path when the first argument is a relative
 /// path (which can occur when manually invoking the test executable.)
-SWT_EXTERN const char *_Nullable swt_getExecutablePath(void);
+SWT_EXTERN const char *_Nullable swt_getEarlyCWD(void);
 #endif
 
 SWT_ASSUME_NONNULL_END


grynspan:
@3405691582 please take a look!

grynspan:
Discussed with @3405691582 in DMs; will need to move the constructor function down to C++ (darn) to ensure it compiles correctly on at least some 6.3-dev toolchains.

3405691582:
Built and tested OK on an OpenBSD 6.3 toolchain.

PR #1390, opened 2025-11-04
Add an upper bound on the number of test cases we run in parallel.

grynspan:
This PR adds an upper bound, `NCORES * 2`, on the number of test cases we run in parallel. Depending on the exact nature of your tests, this can significantly reduce the maximum amount of dirty memory needed, but does not generally impact execution time.

As an example, Swift Testing's own tests go from > 300MB max memory usage to around 60MB.

You can configure a different upper bound by setting the `"SWT_EXPERIMENTAL_MAXIMUM_PARALLELIZATION_WIDTH"` environment variable (look, naming is hard okay?)

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -67,6 +67,14 @@ extension Runner {
   }
 
   /// Context to apply to a test run.
+  ///
+  /// Instances of this type are passed directly to the various functions in
+  /// this file and represent context for the run itself. As such, they are not
+  /// task-local nor are they meant to change as the test run progresses.
+  ///
+  /// This type is distinct from ``Configuration`` which _can_ change on a
+  /// per-test basis. If you find yourself wanting to modify a property of this
+  /// type at runtime, it may be better-suited for ``Configuration`` instead.
   private struct _Context: Sendable {
     /// A serializer used to reduce parallelism among test cases.
     var testCaseSerializer: Serializer?


Sources/Testing/Support/Serializer.swift
```diff
@@ -0,0 +1,65 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// A type whose instances can run a series of work items in strict order.
+///
+/// When a work item is scheduled on an instance of this type, it runs after any
+/// previously-scheduled work items. If it suspends, subsequently-scheduled work
+/// items do not start running; they must wait until the suspended work item
+/// either returns or throws an error.
+final actor Serializer {
```
stmontgomery, line 44:
```swift
/// This type is not part of the public interface of the testing library.
```

Sources/Testing/Running/Runner.swift
```diff
@@ -359,7 +361,11 @@ extension Runner {
     }
 
     await _forEach(in: testCases.enumerated(), namingTasksWith: taskNamer) { _, testCase in
-      await _runTestCase(testCase, within: step)
+      if configuration.isParallelizationEnabled {
+        await configuration.serializer.run { await _runTestCase(testCase, within: step) }
```
stmontgomery: Do we need to go through the serializer when the parallelization width is `Int.max`?

Sources/Testing/Running/Configuration.swift
```diff
@@ -20,6 +22,49 @@ public struct Configuration: Sendable {
...
+  /// cases) will run in parallel. The default value of this property is equal
+  /// to twice the number of CPU cores reported by the operating system, or
+  /// `Int.max` if that value is not available.
+  @_spi(Experimental)
+  public var maximumParallelizationWidth: Int {
+    get {
+      serializer.maximumWidth
+    }
+    set {
+      serializer = Serializer(maximumWidth: newValue)
```
stmontgomery: To me it would make more sense to move the stored `serializer` property to `Runner` and let `Configuration` only keep track of the trivial `maximumParallelizationWidth: Int` value as a stored (instead of computed) property. Then, `maximumParallelizationWidth` could default to `Self._cpuCoreCount.map { max(1, $0) * 2 } ?? .max` to maintain your current behavior.

One reason is that then, the `Serializer` only needs to be kept alive for the duration of a `Runner` instance.

grynspan: The pain-in-the-butt part is that all the `run()` functions are (intentionally) `static`, so we'd need to pass it around as an argument. Thoughts?

Sources/Testing/Running/Runner.swift
grynspan: No, but it's one less `else` branch to worry about.

Sources/Testing/Running/Configuration.swift
grynspan: Made it an argument.

Sources/Testing/Support/Serializer.swift
jerryjrchen: I know we do this throughout the codebase, but isn't this self-evident based on the lack of `public` access modifier on the type?

Sources/Testing/Support/Serializer.swift
```diff
@@ -0,0 +1,86 @@
...
+  nonisolated let maximumWidth: Int
+
+  /// The number of scheduled work items, including any currently running.
+  private var _currentWidth = 0
+
+  /// Continuations for any scheduled work items that haven't started yet.
+  private var _continuations = [CheckedContinuation<Void, Never>]()
+
+  init(maximumWidth: Int = 1) {
+    self.maximumWidth = maximumWidth
```
jerryjrchen, line 56: Should we precondition on `maximumWidth > 0` here? I think `run` blocks forever if the user sets a width of zero, so it doesn't really make sense to allow.

Sources/Testing/Support/Serializer.swift
```diff
@@ -0,0 +1,86 @@
...
+  ///
+  /// - Returns: Whatever is returned from `workItem`.
+  ///
+  /// - Throws: Whatever is thrown by `workItem`.
+  func run<R>(_ workItem: @isolated(any) @Sendable () async throws -> R) async rethrows -> R where R: Sendable {
+    _currentWidth += 1
+    defer {
+      // Resume the next scheduled closure.
+      if !_continuations.isEmpty {
+        let continuation = _continuations.removeFirst()
```
jerryjrchen, line 72: I wonder if this might impact performance if `_continuations.count` is really large since `removeFirst()` is `O(n)`.

Sources/Testing/Support/Serializer.swift
```diff
@@ -0,0 +1,86 @@
...
+
+/// A type whose instances can run a series of work items in strict order.
+///
+/// When a work item is scheduled on an instance of this type, it runs after any
+/// previously-scheduled work items. If it suspends, subsequently-scheduled work
+/// items do not start running; they must wait until the suspended work item
+/// either returns or throws an error.
+///
+/// This type is not part of the public interface of the testing library.
+final actor Serializer {
```
jerryjrchen, line 44: My understanding is that "serial" in this context is referring to the execution order of the tasks, but my initial reaction was that `Serializer` makes it sound like it's related to `.serialized`, i.e. removing parallelism. 

In context at its call site, I think it looks a little contradictory:
```swift
      if configuration.isParallelizationEnabled {
        await context.testCaseSerializer.run {  ... }
```

Maybe we could highlight that this controls the amount of parallelisation, e.g. `WidthLimitedParallelizer`? The execution order of tasks almost feels like an implementation detail - sure, it's FIFO now, but I don't see any reason why we couldn't change it to be LIFO or random or something in the future.

jerryjrchen (approved):
Overall looks good. The only concern I have is over the naming for `Serializer`

Sources/Testing/Support/Serializer.swift
grynspan: Eeeeh okay.

Sources/Testing/Support/Serializer.swift
grynspan: There is no standard FIFO collection type for us to use in Swift and I'm not planning to create one just for this purpose. :) I'd need to see profiling that shows it's an issue before I'm bothered by it.

Sources/Testing/Support/Serializer.swift
grynspan: It is serializing, just potentially partially. Adding extra words to the type name probably won't help much—this code is complex enough that I don't think you can understand it at a glance anyway and would need to read it to figure out what it's doing. 🤷🏻‍♂️

Sources/Testing/Support/Serializer.swift
grynspan: It's a hint to people reading the code that it's _intentionally_ not public.

grynspan:
@stmontgomery Could I get your review please?

Sources/Testing/Support/Serializer.swift
```diff
@@ -0,0 +1,93 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+private import _TestingInternals
+
+/// The number of CPU cores on the current system, or `nil` if that
+/// information is not available.
+var cpuCoreCount: Int? {
```
stmontgomery, line 15:
```suggestion
private var _cpuCoreCount: Int? {
```

stmontgomery (approved):

PR #1394, opened 2025-11-05
Avoid `readToEnd()` on pipes.

grynspan:
Attempting to read to the end of a pipe will block if the writer hasn't finished writing bytes. This can lead to a deadlock (or, at least, pathologically bad read performance) if the writer is _also_ communicating with the reader via another file descriptor or other IPC mechanism. It can also lead to difficult-to-describe issues when the pipe's kernel-side buffer is filled and the writer is blocked waiting for the reader to empty it.

This PR introduces a function on `FileHandle` that reads line-by-line which gives the reader a chance to process data as it comes in rather than waiting for EOF.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/FileHandle.swift
+++ b/Sources/Testing/Support/FileHandle.swift
@@ -211,7 +211,7 @@ struct FileHandle: ~Copyable, Sendable {
   ///
   /// Use this function when calling C I/O interfaces such as `fputs()` on the
   /// underlying C file handle.
-  borrowing func withUnsafeCFILEHandle<R>(_ body: (SWT_FILEHandle) throws -> R) rethrows -> R {
+  borrowing func withUnsafeCFILEHandle<R>(_ body: (SWT_FILEHandle) throws -> R) rethrows -> R where R: ~Copyable {
     try body(_fileHandle)
   }
 
@@ -228,7 +228,7 @@ struct FileHandle: ~Copyable, Sendable {
   /// that require a file descriptor instead of the standard `FILE *`
   /// representation. If the file handle cannot be converted to a file
   /// descriptor, `nil` is passed to `body`.
-  borrowing func withUnsafePOSIXFileDescriptor<R>(_ body: (CInt?) throws -> R) rethrows -> R {
+  borrowing func withUnsafePOSIXFileDescriptor<R>(_ body: (CInt?) throws -> R) rethrows -> R where R: ~Copyable {
     try withUnsafeCFILEHandle { handle in
 #if SWT_TARGET_OS_APPLE || os(Linux) || os(FreeBSD) || os(OpenBSD) || os(Android) || os(WASI)
       let fd = fileno(handle)
@@ -260,7 +260,7 @@ struct FileHandle: ~Copyable, Sendable {
   /// that require the file's `HANDLE` representation instead of the standard
   /// `FILE *` representation. If the file handle cannot be converted to a
   /// Windows handle, `nil` is passed to `body`.
-  borrowing func withUnsafeWindowsHANDLE<R>(_ body: (HANDLE?) throws -> R) rethrows -> R {
+  borrowing func withUnsafeWindowsHANDLE<R>(_ body: (HANDLE?) throws -> R) rethrows -> R where R: ~Copyable {
     try withUnsafePOSIXFileDescriptor { fd in
       guard let fd else {
         return try body(nil)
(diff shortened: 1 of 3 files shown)


Sources/Testing/Support/FileHandle.swift
```diff
@@ -211,7 +211,7 @@ struct FileHandle: ~Copyable, Sendable {
   ///
   /// Use this function when calling C I/O interfaces such as `fputs()` on the
   /// underlying C file handle.
-  borrowing func withUnsafeCFILEHandle<R>(_ body: (SWT_FILEHandle) throws -> R) rethrows -> R {
+  borrowing func withUnsafeCFILEHandle<R>(_ body: (SWT_FILEHandle) throws -> R) rethrows -> R where R: ~Copyable {
```
jerryjrchen, line 214: What's the rationale behind the `R: ~Copyable` in the where clause here? Is it to prevent copying of anything returned by `body` as a performance optimisation?

Sources/Testing/Support/FileHandle.swift
```diff
@@ -354,12 +354,86 @@ extension FileHandle {
...
+#elseif os(Windows)
+          let byteRead = _fgetc_nolock(file)
+#else
+          let byteRead = fgetc(file)
+#endif
+          if byteRead == EOF {
+            if 0 != ferror(file) {
+              throw CError(rawValue: swt_errno())
+            }
+          } else if let byteRead = UInt8(exactly: byteRead) {
```
jerryjrchen, line 395: This is just verifying it really is a char if it's not `EOF`?

Sources/Testing/Support/FileHandle.swift
```diff
@@ -354,12 +354,86 @@ extension FileHandle {
...
+  /// - Returns: A copy of the contents of the file handle starting at the
+  ///   current offset and ending at either the first delimiter byte or the end
+  ///   of the file (whichever comes first). The delimiter byte is not included
+  ///   in the result.
+  ///
+  /// - Throws: Any error that occurred while reading the file or that was
+  ///   thrown by `isDelimiter`.
+  func read(until isDelimiter: (UInt8) throws -> Bool) throws -> [UInt8] {
+    var line = [UInt8]()
+    line.reserveCapacity(1024)
```
jerryjrchen, line 379: What's the reasoning behind this capacity? Seems to me you could get much longer lines for a given delimiter, so is this just based off experience?

Sources/Testing/Support/FileHandle.swift
```diff
@@ -354,12 +354,86 @@ extension FileHandle {
...
+
+    try withUnsafeCFILEHandle { file in
+      try withLock {
```
jerryjrchen, line 382: Random thought... would it make sense for `withLock` to also pass you the file handle?

```swift
try withLock { file in ...
```

I'm thinking it would be slightly more ergonomic for this kind of use case, since it would be a little harder to misuse if the file handler it were limited to the lock scope.

e.g. accidentally using the file handle after the lock is released is possible right now:
```swift
try withUnsafeCFILEHandle { file in
  try withLock {
    // ...
  }
  doStuff(file) // No longer locked here
}
```

jerryjrchen (approved):

Sources/Testing/Support/FileHandle.swift
jerryjrchen: When I wrote my question I forgot this is for lines of the JSON event stream; now the limit makes more sense to me

Sources/Testing/Support/FileHandle.swift
grynspan: I think that's a remnant of an earlier iteration. I'll revert it.

Sources/Testing/Support/FileHandle.swift
grynspan: The value we get back is an `int` or `CInt` and needs to be cast to the correct type. In practice, the `if let` can't fail, so this code is equivalent to `let byteRead = UInt8(byteRead)`.

Sources/Testing/Support/FileHandle.swift
grynspan: Utterly arbitrary. Gotta pick something, no way to know what.

Sources/Testing/Support/FileHandle.swift
grynspan: It's intentional that it doesn't, because then it ought to be `withUnsafeCFILEHandleWhileLocked` or something like that, and ultimately it's not unsafe to acquire the lock _and_ it's valid to get the C `FILE *` without needing to lock it. So we just compose the two.

grynspan: Note that this lock is a POSIX (or Win32) API that doesn't actually bind the `FILE *` to a mutex like `OSAllocatedUnfairLock`, `Mutex`, or `Locked` does, so replicating the semantics of those APIs too closely might be misleading.

PR #1395, opened 2025-11-06
Avoid a potential race condition and use-after-free when calling `Test.cancel()`.

grynspan:
If a test creates an unstructured, non-detached task that continues running after the test has finished and eventually calls `Test.cancel()`, then it may be able to see a reference to the test's (or test case's) task after it has been destroyed by the Swift runtime.

This PR ensures that the infrastructure under `Test.cancel()` clears its reference to the test's task before returning. This then minimizes the risk of observing the task after it has been destroyed.

Further work at the Swift runtime level may be required to completely eliminate this race condition, but this change makes it sufficiently narrow that any example I can come up with is contrived.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Cancellation.swift
+++ b/Sources/Testing/Test+Cancellation.swift
@@ -87,10 +87,19 @@ extension TestCancellable {
   /// the current task, test, or test case is cancelled, it records a
   /// corresponding cancellation event.
   func withCancellationHandling<R>(_ body: () async throws -> R) async rethrows -> R {
+    let taskReference = _TaskReference()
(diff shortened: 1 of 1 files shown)


grynspan:
The dictionary is a task-local, so no lock needed there. The elements are each a struct with a lock that guards the actual `UnsafeCurrentTask` instance. `takeUnsafeCurrentTask()` acquires the lock, swaps the task with `nil`, and returns it to the caller.

grynspan:
Me trying to explain the race condition:

pepe-silvia-v0-dvnwtwovkkke1 jpg

briancroom (approved):

PR #1397, opened 2025-11-06
Remove VERSION.txt workaround for OpenBSD 7.7.

grynspan:
This PR removes a workaround for the lack of `#embed` support in OpenBSD 7.7's version of clang. OpenBSD 7.8 has a newer clang that has full `#embed` support, so we no longer need to special-case OpenBSD here.

Swift Testing will continue to compile for OpenBSD 7.7, but will emit a compile-time warning of the form:

> SWT_TESTING_LIBRARY_VERSION not defined and could not read from VERSION.txt at
> compile time: testing library version is unavailable

### Checklist:


grynspan:
@3405691582 FYI.

3405691582 (approved):

briancroom (approved):

PR #1398, opened 2025-11-06
Remove non-functional `swt_getWASIVersion()`.

grynspan:
A while back I implemented `swt_getWASIVersion()` to speculatively return the value of `WASI_SDK_VERSION` as defined by WASI, but that never landed and the [GitHub issue](https://github.com/WebAssembly/wasi-libc/issues/490) for it was closed. Remove the inoperative helper function here.

### Checklist:


grynspan:
Not sure why that one Android job is failing, but it's unrelated.

PR #1399, opened 2025-11-06
Handle image types derived from the base types.

grynspan:
On Apple platforms, you can declare file types (uniform type identifiers, or "UTTypes" for short) that conform to existing types. That means a developer could conceivably say "save this image as .florb" where `.florb` represents a type that conforms to, but is not identical to, `.jpeg`.

This PR ensures that if a developer passes such a type, we handle it by passing the supported base type to Image I/O. And, if the developer passes some image format that is unsupported by Image I/O, we can throw an error with a more specific diagnostic than just "it failed."

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
+++ b/Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
@@ -50,17 +50,17 @@ package protocol AttachableAsCGImage: AttachableAsImage {
   var attachmentScaleFactor: CGFloat { get }
 }
 
-/// The set of type identifiers supported by Image I/O.
+/// All type identifiers supported by Image I/O.
 @available(_uttypesAPI, *)
 private let _supportedTypeIdentifiers = Set(CGImageDestinationCopyTypeIdentifiers() as? [String] ?? [])
 
-/// The set of content types supported by Image I/O.
+/// All content types supported by Image I/O.
 @available(_uttypesAPI, *)
 private let _supportedContentTypes = {
 #if canImport(UniformTypeIdentifiers_Private)
-  Set(UTType._types(identifiers: _supportedTypeIdentifiers).values)
+  UTType._types(identifiers: _supportedTypeIdentifiers).values
 #else
-  Set(_supportedTypeIdentifiers.compactMap(UTType.init(_:)))
+  _supportedTypeIdentifiers.compactMap(UTType.init(_:))
 #endif
 }()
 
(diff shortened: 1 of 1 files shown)


Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
```diff
@@ -47,6 +50,20 @@ package protocol AttachableAsCGImage: AttachableAsImage {
   var attachmentScaleFactor: CGFloat { get }
 }
 
+/// The set of type identifiers supported by Image I/O.
+@available(_uttypesAPI, *)
+private let _supportedTypeIdentifiers = Set(CGImageDestinationCopyTypeIdentifiers() as? [String] ?? [])
+
+/// The set of content types supported by Image I/O.
+@available(_uttypesAPI, *)
+private let _supportedContentTypes = {
```
jerryjrchen, line 59: Nit: seems like we only use this for linear search later:
```swift 
_supportedContentTypes.first(where: contentType.conforms(to:)
```

Would there be any downside to making this a plain `[UTType]` instead? IMO the intention would be clearer that way.

Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
```diff
@@ -63,8 +80,17 @@ extension AttachableAsCGImage {
     // Convert the image to a CGImage.
     let attachableCGImage = try attachableCGImage
 
+    // Determine the base content type to use.
+    var contentType = imageFormat.contentType
+    if !_supportedTypeIdentifiers.contains(contentType.identifier) {
```
jerryjrchen, line 89: Would it be possible to consistently  check `_supportedContentTypes` where possible? Then it would look like:
```suggestion
    if !_supportedContentTypes.contains(contentType) {
```

Also I realise this is mutually exclusive with my other suggestion to turn `_supportedContentTypes` into an array :)

stmontgomery (approved):

jakepetroules:
I think you need to rebase the PR to pick up the GItHub Actions fixes for Amazon Linux 2 + Android. It appears re-running a failing job will still use the commit of the swiftlang/github-workflows repository at the time the PR was last pushed, rather than when the job was last run.

Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
grynspan: Yeah this can just be inferred type. Will adjust.

Sources/Overlays/_Testing_CoreGraphics/Attachments/AttachableAsCGImage.swift
grynspan: String comparisons are much cheaper to do for the common cases (where the type is either PNG or JPEG). We want the fast path to be fast, after all.

PR #1400, opened 2025-11-06
Fix a regression that prevents saving attachments.

grynspan:
This PR ensures that the attachment-saving logic in `Runner` is applied consistently. I made the mistake of moving that logic to `configureEventHandlerRuntimeState()` because I'd forgotten that that function only has an effect when we're running our own tests. And, as a result, all our tests continued to pass because the code was in place for our tests (and for nobody else.)

I've moved the code to a different location that we will always call when running a `Runner` instance, so the issue is resolved now. I don't have a great way to set up a unit test for this; tested manually at desk.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.RuntimeState.swift
+++ b/Sources/Testing/Running/Runner.RuntimeState.swift
@@ -48,16 +48,6 @@ extension Runner {
     }
 
     configuration.eventHandler = { [oldEventHandler = configuration.eventHandler] event, context in
-#if !SWT_NO_FILE_IO
-      var event = copy event
-      if case .valueAttached = event.kind {
-        guard let configuration = context.configuration,
-              configuration.handleValueAttachedEvent(&event, in: context) else {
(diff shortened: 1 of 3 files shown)


stmontgomery (approved):

Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -48,16 +48,6 @@ extension Runner {
     }
 
     configuration.eventHandler = { [oldEventHandler = configuration.eventHandler] event, context in
-#if !SWT_NO_FILE_IO
```
briancroom, line 51: Do you think it would make sense to further expand on this method's doc comment to help avoid this sort of problem happening again in the future?

grynspan: Maybe. I think this was just me in a sleep-deprived stupor though, and I'm not sure anybody else would fail to read the `guard` statement two lines up.

PR #1403, opened 2025-11-07
Avoid `PATH_MAX` when getting the executable path on Linux and OpenBSD.

grynspan:
This PR adjusts the implementations of `CommandLine.executablePath` on Linux and OpenBSD to avoid an artificial upper bound based on `PATH_MAX`.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/CommandLineAdditions.swift
+++ b/Sources/Testing/Support/Additions/CommandLineAdditions.swift
@@ -47,7 +47,7 @@ extension CommandLine {
           }
           if readCount < buffer.count {
             buffer[readCount] = 0 // NUL-terminate the string.
-            return String(cString: buffer.baseAddress!)
+            result = String(cString: buffer.baseAddress!)
           } else {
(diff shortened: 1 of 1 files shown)


Sources/_TestingInternals/ExecutablePath.cpp
```diff
@@ -24,9 +24,8 @@ static constinit std::atomic<const char *> earlyCWD { nullptr };
 /// path (which can occur when manually invoking the test executable.)
 __attribute__((__constructor__(101), __used__))
 static void captureEarlyCWD(void) {
-  static char buffer[PATH_MAX * 2];
-  if (getcwd(buffer, sizeof(buffer))) {
-    earlyCWD.store(buffer);
+  if (auto cwd = getcwd(nil, 0)) {
```
3405691582, line 27: should be nullptr (buildbreak).

grynspan: D'oh

grynspan: Fixed directly on main.

PR #1404, opened 2025-11-07
When saving a `String` as an attachment, assume `".txt"` as a path extension.

grynspan:
When a string is attached to a test, we encode it as UTF-8 (as is tradition). This PR adjusts `String`'s conformance to `Attachable` so that it uses the `".txt"` extension (unless the user specifies something else). This improves usability in Xcode test reports by making these attachments double-clickable and QuickLookable (is that a word?)

### Checklist:

### DIFF
--- a/Sources/Testing/Attachments/Attachable.swift
+++ b/Sources/Testing/Attachments/Attachable.swift
@@ -103,17 +103,37 @@ public protocol Attachable: ~Copyable {
 
 // MARK: - Default implementations
 
+/// @Metadata {
+///   @Available(Swift, introduced: 6.2)
+///   @Available(Xcode, introduced: 26.0)
+/// }
 extension Attachable where Self: ~Copyable {
+  /// @Metadata {
+  ///   @Available(Swift, introduced: 6.2)
+  ///   @Available(Xcode, introduced: 26.0)
+  /// }
   public var estimatedAttachmentByteCount: Int? {
     nil
   }
 
+  /// @Metadata {
+  ///   @Available(Swift, introduced: 6.2)
+  ///   @Available(Xcode, introduced: 26.0)
+  /// }
   public borrowing func preferredName(for attachment: borrowing Attachment<Self>, basedOn suggestedName: String) -> String {
     suggestedName
   }
 }
 
+/// @Metadata {
+///   @Available(Swift, introduced: 6.2)
+///   @Available(Xcode, introduced: 26.0)
+/// }
 extension Attachable where Self: Collection, Element == UInt8 {
+  /// @Metadata {
+  ///   @Available(Swift, introduced: 6.2)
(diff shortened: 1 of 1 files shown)


Sources/Testing/Attachments/Attachable.swift
```diff
@@ -131,6 +131,13 @@ extension Attachable where Self: StringProtocol {
     // SEE: https://github.com/swiftlang/swift/blob/main/stdlib/public/core/StringUTF8View.swift
     utf8.count
   }
+
+  public borrowing func preferredName(for attachment: borrowing Attachment<Self>, basedOn suggestedName: String) -> String {
+    if suggestedName.contains(".") {
```
jerryjrchen, line 168: I suppose there isn't a consistent way to detect if the `.` precedes a file extension or if the user just wanted to use a namespace or something :/

e.g. `my.cool.file.html` vs `my.cool.file`

Tests/TestingTests/Traits/AttachmentSavingTraitTests.swift
```diff
@@ -130,27 +130,27 @@ extension `AttachmentSavingTrait tests` {
   @Suite(.hidden, currentAttachmentSavingTrait)
   struct FixtureSuite {
     @Test(.hidden) func `Records an attachment (passing)`() {
-      Attachment.record("", named: "PASSING TEST")
+      Attachment.record([], named: "PASSING TEST")
```
jerryjrchen, line 133: Ah, these changes are needed because the old placeholder string would rename the attachment `PASSING TEST.txt`?

Sources/Testing/Attachments/Attachable.swift
grynspan: In terms of filenames, anything after the period character _is_ the extension. It might not be a meaningful one to us, but that's what it is.

Tests/TestingTests/Traits/AttachmentSavingTraitTests.swift
grynspan: Yup

Sources/Testing/Attachments/Attachable.swift
```diff
@@ -125,37 +145,88 @@ extension Attachable where Self: Collection, Element == UInt8 {
...
+  ///   @Available(Xcode, introduced: 26.0)
+  /// }
   public var estimatedAttachmentByteCount: Int? {
     // NOTE: utf8.count may be O(n) for foreign strings.
     // SEE: https://github.com/swiftlang/swift/blob/main/stdlib/public/core/StringUTF8View.swift
     utf8.count
   }
+
+  /// @Metadata {
+  ///   @Available(Swift, introduced: 6.2)
```
grynspan, line 164: The conformance was introduced in 6.2, so yes this is accurate. :)

Issue #1370, opened 2025-10-17
Adopt `#if objectFormat()`


grynspan:
Adopt `#if objectFormat()` in `_TestDiscovery` and in the `@Test`/`@Suite`/`#expect(processExitsWith:)` macro implementations.

We'll need to wait a release (presumably until 6.4) to adopt because we can't write:

```swift
#if objectFormat(ELF)
// ELF implementation
#else ...
```

And have it compile on 6.2 toolchains. Using `os()` is sufficient for now.

See: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0492-section-control.md

grynspan:
See https://github.com/apple/swift-play-experimental/issues/5

PR #1408, opened 2025-11-10
Adopt `@section` and `@used`.

### LINKED ISSUE #1371: Adopt `@section` and `@used`
Adopt the newly-approved `@section` and `@used` attributes. We'll need fallback paths for 6.2 toolchains until 6.3 ships.

See: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0492-section-control.md

grynspan:
This PR adopts the `@section` and `@used` attributes now available in the Swift 6.3 toolchain. These attributes replace the experimental `@_section` and `@_used` attributes verbatim and they are always enabled (so no need to check `hasFeature(SymbolLinkageMarkers)`.)

Because these attributes are only available with the Swift 6.3 toolchain, we do not use them when building with Swift 6.2. Instead, if you build with Swift 6.2 you continue to use the "legacy" test discovery mechanism based on type metadata emission. The "legacy" mechanism will be removed in a future update.

Because the `_TestDiscovery` target needs to continue to compile with Swift 6.2, I have not switched us over to `objectFormat()` yet. See #1370 for details.

Resolves #1371.
Resolves rdar://136789701.

### Checklist:


grynspan:
Blocked by https://github.com/swiftlang/swift/issues/85411.

grynspan:
Windows failure is unrelated. Failure on 6.3-amazonlinux2 is due to an out-of-date toolchain (should self-resolve sooner or later.)

Tests/TestingTests/DiscoveryTests.swift
```diff
@@ -143,21 +143,4 @@ struct DiscoveryTests {
     })
   }
 #endif
-
-#if !SWT_NO_LEGACY_TEST_DISCOVERY && hasFeature(SymbolLinkageMarkers)
-  @Test("Legacy test discovery finds the same number of tests") func discoveredTestCount() async {
```
stmontgomery, line 148: We're intentionally not removing the `SWT_USE_LEGACY_TEST_DISCOVERY` env var and its supporting code, right? Can/should we keep this test to ensure that continues working?

Sources/Testing/Discovery+Macro.swift
```diff
@@ -34,7 +34,7 @@ public typealias __TestContentRecordAccessor = @convention(c) (
 public typealias __TestContentRecord = (
   kind: UInt32,
   reserved1: UInt32,
-  accessor: __TestContentRecordAccessor?,
+  accessor: __TestContentRecordAccessor,
```
stmontgomery, line 37: What's this change for?

grynspan: Optional C functions don't work with the current implementation of `@section`, so we must make them non-optional for now.

Tests/TestingTests/DiscoveryTests.swift
grynspan: This _test_ does not continue working because we stop emitting the legacy metadata. We still _detect_ it though.

Sources/Testing/Discovery+Macro.swift
grynspan: https://github.com/swiftlang/swift/issues/86051

Tests/TestingTests/DiscoveryTests.swift
grynspan: (We still perform legacy discovery on an image-by-image basis, and the environment variable is still honoured on that side.)

stmontgomery: That makes sense, but makes me wonder if we ought to add back a "hardcoded" legacy-style fixture test type solely for the purpose of validating the old lookup mechanism. If we did that, we could presumably preserve this test, albeit with modification to e.g. only look for the one or two hardcoded legacy tests.

If we already have something that you think adequately validates that code path that may be sufficient, feel free to point me elsewhere. I just worry a little about losing coverage of that area.

grynspan: Okay, I can add a test that sniffs a specific test content record created in the old style.

grynspan: Added a test.

grynspan:
Holding until the necessary compiler changes have percolated further.

PR #1410, opened 2025-11-10
Mark `Issue.Severity` cases with availability.

grynspan:
It is necessary to explicitly mark the cases of this enum with their Swift 6.3 availability as it is not inherited when rendered in DocC.

### Checklist:


stmontgomery (approved):

suzannaratcliff:
Thank you!

PR #1436, opened 2025-12-03
Remove `GUID.Wrapper` from the WinSDK overlay.

grynspan:
Now that the `Equatable` and `Hashable` conformances for `GUID` have landed in the 6.3 toolchain (see https://github.com/swiftlang/swift/pull/84792), we can remove our wrapper workaround type.

This should have no other effects on the public-facing interface of the overlay.

### Checklist:


grynspan:
We can also hold off on this until we stop automerging to 6.3 so that main is 6.4(?) and we don't provide 6.2 support anymore.

grynspan:
Superseded by #1481.

PR #1438, opened 2025-12-04
Bump `ABI.CurrentVersion` to 6.3 and ensure minor variances in JSON version numbers are permitted.

grynspan:
This PR ensures that the JSON schema version 6.3 is considered "supported" in Swift Testing 6.3. As well, it suppresses a thrown error when the `version` field of a JSON record is slightly higher than the expected schema. For example, if we ask Swift Testing the ABI version corresponding to `VersionNumber(6, 3, 1)`, we'll get back `v6_3`, but then if we try to decode a JSON record whose `version` field equals `"6.3.1"`, we'll throw a version mismatch error. This error is now suppressed unless the versions really do mismatch.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/ABI.swift
+++ b/Sources/Testing/ABI/ABI.swift
@@ -65,7 +65,7 @@ extension ABI {
   ///   ABI version, or `nil` if no such type exists.
   static func version(
     forVersionNumber versionNumber: VersionNumber,
-    givenSwiftCompilerVersion swiftCompilerVersion: VersionNumber = swiftCompilerVersion
+    givenSwiftCompilerVersion swiftCompilerVersion: @autoclosure () -> VersionNumber = swiftCompilerVersion
   ) -> (any Version.Type)? {
     if versionNumber > ABI.HighestVersion.versionNumber {
       // If the caller requested an ABI version higher than the current Swift
@@ -79,7 +79,7 @@ extension ABI {
       // Note also that building an old version of Swift Testing with a newer
       // compiler may produce incorrect results here. We don't generally support
       // that configuration though.
-      if versionNumber > swiftCompilerVersion {
+      if versionNumber > swiftCompilerVersion() {
         return nil
       }
     }


stmontgomery:
Did you intend to also modify `ABI.CurrentVersion`? I don't see that change included

grynspan:
@stmontgomery Looks like the git commit didn't pick up the change for some reason. Fixed.

grynspan:
Yeah it looks like git ate part of the PR. Am confused as I definitely fixed that test locally.

stmontgomery:
I've pushed a commit which should fix the broken test, and modify the related `ABI.version(forVersionNumber:...)` function to make it more resilient to different compiler versions and thus hopefully less likely to fail for different reasons in the future.

Sources/Testing/ABI/ABI.swift
```diff
@@ -55,10 +55,18 @@ extension ABI {
...
+  ///     to determine whether a version type can be returned. The default value
+  ///     is the version of the Swift compiler which was used to build the
+  ///     testing library.
   ///
   /// - Returns: A type conforming to ``ABI/Version`` that represents the given
   ///   ABI version, or `nil` if no such type exists.
-  static func version(forVersionNumber versionNumber: VersionNumber = ABI.CurrentVersion.versionNumber) -> (any Version.Type)? {
+  static func version(
+    forVersionNumber versionNumber: VersionNumber,
+    givenSwiftCompilerVersion swiftCompilerVersion: VersionNumber = swiftCompilerVersion
```
grynspan: Should this be an autoclosure? It may not be free to compute.

stmontgomery (approved):
I have contributed part of the changes here, but I'm hitting approve on the remainder

Sources/Testing/ABI/ABI.Record.swift
```diff
@@ -70,15 +70,26 @@ extension ABI.Record: Codable {
   init(from decoder: any Decoder) throws {
     let container = try decoder.container(keyedBy: CodingKeys.self)
 
-    let versionNumber = try container.decode(VersionNumber.self, forKey: .version)
-    if versionNumber != V.versionNumber {
+    func validateVersionNumber(_ versionNumber: VersionNumber) throws {
+      if versionNumber == V.versionNumber {
+        return
+      }
+#if !hasFeature(Embedded)
```
jerryjrchen, line 77: Why do we have to strictly match the patch version on embedded?

stmontgomery: I _believe_ it's due to a more fundamental limitation: in Embedded Swift, there are no metatypes at all, so an expression such as `V.self` is not valid

grynspan: Yup.

grynspan: I'm no longer as enamoured of the "versions are types" approach and will probably make this code non-generic in a future release.

PR #1439, opened 2025-12-05
Interop: send events to fallback event handler if active configuration not present

jerryjrchen:
Create helper to post test events to fallback event handler

As part of the strategy for enabling interoperability, unhandled issues
can be turned into events and sent to the fallback event handler.

Also link against the _TestingInterop framework/library (depending on
the platform), which provides a function to lookup the currently
installed fallback event handler.

### DIFF
--- a/Tests/TestingTests/EventHandlingInteropTests.swift
+++ b/Tests/TestingTests/EventHandlingInteropTests.swift
@@ -1,0 +1,67 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+private import _TestingInternals
+@testable @_spi(ForToolsIntegrationOnly) import Testing
+
+#if canImport(Foundation)
+import Foundation
+#endif
+
+#if !SWT_NO_EXIT_TESTS && compiler(>=6.3) && !SWT_NO_INTEROP && canImport(Foundation)
+struct EventHandlingInteropTests {
+  private typealias InstallFallbackEventHandler =
+    @convention(c) (SWTFallbackEventHandler) -> CBool
+  private static let installer: InstallFallbackEventHandler? = {
+    symbol(named: "_swift_testing_installFallbackEventHandler").map {
+      castCFunction(at: $0, to: InstallFallbackEventHandler.self)
+    }
+  }()
+
+  static let handlerContents = Locked<(version: String, record: String?)?>(rawValue: nil)
+
+  private static let capturingHandler: SWTFallbackEventHandler = { schemaVersion, recordJSONBaseAddress, recordJSONByteCount, _ in
+    let version = String(cString: schemaVersion)
+    let record = String(
+      data: Data(bytes: recordJSONBaseAddress, count: recordJSONByteCount),
+      encoding: .utf8)
+    Self.handlerContents.withLock {
+      $0 = (version: version, record: record)
+    }
+  }
+
+  /// This uses an exit test to run in a clean process, ensuring that the
+  /// installed fallback event handler does not affect other tests.
+  ///
+  /// Note this test will no longer work once Swift Testing starts installing
+  /// its own fallback handler.
+  @Test func `Post event without config -> fallback handler`() async throws {
+    await #expect(processExitsWith: .success) {
+      let installer = try #require(Self.installer)
+      try #require(
+        installer(Self.capturingHandler), "Installation of fallback handler should succeed")
+
+      // The detached task forces the event to be posted when Configuration.current
+      // is nil and triggers the post to fallback handler path
+      await Task.detached {
+        Event.post(.issueRecorded(Issue(kind: .system)), configuration: nil)
+      }.value
+
+      // Assert that the expectation failure contents were sent to the fallback event handler
+      try Self.handlerContents.withLock {
+        let contents = try #require(
+          $0, "Fallback should have been called with non nil contents")
+        #expect(contents.version == "6.3")
+        #expect(contents.record?.contains("A system failure occurred") ?? false)
+      }
+    }
+  }
+}
+#endif


Sources/_TestingInternals/include/Stubs.h
```diff
@@ -180,6 +180,17 @@ static int swt_setfdflags(int fd, int flags) {
 }
 #endif
 
+#if !SWT_NO_INTEROP
+
+typedef void (*FallbackEventHandler)(const char *recordJSONSchemaVersionNumber,
```
stmontgomery: This PR's still a draft, but if you stick with this approach, please plan to add some documentation for these (even if it's a cursory explanation of how they shadow functions elsewhere).

Sources/Testing/Events/Event+FallbackHandler.swift
```diff
@@ -0,0 +1,44 @@
...
+  /// Post this event to the currently-installed fallback event handler.
+  ///
+  /// - Parameters:
+  ///   - context: The context associated with this event.
+  ///
+  /// - Returns: Whether or not the fallback event handler was invoked. If the
+  ///   currently-installed handler belongs to the testing library, returns
+  ///   `false`.
+  borrowing func postToFallbackHandler(in context: borrowing Context) -> Bool {
+    #if canImport(_TestingInterop)
```
stmontgomery: Style-wise, we usually outdent `#if` guards to column 0

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -180,6 +180,17 @@ static int swt_setfdflags(int fd, int flags) {
 }
 #endif
 
+#if !SWT_NO_INTEROP
```
grynspan, line 183: Anywhere you're doing this, I suggest changing it to:

```suggestion
#if compiler(>=6.3) && !SWT_NO_INTEROP
```

So that when we build against the 6.2 toolchain, we don't even try to link this function.

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -180,6 +180,17 @@ static int swt_setfdflags(int fd, int flags) {
...
+
+typedef void (*FallbackEventHandler)(const char *recordJSONSchemaVersionNumber,
+                                     const void *recordJSONBaseAddress,
+                                     long recordJSONByteCount,
+                                     const void *_Nullable reserved);
+
+FallbackEventHandler _Nullable _swift_testing_getFallbackEventHandler();
```
grynspan:
```suggestion
SWT_EXTERN FallbackEventHandler _Nullable _swift_testing_getFallbackEventHandler(void);
```

The argument lists `()` and `(void)` don't mean the same thing in C. 😱

Sources/_TestingInternals/include/Stubs.h
grynspan:
```suggestion
typedef void (* FallbackEventHandler)(const char *recordJSONSchemaVersionNumber,
```

The smallest possible nitpick. :)

Sources/Testing/Events/Event+FallbackHandler.swift
```diff
@@ -0,0 +1,44 @@
...
+      guard let fallbackEventHandler = _swift_testing_getFallbackEventHandler() else {
+        return false
+      }
+
+      // Encode the event as JSON and pass it to the handler.
+      let encodeAndInvoke = ABI.CurrentVersion.eventHandler(encodeAsJSONLines: false) {
```
grynspan: You might want to cache this event handler somewhere as there's a bit of runtime overhead to generating it. Remember the fallback event handler can't change after its initial assignment, and will be set by the time an event is generated, so we don't need to worry about it having changed on later calls (i.e. the closure we generate here can permanently capture it.)

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -180,6 +180,31 @@ static int swt_setfdflags(int fd, int flags) {
...
+///
+/// - Parameters:
+///   - recordJSONSchemaVersionNumber: The JSON schema version used to encode
+///     the event record.
+///   - recordJSONBaseAddress: A pointer to the first byte of the encoded event.
+///   - recordJSONByteCount: The size of the encoded event in bytes.
+///   - reserved: Reserved for future use.
+typedef void (* FallbackEventHandler)(const char *recordJSONSchemaVersionNumber,
+                                      const void *recordJSONBaseAddress,
+                                      long long recordJSONByteCount,
```
jerryjrchen: I'm choosing ~`long long`~ `size_t` here because I learned today that Windows is LLP64, so `long` maps to `Int32` and not `Int`

grynspan: Use `size_t`.

Sources/Testing/Events/Event+FallbackHandler.swift
```diff
@@ -0,0 +1,50 @@
...
+  ///   `false`.
+  borrowing func postToFallbackHandler(in context: borrowing Context) -> Bool {
+#if compiler(>=6.3) && !SWT_NO_INTEROP
+      guard let fallbackEventHandler = Self._fallbackEventHandler else {
+        return false
+      }
+
+      // Encode the event as JSON and pass it to the handler.
+      let encodeAndInvoke = ABI.CurrentVersion.eventHandler(encodeAsJSONLines: false) {
+        recordJSON in
```
grynspan: Nitpick, closure headers usually go on the same line as the open brace.

Package.swift
```diff
@@ -142,9 +142,19 @@ let package = Package(
...
-        .linkedLibrary("execinfo", .when(platforms: [.custom("freebsd"), .openbsd]))
-      ]
+      linkerSettings: {
+        var result = [LinkerSetting]()
+        result += [
+          .linkedLibrary("execinfo", .when(platforms: [.custom("freebsd"), .openbsd]))
+        ]
+#if compiler(>=6.3)
+        result += [
+          .linkedLibrary("_TestingInterop"),
```
stmontgomery, line 152: I guess we need to add this linkage to our CMake build too, and without the 6.3 conditional, so that the official copy of Swift Testing in Swift.org toolchains will have it. I imagine that could be added here:

https://github.com/swiftlang/swift-testing/blob/main/Sources/Testing/CMakeLists.txt#L121-L123

Sources/Testing/Events/Event.swift
stmontgomery, line 337: With this change, I wonder what remaining practical usages of `Configuration.all` there will be. @grynspan, do you think it's still worth keeping that?

stmontgomery: It would be nice to update this documentation given the implementation changes, and keep in mind whatever we decide w.r.t. [this comment](https://github.com/swiftlang/swift-testing/pull/1439/changes#r2687991693)

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -180,6 +180,31 @@ static int swt_setfdflags(int fd, int flags) {
...
+/// alternate method of reporting test events to the current test runner.
+/// Shadows the type with the same name in _TestingInterop.
+///
+/// - Parameters:
+///   - recordJSONSchemaVersionNumber: The JSON schema version used to encode
+///     the event record.
+///   - recordJSONBaseAddress: A pointer to the first byte of the encoded event.
+///   - recordJSONByteCount: The size of the encoded event in bytes.
+///   - reserved: Reserved for future use.
+typedef void (* FallbackEventHandler)(const char *recordJSONSchemaVersionNumber,
```
stmontgomery: Namespace prefix
```suggestion
typedef void (* SWTFallbackEventHandler)(const char *recordJSONSchemaVersionNumber,
```

Sources/Testing/Events/Event.swift
grynspan: Yes, we still need it as it provides full-fidelity issue handling when in detached tasks. The fallback handler is orthogonal.

jerryjrchen:
> My original draft of this change included a check whether the fallback event handler was equal to the one installed by Swift Testing because otherwise it might be possible to infinitely recurse into it. Is that not needed?

I think that check is needed, but I'm holding off on it for now since installing a fallback handler is coming in a follow up PR! Currently there is no Swift Testing installed fallback event handler to check against.

Sources/Testing/CMakeLists.txt
```diff
@@ -140,6 +142,8 @@ if(NOT BUILD_SHARED_LIBS)
   target_compile_options(Testing PRIVATE
     "SHELL:-Xfrontend -public-autolink-library -Xfrontend _TestDiscovery"
     "SHELL:-Xfrontend -public-autolink-library -Xfrontend _TestingInternals")
+  # Interop is not supported at this time
+  target_compile_definitions(Testing PRIVATE SWT_NO_INTEROP)
```
jerryjrchen: I just realised that `SWT_NO_INTEROP` is used in `_TestingInternals` and `_TestingInterop` as well. Do I need to add separate `target_compile_definitions` for those libraries too?

Tests/TestingTests/EventHandlingInteropTests.swift
```diff
@@ -0,0 +1,67 @@
...
+      encoding: .utf8)
+    Self.handlerContents.withLock {
+      $0 = (version: version, record: record)
+    }
+  }
+
+  /// This uses an exit test to run in a clean process, ensuring that the
+  /// installed fallback event handler does not affect other tests.
+  ///
+  /// Note this test will no longer work once Swift Testing starts installing
```
jerryjrchen, line 43: ℹ️ This test is a bit janky but I'm also gonna have to rewrite it with the next PR anyway once we start installing the fallback handler in the entry point. In the meantime, this does exercise the fallback code path as expected.

Sources/_TestingInternals/include/Stubs.h
jerryjrchen: Discussed offline: this wasn't possible in Stubs.h since I couldn't check the compiler version. However, the implementations *are* guarded with the complier version check, so it should be fine.

Sources/Testing/CMakeLists.txt
stmontgomery: To be safe, probably yes. We could also hoist this up (along with the `if(NOT BUILD_SHARED_LIBS)` guard which surrounds it) to a higher-level CMake file so it would apply to all targets.

Sources/Testing/Events/Event+FallbackHandler.swift
```diff
@@ -0,0 +1,49 @@
...
+  ///
+  /// - Parameters:
+  ///   - context: The context associated with this event.
+  ///
+  /// - Returns: Whether or not the fallback event handler was invoked. If the
+  ///   currently-installed handler belongs to the testing library, returns
+  ///   `false`.
+  borrowing func postToFallbackHandler(in context: borrowing Context) -> Bool {
+#if compiler(>=6.3) && !SWT_NO_INTEROP
+      guard let fallbackEventHandler = Self._fallbackEventHandler else {
```
stmontgomery: nit: 2-space indentation in this method and the property above

PR #1452, opened 2025-12-16
Document the official distribution locations and the recommendation to use a built-in copy

stmontgomery:
This adds a new documentation article within the top-level `Documentation/` directory describing the various ways Swift Testing is distributed and offers the specific recommendation that most users prefer using a built-in copy instead of a package dependency.

View the [rendered document](https://github.com/stmontgomery/swift-testing/blob/deployment-locations-doc/Documentation/Distributions.md) for more details.

> Note: In this PR the new article is only being placed in the `Documentation/` directory. I chose to place it there, initially, and not within the project's official DocC documentation catalog, but that could potentially happen in the future.

### Motivation:

We have been casually giving users the recommendation to prefer built-in copies of the `Testing` module for a while in settings like the Swift forums, bug reports, and informal chats. But thus far, we have not collected and formally documented the reasons why using a package copy can encounter issues. This discussion recently came up during a [discussion](https://github.com/swiftlang/swift-syntax/pull/3192) about whether to adopt Swift Testing in a test-helper library in SwiftSyntax.

My hope is that adding this document will help clarify expectations around the level of support we offer for the two distribution styles for Swift Testing.

### Checklist:

### DIFF
--- a/Documentation/Distributions.md
+++ b/Documentation/Distributions.md
@@ -55,7 +55,7 @@ because it has several downsides:
   Swift Testing doesn't always declare a dependency on one of the prebuilt tags,
   circumventing this time-saver.) Additionally, the locally-built macro plugin
   and SwiftSyntax will be built for debug, without optimizations.
-* **It may not integrate as well with spporting tools/IDEs as a built-in copy.**
+* **It may not integrate as well with supporting tools/IDEs as a built-in copy.**
   Tools which integrate with Swift Testing such as Swift Package Manager or
   Apple's Xcode IDE often optimize for the copy included in the same
   distribution. Some features may not work as well or be missing entirely when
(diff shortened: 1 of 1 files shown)


Documentation/Distributions.md
```diff
@@ -0,0 +1,100 @@
...
+* **It requires building the Swift Testing runtime library.** This increases
+  your build time, and since builds for testing are typically for debug
+  configuration, it will not include performance optimizations.
+* **It requires building Swift Testing’s macro plugin.** This also increases
+  build time, especially because it often requires building [SwiftSyntax][] as
+  well. (SwiftSyntax now offers [prebuilt copies][prebuilt-swift-syntax], but
+  Swift Testing doesn't always declare a dependency on one of the prebuilt tags,
+  circumventing this time-saver.) Additionally, the locally-built macro plugin
+  and SwiftSyntax will be built for debug, without optimizations.
+* **It may not integrate as well with spporting tools/IDEs as a built-in copy.**
```
heckj:
```suggestion
* **It may not integrate as well with supporting tools/IDEs as a built-in copy.**
```

Documentation/Distributions.md
```diff
@@ -0,0 +1,100 @@
...
+  using Swift Testing as a package.
+* **It may encounter build failures when another package uses Swift
+  Testing.** If you use Swift Testing as a package, but you depend on a library
+  from another package which uses a built-in copy of Swift Testing (as this
+  document recommends), this can cause build failures:
+  * The other package may fail to build non-deterministically due to not having
+    a target dependency on the `Testing` target from the locally-built
+    [swift-testing][] package.
+  * On platforms which don't support a two-level linker namespace, it can fail
+    to link due to duplicate defintions for the symbols in the `Testing` library.
```
heckj:
```suggestion
    to link due to duplicate definitions for the symbols in the `Testing` library.
```

PR #1456, opened 2025-12-21
Correctly block `@Suite` and `@Test` on `XCTest.XCTest` and `XCTestSuite`.

grynspan:
This PR ensures that we diagnose if `@Suite` or `@Test` is applied to a class (or method on such class) that subclasses [`XCTest.XCTest`](https://developer.apple.com/documentation/xctest/xctest) or [`XCTestSuite`](https://developer.apple.com/documentation/xctest/xctestsuite), not just [`XCTestCase`](https://developer.apple.com/documentation/xctest/xctestcase).

This change also suppresses the emitted call to the helper function `__invokeXCTest[Case]Method()` if the suite type is _known_ not to subclass `XCTest.XCTest` (e.g. because it's a `struct`.) This slightly improves the performance of non-class suite types (although probably not measurably so).

Finally, this change improves the emitted diagnostic when `@Test` is used on a method in an `XCTest.XCTest` subclass such that it explicitly states the reason why the attribute can't be used:

> 🛑 Attribute 'Test' cannot be applied to a function within class 'C' because it is a subclass of 'XCTest', 'XCTestCase', or 'XCTestSuite'

### Checklist:

### DIFF
--- a/Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
+++ b/Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
@@ -502,14 +502,6 @@ struct TestDeclarationMacroTests {
     }
   }
 
-  @Test("Self. in @Test attribute is removed")
-  func removeSelfKeyword() throws {
-    let (output, _) = try parse("@Test(arguments: Self.nested.uniqueArgsName, NoTouching.thisOne) func f() {}")
-    #expect(output.contains("nested.uniqueArgsName"))
-    #expect(!output.contains("Self.nested.uniqueArgsName"))
-    #expect(output.contains("NoTouching.thisOne"))
-  }
-
   @Test("Display name is preserved",
     arguments: [
       #"@Test("Display Name") func f() {}"#,
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -18,7 +18,7 @@ on:
 jobs:
   create_merge_pr:
     name: Create PR to merge main into release branch
-    uses: swiftlang/github-workflows/.github/workflows/create_automerge_pr.yml@0.0.3
+    uses: swiftlang/github-workflows/.github/workflows/create_automerge_pr.yml@0.0.4
     with:
       head_branch: main
       base_branch: release/6.3
--- a/Package.swift
+++ b/Package.swift
@@ -369,9 +369,7 @@ extension Array where Element == PackageDescription.SwiftSetting {
   static var packageSettings: Self {
     var result = availabilityMacroSettings
 
-    if buildingForDevelopment {
-      result.append(.unsafeFlags(["-require-explicit-sendable"]))
-    }
+    result.append(.treatWarning("ExplicitSendable", as: .warning))
 
     if buildingForEmbedded {
       result.append(.enableExperimentalFeature("Embedded"))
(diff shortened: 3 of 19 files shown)


Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
```diff
@@ -440,10 +456,10 @@ struct TestDeclarationMacroTests {
       ("@Test @available(*, noasync) func f() {}", nil, "__requiringTry"),
       ("@Test @_unavailableFromAsync func f() {}", nil, "__requiringTry"),
       ("@Test(arguments: []) func f(f: () -> String) {}", "(() -> String).self", nil),
-      ("struct S {\n\t@Test func testF() {} }", nil, "__invokeXCTestCaseMethod"),
-      ("struct S {\n\t@Test func testF() throws {} }", nil, "__invokeXCTestCaseMethod"),
-      ("struct S {\n\t@Test func testF() async {} }", nil, "__invokeXCTestCaseMethod"),
-      ("struct S {\n\t@Test func testF() async throws {} }", nil, "__invokeXCTestCaseMethod"),
+      ("class S {\n\t@Test func testF() {} }", nil, "__invokeXCTestMethod"),
```
jerryjrchen, line 459: I don't totally understand how/why this is intended to be invoked as an XCTest method and probably need some more background on `__invokeXCTestMethod`. When I try the following, it seems to ignore the XCTest assertion:
```swift
class S {
    @Test func testF() {
        XCTAssertTrue(false) // Runs but passes, so we're still in Swift Testing "mode"
    }
}
```

Tests/TestingTests/NonCopyableSuiteTests.swift
```diff
@@ -31,3 +30,7 @@ struct NonCopyableTests: ~Copyable {
     #expect(TypeInfo(describing: Self.self).mangledName != nil)
   }
 }
+
+extension NonCopyableTests {
+  @Test borrowing func testNotAnXCTestCaseMethod() {}
```
jerryjrchen, line 35: This move isn't needed to make it compile, right? Just reorganising the tests?

Tests/TestingTests/TypeNameConflictTests.swift
jerryjrchen, line 14: Don't we need a test here named `testXXX...` to generate a test body with `__invokeXCTestMethod`?

grynspan: That's not what this particular suite is for. :) We have test coverage in several other files.

Tests/TestingTests/NonCopyableSuiteTests.swift
grynspan: This is intentional so we have a test in an extension and can't infer anything about the type from its name alone.

Not needed to make it _compile_, no.

Tests/TestingMacrosTests/TestDeclarationMacroTests.swift
grynspan: `__invokeXCTest[Case]Method()` is a theoretical hook point to let us run test functions in a mixed mode. In practice, we use it instead to generate issues at runtime if you manage to bypass our compile-time filters. (We'll probably want to go back and rename this function to something like `__checkForBadXCTestUsage()` if/when your interop work is done landing.)

PR #1462, opened 2025-12-27
Add `BitwiseCopyable` conformance to test discovery types.

grynspan:
This PR adds `BitwiseCopyable` conformance to those types in the `_TestDiscovery` target that are always POD, namely `SectionBounds`, `TestContentRecord._RecordStorage` and `TestContentKind`.

Explicitly specifying the protocol will cause the compiler to emit diagnostics if any of these types _stop_ being bitwise-copyable (which would not be intentional). Otherwise, these conformances are already inferred by the compiler.

### Checklist:

### DIFF
--- a/Sources/_TestDiscovery/TestContentRecord.swift
+++ b/Sources/_TestDiscovery/TestContentRecord.swift
@@ -64,7 +64,7 @@ extension DiscoverableAsTestContent {
 /// ``DiscoverableAsTestContent/allTestContentRecords()`` on a type that
 /// conforms to ``DiscoverableAsTestContent``.
 @_spi(Experimental) @_spi(ForToolsIntegrationOnly)
-public struct TestContentRecord<T>: BitwiseCopyable where T: DiscoverableAsTestContent {
+public struct TestContentRecord<T> where T: DiscoverableAsTestContent {
   /// The base address of the image containing this instance, if known.
   ///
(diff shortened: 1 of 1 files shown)


Sources/_TestDiscovery/TestContentKind.swift
```diff
@@ -32,7 +32,7 @@ private import _TestingInternals
 ///   in the compiler.
 /// }
 @_spi(Experimental) @_spi(ForToolsIntegrationOnly)
-@frozen public struct TestContentKind: Sendable, RawRepresentable {
+@frozen public struct TestContentKind: Sendable, BitwiseCopyable, RawRepresentable {
```
stmontgomery, line 35: In the PR description you mentioned `BitwiseCopyable` is inferred currently for these types, but this one is `public` so am I correct that it's _not_ inferred currently? It seems good either way, but just clarifying that this one is technically a change

grynspan: It's `@frozen` so it is still inferred.

jerryjrchen:
> This PR adds BitwiseCopyable conformance to those types in the _TestDiscovery target that are always POD

What does POD refer to?

grynspan:
"Plain Old Data". See https://en.cppreference.com/w/cpp/types/is_pod.html

PR #1463, opened 2025-12-27
Include the API level in the Android OS version.

grynspan:
When a test run starts, we emit some metadata including the OS version for diagnostic purposes. This PR adds Android's API level (SDK version) to the OS version info we emit on that platform. This is an Android-specific concept and is used in Swift for `{#,@}available(Android)` expressions and attributes.

### Checklist:


grynspan:
@finagolfin @compnerd FYI

grynspan:
Oh sure, it just saves a step. :)

PR #1464, opened 2025-12-30
Avoid hashtable overhead during Mach-O section discovery.

grynspan:
This PR removes a dictionary from the Mach-O section discovery path. Hashing has a non-zero cost vs. indexed lookups. (The expected performance win here is negligible, but this is low-hanging fruit.)

### Checklist:

### DIFF
--- a/Sources/_TestDiscovery/SectionBounds.swift
+++ b/Sources/_TestDiscovery/SectionBounds.swift
@@ -71,7 +71,15 @@ extension SectionBounds.Kind {
 
 /// An array containing all of the test content section bounds known to the
 /// testing library.
+///
+/// Indices into this array are equivalent to the `rawValue` values of instances
+/// of ``SectionBounds/Kind``.
 private nonisolated(unsafe) let _sectionBounds = {
+  // We generate a contiguous array here rather than a dictionary because the
+  // former has less overall bridging with the Objective-C runtime (reducing the
+  // risk of reentrance while holding the libobjc lock) and because the set of
+  // keys or indices is closed, so an array lookup is always more efficient than
+  // a hashtable lookup.
   let kindCount = SectionBounds.Kind.allCases.count
   let result = ManagedBuffer<ContiguousArray<ContiguousArray<SectionBounds>>, pthread_mutex_t>.create(
     minimumCapacity: 1,


Sources/_TestDiscovery/SectionBounds.swift
```diff
@@ -23,7 +23,7 @@ struct SectionBounds: Sendable {
 
   /// An enumeration describing the different sections discoverable by the
   /// testing library.
-  enum Kind: Equatable, Hashable, CaseIterable {
+  enum Kind: Int, Equatable, Hashable, CaseIterable {
     /// The test content metadata section.
     case testContent
```
jerryjrchen, line 28: For clarity since these will be used as indices now, how about explicitly specifying the first value?

```suggestion
    case testContent = 0
```

Sources/_TestDiscovery/SectionBounds.swift
```diff
@@ -72,17 +72,18 @@ extension SectionBounds.Kind {
 /// An array containing all of the test content section bounds known to the
 /// testing library.
 private nonisolated(unsafe) let _sectionBounds = {
-  let result = ManagedBuffer<[SectionBounds.Kind: [SectionBounds]], pthread_mutex_t>.create(
+  let kindCount = SectionBounds.Kind.allCases.count
+  let result = ManagedBuffer<ContiguousArray<ContiguousArray<SectionBounds>>, pthread_mutex_t>.create(
```
jerryjrchen, line 84: I kinda preferred the previous code, even if it was less performant. Comparing `[SectionBounds.Kind: [SectionBounds]]` to `ContiguousArray<ContiguousArray<SectionBounds>>`, the former better expresses through the type what to expect.

Maybe it would help to have a comment saying each index i of the top level array stores only one kind of SectionBounds?

jerryjrchen (approved):

Sources/_TestDiscovery/SectionBounds.swift
grynspan: They're equivalent, of course. Do you feel strongly?

Sources/_TestDiscovery/SectionBounds.swift
grynspan: I want to move us away from
`Dictionary` here because it has more bridging with Objective-C and thus more opportunities to explode in our faces if the implementation in the stdlib or Foundation changes and causes us to recurse back into libobjc when we use it.

I can certainly add a comment though.

stmontgomery: Agreed, a comment explaining the intended usage would be helpful

grynspan: Added!

Sources/_TestDiscovery/SectionBounds.swift
grynspan: Leaving as-is for now, but we can trivially add later.

jerryjrchen: I don't feel strongly either way. I just think it would benefit from being a little clearer that this is used as an index later.

PR #1467, opened 2026-01-04
Remove `_SelfRemover` as obsolete.

grynspan:
The `_SelfRemover` syntax rewriter is used during expansion of `@Test` and `@Suite` to rewrite `Self` as the name of the enclosing type so that the nested helper type we generate doesn't interfere with resolution of said keyword. Because the helper type now only contains `__testContentRecord` (with the bulk of the macro expansion being direct peers of the `@Test` or `@Suite` macro), this work is now redundant, so we can remove it.

See `TestsWithStaticMemberAccessBySelfKeyword` for existing test coverage.

### Checklist:


stmontgomery (approved):
Nice cleanup, thanks

PR #1468, opened 2026-01-04
Allow non-escapable types as suites.

grynspan:
This PR enables `@Suite` on non-escapable types and `@Test` on functions within them. We need to use the experimental `@_lifetime` attribute on some helper functions to ensure this works. Developers who want to write tests on `~Escapable` types will also need to enable the experimental "Lifetimes" feature.

### Checklist:

### DIFF
--- a/Sources/Testing/Test+Macro.swift
+++ b/Sources/Testing/Test+Macro.swift
@@ -53,9 +53,8 @@ public typealias __XCTestCompatibleSelector = Never
 /// - Parameters:
 ///   - traits: Zero or more traits to apply to this test suite.
 ///
-/// A test suite is a type that contains one or more test functions. Any
-/// escapable type (that is, any type that is not marked `~Escapable`) may be a
-/// test suite.
+/// A test suite is a type that contains one or more test functions. Any type
+/// may be a test suite.
 ///
 /// The use of the `@Suite` attribute is optional; types are recognized as test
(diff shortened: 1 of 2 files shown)


Sources/Testing/Test+Macro.swift
plemarquand: This is no longer accurate 🎉

Tests/TestingTests/NonCopyableSuiteTests.swift
stmontgomery, line 1: Filename could be generalized now

Package.swift
```diff
@@ -390,6 +390,9 @@ extension Array where Element == PackageDescription.SwiftSetting {
       // proposal via Swift Evolution.
       .enableExperimentalFeature("SymbolLinkageMarkers"),
 
+      // Enabled to allow tests to be added to ~Escapable suites.
+      .enableExperimentalFeature("Lifetimes"),
```
stmontgomery, line 394: Don't forget to add this to `CompilerSettings.cmake` too [P1]

Sources/Testing/Test+Macro.swift
grynspan: I forgot to delete this commeeeeeeent

Package.swift
grynspan: Right thanks :)

stmontgomery (approved):

PR #1470, opened 2026-01-05
Adopt `objectFormat()`.

### LINKED ISSUE #1370: Adopt `#if objectFormat()`

grynspan:
This PR adopts `objectFormat()` in test content emission, but not in discovery. Test content discovery will continue to use `#if os(...)` until swift-playgrounds' minimum toolchain requirement is 6.3 or newer.

Resolves #1370.
Resolves rdar://162878021.

### Checklist:

### DIFF
--- a/Documentation/Porting.md
+++ b/Documentation/Porting.md
@@ -205,18 +205,14 @@ platform if it does not use an image format already supported by Swift Testing:
 ```diff
 --- a/Sources/TestingMacros/Support/TestContentGeneration.swift
 +++ b/Sources/TestingMacros/Support/TestContentGeneration.swift
-   // ...
-+  #elseif objectFormat(CFM)
-+  @section(".rsrc,swft,__swift5_tests")
-   #else
-   @__testing(warning: "Platform-specific implementation missing: test content section name unavailable")
-   #endif
+   let objectFormatsAndSectionNames: [(objectFormat: String, sectionName: String)] = [
+     ("MachO", "__DATA_CONST,__swift5_tests"),
+     ("ELF", "swift5_tests"),
+     ("COFF", ".sw5test$B"),
+     ("Wasm", "swift5_tests"),
++    ("CFM", ".rsrc,swft,__swift5_tests"),
+   ]
 ```
-
-Keep in mind that this code is emitted by the `@Test` and `@Suite` macros
-directly into test authors' test targets, so you will not be able to use
-compiler conditionals defined in the Swift Testing package (including those that
-start with `"SWT_"`).
 
 ## Runtime test discovery with static linkage
 
--- a/Documentation/ABI/TestContent.md
+++ b/Documentation/ABI/TestContent.md
@@ -13,11 +13,6 @@ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 This document describes the format and location of test content that the testing
 library emits at compile time and can discover at runtime.
 
-> [!WARNING]
-> The content of this document is subject to change pending efforts to define a
-> Swift-wide standard mechanism for runtime metadata emission and discovery.
-> Treat the information in this document as experimental.
-
 ## Basic format
 
 Swift Testing stores test content records in a dedicated platform-specific


Sources/TestingMacros/Support/TestContentGeneration.swift
```diff
@@ -76,18 +78,44 @@ func makeTestContentRecordDecl(named name: TokenSyntax, in typeName: TypeSyntax?
...
-  @Testing.__testing(warning: "Platform-specific implementation missing: test content section name unavailable")
-  #endif
   @used
   \(result)
   """
+
+  let objectFormatsAndSectionNames: [(objectFormat: String, sectionName: String)] = [
+    ("MachO", "__DATA_CONST,__swift5_tests"),
+    ("ELF", "swift5_tests"),
+    ("COFF", ".sw5test$B"),
```
jerryjrchen, line 88: Do we need to specify PE here as well? All I know is that it is "extended from" COFF. Does that mean it's automatically included under the COFF object format?

Documentation/Porting.md
```diff
@@ -192,33 +198,22 @@ to load that information:
...
-+  @section(".rsrc,swft,__swift5_tests")
-   #else
-   @__testing(warning: "Platform-specific implementation missing: test content section name unavailable")
-   #endif
+   let objectFormatsAndSectionNames: [(objectFormat: String, sectionName: String)] = [
+     ("MachO", "__DATA_CONST,__swift5_tests"),
+     ("ELF", "swift5_tests"),
+     ("COFF", ".sw5test$B"),
+     ("Wasm", "swift5_tests"),
++    ("CFM", ".rsrc,swft,__swift5_tests"),
```
jerryjrchen, line 213: Is CFM just a fictitious object format provided as an example here? I can't find anything relevant online :P

Sources/TestingMacros/Support/TestContentGeneration.swift
```diff
@@ -76,18 +78,44 @@ func makeTestContentRecordDecl(named name: TokenSyntax, in typeName: TypeSyntax?
...
+      try buildConfiguration.isActiveTargetObjectFormat(name: objectFormat)
+    }
+    if let (_, sectionName) = objectFormatAndSectionName {
+      result = """
+      @section(\(literal: sectionName))
+      \(result)
+      """
+    } else {
+      result = """
+      @Testing.__testing(warning: "Platform-specific implementation missing: test content section name unavailable")
```
jerryjrchen, line 103: Is there any way to include the object format name here? Thinking from the perspective of a user, it might be easier to debug if we could say "Could not create test content section name for unrecognized object format 'XXX'".

Sources/TestingMacros/Support/TestContentGeneration.swift
```diff
@@ -76,18 +78,44 @@ func makeTestContentRecordDecl(named name: TokenSyntax, in typeName: TypeSyntax?
 
 #if compiler(>=6.3)
   result = """
```
jerryjrchen, line 80: What's the motivation for the extra logic in the section to try and only include a single `@section` declaration in the generated code? Is it to try and reduce the size of the generated code?

It seems to me the simpler solution would be replacing `#if os(...)` checks directly with `#if objectFormat(...)` here.

Sources/TestingMacros/Support/TestContentGeneration.swift
```diff
@@ -49,18 +50,19 @@ enum TestContentKind: UInt32 {
 ///   - kind: The kind of test content record being emitted.
 ///   - accessorName: The Swift name of an `@convention(c)` function to emit
 ///     into the resulting record.
-///   - context: A value to emit as the `context` field of the test content
-///     record.
+///   - contextFieldValue: A value to emit as the `context` field of the test
+///     content record.
+///   - context: The macro context in which the expression is being parsed.
```
jerryjrchen, line 55: Nit: maybe `macroContext` would be clearer?

Sources/TestingMacros/Support/TestContentGeneration.swift
grynspan: Yes, it produces cleaner macro output.

Sources/TestingMacros/Support/TestContentGeneration.swift
grynspan: No, we don't know what it is.

Documentation/Porting.md
grynspan: https://developer.apple.com/library/archive/documentation/mac/pdf/PPC_System_Software/Code_Frag_Mgr.pdf

Classic here is our example platform for porting purposes, and Classic used the Code Fragment Manager (well… it's _complicated_).

Sources/TestingMacros/Support/TestContentGeneration.swift
grynspan: No, Windows uses COFF as far as the Swift compiler is concerned.

Sources/TestingMacros/Support/TestContentGeneration.swift
grynspan: I got lazy copying and pasting. :)

grynspan:
As far as I can tell, the Xcode-based builds are hanging right now for external reasons. Merging.

PR #1471, opened 2026-01-05
Avoid runtime collisions when multiple copies of the testing library are loaded

stmontgomery:
This PR avoids runtime collisions when multiple copies of the testing library are loaded into a given process by introducing the ability to customize the module ABI name of each copy.

### Motivation:

In some usage scenarios there can be two or more copies of the testing library loaded into a runner process. When this happens on platforms such as Darwin which use the Objective-C runtime, this can cause duplicate class definition warnings to be logged to the console because (non-generic) class and actor types are implemented in the Objective-C runtime as classes. For a recent example where this occurred and was worked around, see #1444. Although we have not seen practical reports of this, a similar problem could arise at build/link time on some platforms due to exported symbols from the testing library having multiple definitions.

Resolves rdar://148912491

### Modifications:

This PR specifies a unique module ABI name for each public library module by passing `-module-abi-name <name>` to the Swift compiler. It does this differently depending on the build context:

- When building as a Swift package, this PR appends a hardcoded suffix (`_package`) to the name of each module to form its ABI name. It does this by default for package builds because package copies of the testing library are virtually always loaded by runner processes which also load a prebuilt copy of the testing library, meaning that a package copy is most often considered a "secondary" copy which needs to avoid runtime conflicts with a primary copy.
- When building via CMake, this PR adds the _ability_ to define a custom ABI name suffix, and this allows the client performing the build to control the ABI name. In particular, this allows a build intended for inclusion in an official Swift toolchain to specify a suffix, but allows custom CMake builds to remain unsuffixed.

### Checklist:

### DIFF
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -77,7 +77,7 @@ extension Runner {
   /// type at runtime, it may be better-suited for ``Configuration`` instead.
   private struct _Context: Sendable {
     /// A serializer used to reduce parallelism among test cases.
-    var testCaseSerializer: Serializer<Void>?
+    var testCaseSerializer: Serializer?
   }
 
   /// Apply the custom scope for any test scope providers of the traits
--- a/Sources/Testing/Support/Serializer.swift
+++ b/Sources/Testing/Support/Serializer.swift
@@ -42,13 +42,8 @@ var defaultParallelizationWidth: Int {
 /// items do not start running; they must wait until the suspended work item
 /// either returns or throws an error.
 ///
-/// The generic type parameter `T` is unused. It avoids warnings when multiple
-/// copies of the testing library are loaded into a runner process on platforms
-/// which use the Objective-C runtime, due to non-generic actor types being
-/// implemented as classes there.
-///
 /// This type is not part of the public interface of the testing library.
-final actor Serializer<T> {
+final actor Serializer {
   /// The maximum number of work items that may run concurrently.
   nonisolated let maximumWidth: Int
 


Package.swift
```diff
@@ -189,21 +189,15 @@ let package = Package(
...
         // Testing module. This target's module is never distributed to users,
         // but as an additional guard against accidental misuse, this specifies
         // the unit test target as the only allowable client.
-        if buildingForDevelopment {
-          result.append(.unsafeFlags(["-Xfrontend", "-allowable-client", "-Xfrontend", "TestingMacrosTests"]))
-        }
-
-        return result
-      }()
+        .unsafeFlags(["-Xfrontend", "-allowable-client", "-Xfrontend", "TestingMacrosTests"]),
```
grynspan, line 199: Is it safe to use `.unsafeFlags()` with the 6.2 toolchain?

cmake/modules/shared/ModuleABIName.cmake
```diff
@@ -0,0 +1,27 @@
...
+# `SwiftTesting_MODULE_ABI_NAME_SUFFIX`. By default, no suffix is included but
+# builds of the testing library intended for a toolchain can specify a suffix
+# so that its symbols will not conflict with either a copy built by a client as
+# a Swift package or a precompiled vendor copy.
+#
+# When determining the base module ABI name to specify, this prefers the
+# `Swift_MODULE_NAME` property of the target if it's set, but since that's
+# uncommon, it uses its `NAME` property by default.
+add_compile_options(
+  "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-module-abi-name $<IF:$<BOOL:$<TARGET_PROPERTY:Swift_MODULE_NAME>>,$<TARGET_PROPERTY:Swift_MODULE_NAME>,$<TARGET_PROPERTY:NAME>>${SwiftTesting_MODULE_ABI_NAME_SUFFIX}>")
```
grynspan, line 27: I'm unsure if any changes to CMake are needed here? We only have the one workflow that uses it.

Tests/TestingTests/TestSupport/TestingAdditions.swift
```diff
@@ -18,6 +18,13 @@ import XCTest
 import Foundation
 #endif
 
+/// The ABI name of the testing library's main module.
+///
+/// This can be different than the target name ("Testing") whenever the module
+/// ABI name is customized at build time, as it is in certain deployment
+/// contexts.
+let testingModuleABIName = TypeInfo(describing: Test.self).fullyQualifiedNameComponents.first!
```
grynspan, line 26: Is the custom name present in demangled FQNs? That's not great.

Package.swift
stmontgomery: Yes! I confirmed this locally

cmake/modules/shared/ModuleABIName.cmake
stmontgomery: The CMake changes are needed primarily so that the copy included in Swift.org toolchains for Darwin can be given a unique suffix, and that avoids conflicts with the Xcode copy. As the comment in this file and the PR description describe, this suffix is optional and off by default for CMake builds.

Even if you ignore the Xcode case though, this is a generally useful capability: it allows clients who build via CMake to supply their own custom suffix if they need or want to.

Tests/TestingTests/TestSupport/TestingAdditions.swift
stmontgomery: Yeah, and that's a symptom of a larger, pre-existing issue: that whenever multiple copies of the testing library are loaded into a process, they aren't separated from each other very clearly and can cause conflicts. I actually think it could wind up being helpful to us when troubleshooting issues for these to be distinguished in demangled symbol names.

The end result of the changes in this and a [related PR I've posted](https://github.com/swiftlang/swift/pull/86305) will be that the "canonical" copy of the testing library in each situation will maintain the status quo: an unsuffixed module ABI name. Only optional, secondary copies will have a distinct name.

stmontgomery:
I confirmed via a CI toolchain build (in https://github.com/swiftlang/swift/pull/86305) that the CMake changes have the intended effect (at least for macOS which is the only platform made available to download and inspect).

Tests/TestingTests/TestSupport/TestingAdditions.swift
grynspan: If a developer wants to write `Testing.Test`, that's still valid in source, right? This is invisible to the Swift side at compile time?

stmontgomery: Correct, there's no change at compile time

grynspan:
I want to bikeshed a bit on the way we adjust the module name, but just a little bit. :)

PR #1475, opened 2026-01-07
Add revision notes to JSON.md.

grynspan:
This PR adds a section to JSON.md listing Swift Evolution proposals that have modified the JSON schema.

### Checklist:

### DIFF
--- a/Documentation/ABI/JSON.md
+++ b/Documentation/ABI/JSON.md
@@ -219,3 +219,14 @@ sufficient information to display the event in a human-readable format.
   ["testID": <test-id>,
     ["testCase": <test-case>]]
 -->
+
+## See Also
+
+### Relevant Swift Evolution proposals
+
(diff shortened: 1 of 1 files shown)


Documentation/ABI/JSON.md
```diff
@@ -219,3 +219,14 @@ sufficient information to display the event in a human-readable format.
   ["testID": <test-id>,
     ["testCase": <test-case>]]
 -->
+
+## See Also
```
jerryjrchen, line 223: (very nitpicky) Would it make sense to just promote the H3 header below to an H2 given that this "see also" section only has one entry?

And instead of "Relevant Swift Evolution proposals", I think "Revision notes" as stated in the PR title is actually a little clearer.

grynspan: I will probably add more content to this section in the future. :)

PR #1478, opened 2026-01-12
Need to initialize os_unfair_lock guarding the fallback event handler

jerryjrchen:
Prior to this fix, we didn't have an explicit init which could lead to garbage values.

Resolves rdar://167861118

### Checklist:

### DIFF
--- a/Sources/_TestingInterop/FallbackEventHandler.swift
+++ b/Sources/_TestingInterop/FallbackEventHandler.swift
@@ -22,7 +22,9 @@ private nonisolated(unsafe) let _fallbackEventHandler = {
     minimumCapacity: 1,
     makingHeaderWith: { _ in nil }
   )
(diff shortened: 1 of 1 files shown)


Sources/_TestingInterop/FallbackEventHandler.swift
```diff
@@ -22,7 +22,9 @@ private nonisolated(unsafe) let _fallbackEventHandler = {
     minimumCapacity: 1,
     makingHeaderWith: { _ in nil }
   )
-  result.withUnsafeMutablePointerToHeader { $0.initialize(to: nil) }
```
jerryjrchen, line 25: I think the original header init here was redundant since `    makingHeaderWith: { _ in nil }` should already fill that in as nil.

grynspan: Correct, it should have always been initializing the elements.

PR #1483, opened 2026-01-14
Adjust compiler version checks around `WinSDK.GUID.Wrapper`.

grynspan:
This PR ensures we don't start spuriously warning if Swift Testing is built as a package with the (eventual/hypothetical) Swift 6.3.1 toolchain, and uses the built-in `Equatable` and `Hashable` conformances on `GUID` when built with 6.3.0 (because those conformances have landed everywhere now.)

### Checklist:


grynspan:
Paging @compnerd to the courtesy phone.

compnerd (approved):
I like this approach for encouraging the removal of old types! I will have to remember it.

PR #1492, opened 2026-01-21
Extend `CustomTestStringConvertible` to `~Copyable` and `~Escapable` values.

grynspan:
This PR extends `CustomTestStringConvertible` to cover values with suppressed conformances, and extends string interpolation for `Comment` similarly.

See also [SE-0499](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0499-support-non-copyable-simple-protocols.md) for where `CustomStringConvertible` was extended in this fashion.

### Checklist:


grynspan:
Needs adjustments to work with the various string protocols.

PR #1493, opened 2026-01-21
Prepare for potential nonisolated(nonsending) source break in stdlib

ktoso:
We need to move all async "with..." methods in stdlib to nonisolated(nonsending), this turns out can cause source breakage...

These changes would have to be made in swift-testing for example 

See: https://github.com/swiftlang/swift/pull/80753


ktoso:
@swift-ci please smoke test

grynspan:
We use GHA in this repo now.

PR #1495, opened 2026-01-21
`Self` is broken when used within `@Suite` (e.g. in traits).

grynspan:
This PR fixes a regression caused by #1467 where the `Self` token in a `@Suite` attribute is resolved to the attached type, not the containing type. For example:

```swift
struct Outer {
  @EnvironmentVariable("TERM") var terminalName: String?

  @Suite(.enabled(if: Self.terminalName == "xterm"))
  struct Inner { /*...*/ }
}
```

Without this fix, the above fails to compile with:

> 🛑 Type 'Outer.Inner' has no member 'terminalName'

Where it previously succeeded.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/SuiteDeclarationMacro.swift
+++ b/Sources/TestingMacros/SuiteDeclarationMacro.swift
@@ -20,15 +20,6 @@ public import SwiftSyntaxMacros
 public struct SuiteDeclarationMacro: PeerMacro, Sendable {
   public static func expansion(
     of node: AttributeSyntax,
-    providingMembersOf declaration: some DeclGroupSyntax,
-    conformingTo protocols: [TypeSyntax],
-    in context: some MacroExpansionContext
-  ) throws -> [DeclSyntax] {
-    []
-  }
-
-  public static func expansion(
-    of node: AttributeSyntax,
     providingPeersOf declaration: some DeclSyntaxProtocol,
     in context: some MacroExpansionContext
   ) throws -> [DeclSyntax] {


Sources/Testing/Test+Macro.swift
```diff
@@ -66,7 +66,7 @@ public typealias __XCTestCompatibleSelector = Never
 /// ## See Also
 ///
 /// - <doc:OrganizingTests>
-@attached(member) @attached(peer)
+@attached(peer)
```
grynspan, line 69: This has no effect on ABI.

Sources/TestingMacros/SuiteDeclarationMacro.swift
```diff
@@ -17,30 +17,26 @@ public import SwiftSyntaxMacros
...
 /// This type is used to implement the `@Suite` attribute macro. Do not use it
 /// directly.
-public struct SuiteDeclarationMacro: MemberMacro, PeerMacro, Sendable {
+public struct SuiteDeclarationMacro: PeerMacro, Sendable {
   public static func expansion(
     of node: AttributeSyntax,
     providingMembersOf declaration: some DeclGroupSyntax,
     conformingTo protocols: [TypeSyntax],
     in context: some MacroExpansionContext
   ) throws -> [DeclSyntax] {
```
stmontgomery: If we're removing the `MemberMacro` conformance, why do we still need this method?

Sources/TestingMacros/SuiteDeclarationMacro.swift
```diff
@@ -17,32 +17,19 @@ public import SwiftSyntaxMacros
 ///
 /// This type is used to implement the `@Suite` attribute macro. Do not use it
 /// directly.
-public struct SuiteDeclarationMacro: MemberMacro, PeerMacro, Sendable {
+public struct SuiteDeclarationMacro: PeerMacro, Sendable {
   public static func expansion(
     of node: AttributeSyntax,
-    providingMembersOf declaration: some DeclGroupSyntax,
```
grynspan, line 23: The diff here is confused: I removed the first overload of `expansion(of:...)`. This is ABI-safe since the ABI for macro plugins does not need to be stable at runtime.

Sources/TestingMacros/SuiteDeclarationMacro.swift
grynspan: Removed.

stmontgomery (approved):

PR #1498, opened 2026-01-23
Add the ability to attribute issues to tests even when run in detached tasks.

grynspan:
Let's say you have code like this:

```swift
@Test func foo() async {
  await Task.detached {
    let x = ...
    #expect(x.y == z)
  }.value
}
```

(Swap in `DispatchQueue.async()`, `Thread.detachNewThread()`, etc. as desired.)

Today, if the expectation in the middle there fails, we have no way to attribute it back to the test function `foo()`.

This PR modifies the `@Test` macro so that it captures the full source range of a test function (not just its starting source location). We do so using a new structure called `__SourceBounds` that is functionally equivalent to `Range<SourceLocation>` but which avoids redundant string storage and comparisons for `fileID` and `filePath`.

This PR also modifies `Runner` so that when it starts running tests, it stores references to them in a global (lock-guarded) location. This storage is always sorted by the tests' source locations (i.e. the lower bound of the range).

Then, when an expectation like the one above fails, we can inspect its source location (which is always captured by the expansion of the `#expect()` macro) and compare it against the source bounds of every test reported by `Runner`. If it lines up with a known test function, we can associate the issue with it.

The new `__SourceBounds` type isn't API at this stage since it's only consumed, by Swift Testing itself, although we could expose it as API if needed.

This change has no effect for issues without source location information (such as unhandled thrown errors).

## Memory usage

Because we're now keeping a larger number of `Test` instances in memory, I took this opportunity to move `Test`'s stored property into a class to implement copy-on-write semantics. I didn't go so far as to use [`isKnownUniquelyReferenced(_:)`](https://developer.apple.com/documentation/swift/isknownuniquelyreferenced(_:)-5kvtu) because that incurs additional non-zero overhead and my expectation is that mutations of instances of `Test` are relatively rare after planning concludes.

The immediate size of the `Test` structure has been reduced from 201 bytes per instance to 4 or 8 per instance (depending on target pointer width).

This change lowered Swift Testing's own peak memory footprint (per `vmmam` on macOS) from about 478MB before this PR to 435MB after, a savings of 44MB or about 9%[^ymmv] _despite increasing the number of `Test` instances stored in memory_. Not bad!

[^ymmv]: The exact amount of memory used when testing Swift Testing varies significantly. Your mileage may vary.

Resolves rdar://169036231.

### Checklist:

### DIFF
--- a/Tests/TestingTests/IssueTests.swift
+++ b/Tests/TestingTests/IssueTests.swift
@@ -1045,15 +1045,27 @@ final class IssueTests: XCTestCase {
   }
 
 #if !SWT_NO_UNSTRUCTURED_TASKS
+  // Positioned outside the bounds of the test function. Checks that source
+  // location info can be successfully propagated to a callee and recovered from
+  // the issue.
+  @Sendable private static func recordIssue(sourceLocation: SourceLocation) {
+    _ = Issue.record(sourceLocation: sourceLocation)
+  }
+
   func testFailWithoutCurrentTest() async throws {
+    let issueRecorded = expectation(description: "Issue recorded")
+    issueRecorded.expectedFulfillmentCount = 2
+
     let lowerBound = #_sourceLocation
+    let upperBound = Self.testFailWithoutCurrentTestEnd
     let sourceBounds = __SourceBounds(
       __uncheckedLowerBound: lowerBound,
-      upperBound: (lowerBound.line + 10, lowerBound.column) // ballpark number
+      upperBound: (upperBound.line, upperBound.column)
     )
     let test = Test(sourceBounds: sourceBounds) {
       await Task.detached {
         _ = Issue.record()
+        Self.recordIssue(sourceLocation: #_sourceLocation)
       }.value
     }
 
@@ -1065,10 +1077,14 @@ final class IssueTests: XCTestCase {
       XCTAssertFalse(issue.isKnown)
       XCTAssertNotNil(event.testID)
       XCTAssertEqual(event.testID, test.id)
+      issueRecorded.fulfill()
     }
 
     await test.run(configuration: configuration)
+
+    await fulfillment(of: [issueRecorded], timeout: 0.0)
   }
+  private static let testFailWithoutCurrentTestEnd = #_sourceLocation
 
   func testFailWithoutCurrentTestAndNoSourceLocation() async throws {
     var configuration = Configuration()
@@ -1080,13 +1096,13 @@ final class IssueTests: XCTestCase {
       XCTAssertNil(event.testID)
     }
 
-    @Sendable func body() {
+    @Sendable func helper() {
       _ = Issue.record()
     }
 
     await Test {
       await Task.detached {
-        body()
+        helper()
       }.value
     }.run(configuration: configuration)
   }


Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -260,3 +260,69 @@ func currentTestAndTestCase() -> (Test?, Test.Case?) {
   }
   return (state.test, state.testCase)
 }
+
+// MARK: - Finding running tests by source location
+
+extension Runner {
+  /// Storage for ``scheduledTests``.
+  private static let _scheduledTests = Locked<[Test]>()
```
grynspan, line 268: I'm not terribly happy with how I'm storing tests here. I'd like to use the same basic mechanism as for `Configuration` where we add/remove test arrays. But for users calling `swift test` or using Xcode, there's no practical difference.

Sources/Testing/Test.swift
```diff
@@ -23,20 +23,86 @@ public import ObjectiveC
...
+  /// - Parameters:
+  ///   - newValue: The new value to assign.
+  ///   - keyPath: The key path to modify.
+  private mutating func _setValue<V>(_ newValue: consuming V, forKeyPath keyPath: WritableKeyPath<_Storage.Properties, V>) {
+    var properties = _storage.properties
+    properties[keyPath: keyPath] = newValue
+#if DEBUG
+    properties.mutationCount += 1
+#endif
+    _storage = _Storage(properties)
```
harlanhaskins, line 72: Part of me thinks it would be worth implementing COW and using @unchecked Sendable here (or at least putting Properties in a Mutex or something) to avoid this unboxing and reboxing for all property accesses, but also it seems unlikely that these values would change much during execution and as such it is something we can do later

Sources/Testing/Test.swift
```diff
@@ -47,26 +113,61 @@ public struct Test: Sendable {
         }
       }
       precondition(traitsAreCorrectlyTyped(), "Programmatically added an inapplicable trait to test \(self)")
+      _setValue(newValue, forKeyPath: \.traits)
     }
   }
 
   /// The source location of this test.
-  public var sourceLocation: SourceLocation
+  public var sourceLocation: SourceLocation {
```
harlanhaskins, line 121: It's a little strange to me that this is a lossy value; that if you've set sourceBounds before and you set sourceLocation, you lose the upperBound. Does that seem like it would be an issue in practice?

Sources/TestingMacros/Support/SourceLocationGeneration.swift
```diff
@@ -39,3 +39,13 @@ func createSourceLocationExpr(of expr: some SyntaxProtocol, context: some MacroE
...
 }
+
+func createSourceBoundsExpr(from lowerBoundNode: some SyntaxProtocol, to upperBoundNode: some SyntaxProtocol, context: some MacroExpansionContext) -> ExprSyntax {
+  let lowerBoundExpr = createSourceLocationExpr(of: lowerBoundNode, context: context)
+  let upperBoundExpr: ExprSyntax = if let upperBoundSourceLoc = context.location(of: upperBoundNode, at: .beforeTrailingTrivia, filePathMode: .fileID) {
+    "(\(upperBoundSourceLoc.line), \(upperBoundSourceLoc.column))"
+  } else {
+    "(.max, .max)"
+  }
+  return "Testing.__SourceBounds(__uncheckedLowerBound: \(lowerBoundExpr), upperBound: \(upperBoundExpr))"
```
harlanhaskins, line 68: Is `(.max, .max)` a good default to use here? Would we not instead want to use the `__SourceBounds(lowerBoundOnly:)` initializer?

Sources/Testing/Test.swift
grynspan: My assertion (heh) here is that the overhead required for proper COW support (i.e. a lock) exceeds that of just creating a new object as it's rare we modify a `Test` instance after creating it, but reading properties is very common.

Sources/Testing/Test.swift
grynspan: No, because the `sourceBounds` property isn't public. If we decide to expose it, we would likely want to store it separately from `sourceLocation` for that reason, but right now it's redundant information so we're saving `Int+Int+String+String` bytes per instance this way.

Sources/TestingMacros/Support/SourceLocationGeneration.swift
grynspan: We'll only hit the `else` clause when running our own macro expansion tests, never when running actual production code, so it's fundamentally shorthand for "it doesn't matter". Does that make sense?

harlanhaskins: Yep, makes total sense to me.

Sources/Testing/Events/Event.swift
```diff
@@ -245,7 +245,18 @@ public struct Event: Sendable {
...
     // reset it to the actual configuration that handles the event when we call
     // handleEvent() later, so there's no need to make a copy of it yet.
-    let (test, testCase) = testAndTestCase
+    var (test, testCase) = testAndTestCase
+    if test == nil && testCase == nil,
+       case let .issueRecorded(issue) = kind,
+       let issueSourceLocation = issue.sourceLocation {
+      // There was no test on the current task, but an issue was recorded. Check
+      // if its source location lines up with the bounds of any known test and
+      // attribute it to that test (but no test case) if so.
```
jerryjrchen: I don't get this part of the comment:
> (but no test case)

Aren't we assigning `testCase = test.testCases?.first { _ in true }` below for non parameterized tests?

Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -260,3 +260,74 @@ func currentTestAndTestCase() -> (Test?, Test.Case?) {
...
+
+  /// Report a set of tests that some instance of ``Runner`` will run.
+  ///
+  /// - Parameters:
+  ///   - testsToSchedule: The set of tests that the caller is about to run. If
+  ///     a test is passed to this function more than once (as identified by its
+  ///     ``Test/id`` property), it is only tracked once.
+  ///
+  /// To get a list of all tests that have been passed to this function, get the
+  /// value of the ``testsThatHaveRun`` property.
```
jerryjrchen: Should this be referring to `scheduledTests`?

Sources/Testing/SourceAttribution/SourceBounds.swift
```diff
@@ -0,0 +1,103 @@
...
+    )
+  }
+}
+
+// MARK: -
+
+@_spi(Experimental)
+extension __SourceBounds {
+  public init(lowerBound: SourceLocation, upperBound: SourceLocation) {
+#if DEBUG
```
jerryjrchen: Could we just use `assert` instead of `precondition` here to avoid the checks outside of debug builds?

Sources/Testing/SourceAttribution/SourceBounds.swift
```diff
@@ -0,0 +1,103 @@
...
+  }
+  
+  public func contains(_ element: SourceLocation) -> Bool {
+    // This function can also be implemented more simply as:
+    //
+    // ```swift
+    // lowerBound <= element && element < _upperBound
+    // ```
+    //
+    // However that implementation produces extra redundant string comparisons.
```
jerryjrchen: Is this premature optimisation, or are extra string comparisons really that bad? :P

I figure that this code path will only be hit for issues created without an associated test, so it should also be relatively rare (?) in the wild. In return, it would be easier to verify that the simple yet less performant implementation is correct.

Sources/Testing/SourceAttribution/SourceBounds.swift
```diff
@@ -0,0 +1,103 @@
...
+
+// MARK: - RangeExpression
+
+@_spi(Experimental)
+extension __SourceBounds: RangeExpression {
+  public func relative<C>(to collection: C) -> Range<SourceLocation> where C : Collection, SourceLocation == C.Index {
+    // I'm honestly not sure how I'm supposed to implement this function, but
+    // this matches what Range<T> does (and both Range<T> and this type
+    // represent half-open ranges, so it must be right, right?)
+    lowerBound ..< upperBound
```
jerryjrchen, line 59: Seems right to me since `__SourceBounds` endpoints are fixed, so the relative representation is the same. A partial range has the endpoints depend on the collection bounds.

Sources/Testing/Test.swift
```diff
@@ -47,26 +113,61 @@ public struct Test: Sendable {
...
+    }
+    set {
+      sourceBounds = __SourceBounds(lowerBoundOnly: newValue)
+    }
+  }
+
+  /// The source bounds of this test.
+  ///
+  /// If this instance represents a test suite, the value of this property is
+  /// `nil` because test suites can span multiple source files.
```
stmontgomery: This documentation says the value could be `nil` but it's not an `Optional`

Sources/Testing/Test.swift
```diff
@@ -47,26 +113,61 @@ public struct Test: Sendable {
...
+  }
 
   /// The XCTest-compatible Objective-C selector corresponding to this
   /// instance's underlying test function.
   ///
   /// On platforms that do not support Objective-C interop, the value of this
   /// property is always `nil`.
   @_spi(ForToolsIntegrationOnly)
-  public nonisolated(unsafe) var xcTestCompatibleSelector: __XCTestCompatibleSelector?
+  public var xcTestCompatibleSelector: __XCTestCompatibleSelector? {
```
stmontgomery, line 160: I'm wondering why the `_Storage.Properties` struct doesn't need `nonisolated(unsafe)` on its `xcTestCompatibleSelector` stored property like this one needed?

Sources/TestingMacros/Support/SourceLocationGeneration.swift
```diff
@@ -39,3 +39,13 @@ func createSourceLocationExpr(of expr: some SyntaxProtocol, context: some MacroE
 
   return "Testing.SourceLocation(fileID: \(fileIDSourceLoc.file), filePath: \(filePathSourceLoc.file), line: \(fileIDSourceLoc.line), column: \(fileIDSourceLoc.column))"
 }
+
+func createSourceBoundsExpr(from lowerBoundNode: some SyntaxProtocol, to upperBoundNode: some SyntaxProtocol, context: some MacroExpansionContext) -> ExprSyntax {
```
stmontgomery: nit: documentation

Sources/Testing/Running/Runner.RuntimeState.swift
stmontgomery: This may not be the aspect you're referring to, but I notice the logic below only ever _adds_ to the array of known "scheduled" tests; it's an only-increasing data structure, and it doesn't appear to remove the batch of tests which finished after `Runner` finishes running them.

Even though it may not matter a lot in production testing workflows, that part bothers me some on an intellectual level :) Comparing to `Configuration`, we do clean up the stored `all` property (via `_removeFromAll(identifiedBy:)`) so it seems like we ought to do the same thing here.

grynspan: That's a bit hard to solve. In the common/external case, it's moot because a runner runs once, but in our own tests, removing sets is destructive. Each test in the set must be unique (or rather, have source bounds disjoint with all other source bounds). That means that if `Runner` A adds `Test` B, then `Runner` C adds another copy of `Test` B, we need to dedup the two copies of `B`. Then, if `Runner` C tears down and removes the tests it added, `Test` B disappears from the set.

So for the moment I'm inclined to leave it as-is and revisit in a follow-up when I've had more time to consider the optimal implementation of a `remove()` function.

Sources/Testing/Events/Event.swift
grynspan: The comment is unclear&hellip;I'm specifically referring to parameterized tests there. In the case of a non-parameterized test, it only ever has one test case so it's "safe" to assume that the issue belongs to that test case. I hope that makes sense!

grynspan: (I'll revise the comment.)

Sources/Testing/Running/Runner.RuntimeState.swift
grynspan: Yep, old name from a draft :) Will fix, thanks!

Sources/Testing/SourceAttribution/SourceBounds.swift
grynspan: It's subtle, but I prefer to use `precondition()` in code because it specifically checks for the violation of a precondition rather than a logic error inside the function.

A complete affectation, I know.

Sources/Testing/SourceAttribution/SourceBounds.swift
grynspan: I am anticipating using this structure in other places and file paths can be arbitrarily long, so I don't want to tempt fate. But this _is_ currently a cold path, so if you feel strongly I can switch implementations until it becomes a hot(ter) path.

Sources/Testing/Test.swift
grynspan: Will fix.

Sources/Testing/Test.swift
grynspan: We actually don't need it on the old property either. There was a bug in Apple's SDKs a few years back where `Selector` was marked non-`Sendable` by mistake. The bug's been resolved for a while now, so `nonisolated(unsafe)` here is just tech debt.

Sources/Testing/SourceAttribution/SourceBounds.swift
jerryjrchen: Looks fine tbh, and it's a fairly simple implementation so I'm not too worried.

PR #1500, opened 2026-01-26
Adopt `Span`/`RawSpan` almost everywhere.

grynspan:
This PR adopts `Span` and `RawSpan` throughout Swift Testing. Unsafe buffer pointers and the like are replaced with spans wherever possible. A few carve-outs:

- `Attachment`/`Attachable` are untouched for now as we need to think through the correct API shape for them.
- Code that interacts with C functions (including code that works with C strings) generally also must still use unsafe pointers, but the scope of those pointers is generally constrained to some local context.
- The ABI entry point function and test content section must continue to use unsafe pointers as they must remain broadly compatible with C.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/ArrayAdditions.swift
+++ b/Sources/Testing/Support/Additions/ArrayAdditions.swift
@@ -100,10 +100,6 @@ extension String.UTF8View {
   ///
   /// This property is equivalent to the `span` property in the Swift standard
   /// library, but is available on earlier Apple platforms.
-  ///
-  /// For strings with native UTF-8 storage, getting the value of this property
-  /// is an _O_(1) operation. For strings with non-native or non-UTF-8 storage
-  /// (i.e. bridged from Objective-C), the operation may be up to _O_(_n_).
   var span: Span<Element> {
     @_lifetime(borrow self)
     _read {


Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -873,7 +871,7 @@ extension ExitTest {
       // Insert a specific variable that tells the child process which exit test
       // to run.
       try JSON.withEncoding(of: exitTest.id) { json in
-        childEnvironment[Self._idEnvironmentVariableName] = String(decoding: json, as: UTF8.self)
+        childEnvironment[Self._idEnvironmentVariableName] = String(decoding: Array(json), as: UTF8.self)
```
harlanhaskins, line 871: This one seems a little unfortunate, does `RawSpan` not conform to `Sequence<UInt8>`?

harlanhaskins: Oh right, we don't have BorrowingSequence yet, so it can't.

Sources/Testing/Support/Additions/ArrayAdditions.swift
```diff
@@ -67,6 +67,48 @@ extension Array {
...
+extension String.UTF8View {
+  /// A raw span representing this string as UTF-8, not including a trailing
+  /// null character.
+  ///
+  /// This property is equivalent to the `span` property in the Swift standard
+  /// library, but is available on earlier Apple platforms.
+  var span: Span<Element> {
+    @_lifetime(borrow self)
+    _read {
+      // This implementation incurs a copy even for native Swift strings. This
```
harlanhaskins, line 106: Kinda unfortunate that we have to pay the perf penalty even if the `span` property is available, but I am sure we could figure out a way to use the proper one without an onerous availability check if this becomes a hot path

Sources/Testing/Support/FileHandle.swift
```diff
@@ -376,57 +376,26 @@ extension FileHandle {
...
+      let countWritten = bytes.withUnsafeBytes { bytes in
+        fwrite(bytes.baseAddress!, MemoryLayout<UInt8>.stride, bytes.count, file)
+      }
+      if countWritten < bytes.byteCount {
         throw CError(rawValue: swt_errno())
       }
     }
   }
 
-  /// Write a sequence of bytes to this file handle.
```
harlanhaskins, line 394: _very_ nice to see all these overloads consolidated.

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
```diff
@@ -110,7 +110,9 @@ extension Test.Case.Argument.ID {
   ///
   /// - Throws: Any error encountered during encoding.
   private static func _encode(_ value: some Encodable, parameter: Test.Parameter) throws -> [UInt8] {
-    try JSON.withEncoding(of: value, userInfo: [._testParameterUserInfoKey: parameter], Array.init)
+    try JSON.withEncoding(of: value, userInfo: [._testParameterUserInfoKey: parameter]) { json in
+      Array(json)
```
stmontgomery: Did this code need to change, strictly speaking? Not opposed, just curious why it wouldn't have worked as written

harlanhaskins (approved):
Looks great overall. Really nice to see how many nested `withUnsafeX` or `withUTF8` calls can just use non escaping accessors.

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -1007,9 +1005,7 @@ extension ExitTest {
 
     for recordJSON in bytes.split(whereSeparator: \.isASCIINewline) where !recordJSON.isEmpty {
       do {
-        try recordJSON.withUnsafeBufferPointer { recordJSON in
-          try Self._processRecord(.init(recordJSON), fromBackChannel: backChannel)
-        }
+        return try Self._processRecord(recordJSON.span.bytes, fromBackChannel: backChannel)
```
stmontgomery: Why does this now have an explicit `return`, whereas the outer `try ...` didn't before?

grynspan: No practical reason, just what I typed as I was translating things. I can remove it if it bothers you?

Sources/Testing/Support/Additions/ArrayAdditions.swift
grynspan: No way without the availability check, and the semantics of the accessors differ (we need a `_read` because of the copy, while the "real" property is `borrowing get`.)

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
grynspan: Let's find out!

Sources/Testing/ExitTests/ExitTest.swift
grynspan: Oh yeah I think this is probably wrong anyway :)

grynspan:
Wasm failures are known/unrelated.

PR #1503, opened 2026-01-28
[6.3] Interop: send events to fallback event handler if active configuration not present

jerryjrchen:
- **Explanation**:
  Turn unhandled issues into events and sent to the fallback event handler if
  one has been installed.

- **Scope**:
  Users can only execute this new code path if:
  1. they install a fallback event handler and
  2. record an unhandled issue (e.g. `XCTAssert` in the body of a Swift Testing
     test case).

  Existing tests may satisfy condition 2, but neither swift-corelibs-xctest nor
  Swift Testing install a fallback event handler. When that happens, the
  interoperability functionality will initially be opt-in. Details about the
  phased rollout of interoperability are included in
  https://github.com/jerryjrchen/swift-evolution/blob/swt-xct-interop/proposals/testing/NNNN-targeted-interoperability-swift-testing-and-xctest.md#interoperability-modes

- **Issues**:
  n/a

- **Original PRs**:
  #1439

- **Risk**:
  Links a new dylib, _TestingInterop. Uses the fallback event handler getter
  from this library, but looks it up using dlsym/GetProcAddress to avoid crashing due to missing symbols at runtime.

- **Testing**:
  Validate in CI that we can link against _TestingInterop.
  Automated testing to validate that we are forwarding unhandled events to the
  new fallback event handler.

- **Reviewers**:
  @stmontgomery 

### Checklist:

### DIFF
--- a/Sources/_TestingInternals/include/Stubs.h
+++ b/Sources/_TestingInternals/include/Stubs.h
@@ -180,6 +180,31 @@ static int swt_setfdflags(int fd, int flags) {
 }
 #endif
 
+#if !SWT_NO_INTEROP
+
+/// A type describing a fallback event handler that testing API can invoke as an
+/// alternate method of reporting test events to the current test runner.
+/// Shadows the type with the same name in _TestingInterop.
+///
+/// - Parameters:
+///   - recordJSONSchemaVersionNumber: The JSON schema version used to encode
+///     the event record.
+///   - recordJSONBaseAddress: A pointer to the first byte of the encoded event.
+///   - recordJSONByteCount: The size of the encoded event in bytes.
+///   - reserved: Reserved for future use.
+typedef void (* SWTFallbackEventHandler)(const char *recordJSONSchemaVersionNumber,
+                                      const void *recordJSONBaseAddress,
+                                      size_t recordJSONByteCount,
+                                      const void *_Nullable reserved);
+
+/// Get the current fallback event handler.
+/// Shadows the function with the same name in _TestingInterop.
+///
+/// - Returns: The currently-set handler function, if any.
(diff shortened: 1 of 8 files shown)


jerryjrchen:
@swift-ci test

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -180,6 +180,31 @@ static int swt_setfdflags(int fd, int flags) {
 }
 #endif
 
+#if !SWT_NO_INTEROP
+
+/// A type describing a fallback event handler that testing API can invoke as an
+/// alternate method of reporting test events to the current test runner.
+/// Shadows the type with the same name in _TestingInterop.
```
grynspan, line 187: We _could_ put a header in `_TestingInternals` for this, and just have `_TestingInterop` depend on that header (without needing to link any sources in CMake, just see the header).

I don't feel too too strongly, just a thought.

grynspan:
Please use the nomination template [here](https://github.com/swiftlang/.github/blob/main/PULL_REQUEST_TEMPLATE/release.md?plain=1).

Sources/_TestingInternals/include/Stubs.h
grynspan: (Not for the cherry-pick obviously, I mean on main)

jerryjrchen: Ah, would the benefit there be that `FallbackEventHandler.swift` would depend on the same function definition in some `FallbackEventHandlerStubs.h`? Would that still work for building `_TestingInterop_DO_NOT_USE` via the package?

grynspan: I think so. Let's discuss privately.

PR #1504, opened 2026-01-28
Add documentation for debugging Swift Testing tests with LLDB from the command line

gottesmm:
This document explains how to attach LLDB to debug test code directly when `swift test` alone is insufficient. It targets compiler engineers who need to debug Swift Testing executables.

This documentation lives in the swift-testing repository so Swift Testing engineers can update it as the project evolves. A subsequent commit to the swift repository will add a link from DebuggingTheCompiler.md to help Swift engineers discover it.

### DIFF
--- a/Documentation/CommandlineDebugging.md
+++ b/Documentation/CommandlineDebugging.md
@@ -1,0 +1,100 @@
+# Debugging Swift Testing on the Commandline using LLDB
+
+<!--
+This source file is part of the Swift.org open source project
+
+Copyright (c) 2026 Apple Inc. and the Swift project authors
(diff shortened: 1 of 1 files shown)


stmontgomery:
Thanks! I'll review in depth, but first wanted to note that an anticipated SwiftPM feature will hopefully simplify this workflow and likely mean changes to this document once it lands. See https://github.com/swiftlang/swift-package-manager/pull/9085. cc @plemarquand 

This is still probably helpful in the interim though

gottesmm:
@stmontgomery Yes. Please change it when your change lands. I am trying to just start this so that compiler engineers have the tools needed to debug this stuff without needing to bug you all

PR #1505, opened 2026-01-28
FreeBSD: bsearch pointer is Nonnull

etcwilde:
The pointer passed into the compare function is marked Nonnull in the FreeBSD stdlib header. This causes a build failure due to the unwrap.

```
error: cannot force unwrap value of non-optional type 'UnsafeRawPointer'
```

Fixing it for FreeBSD.

### DIFF
--- a/Sources/Testing/Support/Additions/ArrayAdditions.swift
+++ b/Sources/Testing/Support/Additions/ArrayAdditions.swift
@@ -54,7 +54,7 @@ extension Array {
       return withUnsafePointer(to: context) { context in
         self.withUnsafeBufferPointer { elements in
           let result = bsearch(context, elements.baseAddress!, elements.count, MemoryLayout<Element>.stride) { contextAddress, elementAddress in
-#if os(Android)
+#if os(Android) || os(FreeBSD)
             let contextAddress = contextAddress as UnsafeRawPointer?
 #endif
             let context = contextAddress!.load(as: _BinarySearchContext.self)


Sources/Testing/Support/Additions/ArrayAdditions.swift
```diff
@@ -57,7 +57,11 @@ extension Array {
 #if os(Android)
             let contextAddress = contextAddress as UnsafeRawPointer?
```
harlanhaskins, line 58: Should we actually just merge this with the android check above?

grynspan: Yes.

etcwilde: Convert to an optional to force-unwrap it? Sure.

etcwilde: I've confirmed that the `as` gets optimized out correctly and doesn't try materializing any runtime metadata or making extra allocations. I'm fine with the round-about. Just a source-level oddity.

grynspan: For those watching at home, we're doing this sorta "backwards" because different Android NDKs have different nullability.

etcwilde: And for those watching on pins and needles, the nullability annotations for this function were added to Android in 2022 in bionic commit 5a918d922b.

PR #1511, opened 2026-01-30
Adopt `Mutex` (again).

grynspan:
This PR adopts the [`Synchronization.Mutex`](https://developer.apple.com/documentation/synchronization/mutex) type with a pretty large caveat: on Apple platforms, because we back-deploy earlier than when `Mutex` is available, we must provide our own implementation of `Mutex` there.

## Apple

The Apple-specific implementation is, effectively, the same as what we had before. We directly allocate storage for our unfair lock and the guarded value instead of using `ManagedBuffer` (in particular because `ManagedBuffer` does not allow for move-only values). Because our `Mutex` is `~Copyable`, it has a `deinit` member that deinitializes and deallocates this storage (instead of it being deallocated as part of `ManagedBuffer` being released).

This is actually slightly _more_ performant than before (though not measurably so) because the raw storage does not need to be reference-counted.

This implementation can be removed if/when our minimum deployment target on Apple platforms is high enough to let us use the _real_ `Mutex` type.

## Elsewhere

On non-Apple platforms, we use `Mutex` directly. This is more efficient than our current implementation (which stuffs a `Mutex` into a heap-allocated box) in the scenarios where the mutex need not be heap-allocated.

## Both

I've written a simple `Allocated<T>` box type that provides the appropriate storage and lifetime management for when a mutex _does_ need to be heap-allocated (primarily when used as an instance property of a value type). This type provides equivalent performance to the "old" implementation on non-Apple platforms. On Apple platforms, it does result in a double-allocation scenario, but I've made the Executive Decision™ that this is worth it for getting us closer to adopting `Mutex`.

See also #538.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/MutexAdditions.swift
+++ b/Sources/Testing/Support/Additions/MutexAdditions.swift
@@ -14,7 +14,15 @@ internal import _TestingInternals
 internal import Synchronization
 #endif
 
-/// # Q: When should I use `Mutex<T>` vs. `Allocated<Mutex<T>>`?
+#if SWT_TARGET_OS_APPLE
+/// A type that replicates the interface of ``Synchronization/Mutex``.
+///
+/// This type is used on Apple platforms because our deployment target there is
+/// earlier than the availability of the ``Synchronization/Mutex`` type. It
+/// replicates the interface of that type but is implemented differently (using
+/// heap-allocated storage for the underlying lock and the value it guards).
+///
+/// **Q:** When should I use `Mutex<T>` vs. `Allocated<Mutex<T>>`?
 ///
 /// **A (short):** Whenever the compiler lets you use `Mutex<T>`, use that.
 ///
@@ -24,14 +32,6 @@ internal import Synchronization
 ///   `Allocated`. If, however, you need a mutex to be an instance member of a
 ///   copyable value type (a structure or enumeration), then it _must_ be boxed
 ///   with `Allocated` (or something else that moves its storage onto the heap).
-
-#if SWT_TARGET_OS_APPLE
-/// A type that replicates the interface of ``Synchronization/Mutex``.
-///
-/// This type is used on Apple platforms because our deployment target there is
-/// earlier than the availability of the ``Synchronization/Mutex`` type. It
-/// replicates the interface of that type but is implemented differently (using
-/// heap-allocated storage for the underlying lock and the value it guards).
 struct Mutex<Value>: Sendable, ~Copyable where Value: ~Copyable {
   /// The underlying lock type.
 #if !SWT_NO_OS_UNFAIR_LOCK


grynspan:
The Wasm-specific failures are due to #1512.

Sources/Testing/Attachments/Attachment.swift
```diff
@@ -24,22 +24,8 @@ private import _TestingInternals
 ///   @Available(Xcode, introduced: 26.0)
 /// }
 public struct Attachment<AttachableValue> where AttachableValue: Attachable & ~Copyable {
-  /// A class that stores an attachment's (potentially move-only) attachable
-  /// value.
-  ///
-  /// We use a class to store the attachable value so that ``Attachment`` can
-  /// conform to `Copyable` even if `AttachableValue` doesn't.
-  fileprivate final class Storage {
```
stmontgomery, line 32: Removing the need for this and consolidating into `Allocated<T>` is a nice cleanup

Tests/TestingTests/Support/LockTests.swift
```diff
@@ -11,11 +11,16 @@
...
 private import _TestingInternals
 
-@Suite("Locked Tests")
-struct LockTests {
+#if !SWT_TARGET_OS_APPLE && canImport(Synchronization)
+import Synchronization
+#endif
+
+@Suite("Mutex Tests")
+final class LockTests: Sendable {
```
stmontgomery:
```suggestion
final class MutexTests: Sendable {
```

Tests/TestingTests/Support/MutexTests.swift
stmontgomery, line 1: `MutexTests.swift`

stmontgomery:
No major comments, looks helpful! I would encourage running a toolchain build before landing

Tests/TestingTests/Support/MutexTests.swift
```diff
@@ -11,25 +11,26 @@
 @testable import Testing
 private import _TestingInternals
 
-@Suite("Locked Tests")
-struct LockTests {
-  @Test("Locking and unlocking")
-  func locking() {
-    let lock = Locked(rawValue: 0)
+#if !SWT_TARGET_OS_APPLE && canImport(Synchronization)
```
jerryjrchen, line 14: This is quite nitpicky, but would we want to update the other `canImport(Synchronization)
` checks in this PR to be `#if !SWT_TARGET_OS_APPLE && canImport(Synchronization)`? Seems like we'd be using our custom implementation of `Mutex` when `SWT_TARGET_OS_APPLE` is set.

grynspan: They're only necessary in the test target. In `Testing`, the type checker will prefer the local type.

grynspan:
Toolchain builds passed.

PR #1512, opened 2026-01-30
Don't set `SWT_NO_INTEROP` on WASI.

grynspan:
The toolchain on the release/6.3 branch has regressed because we are inadvertently setting `SWT_NO_INTEROP` on WASI in the CMake build but not in the SwiftPM manifest. This change was unintentional.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -411,18 +411,6 @@ extension Array where Element == PackageDescription.SwiftSetting {
       .define("SWT_NO_LIBDISPATCH", .whenEmbedded()),
     ]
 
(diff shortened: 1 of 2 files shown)


grynspan:
This error is expected until the change actually lands in the toolchain:

```
error: link command failed with exit code 1 (use -v to see invocation)
wasm-ld: error: /__w/swift-testing/swift-testing/.build/wasm32-unknown-wasip1/debug/Testing.build/Event+FallbackHandler.swift.o: undefined symbol: _swift_testing_getFallbackEventHandler
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

PR #1516, opened 2026-02-02
CMake: Add autolink entry for TestingInterop library

kateinoigakukun:
The `lib_TestingInterop.a` library should be autolinked when linking the Testing library in static builds.

### Motivation:

 This patch fixes the linkage issue on the latest main snapshot of Swift SDK for Wasm:

```
wasm-ld: error: path/to/swift_static/wasi/libTesting.a(ABIEntryPoint.swift.obj): undefined symbol: _swift_testing_getFallbackEventHandler
```

### Modifications:

Just declare `_TestingInterop` as an auto-linked library

### Checklist:


stmontgomery (approved):
Thanks @kateinoigakukun!

PR #1521, opened 2026-02-03
Workaround for CI failure due to `Data.bytes` not being present.

grynspan:
swift-ci still uses an older Xcode version with an older OS SDK that does not include `Data.bytes`. As a result, Swift Testing fails to build for Apple platforms. Work around the issue by getting an unsafe buffer from the `Data` and using the buffer's `RawSpan` instead of the data's.

Resolves rdar://169480914.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/JSON.swift
+++ b/Sources/Testing/Support/JSON.swift
@@ -44,7 +44,10 @@ enum JSON {
     encoder.userInfo.merge(userInfo, uniquingKeysWith: { _, rhs in rhs})
 
     let data = try encoder.encode(value)
-    return try body(data.bytes)
+    // WORKAROUND for older SDK on swift-ci (rdar://169480914)
+    return try data.withUnsafeBytes { data in
+      try body(data.bytes)
+    }
 #else
     throw SystemError(description: "JSON encoding requires Foundation which is not available in this environment.")
 #endif


grynspan:
https://github.com/swiftlang/swift/pull/76034#issuecomment-3842439073

Sources/Testing/Support/JSON.swift
```diff
@@ -44,7 +44,10 @@ enum JSON {
     encoder.userInfo.merge(userInfo, uniquingKeysWith: { _, rhs in rhs})
 
     let data = try encoder.encode(value)
-    return try body(data.bytes)
+    // WORKAROUND for older SDK on swift-ci (rdar://169480914)
```
harlanhaskins, line 47: Is there a bug to remove this once swift-ci updates?

harlanhaskins (approved):

grynspan:
Wasm issues are known/unrelated.

grynspan:
That's the only use of `Data.bytes` specifically (I checked) outside our test target (which doesn't need to build in the affected CI context).

AFAIK other stuff should be fine, but we'll find out.

grynspan:
Toolchain build isn't 100% complete, but Swift Testing built successfully so that's all we were looking for.

Sources/Testing/Support/JSON.swift
grynspan: I honestly don't know if/when that is planned, so I didn't bother, but we can check in with @shahmishal who might know.

PR #1523, opened 2026-02-03
Install fallback event handler before running tests

jerryjrchen:
Swift Testing now participates in both directions of interoperability. Currently, this is opt-in via an environment variable at runtime.

### Motivation:

Allows Swift Testing to handle foreign test events (e.g. XCTAssert).

### Modifications:

Install fallback event handler before running tests.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Event+FallbackEventHandler.swift
+++ b/Sources/Testing/Events/Event+FallbackEventHandler.swift
@@ -22,8 +22,8 @@ extension Issue {
   /// represented by the schema.
   ///
   /// - Parameter event: The `EncodedIssue` wrapped in an `EncodedEvent`.
-  /// Return `nil` if this is not an issueRecorded kind of event, or if the
-  /// event doesn't include an `EncodedIssue`.
+  /// - Returns: `nil` if this is not an issueRecorded kind of event, or if the
+  ///   event doesn't include an `EncodedIssue`.
   init?<V>(event: ABI.EncodedEvent<V>) where V: ABI.Version {
     switch event.kind {
     case .issueRecorded:
@@ -118,13 +118,27 @@ extension Event {
   /// testing library.
   private static let _ourFallbackEventHandler: SWTFallbackEventHandler = {
     recordJSONSchemaVersionNumber, recordJSONBaseAddress, recordJSONByteCount, _ in
-    let abi = String(validatingCString: recordJSONSchemaVersionNumber)
+    let version = String(validatingCString: recordJSONSchemaVersionNumber)
       .flatMap(VersionNumber.init)
       .flatMap { ABI.version(forVersionNumber: $0) }
-    if let abi {
+    if let version {
       let recordJSON = UnsafeRawBufferPointer(
         start: recordJSONBaseAddress, count: recordJSONByteCount)
-      try! Self.handle(recordJSON, encodedWith: abi)
+      do {
+        try Self.handle(recordJSON, encodedWith: version)
+      } catch {
+        // Surface otherwise "unhandleable" records instead of dropping them silently
+        let errorContext: Comment = """
+        Another test library reported a test event that Swift Testing could not decode. Inspect the payload to determine if this was a test assertion failure.
+
+        Error:
+        \(error)
+
+        Raw payload:
+        \(recordJSON)
+        """
+        Issue.record(errorContext)
+      }
     }
   }
 #endif
@@ -132,10 +146,7 @@ extension Event {
   /// The implementation of ``installFallbackEventHandler()``.
   private static let _installFallbackEventHandler: Bool = {
 #if !SWT_NO_INTEROP
-    if let environmentValue = Environment.variable(named: "SWT_EXPERIMENTAL_INTEROP_ENABLED")
-      .flatMap(Int.init),
-      environmentValue > 0
-    {
+    if Environment.flag(named: "SWT_EXPERIMENTAL_INTEROP_ENABLED") == true {
       return _swift_testing_installFallbackEventHandler(Self._ourFallbackEventHandler)
     }
 #endif
--- a/Sources/_TestingInternals/include/Stubs.h
+++ b/Sources/_TestingInternals/include/Stubs.h
@@ -302,7 +302,7 @@ typedef void (* SWTFallbackEventHandler)(const char *recordJSONSchemaVersionNumb
 /// The fallback event handler can only be installed once per process, typically
 /// by the first testing library to run. If this function has already been
 /// called and the handler set, it does not replace the previous handler.
-SWT_EXTERN bool _swift_testing_installFallbackEventHandler(SWTFallbackEventHandler);
+SWT_EXTERN bool _swift_testing_installFallbackEventHandler(SWTFallbackEventHandler handler);
 
 /// Get the current fallback event handler.
 /// Shadows the function with the same name in _TestingInterop.
--- a/Tests/TestingTests/EventHandlingInteropTests.swift
+++ b/Tests/TestingTests/EventHandlingInteropTests.swift
@@ -148,6 +148,43 @@ struct EventHandlingInteropTests {
     }
   }
 
+  @Test func `Fallback handler records an issue if invalid event provided`() async throws {
+    await #expect(processExitsWith: .success) {
+      // Install and retrieve the fallback event handler.
+      // Capture all observed issues when running tests.
+      Self.enableExperimentalInterop()
+      try #require(Event.installFallbackEventHandler(), "Should successfully install a new handler")
+      let currentHandler = try #require(
+        _swift_testing_getFallbackEventHandler(), "Should successfully retrieve installed handler")
+
+      let issues = Mutex<[Issue]>()
+      var configuration = Configuration()
+      configuration.eventHandler = { event, _ in
+        if case .issueRecorded(let issue) = event.kind {
+          issues.withLock { $0.append(issue) }
+        }
+      }
+
+      // Pass an invalid record JSON to the event handler
+      struct Empty: Encodable {}
+      await Test {
+        try JSONEncoder().encode(Empty()).withUnsafeBytes { ptr in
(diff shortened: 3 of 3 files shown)


Tests/TestingTests/EventHandlingInteropTests.swift
```diff
@@ -9,6 +9,7 @@
 //
 
 private import _TestingInternals
+
```
grynspan: Stray newline

Sources/Testing/Events/Event+FallbackHandler.swift
```diff
@@ -10,13 +10,133 @@
...
+
+      let severity: Issue.Severity =
+        switch issue.severity {
+        case .warning: .warning
+        case nil, .error: .error
+        }
+
+      let comments: [Comment] = event._comments?.map(Comment.init(rawValue:)) ?? []
+      let sourceContext = SourceContext(
+        backtrace: nil,  // A backtrace from the child process will have the wrong address space.
```
stmontgomery:
```swift
// A backtrace from the child process will have the wrong address space.
```
This is true for exit tests, which operate in a separate process, but in this scenario I think the backtrace should be valid and useful to maintain

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -10,13 +10,133 @@
 
 private import _TestingInternals
 
+extension Issue {
+  init?<V>(event: ABI.EncodedEvent<V>) where V: ABI.Version {
```
stmontgomery, line 27: Mind documenting this? It'd be useful to explain that it returns `nil` for events whose kinds don't represent issues. (And in general, we try to document initializers, except the simplest, pure memberwise assignment ones.)

Sources/Testing/Events/Event+FallbackHandler.swift
```diff
@@ -10,13 +10,133 @@
...
 extension Event {
+  /// Attempt to handle an event encoded as JSON as if it had been generated in
+  /// the current testing context.
+  ///
+  /// If the event contains an issue, handle it, but also record a warning issue
+  /// notifying the user that interop was performed.
+  ///
+  /// - Parameters:
+  ///   - recordJSON: The JSON encoding of an event record.
+  ///   - abi: The ABI version to use for decoding `recordJSON`.
```
stmontgomery: bikeshedding
```suggestion
  ///   - version: The ABI version to use for decoding `recordJSON`.
```
And same with the param itself, below.

Sources/Testing/Events/Event+FallbackHandler.swift
```diff
@@ -10,13 +10,133 @@
...
+      case .event(let event) = record.kind,
+      let issue = Issue(event: event)
+    else {
+      return
+    }
+
+    // For the time being, assume that foreign test events originate from XCTest
+    let warnForXCTestUsageIssue = {
+      let sourceContext = SourceContext(
+        backtrace: nil,  // A backtrace from the child process will have the wrong address space.
```
stmontgomery: Same comment as [earlier](https://github.com/swiftlang/swift-testing/pull/1523/changes#r2865729680) about the backtrace

Sources/Testing/Events/Event+FallbackHandler.swift
jerryjrchen: I can update the comment, but since backtraces aren't formally part of the JSON ABI yet, I'm still planning on leaving the backtrace as nil here

Sources/Testing/Events/Event+FallbackHandler.swift
jerryjrchen: Hmm, I guess this can just inherit the backtrace from the original issue if one exists?

Sources/Testing/Events/Event+FallbackEventHandler.swift
jerryjrchen: Good reminder, thanks!

grynspan: Why not just have it take an instance of `EncodedIssue` instead?

jerryjrchen: I'm filling in comments from ~`event._comments`~ actually changed this, but still need `event.messages` to fill in issue details

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -263,6 +263,18 @@ typedef void (* SWTFallbackEventHandler)(const char *recordJSONSchemaVersionNumb
...
+///
+/// - Parameters:
+///   - handler: The handler function to set.
+///
+/// - Returns: Whether or not `handler` was installed.
+///
+/// The fallback event handler can only be installed once per process, typically
+/// by the first testing library to run. If this function has already been
+/// called and the handler set, it does not replace the previous handler.
+SWT_EXTERN bool _swift_testing_installFallbackEventHandler(SWTFallbackEventHandler);
```
stmontgomery:
```suggestion
SWT_EXTERN bool _swift_testing_installFallbackEventHandler(SWTFallbackEventHandler handler);
```
Since the docs reference that parameter

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+  /// `EncodedEvent` format and then passes it through Swift Testing's installed
+  /// fallback event handler to be converted into an `Issue`.
+  ///
+  /// The fidelity of this conversion is limited by the fields present in
+  /// `EncodedIssue`, in addition to how well the foreign test issue is
+  /// represented by the schema.
+  ///
+  /// - Parameter event: The `EncodedIssue` wrapped in an `EncodedEvent`.
+  /// Return `nil` if this is not an issueRecorded kind of event, or if the
+  /// event doesn't include an `EncodedIssue`.
```
stmontgomery:
```suggestion
  ///
  /// - Returns: `nil` if this is not an issueRecorded kind of event, or if the
  ///   event doesn't include an `EncodedIssue`.
```

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+  /// The fallback event handler that was installed in the current test process.
+  private static let _activeFallbackEventHandler: SWTFallbackEventHandler? = {
+    _swift_testing_getFallbackEventHandler()
+  }()
+
+  /// The fallback event handler to install when Swift Testing is the active
+  /// testing library.
+  private static let _ourFallbackEventHandler: SWTFallbackEventHandler = {
+    recordJSONSchemaVersionNumber, recordJSONBaseAddress, recordJSONByteCount, _ in
+    let abi = String(validatingCString: recordJSONSchemaVersionNumber)
```
stmontgomery: I'd suggest naming this `version` since it's a version metatype
```suggestion
    let version = String(validatingCString: recordJSONSchemaVersionNumber)
```

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+      .flatMap(VersionNumber.init)
+      .flatMap { ABI.version(forVersionNumber: $0) }
+    if let abi {
+      let recordJSON = UnsafeRawBufferPointer(
+        start: recordJSONBaseAddress, count: recordJSONByteCount)
+      try! Self.handle(recordJSON, encodedWith: abi)
```
stmontgomery: Can you talk about why you believe it's safe/appropriate to `try!` here? Would be nice to capture that in a comment (though not critical), but my main question is whether it's actually safe

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -0,0 +1,199 @@
...
+    }
+  }
+#endif
+
+  /// The implementation of ``installFallbackEventHandler()``.
+  private static let _installFallbackEventHandler: Bool = {
+#if compiler(>=6.3) && !SWT_NO_INTEROP
+    if let environmentValue = Environment.variable(named: "SWT_EXPERIMENTAL_INTEROP_ENABLED")
+      .flatMap(Int.init),
+      environmentValue > 0
```
stmontgomery: We have `Environment.flag(named:)` which handles the `Bool` conversion

Sources/Testing/Running/Runner.swift
```diff
@@ -452,6 +452,7 @@ extension Runner {
 #if !SWT_NO_FILE_IO
     runner.configureAttachmentHandling()
 #endif
+    _ = Event.installFallbackEventHandler()
```
stmontgomery, line 455: I wonder if we should add a `Configuration` property which specifies whether to _skip_ installing the fallback event handler. That way, an integrated tool (such as Xcode, but potentially others) could configure that to avoid that step and the expectations around it would become more deterministic. I understand subsequent installation attempts may be a no-op, so this could be a follow-on improvement. Just an idea to consider

Sources/Testing/Events/Event+FallbackEventHandler.swift
jerryjrchen: Hmm, I think I actually want to do some logging here instead of the `try!`. My initial intention was that we didn't want to let errors silently slip by because the handler failed to decode the message. I'm going to record an issue with the error (and add a test for that behaviour too)

Sources/Testing/Running/Runner.swift
jerryjrchen: I'll make a note to do this as a follow up improvement

Sources/Testing/Events/Event+FallbackEventHandler.swift
jerryjrchen: I came up with a somewhat clunky error message, but at the very least it should provide plenty of context for this rare failure mode.

PR #1527, opened 2026-02-04
Always cast the `bsearch()` context argument to optional.

grynspan:
Different platforms assign different nullability to the `context` argument of `bsearch()`'s callback function. For consistency's sake and to avoid problems porting to future platforms, always cast the value to an optional before loading from it. The compiler is smart enough to eliminate the copy/cast, so this is ultimately a zero-cost no-op.

### Checklist:


grynspan:
Wasm failures are a known issue.

PR #1528, opened 2026-02-05
Move test repetition down to the test case level

### LINKED ISSUE #1392: Support retry-on-failure for only the failing @Test cases
### Motivation

`xcodebuild` supports the option `-retry-tests-on-failure` which can be passed to automatically retry tests that fail. Swift Testing is partially supported with this option, but _all_ Swift Testing `@Test` are retried, even if they had already passed.

Consider the following tests;

```swift
struct SwiftTestingA {
  @Test
  func successA(){
    #expect(Bool(true))
  }

 @Test
  func failureA() {
    #expect(Bool(false))
  }
}

struct SwiftTestingB {
  @Test
  func successB() {
    #expect(Bool(true))
  }
}

final class XCTestA: XCTestCase {

  func testSuccess() {
    XCTAssertTrue(true)
  }

  func testFailure() {
    XCTAssertTrue(false)
  }
}
```

When executing with `xcodebuild test ...`
- XCTestA.testSuccess() is executed and passes 1 time only (expected).
- SwiftTestingA.success() is executed and passes 3 times (1 expected).
- SwiftTestingB.success() is executed and passes 3 times (1 expected).

```shell
% xcodebuild test -scheme TestLibrary -retry-tests-on-failure

Test suite 'XCTestA' started
Test suite 'SwiftTestingB' started
Test suite 'SwiftTestingA' started
Test case 'SwiftTestingB/successB()' passed (0.000 seconds)
Test case 'SwiftTestingA/successA()' passed (0.000 seconds)
Test case 'XCTestA.testFailure()' failed (0.284 seconds)
Test case 'XCTestA.testFailure()' failed (0.001 seconds)
Test case 'XCTestA.testFailure()' failed (0.001 seconds)
Test case 'SwiftTestingA/failureA()' failed (0.000 seconds)
Test suite 'SwiftTestingB' started
Test suite 'SwiftTestingA' started
Test case 'SwiftTestingB/successB()' passed (0.000 seconds)
Test case 'SwiftTestingA/successA()' passed (0.000 seconds)
Test case 'SwiftTestingA/failureA()' failed (0.000 seconds)
Test suite 'SwiftTestingB' started
Test suite 'SwiftTestingA' started
Test case 'SwiftTestingA/successA()' passed (0.000 seconds)
Test case 'SwiftTestingB/successB()' passed (0.000 seconds)
Test case 'SwiftTestingA/failureA()' failed (0.000 seconds)
Test case 'XCTestA.testSuccess()' passed (0.001 seconds)
```

### Proposed solution

With `xcodebuild` aside, it would be useful for `swift test` to natively support an option like `-retry-tests-on-failure` to automatically retry tests that fail;

```
% swift test -retry-tests-on-failure

◇ Suite SwiftTestingA started.
◇ Suite SwiftTestingB started.
✔ Test successA() passed after 0.001 seconds.
✔ Test successB() passed after 0.001 seconds.
✔ Suite SwiftTestingB passed after 0.001 seconds.
✘ Test failureA() recorded an issue at Tests.swift:17:5: Expectation failed: Bool(false)
✘ Test failureA() iteration: 1 failed after 0.001 seconds with 1 issue.
✘ Test failureA() recorded an issue at Tests.swift:17:5: Expectation failed: Bool(false)
✘ Test failureA() iteration: 2 failed after 0.001 seconds with 1 issue.
✘ Test failureA() recorded an issue at Tests.swift:17:5: Expectation failed: Bool(false)
✘ Test failureA() iteration: 3 failed after 0.001 seconds with 1 issue.
✘ Suite SwiftTestingA failed after 0.001 seconds with 3 issues.
✘ Test run with 3 tests failed after 0.001 seconds with 3 issues.
```

Tests that pass `successA()` / `successB()` would not be retried and would only be executed once.

While parametrized tests could potentially retry just the parameters that fail — the CLI output reports these as a single test so it may be fine to just retry all parameters if a single parameter fails.

### Alternatives considered

_No response_

### Additional information

https://forums.swift.org/t/retry-tests-on-failure-retries-test-that-have-already-passed/83031

harlanhaskins:
This changes the runner such that "repeat until success" or the other repetition policies repeat only the affected test case, rather than repeating the entire test plan.

Fixes #1392
Fixes rdar://130508488

### Motivation:

The previous behavior was unexpected and does not match the behavior of XCTest. This changes the behavior to align the behaviors (in the way that users expect).

### Modifications:

- Adds a configuration option requiring clients to opt in to this behavior (for now, until we change the default).
- Updates the Runner to move the iteration into a helper function
- Adds this helper function within _runTestCase (if opted in)
- Replaces the plan-level iteration with the helper function (if opted out)
- Mirrors the plan-level iteration tests to test-case level iteration tests that test the same conditions with the new behavior

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
@@ -117,10 +117,18 @@ extension ABI {
       case .testStarted:
         kind = .testStarted
       case .testCaseStarted:
+        // For non-parameterized tests, we elide `testCaseStarted` calls because it would be
+        // redundant. However, for multiple iterations of a test case within a non-parameterized
+        // function, we need to emit another `testStarted` event.
         if eventContext.test?.isParameterized == false {
-          return nil
+          if let iteration = eventContext.iteration, iteration > 1 {
+            kind = .testStarted
+          } else {
+            return nil
+          }
+        } else {
+          kind = .testCaseStarted
         }
-        kind = .testCaseStarted
       case let .issueRecorded(recordedIssue):
         kind = .issueRecorded
         issue = EncodedIssue(encoding: recordedIssue, in: eventContext)
@@ -129,9 +137,14 @@ extension ABI {
         self.attachment = EncodedAttachment(encoding: attachment)
       case .testCaseEnded:
         if eventContext.test?.isParameterized == false {
-          return nil
+          if let iteration = eventContext.iteration, iteration > 1 {
+            kind = .testEnded
+          } else {
+            return nil
+          }
+        } else {
+          kind = .testCaseEnded
         }
-        kind = .testCaseEnded
       case .testCaseCancelled:
         kind = .testCaseCancelled
       case .testEnded:
@@ -164,6 +177,7 @@ extension ABI {
           let .testCancelled(skipInfo):
           _comments = Array(skipInfo.comment).map(\.rawValue)
           _sourceLocation = skipInfo.sourceLocation.map { EncodedSourceLocation(encoding: $0) }
+          _iteration = eventContext.iteration
         default:
           break
         }
--- a/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
@@ -333,6 +333,9 @@ public struct __CommandLineArguments_v0: Sendable {
   /// The value of the `--repeat-until` argument.
   public var repeatUntil: String?
 
+  /// Whether or not to use the per-test-case repetition mode.
+  var usePerTestCaseRepetition: Bool = false
+
   /// The value of the `--attachments-path` argument.
   public var attachmentsPath: String?
 }
@@ -537,6 +540,9 @@ func parseCommandLineArguments(from args: [String]) throws -> __CommandLineArgum
   if let repeatUntil = args.argumentValue(forLabel: "--repeat-until") {
     result.repeatUntil = repeatUntil
   }
+  if args.contains("--experimental-per-test-case-repetition") {
+    result.usePerTestCaseRepetition = true
+  }
 
   return result
 }
@@ -664,6 +670,9 @@ public func configurationForEntryPoint(from args: __CommandLineArguments_v0) thr
     }
   }
   configuration.repetitionPolicy = repetitionPolicy
+
+  // Opt in to per-test-case repetition
+  configuration.shouldUseLegacyPlanLevelRepetition = !args.usePerTestCaseRepetition
 
 #if !SWT_NO_EXIT_TESTS
   // Enable exit test handling via __swiftPMEntryPoint().
--- a/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
+++ b/Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
@@ -365,6 +365,10 @@ extension Event.HumanReadableOutputRecorder {
       String(describing: TimeValue(rawValue: end.suspending.rawValue - start.suspending.rawValue))
     }
 
+    func testStartedMessage(for test: Test) -> String {
+      "\(_capitalizedTitle(for: test)) \(testName) started"
+    }
+
     // Finally, produce any messages for the event.
     switch event.kind {
     case .testDiscovered:
@@ -425,7 +429,7 @@ extension Event.HumanReadableOutputRecorder {
       return [
         Message(
           symbol: .default,
-          stringValue: "\(_capitalizedTitle(for: test)) \(testName) started."
+          stringValue: "\(testStartedMessage(for: test))."
         )
       ]
 
@@ -565,15 +569,26 @@ extension Event.HumanReadableOutputRecorder {
       return result
 
     case .testCaseStarted:
-      guard let testCase, testCase.isParameterized, let arguments = testCase.arguments else {
+      guard let test, let testCase else { break }
+      let iteration = eventContext.iteration ?? 1
+
+      var message: String
+      if testCase.isParameterized, let arguments = testCase.arguments {
+        message = "Test case passing \(arguments.count.counting("argument")) \(testCase.labeledArguments(includingQualifiedTypeNames: verbosity > 0)) to \(testName) started"
+      } else if iteration > 1 {
+        message = testStartedMessage(for: test)
+      } else {
         break
       }
 
+      if iteration > 1 {
+        message += " (repetition \(iteration))"
+      }
+
+      message += "."
+
       return [
-        Message(
-          symbol: .default,
-          stringValue: "Test case passing \(arguments.count.counting("argument")) \(testCase.labeledArguments(includingQualifiedTypeNames: verbosity > 0)) to \(testName) started."
-        )
+        Message(symbol: .default, stringValue: message)
       ]
 
     case .testCaseEnded:
--- a/Sources/Testing/Running/Configuration.swift
+++ b/Sources/Testing/Running/Configuration.swift
@@ -132,6 +132,11 @@ public struct Configuration: Sendable {
     }
   }
 
+  /// Whether to perform test repetition at the plan level or on a per-test-
+  /// case basis.
+  @_spi(Experimental)
+  public var shouldUseLegacyPlanLevelRepetition: Bool = true
+
   /// Whether or not, and how, to iterate the test plan repeatedly.
   ///
   /// By default, the value of this property allows for a single iteration.
--- a/Sources/Testing/Running/Runner+Repetition.swift
+++ b/Sources/Testing/Running/Runner+Repetition.swift
@@ -8,6 +8,10 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
+#if canImport(Synchronization)
+private import Synchronization
+#endif
+
 extension Runner {
   /// A thread-safe set of test+case IDs that have recorded issues.
   /// This keeps track of tests that recorded an issue during a test repetition
--- a/Sources/Testing/Running/Runner+Repetition.swift
+++ b/Sources/Testing/Running/Runner+Repetition.swift
@@ -18,11 +18,11 @@ extension Runner {
   /// and is used by the repetition machinery to determine if an issue was recorded
   /// during a run.
   final class TestIssueRecorder: Sendable {
-    struct ID: Hashable {
+    private struct ID: Hashable {
       var testID: Test.ID
       var testCaseID: Test.Case.ID
     }
-    let ids = Mutex<Set<ID>>([])
+    private let ids = Mutex<Set<ID>>([])
 
     var hasIssues: Bool {
       ids.withLock { !$0.isEmpty }
--- a/Sources/Testing/Running/Runner+Repetition.swift
+++ b/Sources/Testing/Running/Runner+Repetition.swift
@@ -1,0 +1,92 @@
+//
+//  Runner+Repetition.swift
+//  swift-testing
+//
+//  Created by Harlan Haskins on 5/5/26.
+//
+
+extension Runner {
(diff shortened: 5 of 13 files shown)


harlanhaskins:
Converting to a draft for now; there's follow-on discussions we need to have about behavioral changes before we can actually sink this repetition down

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -333,6 +333,8 @@ public struct __CommandLineArguments_v0: Sendable {
   /// The value of the `--repeat-until` argument.
   public var repeatUntil: String?
 
+  var usePerTestCaseRepetition: Bool = false
```
stmontgomery, line 337: nit: Can we include documentation aligning with the other Bool properties in this type?

Sources/Testing/Running/Configuration.swift
```diff
@@ -132,6 +132,10 @@ public struct Configuration: Sendable {
     }
   }
 
+  /// Whether to perform test repetition at the plan level or on a per-test-
+  /// case basis.
+  public var shouldUseLegacyPlanLevelRepetition: Bool = true
```
stmontgomery, line 138: Let's include `@_spi(Experimental)` too, and that would be something we remove if the proposal becomes accepted. (This will mean initially, it will have that _plus_ inherited `@_spi(ForToolsIntegrationOnly)`, and the latter will remain even post-acceptance.)

Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -29,6 +29,9 @@ extension Runner {
     /// The test case that is running on the current task, if any.
     var testCase: Test.Case?
 
+    /// The current iteration of the test repetition policy.
```
stmontgomery:
```suggestion
    /// The current iteration of the test repetition policy, if any.
```

Sources/Testing/Running/Runner.RuntimeState.swift
```diff
@@ -233,6 +236,27 @@ extension Test {
       try await test.withCancellationHandling(body)
     }
   }
+
+  static var currentIteration: Int? {
```
stmontgomery, line 247: Can you add documentation for this property?

Sources/Testing/Running/Runner.RuntimeState.swift
stmontgomery:
```suggestion
// MARK: - Current test, test case, and iteration
```

Sources/Testing/Running/Runner.swift
```diff
@@ -400,16 +397,37 @@ extension Runner {
   /// - Parameters:
   ///   - testCase: The test case to run.
   ///   - step: The runner plan step associated with this test case.
-  ///   - context: Context for the test run.
```
stmontgomery, line 441: Looks like the `context` parameter still exists on this overload, might be an unintentional copy/paste

stmontgomery: Or perhaps, you meant to remove the parameter from the function signature?

Sources/Testing/Running/Runner.swift
```diff
@@ -434,6 +452,49 @@ extension Runner {
     }
   }
 
+  /// Applies the repetition policy specified in the current configuration by running the provided test case
+  /// repeatedly until the continuation condition is satisfied.
+  ///
+  /// - Parameters:
+  ///   - test: The test being executed.
+  ///   - testCase: The test case being iterated.
```
stmontgomery: Params aren't in this function

Sources/Testing/Running/Runner.swift
```diff
@@ -492,45 +544,30 @@ extension Runner {
...
         }
+      } else {
+        await runner.runAll(context: context)
+      }
+    }
+  }
 
-        // Reset the run-wide "issue was recorded" flag for this iteration.
-        issueRecorded.store(false, ordering: .sequentiallyConsistent)
+  private func runAll(context: _Context) async {
```
stmontgomery: nit: Underscore prefix. But, I also feel like its name is missing a noun: run all _what_? And once the name is more clear, could we add light documentation? I 
```suggestion
  private func _runAll(context: _Context) async {
```

Tests/TestingTests/TestCaseIterationTests.swift
```diff
@@ -0,0 +1,122 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
```
stmontgomery:
```suggestion
// Copyright (c) 2026 Apple Inc. and the Swift project authors
```

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -565,15 +569,29 @@ extension Event.HumanReadableOutputRecorder {
...
     case .testCaseStarted:
-      guard let testCase, testCase.isParameterized, let arguments = testCase.arguments else {
+      guard let test, let testCase else { break }
+      let iteration = eventContext.iteration ?? 1
+
+      let shouldShowMessage = testCase.isParameterized || iteration > 1
+
+      guard shouldShowMessage else {
         break
       }
```
stmontgomery: I did a double take below, thinking this logic would mean we would be printing the "Test <name> started." message twice, but then I saw this `guard` and understand that isn't the case.

I'd find it more clear if we moved this check down to the `else` clause below, and removed this `guard`. The `if` conditional already checks `testCase.isParameterized`, so it seems like if the `else` became `else if iteration > 1`, that would maintain the behavior, and then you could add a final `else { break }` to avoid output in the base case.

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -118,9 +118,14 @@ extension ABI {
...
       case .testCaseStarted:
         if eventContext.test?.isParameterized == false {
-          return nil
+          if let iteration = eventContext.iteration, iteration > 1 {
+            kind = .testStarted
+          } else {
+            return nil
+          }
+        } else {
+          kind = .testCaseStarted
```
stmontgomery, line 130: I'm a little unclear on the need for this part — would you mind capturing the rationale in a code comment? Similar for the chunk below

harlanhaskins:
@grynspan I resolved the `testFailWithoutCurrentTest` failure in a way that I'm somewhat dubious about, by making it so the `Configuration.withCurrent` that's used within the iteration helper does not add that to `Configuration._all`, meaning the iterations `issueRecorded` detection won't detect Issues from detached tasks. I am pretty sure that's not desirable, but I can't think of another way to rework this test where we can ensure the issue recorded callback is only called twice.

grynspan:
We have a `Configuration.removeAll()` function you could call from an exit test, maybe?

harlanhaskins:
I don't want to remove _all_ of them (this test specifically is testing that the issue handling traits still get the issue), the issue is that there are multiple of them, one of which is a copy of the other:

```swift
var topLevelConfig = // comes from the entry point

func _applyRepetitionPolicy(...) {
    var innerConfig = topLevelConfig
    innerConfig.eventHandler = { event, context in
        // check for issue, then:
        topLevelConfig.eventHandler(event, context)
    }
    return try await Configuration.withCurrent(innerConfig) {
        try await body()
    }
}

await Configuration.withCurrent(topLevelConfiguration) {
    _applyRepetitionPolicy {
        _runTest()
    }
}
```

In this situation, both `innerConfig` and `topLevelConfig` are pushed into this stack, which means `topLevelConfig`'s event handlers are called twice. I'm not sure how _all is meant to behave with multiple entries...

harlanhaskins:
@grynspan Per our offline discussion, I updated it to instead store compound test+case IDs in a mutex-guarded Set in _Context and then use that to determine if issues are recorded.

Sources/Testing/Running/Runner+Repetition.swift
stmontgomery, line 1: This new file will need to be added to `CMakeLists.txt`

Sources/Testing/Running/Runner+Repetition.swift
```diff
@@ -0,0 +1,91 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
```
stmontgomery:
```suggestion
// Copyright (c) 2026 Apple Inc. and the Swift project authors
```

Sources/Testing/Running/Runner.swift
```diff
@@ -83,8 +83,9 @@ extension Runner {
     /// A serializer used to reduce parallelism among test cases.
     var testCaseSerializer: Serializer<Void>?
 
-    /// Which iteration of the test plan is being executed.
-    var iteration: Int
+    /// A set of test+case IDs that have recorded at least one issue during a test run.
+    /// This is consumed
```
stmontgomery: Is this sentence incomplete?

Sources/Testing/Running/Runner+Repetition.swift
```diff
@@ -0,0 +1,91 @@
...
+      var testID: Test.ID
+      var testCaseID: Test.Case.ID
+    }
+    private let ids = Mutex<Set<ID>>([])
+
+    var hasIssues: Bool {
+      ids.withLock { !$0.isEmpty }
+    }
+
+    func recordIssue(for test: Test.ID, testCase: Test.Case.ID) {
```
stmontgomery, line 40: I think this one could be `private`?
```suggestion
    private func _recordIssue(for test: Test.ID, testCase: Test.Case.ID) {
```

Sources/Testing/Running/Runner+Repetition.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    func clear() {
+      ids.withLock { $0.removeAll() }
+    }
+  }
+
+  /// Applies the repetition policy specified in the current configuration by running the provided test case
+  /// repeatedly until the continuation condition is satisfied.
+  ///
+  /// - Parameters:
```
stmontgomery, line 69: nit: missing `- policy`

Sources/Testing/Running/Runner.swift
stmontgomery: I think this `context` parameter is still used

Sources/Testing/Running/Runner+Repetition.swift
```diff
@@ -0,0 +1,91 @@
...
+#if canImport(Synchronization)
+private import Synchronization
+#endif
+
+extension Runner {
+  /// A thread-safe set of test+case IDs that have recorded issues.
+  /// This keeps track of tests that recorded an issue during a test repetition
+  /// and is used by the repetition machinery to determine if an issue was recorded
+  /// during a run.
+  final class TestIssueRecorder: Sendable {
```
stmontgomery, line 20: Good thing to factor out 👍

Sources/Testing/Running/Runner+Repetition.swift
```diff
@@ -0,0 +1,91 @@
...
+    private struct ID: Hashable {
```
stmontgomery: style: We use underscore prefix on `private` members
```suggestion
    private struct _ID: Hashable {
```

Sources/Testing/Running/Runner+Repetition.swift
```diff
@@ -0,0 +1,91 @@
...
+      var testID: Test.ID
+      var testCaseID: Test.Case.ID
+    }
+    private let ids = Mutex<Set<ID>>([])
```
stmontgomery:
```suggestion
    private let _ids = Mutex<Set<ID>>([])
```

Sources/Testing/Running/Runner+Repetition.swift
stmontgomery, line 1: Several things in this file don't have doc comments, could we add those before landing?

stmontgomery (approved):

PR #1529, opened 2026-02-05
Ensure issues with warning severity are delivered to Xcode.

grynspan:
This PR adjusts the logic we use to decide whether or not to suppress issues with warning severity so that Swift Testing, when built as a package, reports them to Xcode. Xcode 26.0 and later support different issue severities (this branch of Swift Testing no longer supports earlier versions of Xcode.)

Resolves rdar://169741898.

### Checklist:


grynspan:
Wasm failures are a known issue.

jerryjrchen (approved):

PR #1533, opened 2026-02-06
Drop support for older Apple releases.

grynspan:
This PR drops support for older Apple releases when Swift Testing is built as a package.

See Also: https://github.com/swiftlang/swift/pull/87042

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -424,6 +424,7 @@ extension Array where Element == PackageDescription.SwiftSetting {
   private static var availabilityMacroSettings: Self {
     [
       .enableExperimentalFeature("AvailabilityMacro=_uttypesAPI:macOS 11.0, iOS 14.0, watchOS 7.0, tvOS 14.0"),
+      .enableExperimentalFeature("AvailabilityMacro=_clockAPI:macOS 13.0, iOS 16.0, watchOS 9.0, tvOS 16.0"),
       .enableExperimentalFeature("AvailabilityMacro=_typedThrowsAPI:macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0"),
       .enableExperimentalFeature("AvailabilityMacro=_castingWithNonCopyableGenerics:macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0"),
 
--- a/Sources/Testing/Traits/TimeLimitTrait.swift
+++ b/Sources/Testing/Traits/TimeLimitTrait.swift
@@ -13,6 +13,7 @@ private import _TestingInternals
 /// A type that defines a time limit to apply to a test.
 ///
 /// To add this trait to a test, use ``Trait/timeLimit(_:)-4kzjp``.
+@available(_clockAPI, *) // For DocC
 public struct TimeLimitTrait: TestTrait, SuiteTrait {
   /// A type representing the duration of a time limit applied to a test.
   ///
@@ -51,6 +52,7 @@ public struct TimeLimitTrait: TestTrait, SuiteTrait {
 
 // MARK: -
 
+@available(_clockAPI, *) // For DocC
 extension Trait where Self == TimeLimitTrait {
   /// Construct a time limit trait that causes a test to time out if it runs for
   /// too long.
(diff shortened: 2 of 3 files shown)


Sources/Testing/Events/Clock.swift
```diff
@@ -136,8 +134,7 @@ extension Test.Clock {
   /// This function is not part of the public interface of the testing library.
   /// It is primarily used by the testing library's own tests. External clients
   /// can use ``sleep(for:tolerance:)`` or ``sleep(until:tolerance:)`` instead.
-  @available(_clockAPI, *)
-  static func sleep(for duration: Duration) async throws {
+    static func sleep(for duration: Duration) async throws {
```
stmontgomery: accidental indent

Sources/Testing/Events/TimeValue.swift
```diff
@@ -47,13 +47,11 @@ struct TimeValue: Sendable {
...
 
-  @available(_clockAPI, *)
-  init(_ duration: Duration) {
+    init(_ duration: Duration) {
     self.init(duration.components)
   }
 
-  @available(_clockAPI, *)
-  init(_ instant: SuspendingClock.Instant) {
+    init(_ instant: SuspendingClock.Instant) {
```
stmontgomery: accidental indents

Sources/Testing/Attachments/Images/AttachableAsImage.swift
```diff
@@ -55,7 +55,6 @@
 #if SWT_NO_IMAGE_ATTACHMENTS
 @available(*, unavailable, message: "Image attachments are not available on this platform.")
 #endif
-@available(_uttypesAPI, *)
 public protocol AttachableAsImage {
```
stmontgomery, line 59: I think it would be nice to _keep_ these `@available` attributes which are attached to `public`, non-`@_spi` declarations so that in our rendered documentation we'll continue to show the OS versions. I'd find that useful purely for historical documentation purposes, even though the raised deployment targets will mean they're no longer necessary to build the project successfully.

grynspan: I can add them back. I'll note we don't list similar requirements for other bits of platform API that were available with Swift Testing's original deployment targets, and the presence here would be moot since you won't be able to use an older Darwin OS with this version of Swift Testing anyway.

Sources/Overlays/_Testing_WinSDK/Attachments/AttachableAsIWICBitmapSource.swift
```diff
@@ -104,7 +104,7 @@ extension AttachableAsIWICBitmapSource {
   /// used by types that conform to ``AttachableAsIWICBitmapSource``.
   ///
   /// For documentation, see ``AttachableAsImage/withUnsafeBytes(as:_:)``.
-  public func withUnsafeBytesImpl<R>(as imageFormat: AttachableImageFormat, _ body: (UnsafeRawBufferPointer) throws -> R) throws -> R {
+  func withUnsafeBytesImpl<R>(as imageFormat: AttachableImageFormat, _ body: (UnsafeRawBufferPointer) throws -> R) throws -> R {
```
grynspan, line 107: Drive-by fix: this function is Windows-only, is an implementation detail, and need not be `public`.

Sources/Testing/Events/Clock.swift
```diff
@@ -21,21 +21,7 @@ extension Test {
...
-        // SEE: https://github.com/swiftlang/swift/blob/main/stdlib/public/Concurrency/Clock.cpp
-        var uptime = timespec()
-        _ = clock_gettime(CLOCK_UPTIME_RAW, &uptime)
-        return TimeValue(uptime)
-#else
-        /// The corresponding suspending-clock time.
-        TimeValue(SuspendingClock.Instant.now)
-#endif
-      }()
+      fileprivate(set) var suspending = TimeValue(SuspendingClock.Instant.now)
```
stmontgomery, line 24: Yay!

Sources/Testing/Test.swift
```diff
@@ -465,7 +465,6 @@ extension Test {
     private var _timeLimit: TimeValue?
 
     /// The maximum amount of time a test may run for before timing out.
-    @available(_clockAPI, *)
     public var timeLimit: Duration? {
```
stmontgomery, line 468: Keep this one too? `// For DocC`

grynspan: Sure

PR #1535, opened 2026-02-08
Where possible, use `clonefile()` to save attachments.

grynspan:
This PR introduces a new underscored/default-implemented protocol requirement on `Attachable` that allows us to customize how we save attachable values to disk. Right now, the only custom implementation is on `_AttachableURLWrapper`:

- On Darwin, we call `clonefile()`
- On Linux, we call `ioctl(FICLONE)`
- On FreeBSD, we call `copy_file_range(COPY_FILE_RANGE_CLONE)`

On success, these operations reduce disk usage. On failure, we fall back to the default implementation using `fopen()` and `fwrite()`.

See also https://github.com/swiftlang/swift-foundation/issues/1727.

At this time, Xcode does not use this code path when saving attachments, but `swift test` and VS Code do use it.

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_Foundation/Attachments/_AttachableURLWrapper.swift
+++ b/Sources/Overlays/_Testing_Foundation/Attachments/_AttachableURLWrapper.swift
@@ -76,6 +76,7 @@ extension _AttachableURLWrapper: AttachableWrapper {
     try data.withUnsafeBytes(body)
   }
 
+#if !SWT_NO_FILE_CLONING
   /// Use platform-specific file-cloning API to create a copy-on-write copy of
   /// the represented file.
   ///
@@ -86,7 +87,6 @@ extension _AttachableURLWrapper: AttachableWrapper {
   ///
   /// - Throws: If a file exists at `filePath`, throws `EEXIST`.
   private func _clone(toFileAtPath filePath: String) throws -> Bool {
-#if !SWT_NO_FILE_CLONING
     return try filePath.withCString { destinationPath throws in
       var fileCloned = false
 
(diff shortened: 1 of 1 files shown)


grynspan:
@michael-yuji Could you please verify the FreeBSD code here builds?

michael-yuji:
`COPY_FILE_RANGE_CLONE` only appears in FreeBSD 15 afaik, we probably need to find a fallback for 14.3

grynspan:
That's okay—I can put it in a stub function in C with a version check and just bail to the existing fallback path on 14.3.

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -126,6 +126,33 @@ static char *_Nullable *_Null_unspecified swt_environ(void) {
...
+///
+/// This function is provided because `COPY_FILE_RANGE_CLONE` is not available
+/// prior to FreeBSD 15.0. The caller should check `getosreldate()` before using
+/// this flag.
+static unsigned int swt_COPY_FILE_RANGE_CLONE(void) {
+#if defined(COPY_FILE_RANGE_CLONE)
+  return COPY_FILE_RANGE_CLONE;
+#else
+  // Compiled against an older unistd.h, but presumably running on FreeBSD 15.0
+  // or newer. SEE: https://github.com/freebsd/freebsd-src/blob/main/sys/sys/unistd.h
```
harlanhaskins, line 150: 15.0 seems pretty new, does Swift only support 15+? What happens if this flag is passed to pre-15?

grynspan: We currently build for 14.3. 15.0 is very new. If you pass the flag in earlier versions, the result is unspecified, hence the runtime check on the Swift side.

Sources/Overlays/_Testing_Foundation/Attachments/_AttachableURLWrapper.swift
```diff
@@ -26,6 +28,23 @@ public struct _AttachableURLWrapper: Sendable {
...
+
+#if !SWT_NO_FILE_CLONING
+  /// A file handle that refers to the original file (or, if a directory, the
+  /// compressed copy thereof).
+  ///
+  /// This file handle is used when cloning the represented file.
+  private var _fileHandle: FileHandle
+#endif
+
+  init(url: URL, copiedToFileAt copyURL: URL? = nil, isCompressedDirectory: Bool) throws {
```
stmontgomery, line 51: Documentation?

Sources/Testing/Test+Macro.swift
```diff
@@ -90,9 +89,21 @@ public macro Suite(
...
+///   @Suite struct `Food Truck Tests` {
+///     // ...
+///   }
+///   ```
+/// }
+///
 /// ## See Also
 ///
 /// - <doc:OrganizingTests>
+@available(swift, deprecated: 100000.0, message: "")
```
stmontgomery: Are these changes to the `Test` macro declaration meant for another PR?

Sources/Testing/Attachments/Attachment.swift
```diff
@@ -429,13 +442,12 @@ extension Attachment where AttachableValue: ~Copyable {
 
     let preferredName = usingPreferredName ? preferredName : Self.defaultPreferredName
 
-    var file: FileHandle?
     do {
       // First, attempt to create the file with the exact preferred name. If a
       // file exists at this path (note "x" in the mode string), an error will
       // be thrown and we'll try again by adding a suffix.
```
stmontgomery: This comment about file mode strings may need to be updated

Sources/Testing/Test+Macro.swift
grynspan: I'm actually not sure where they came from!

Sources/Testing/Attachments/Attachment.swift
grynspan: How so?

stmontgomery: `(note "x" in the mode string)` for example, there's no longer a mode string in the code below

PR #1537, opened 2026-02-10
Squash the Foundation and `swift_willThrow` backtraces together if we have both.

grynspan:
This PR does something a little weird: on Darwin, for `NSError` instances that are created (and thus have a backtrace captured by Foundation) and later thrown through Swift (and thus have one captured by `swift_willThrow`), squash them together to produce a combined franketrace.

We have no way to know at our layer which backtrace is more salient. In some cases, test authors want to see in Swift where an error originated, while in other cases that backtrace is heavily truncated and the source location down in Foundation is more useful.

So provide both?

If the two backtraces have a common suffix (likely if they occur in a synchronous call), we deduplicate it so that you end up with a backtrace that looks like:

```
SWIFT_WILLTHROW_BASED
SWIFT_WILLTHROW_BASED
SWIFT_WILLTHROW_BASED
FOUNDATION_BASED
FOUNDATION_BASED
COMMON
COMMON
COMMON
...
```

This actually makes sense when you see it, because you see a backtrace that shows the error being created and then being thrown.

If the two operations occur asynchronously in a way that doesn't use Swift concurrency (e.g. using a dispatch queue or good ol' `NSThread`), the backtraces likely won't have a common prefix of more than a few Mach-level symbols, so you end up with, chronologically, the Foundation backtrace followed by the `swift_willThrow` backtrace in a single array.

And finally, if only one backtrace or the other is available, you end up with that backtrace verbatim.

Resolves rdar://170081869.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/Backtrace.swift
+++ b/Sources/Testing/SourceAttribution/Backtrace.swift
@@ -440,6 +440,11 @@ extension Backtrace {
 #endif
   }
 
+#if !hasFeature(Embedded) && SWT_TARGET_OS_APPLE && !SWT_NO_DYNAMIC_LINKING
+  private static var _markerAddressBetweenSwiftWillThrowAndFoundation: Address {
+    0x6C7B8BEB4B611F10 // randomly generated
+  }
+#endif
   /// Initialize an instance of this type with the previously-cached backtrace
   /// for a given error.
   ///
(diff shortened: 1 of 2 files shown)


Sources/Testing/SourceAttribution/Backtrace.swift
```diff
@@ -440,6 +440,11 @@ extension Backtrace {
 #endif
   }
 
+#if !hasFeature(Embedded) && SWT_TARGET_OS_APPLE && !SWT_NO_DYNAMIC_LINKING
+  private static var _markerAddressBetweenSwiftWillThrowAndFoundation: Address {
```
briancroom, line 444: Unused?

grynspan: Yeah I ended up not needing it.

PR #1541, opened 2026-02-11
Revert `RawSpan` adoption again.

grynspan:
We've been advised that libraries that back-deploy cannot safely adopt `RawSpan` because the compiler doesn't have a way to convey to clients that they also have a dependency on it and might require the compatibility library.

Reverting for now until we can resolve this issue.

### Checklist:


grynspan:
Squashing rather than merging so that in the future a revert of the revert is easier I guess. The repo history is already iffy here.

PR #1542, opened 2026-02-11
Check the XCTest fallback handler *after* checking for other configurations.

grynspan:
The new interop code orders the fallback handler _before_ the `Configuration.all` check, meaning that issues recorded in detached tasks _always_ fall back to XCTest. This can cause spurious test passage because these issues get recorded as warnings by default. (Our own tests are broken as a result).

Resolves rdar://170161483.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Event.swift
+++ b/Sources/Testing/Events/Event.swift
@@ -351,7 +351,7 @@ extension Event {
         // handler instead.
         _ = postToFallbackHandler(in: context)
       } else {
-        for configuration in Configuration.all {
+        for configuration in configurations {
           _post(in: context, configuration: configuration)
         }
       }


Sources/Testing/Events/Event.swift
```diff
@@ -342,13 +342,18 @@ extension Event {
...
       // will be lost! Post it to every registered event handler to avoid that.
-      for configuration in Configuration.all {
-        _post(in: context, configuration: configuration)
+      let configurations = Configuration.all
+      if configurations.isEmpty {
+        // There are no registered event handlers. Use the fallback event
+        // handler instead.
+        _ = postToFallbackHandler(in: context)
+      } else {
+        for configuration in Configuration.all {
```
jerryjrchen: Nit: would it be better to use the same `configurations` checked above?
```suggestion
        for configuration in configurations {
```

jerryjrchen (approved):

Sources/Testing/Events/Event.swift
grynspan: D'oh

PR #1547, opened 2026-02-12
Remove `Attachable._write()`.

grynspan:
This PR removes the recently-introduced experimental `_write()` requirement on `Attachable` and replaces it with a package-only `FileClonable` protocol that handles the clone operation. We're not prepared to expose a `write()` method on `Attachable` at this time and there are platform inconsistencies that make it difficult to support file cloning with a consistent API.

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_Foundation/Attachments/_AttachableURLWrapper.swift
+++ b/Sources/Overlays/_Testing_Foundation/Attachments/_AttachableURLWrapper.swift
@@ -126,7 +126,7 @@ extension _AttachableURLWrapper: FileClonable {
     return 0 == fclonefileat(srcFD, AT_FDCWD, filePath, 0)
 #elseif os(Linux) || os(FreeBSD)
     // Open the destination file for exclusive writing.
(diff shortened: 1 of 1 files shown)


Sources/Overlays/_Testing_Foundation/Attachments/_AttachableURLWrapper.swift
```diff
@@ -55,8 +54,11 @@ public struct _AttachableURLWrapper: Sendable {
     self.url = url
     self.data = try Data(contentsOf: copyURL ?? url, options: [.mappedIfSafe])
     self.isCompressedDirectory = isCompressedDirectory
-#if !SWT_NO_FILE_CLONING
-    self._fileHandle = try FileHandle(forReadingFrom: copyURL ?? url)
+#if !SWT_NO_FILE_CLONING && !os(Windows)
+    if let fileHandle = try? FileHandle(forReadingFrom: copyURL ?? url) {
```
grynspan: If we fail to create the file handle, okay, not the end of the world, we can probably still do non-clone-based writes. (I do throw if `setFD_CLOEXEC()` fails because that's really an extraordinary condition and I don't think can happen in practice.)

PR #1531, opened 2026-02-05
Add ability to filter/skip by tags using a special `tag:` prefix

gmedori:
While Xcode has had the ability to filter based on tags, this ability has been missing from the command line. This PR introduces a special prefix you can apply to your `--filter` and `--skip` option arguments: `tag:`.

```swift
extension Tag {
  @Tag static var integrationTest: Self
}
```

If you then add that tag to some suites or tests, you'll be able to run those tests exclusively by running:

```
swift test --filter tag:integrationTest
```

Resolves #591.
Resolves rdar://132989780.

https://github.com/user-attachments/assets/77dca6df-f010-4438-9016-7a25b3511426

### Modifications:

This PR enhances the nested `testFilter` function inside of `configurationForEntryPoint(from:)` function. The `testFilter` function now:

- Processes any arguments prefixed with `tag:` as regexes to be matched against tags
- Handles escaping the `tag:` prefix with the following syntax: `tag\:myRegex` to force the entire string to be read as a regex (as it would be normally prior to this change). This allows for filtering on test names that contain `tag:`.
- Continues to process regexes as it did before this change, once all the tag-based arguments are processed.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1548, opened 2026-02-13
Update `Test.traits` setter to filter applicable values rather than crashing when non-applicable values are encountered

### LINKED ISSUE #1048: Recursive Suite Traits must conform to TestTrait or else they crash
### Description

If a recursive SuiteTrait type fails to also conform to TestSuite, you get a runtime crash (apparently an assertion failure)

See [this Swift.org discussion](https://forums.swift.org/t/recursive-suite-traits-must-conform-to-testtrait-or-else-they-crash/79037)

### Reproduction

```swift
// DANGER: missing TestSuite conformance
struct TestConfiguration: SuiteTrait, TestScoping {
    let isRecursive: Bool = true
    func provideScope(for test: Test, testCase: Test.Case?, performing function: @Sendable () async throws -> Void) async throws {
        try await function()
    }
}

extension Trait where Self == TestConfiguration {
    static var testConfiguration: Self { Self() }
}

@Suite(.testConfiguration)
struct DemoSuite {
    @Test func demo1() {}
```

### Expected behavior

Ideally, some sort of catastrophic but non-crashing result. Or, at least, a clearer indication as to what's wrong

### Environment

Xcode 16.3

Testing Library Version: 124

### Additional information

_No response_

gmedori:
Prior to this PR, when a developer would define their own `SuiteTrait` and provide an implementation for `isRecursive` that resolved to `true`, the test process would crash unless they also had the trait conform to `TestTrait`. This happened because we consider a non-suite `Test` having a `SuiteTrait` to be invalid, so we placed a precondition failure in the setter for `Test.traits`. When we would recursively attempt to apply the trait, it would do so for all sub-suites _and_ tests, tripping the precondition.

While this is technically correct behavior, the failure mode is a poor experience that doesn't adequately communicate what went wrong. Here's an example of the log when the crash happens:

```
Build complete! (0.79s)
Test Suite 'All tests' started at 2026-02-12 17:40:55.984.
Test Suite 'All tests' passed at 2026-02-12 17:40:55.985.
         Executed 0 tests, with 0 failures (0 unexpected) in 0.000 (0.002) seconds
error: Process 'swiftpm-testing-helper --test-bundle-path .build/arm64-apple-macosx/debug/recursive-suite-traitsPackageTests.xctest/Contents/MacOS/recursive-suite-traitsPackageTests .build/arm64-apple-macosx/debug/recursive-suite-traitsPackageTests.xctest/Contents/MacOS/recursive-suite-traitsPackageTests --testing-library swift-testing' exited with unexpected signal code 5
```

To avoid this crash and to allow for previously infeasible trait configurations, this change updates the `Test.traits` setter to:

1. Ditch the precondition failure
2. Instead, only _actually_ set valid values

As a short cookbook, we then have the following:

| As a developer I want to… | Recipe |
| --- | --- |
| … write a trait that only applies to a suite and its sub-suites | Have your trait conform to `SuiteTrait` and set `isRecursive` to `true` (the previous crash scenario). The nested tests will ignore the trait when we attempt to set it on them. |
| … write a trait that applies to a suite, its sub-suites, _and_ tests | Have your trait conform to both `SuiteTrait` **and** `TestTrait`, and set `isRecursive` to `true`. With the extra conformance, nested tests will pick up the trait. |

Resolves #1048 

### Checklist:

### DIFF
--- a/Tests/TestingTests/Test.TraitTests.swift
+++ b/Tests/TestingTests/Test.TraitTests.swift
@@ -1,7 +1,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
--- a/Tests/TestingTests/Test.TraitTests.swift
+++ b/Tests/TestingTests/Test.TraitTests.swift
@@ -19,16 +19,17 @@ private struct TestOnlyTrait: TestTrait {}
 /// A trait that applies to both
 private struct TestAndSuiteTrait: SuiteTrait, TestTrait {}
 
-/// A stub test suite.
-@Suite
-private struct DemoTestSuite {}
-
 @Suite("Test.Trait Tests")
 struct Test_TraitTests {
   @Test("Setting traits on a suite filters out non-suite traits", .tags(.traitRelated))
   func filterTraitsOnSuites() async throws {
     let mixedTraits: [any Trait] = [SuiteOnlyTrait(), TestOnlyTrait(), TestAndSuiteTrait()]
-    var suite = try #require(await test(for: DemoTestSuite.self))
+    var suite = Test(
+      displayName: "Fake Suite",
+      traits: [],
+      sourceLocation: #_sourceLocation,
+      containingTypeInfo: TypeInfo(describing: Void.self),
+    )
 
     // Setting the traits property should filter out non-suite traits
     suite.traits = mixedTraits
--- a/Tests/TestingTests/Test.TraitTests.swift
+++ b/Tests/TestingTests/Test.TraitTests.swift
@@ -1,0 +1,52 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+@testable @_spi(Experimental) @_spi(ForToolsIntegrationOnly) import Testing
+
+/// A trait that only applies to suites
+private struct SuiteOnlyTrait: SuiteTrait {}
+
+/// A trait that only applies to test functions
+private struct TestOnlyTrait: TestTrait {}
+
+/// A trait that applies to both
+private struct TestAndSuiteTrait: SuiteTrait, TestTrait {}
+
+/// A stub test suite.
+@Suite
+private struct DemoTestSuite {}
+
+@Suite("Test.Trait Tests")
+struct Test_TraitTests {
+  @Test("Setting traits on a suite filters out non-suite traits", .tags(.traitRelated))
+  func filterTraitsOnSuites() async throws {
+    let mixedTraits: [any Trait] = [SuiteOnlyTrait(), TestOnlyTrait(), TestAndSuiteTrait()]
+    var suite = try #require(await test(for: DemoTestSuite.self))
+
+    // Setting the traits property should filter out non-suite traits
+    suite.traits = mixedTraits
+    #expect(suite.traits.count == 2)
+    #expect(suite.traits[0] is SuiteOnlyTrait)
+    #expect(suite.traits[1] is TestAndSuiteTrait)
+  }
+
+  @Test("Setting traits on a test filters out non-test traits", .tags(.traitRelated))
+  func filterTraitsOnTests() async throws {
+    let mixedTraits: [any Trait] = [SuiteOnlyTrait(), TestOnlyTrait(), TestAndSuiteTrait()]
+    var test = Test {}
+
+    // Setting the traits property should filter out non-test traits
+    test.traits = mixedTraits
+    #expect(test.traits.count == 2)
+    #expect(test.traits[0] is TestOnlyTrait)
+    #expect(test.traits[1] is TestAndSuiteTrait)
+  }
+
+}
--- a/Tests/TestingTests/Test.TraitTests.swift
+++ b/Tests/TestingTests/Test.TraitTests.swift
@@ -19,10 +19,10 @@ private struct TestOnlyTrait: TestTrait {}
 /// A trait that applies to both
 private struct TestAndSuiteTrait: SuiteTrait, TestTrait {}
 
-@Suite("Test.Trait Tests")
-struct Test_TraitTests {
-  @Test("Setting traits on a suite filters out non-suite traits", .tags(.traitRelated))
-  func filterTraitsOnSuites() async throws {
+@Suite
+struct `Test.Trait Tests` {
+  @Test(.tags(.traitRelated))
+  func `Setting traits on a suite filters out non-suite traits`() async throws {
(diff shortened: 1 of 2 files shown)


grynspan (approved):
Works for me. I'd like at least one of my colleagues to also review before we merge it. :)

grynspan (commented):
Oh, we need some test coverage!

gmedori:
Good shout! I'm also adding some test coverage over in #1531. I'll get cooking on those.

stmontgomery:
Thank you @gmedori for taking this on!

I recall @grynspan proposing this solution once in the past, and back then, I raised a concern that there was one potential use case that this solution would make impossible:

> **As a developer I want too…**
> …write a trait that can be **explicitly** applied to a suite, will be inherited by its sub-suites and tests, _but_ **cannot** be explicitly applied to a test.

It’s fair to ask whether that's a use case that matters or is useful. So far, the only speculative use case I have come up with is a trait which only conforms to `SuiteTrait` and `TestScoping` and is written such that it performs some more expensive scoping work once for the entire suite, but then also does some other, cheaper work once per child test. The author of such a trait might see it as useful to expressly prohibit it from being written explicitly on a test, forcing it to always be written on a suite even though it will inherit to everything.

Some time has passed since `TestScoping` was introduced though and this need hasn't come up organically, so I think we can reasonably conclude that use case isn't that important. I did want to at least describe it here for awareness in case it comes up in the future though. I'll keep thinking about this a bit more but at the moment I'm inclined to proceed with this approach (and finally put this issue to rest!)

grynspan:
@gmedori actually independently raised that question with me in DMs. As I replied there:


> I would say this behaviour is unintentional but I don't think we designed for it since this would be a broken trait previously.

gmedori:
@stmontgomery thanks for the thoughtful comment!

> So far, the only speculative use case I have come up with is a trait which only conforms to `SuiteTrait` and `TestScoping` and is written such that it performs some more expensive scoping work once for the entire suite, but then also does some other, cheaper work once per child test.

Yup as Jonathan said, this was the exact scenario I thought of that is basically unsupported by this change. This scenario was described in [a forum post](https://forums.swift.org/t/recursive-suite-traits-must-conform-to-testtrait-or-else-they-crash/79037) and was the impetus for the original issue being filed, so I wanted to at least consider it. When I brought it up, I was actually concerned that where we used to crash, now we just silently ignore. But I convinced myself that the behavior merely matches the semantics of the declaration. That is, if you make your trait conform to `SuiteTrait`, then it applies to suites. If you make it conform to `TestTrait` then it applies to tests. And that just kind of makes sense, and no longer appears to me as "failing silently".

When reading the original forum post, it seemed like the original issue author brought it up not as a production issue, but rather an educational one. And if we haven't seen more concern/discussion about it, I'd agree that this use case is unlikely to be important, which makes me more bullish on this approach.

grynspan:
If we think there's a practical issue here, we could generate a `.apiMisused` issue (with `.warning` severity, I think) about it when we detect it has happened.

gmedori:
Yeah that was what I was thinking when I posted https://github.com/swiftlang/swift-testing/issues/1048#issuecomment-3887855206 in the original issue. The more that I think about it, though, the more that I'm feeling that the thing we want to detect is not really detectable. Specifically, this is the scenario that we would consider "invalid":

```swift
struct RecursiveTrait: SuiteTrait, TestScoping { // <-- ⚠️ No TestTrait conformance
  let isRecursive: Bool = true

  func provideScope(for test: Test, testCase: Test.Case?, performing function: () async throws -> Void) async throws {
    if test.isSuite {
      // 1. Do expensive work before the suite
      // ...
      // 2. Then run the test suite
      try await function()
    } else {
      // Do cheap work before each test
      // ❌ Will never happen because this trait is never applied to tests
    }
  }

  func scopeProvider(for test: Test, testCase: Test.Case?) -> RecursiveTrait? {
    return Self() // Provide a non-nil value always, regardless of suite or test
  }
}
```

I'm still a bit new to all this so please correct me if I'm missing something, but my understanding is that it's perfectly valid to have a `SuiteTrait` that conforms to `TestScoping` without also conforming to `TestTrait`. The issue is when, at runtime, the developer might want to do something when `test.isSuite == false`, not realizing it won't fire without the trait conforming to `TestTrait` and we can't really detect that.

...

Can we? (⚠️ WARNING: Idle musings follow below, feel free to ignore)

So I'll preface this by saying I don't think the juice is worth the squeeze, but purely as a thought exercise, we could build in some sort of alerting mechanism inside of the getter for `Test.isSuite` that we prime right before we're about to call `provideScope`. Then we check if it was tripped on any traits that conform to the types we care about. 

But that sounds super complex, hacky, and fragile. Furthermore the most refined signal we could get is whether `isSuite` was accessed, not whether it was negated, part of an if/else, or any information that would tell us how the user branched. We would effectively be saying "don't check `isSuite` inside of `provideScope`", which I don't think is fair.

Tests/TestingTests/Test.TraitTests.swift
```diff
@@ -0,0 +1,52 @@
...
+/// A stub test suite.
+@Suite
+private struct DemoTestSuite {}
+
+@Suite("Test.Trait Tests")
+struct Test_TraitTests {
+  @Test("Setting traits on a suite filters out non-suite traits", .tags(.traitRelated))
+  func filterTraitsOnSuites() async throws {
+    let mixedTraits: [any Trait] = [SuiteOnlyTrait(), TestOnlyTrait(), TestAndSuiteTrait()]
+    var suite = try #require(await test(for: DemoTestSuite.self))
```
gmedori: I wasn't sure what the right incantation was for getting an instance of a test suite, then stumbled across this function. Is this the way?

grynspan: Depends what you need it for. If you just want to mess with the instance, you can create one with:

```swift
var suite = Test(
  displayName: "Fake Suite",
  traits: [],
  sourceLocation: #_sourceLocation,
  containingTypeInfo: TypeInfo(describing: Void.self)
)
```

Tests/TestingTests/Test.TraitTests.swift
```diff
@@ -0,0 +1,52 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
```
grynspan: Nitpick: 2026

Tests/TestingTests/Test.TraitTests.swift
gmedori: I like this inline approach better than declaring a demo suite at the top level just to get an instance of it one time.

9432a50

Tests/TestingTests/Test.TraitTests.swift
gmedori: Resolved in 6853de3

grynspan:
@stmontgomery You okay with merging this PR once conflicts are resolved?

Tests/TestingTests/Test.TraitTests.swift
```diff
@@ -0,0 +1,53 @@
...
+/// A trait that only applies to suites
+private struct SuiteOnlyTrait: SuiteTrait {}
+
+/// A trait that only applies to test functions
+private struct TestOnlyTrait: TestTrait {}
+
+/// A trait that applies to both
+private struct TestAndSuiteTrait: SuiteTrait, TestTrait {}
+
+@Suite("Test.Trait Tests")
```
grynspan: Nitpick: let's use raw identifiers for new test suites going forward.

gmedori: Resolved in 38daf97

grynspan:
Where are we with this PR? Do we want to move forward?

gmedori:
Happy to go back and address any issues!

PR #1550, opened 2026-02-13
Make `swt_getTestingLibraryVersion()` `constexpr`.

grynspan:
This PR changes the implementation of `swt_getTestingLibraryVersion()` so that it computes the end of the string (read from a file using `#embed`) at compile time using `constexpr` instead of at runtime using `std::call_once()`. Minor performance win avoiding a heavy-ish synchronization primitive on data that never changes.

This implementation is only used when we build as a package. When building in the toolchain, the version string is supplied by CMake.

### Checklist:

### DIFF
--- a/Sources/_TestingInternals/Versions.cpp
+++ b/Sources/_TestingInternals/Versions.cpp
@@ -22,24 +22,29 @@ const char *swt_getTestingLibraryVersion(void) {
   return SWT_TESTING_LIBRARY_VERSION;
 #elif __clang_major__ >= 17 && defined(__has_embed)
 #if __has_embed("../../VERSION.txt")
-  // Read the version from version.txt at the root of the package's repo.
-  static char version[] = {
+  static constexpr std::array result = [] () {
+    // Read the version from version.txt at the root of the package's repo.
+    constexpr const char version[] = {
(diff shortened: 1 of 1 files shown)


Sources/_TestingInternals/Versions.cpp
```diff
@@ -22,24 +22,29 @@ const char *swt_getTestingLibraryVersion(void) {
...
-  });
+    // Copy from the C string into a C++ array, stopping at the first newline if
+    // one is present.
+    std::array<char, std::size(version)> result {};
+    for (size_t i = 0; i < std::size(version); i++) {
+      char c = version[i];
+      if (c == '\r' || c == '\n') {
+        break;
+      }
+      result[i] = c;
```
harlanhaskins, line 42: Where does the NUL terminator get inserted?

harlanhaskins: Ah, std::array's initializer-list constructor defaults values, in this case zero. Neat.

grynspan: Fun fact, because this code runs at compile time, clang will actually emit an error if a value in the array is uninitialized.

PR #1552, opened 2026-02-13
Fix file cloning in attachments broken by #1547.

grynspan:
I broke file cloning in #1547, but I did it in a bunch of different ways that mostly cancelled each other out, so tests ended up passing spuriously.

### Checklist:


stmontgomery:
(For tracking) Caused by https://github.com/swiftlang/swift-testing/pull/1547

Sources/Overlays/_Testing_Foundation/Attachments/_AttachableURLWrapper.swift
```diff
@@ -116,6 +116,11 @@ extension _AttachableURLWrapper: AttachableWrapper {
 
 #if !SWT_NO_FILE_CLONING
 extension _AttachableURLWrapper: FileClonable {
+#if os(FreeBSD)
```
harlanhaskins, line 119: Where was this before?

grynspan: It wasn't a `static let`, it was a local call, but it's expensive.

PR #1554, opened 2026-02-16
Provide iteration index along with test[Case]{Started|Ended} events

harlanhaskins:
This adds an optional `_iteration` field to the `EncodedEvent` struct which currently corresponds to the value provided via `iterationStarted`/`iterationEnded` events.

### Motivation:

Essentially, this provides the current iteration of the `testStarted`, `testCaseStarted`, `testEnded`, and `testCaseEnded` events in the event context as an optional, experimental field. Clients that previously read `iterationStarted`/`iterationEnded` events can now just read the iteration index directly from the test event.

This is groundwork for being able to move test iteration down to the test case level, rather than re-iterating the whole test plan. Regardless which level of the runner the iteration happens, the value will be reported alongside the test case.

When the repetition behavior changes such that individual test cases are re-run, it should not need any changes to the test format (other than making the iteration value non-experimental). Ideally, once we make that change we can deprecate and stop sending `iterationStarted`/`iterationEnded` events.

### Modifications:

- Updates `Event.Context` to have an iteration value 
- Encodes that value in `EncodedEvent` when experimental fields are enabled
- Provides the global iteration value when `test[Case]{Started|Ended}` events are posted

### Checklist:

Fixes rdar://170345148

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
@@ -100,7 +100,7 @@ extension ABI {
 
     /// The iteration of the `testID` being executed.
     ///
-    /// Currently, this value is identical to the value provided via the most recent `iterationStarted` event.
+    /// This value is one-indexed; the first iteration is `1`.
     ///
     /// - Warning: Iteration indices are not yet part of the JSON schema.
     var _iteration: Int?
--- a/Sources/Testing/Events/Event.swift
+++ b/Sources/Testing/Events/Event.swift
@@ -299,7 +299,7 @@ extension Event {
     public var testCase: Test.Case?
 
     /// The iteration of the current test and test case.
-    /// The first iteration has an index of `0`.
+    /// This value is one-indexed; the first iteration is `1`.
     ///
     /// This value is not guaranteed to be provided for all events, but if this
     /// is provided, the `test` will be non-`nil`.
--- a/Sources/Testing/Running/Runner.swift
+++ b/Sources/Testing/Running/Runner.swift
@@ -508,7 +508,8 @@ extension Runner {
           }
           _ = taskGroup.addTaskUnlessCancelled(name: decorateTaskName("test run", withAction: taskAction)) {
             var iterationContext = context
-            iterationContext.iteration = iterationIndex
+            // `iteration` is one-indexed, so offset that here.
+            iterationContext.iteration = iterationIndex + 1
             try? await _runStep(atRootOf: runner.plan.stepGraph, context: iterationContext)
           }
           await taskGroup.waitForAll()
--- a/Tests/TestingTests/EventIterationTests.swift
+++ b/Tests/TestingTests/EventIterationTests.swift
@@ -24,14 +24,14 @@ struct EventIterationTests {
     testBody: @escaping @Sendable (Int) -> Void,
     location: SourceLocation = #_sourceLocation
   ) async {
-    let recordedIterationIndex = Mutex<Int>(0)
+    let recordedIteration = Mutex<Int>(0)
 
     await confirmation("Events received", expectedCount: expectedIterations * eventKinds.count, sourceLocation: location) { eventReceived in
       var configuration = Configuration()
       configuration.eventHandler = { event, context in
         if eventKinds.contains(where: { Self.matchesTestLifetimeEventKind($0, event.kind) }) {
           if let iteration = context.iteration {
-            recordedIterationIndex.withLock { $0 = iteration }
+            recordedIteration.withLock { $0 = iteration }
           }
           eventReceived()
         }
@@ -39,12 +39,12 @@ struct EventIterationTests {
       configuration.repetitionPolicy = repetitionPolicy
 
       await Test {
-        testBody(recordedIterationIndex.rawValue)
+        testBody(recordedIteration.rawValue)
       }.run(configuration: configuration)
 
       // Verify all expected iterations were recorded
-      let iterations = recordedIterationIndex.rawValue
-      #expect(iterations == expectedIterations - 1, "Final observed iteration index did not match expected number of iterations", sourceLocation: location)
+      let iteration = recordedIteration.rawValue
+      #expect(iteration == expectedIterations, "Final observed iteration did not match expected number of iterations", sourceLocation: location)
     }
   }
 
@@ -95,8 +95,8 @@ struct EventIterationTests {
       for: [.testCaseStarted, .testCaseEnded],
       repetitionPolicy: policy,
       expectedIterations: expectedIterations
-    ) { index in
-      if index < 2 {
+    ) { iteration in
+      if iteration < 3 {
         Issue.record("Failure")
       }
     }


Sources/Testing/Events/Event.swift
```diff
@@ -303,17 +304,31 @@ extension Event {
     /// ``Configuration/eventHandler`` property of this instance is cleared.
     public var configuration: Configuration?
 
+    /// The iteration of the current test and test case.
+    ///
+    /// This value is not guaranteed to be provided for all events, but if this
+    /// is provided, the `test` will be non-`nil`.
+    public var iteration: Int?
```
jerryjrchen, line 306: Nit: we could programatically check this in the init: `precondition(iteration == nil || (iteration != nil && test != nil))`

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -98,6 +98,13 @@ extension ABI {
     ///   part of said JSON schema.
     var _sourceLocation: EncodedSourceLocation<V>?
 
+    /// The iteration of the `testID` being executed.
+    ///
+    /// Currently, this value is identical to the value provided via the most recent `iterationStarted` event.
```
jerryjrchen: Could we inline the docs from `iterationStarted`? At least the part about this being zero-indexed:
```
    ///   - index: The index of the iteration. The first iteration has an index
    ///     of `0`.
```

Tests/TestingTests/EventIterationTests.swift
```diff
@@ -0,0 +1,106 @@
...
+         (.testCaseStarted, .testCaseStarted),
+         (.testCaseEnded, .testCaseEnded):
+      return true
+    default:
+      return false
+    }
+  }
+
+  @Test("testStarted and testEnded events include iteration in context")
+  func testEventsIncludeIteration() async {
```
jerryjrchen: Nit: I think we've been doing raw identifiers for tests in new test suites:
```suggestion
  func `testStarted and testEnded events include iteration in context`() async {
```

Tests/TestingTests/EventIterationTests.swift
```diff
@@ -0,0 +1,106 @@
...
+      }
+      configuration.repetitionPolicy = repetitionPolicy
+
+      await Test {
+        testBody(recordedIterationIndex.rawValue)
+      }.run(configuration: configuration)
+
+      // Verify all expected iterations were recorded
+      let iterations = recordedIterationIndex.rawValue
+      #expect(iterations == expectedIterations - 1, sourceLocation: location)
```
jerryjrchen: Comment suggestion to make it easier to triage this failure:
```suggestion
      #expect(iterations == expectedIterations - 1, sourceLocation: location, "Final observed iteration index did not match expected number of iterations")
```

Tests/TestingTests/EventIterationTests.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  private static func matchesKind(_ expected: Event.Kind, _ actual: Event.Kind) -> Bool {
```
jerryjrchen: 🥲

Tests/TestingTests/EventIterationTests.swift
harlanhaskins: Oh, nice!

Tests/TestingTests/EventIterationTests.swift
harlanhaskins: I could sink this down into a general utility on Event.Kind but it seemed like a lot of surface area

jerryjrchen: Just lamenting the current state of enum matching in Swift :P

Sources/Testing/Events/Event.swift
```diff
@@ -303,17 +304,34 @@ extension Event {
     /// ``Configuration/eventHandler`` property of this instance is cleared.
     public var configuration: Configuration?
 
+    /// The iteration of the current test and test case.
+    /// The first iteration has an index of `0`.
+    ///
+    /// This value is not guaranteed to be provided for all events, but if this
+    /// is provided, the `test` will be non-`nil`.
+    public var iteration: Int?
```
stmontgomery, line 306: This might seem arbitrary/superficial but I would place this new property after the `test` and `testCase` properties but before `configuration`. (In other words, shift it one property up in the list.) It seems more closely aligned with those two earlier properties which are concerned with identifying the "subject" of the event; whereas in contrast, the `configuration` property is providing more general info about the configuration in effect when the event was posted, rather than identifying any particular subject. To me, the ordering of the properties could help subtly reinforce that connection.

Sources/Testing/Events/Event.swift
```diff
@@ -303,17 +304,34 @@ extension Event {
...
     /// Initialize a new instance of this type.
     ///
     /// - Parameters:
     ///   - test: The test for which this instance's associated event occurred,
     ///     if any.
     ///   - testCase: The test case for which this instance's associated event
     ///     occurred, if any.
-    init(test: Test?, testCase: Test.Case?, configuration: Configuration?) {
+    ///   - configuration: The ``Configuration`` of the current test run.
+    ///   - iteration: The iteration of the associated test and case, if any.
```
stmontgomery, line 322: If you take the [comment](https://github.com/swiftlang/swift-testing/pull/1554/changes#r2814947771) about reordering the properties, I'd also match that order here in the docs, and in the parameter list.

Sources/Testing/Events/Event.swift
```diff
@@ -237,6 +237,7 @@ public struct Event: Sendable {
   static func post(
     _ kind: Kind,
     for testAndTestCase: (Test?, Test.Case?) = currentTestAndTestCase(),
+    iteration: Int? = nil,
```
stmontgomery, line 241: nit: Documentation for the new parameter

Sources/Testing/Events/Event.swift
```diff
@@ -296,6 +298,13 @@ extension Event {
     /// functions), the value of this property is `nil`.
     public var testCase: Test.Case?
 
+    /// The iteration of the current test and test case.
+    /// The first iteration has an index of `0`.
+    ///
+    /// This value is not guaranteed to be provided for all events, but if this
+    /// is provided, the `test` will be non-`nil`.
+    public var iteration: Int?
```
grynspan, line 306: This maybe needs `@_spi(Experimental)`? If you're not sure you want this to be semi-stable tools SPI.

Sources/Testing/Events/Event.swift
```diff
@@ -608,13 +627,19 @@ extension Event.Context {
     /// functions), the value of this property is `nil`.
     public var testCase: Test.Case.Snapshot?
 
+    /// A snapshot of the test iteration when the associated ``Event``
+    /// occurred, if any. This value is determined by the `repetitionPolicy`
+    /// applied to the test plan.
+    public var iteration: Int?
```
grynspan, line 633: As above, experimental?

Sources/Testing/Events/Event.swift
harlanhaskins: Good catch! I'll add this

harlanhaskins: Ah, since these are ForToolsIntegrationOnly, I think that implies Experimental

Sources/Testing/Running/Runner.swift
```diff
@@ -466,7 +469,7 @@ extension Runner {
     // accidentally depend on e.g. the `configuration` property rather than the
     // current configuration.
     let context: _Context = {
-      var context = _Context()
+      var context = _Context(iteration: 0)
```
stmontgomery, line 472: Should the iteration counting be 0- or 1-based? Only reason I ask is it could affect downstream clients; if it's 0-based and they intend to present it in UI showing test results, they may always need to increment by one.

harlanhaskins: I wanted to preserve the existing behavior from iterationStarted/iterationEnded, but if we think it would be good to switch the behavior to be more aligned with XCTest I am not opposed; this is a new field that can have new semantics

stmontgomery: To me it's less about aligning with XCTest and more about general pragmatism w.r.t. how the field will be used and interpreted. And to confirm, the old code didn't propagate a specific iteration _number_ to clients, right? So as you say, this is entirely new. It's not a strong opinion but I'd slightly lean towards having the count be 1-based.

harlanhaskins: I do agree that 1-indexing aligns better with the user's mental model ("this is the first iteration")

stmontgomery (approved):

PR #1555, opened 2026-02-17
Mark FileClonable behind SWT_NO_FILE_CLONING.

3405691582:
This protocol is guarded by SWT_NO_FILE_CLONING and so requires qualification before use, otherwise results in a build error.

### Motivation:

Buildfixes tests for OpenBSD.

### Modifications:

Added requisite guards for uses of FileClonable.

### Checklist:

- [ ] If public symbols are renamed or modified, DocC references should be updated. (n/a)


stmontgomery (approved):
Thanks!

PR #1560, opened 2026-02-17
Make `#expect()` ∞% faster.

grynspan:
This PR improves the performance of the `SourceLocation.__here()` macro-only function (as used by `#expect()`) by avoiding runtime precondition checks since the inputs are known to be generated by the compiler and therefore "good".

Given the following test function:

```swift
@Test func bigLoop() {
  for _ in 0 ..< 100_000_000 {
    _ = SourceLocation.__here()
  }
}
```

Prior to this change, running it produced this result:

> Test bigLoop() started.
> Test bigLoop() passed after 22.089 seconds.
> Test run with 1 test in 0 suites passed after 22.089 seconds.

After this change, it produced this result:

> Test bigLoop() started.
> Test bigLoop() passed after 10.807 seconds.
> Test run with 1 test in 0 suites passed after 10.807 seconds.

When profiled, the entire measurable time spent inside the function is just spent in the loop (since the test was built with `-Onone`):

```
819  bigLoop() + 232 (ZipTests.swift in TestingTests + 4796684) [0x10ccc710c]
  598  IndexingIterator.next() + 620 (Collection.swift:128 in libswiftCore.dylib + 1286080) [0x1952eefc0]
  59   IndexingIterator.next() + 584 (<compiler-generated> in libswiftCore.dylib + 1286044) [0x1952eef9c]
  51   IndexingIterator.next() + 452 (Collection.swift:127 in libswiftCore.dylib + 1285912) [0x1952eef18]
  6    Range<>.endIndex.getter + 28 (Range.swift:234 in libswiftCore.dylib + 2918772) [0x19547d974] (running)
  6    swift::metadataimpl::ValueWitnesses<swift::metadataimpl::NativeBox<unsigned long long, 8ul, 8ul, 8ul>>::initializeWithCopy(swift::OpaqueValue*, swift::OpaqueValue*, swift::TargetMetadata<swift::InProcess> const*) + 4 (MetadataImpl.h:788 in libswiftCore.dylib + 60416) [0x1951c3c00] (running)
  6    swift_getAssociatedTypeWitness + 28 (Metadata.cpp:7170 in libswiftCore.dylib + 41868) [0x1951bf38c] (running)
  // ...
```

Rewriting the test to avoid unoptimized `IteratorProtocol` overhead (but still `-Onone`) as:

```swift
@Test func bigLoop() {
  var i = 0
  while i < 100_000_000 {
    _ = SourceLocation.__here()
    i += 1
  }
}
```

Produced this output:

> Test bigLoop() started.
> Test bigLoop() passed after 1.032 seconds.
> Test run with 1 test in 0 suites passed after 1.033 seconds.

With the remaining time spent in mandatory retain/release traffic.

Resolves rdar://150474666.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/SourceLocation.swift
+++ b/Sources/Testing/SourceAttribution/SourceLocation.swift
@@ -24,7 +24,7 @@ public struct SourceLocation: Sendable {
   public var fileID: String {
     willSet {
       precondition(!newValue.isEmpty, "SourceLocation.fileID must not be empty (was \(newValue))")
-      precondition(newValue.contains("/"), "SourceLocation.fileID must be a well-formed file ID (was \(newValue))")
+      precondition(newValue.utf8.contains(UInt8(ascii: "/")), "SourceLocation.fileID must be a well-formed file ID (was \(newValue))")
     }
   }
 


harlanhaskins:
Very nice!

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
```diff
@@ -30,6 +30,17 @@ extension SourceLocation {
     line: Int = #line,
     column: Int = #column
   ) -> Self {
-    Self(fileID: fileID, filePath: filePath, line: line, column: column)
+    Self(__uncheckedFileID: fileID, filePath: filePath, line: line, column: column)
+  }
+
+  /// Initialize an instance of this type without validating any arguments.
+  ///
+  /// - Warning: This initializer is used to implement the `#_sourceLocation`
```
stmontgomery, line 38: Is this actually used for `#_sourceLocation`, or for other macro expansions?
```
This initializer is used to implement the `#_sourceLocation`
```

grynspan: Both.

SeanROlszewski:
prettttty spiffy @grynspan

Sources/Testing/SourceAttribution/SourceLocation.swift
```diff
@@ -113,14 +113,11 @@ public struct SourceLocation: Sendable {
   /// - Precondition: `column` must be greater than `0`.
   public init(fileID: String, filePath: String, line: Int, column: Int) {
     precondition(!fileID.isEmpty, "SourceLocation.fileID must not be empty (was \(fileID))")
-    precondition(fileID.contains("/"), "SourceLocation.fileID must be a well-formed file ID (was \(fileID))")
+    precondition(fileID.utf8.contains(UInt8(ascii: "/")), "SourceLocation.fileID must be a well-formed file ID (was \(fileID))")
```
younata, line 116: This appears to be an unrelated optimization to the `init(__uncheckedFileID:...)` that only macros are supposed to use. Is that reading correct?

I assume this also improves the speed of this initializer?

jerryjrchen (approved):

Sources/Testing/SourceAttribution/SourceLocation.swift
grynspan: This improves performance significantly when we still want to do the precondition checks, yeah. We discussed privately with some core Swift folks and we're fairly certain that the distinction between UTF-8 `"/"` and grapheme cluster `"/"` is not going to be important here (i.e. if you have `"/"` with a combining character so that it passes this check, it'll end up failing later anyway, and also nobody constructs module names or file names that way in the real world.)

stephentyrone:
Thanks!

PR #1566, opened 2026-02-17
Ensure the old overload of `Issue.record()` is still documented.

grynspan:
The new overload of `Issue.record()` that takes a `severity` argument was introduced in Swift 6.3. We need to continue to document the older overload.

### Checklist:


stmontgomery (approved):
Fine with me, although can you elaborate on the justification? Is it so that the prior overload remains in the documentation so that its Swift version API availability (or lack thereof) is shown? That makes sense, just confirming

grynspan:
Yeah, right now the published docs make it look like issue recording was added in Swift 6.3. :)

PR #1572, opened 2026-02-18
Show subexpressions separately in expectation failures.

grynspan:
This PR refactors how we capture and display subexpressions when an expectation fails. For example, given the following test:

```swift
@Test func foo() {
  let lhs = "abc123"
  let rhs = "d"
  #expect(lhs.contains(rhs))
}
```

We currently output:

```
< > Test foo() started.
<X> Test foo() recorded an issue at FileName.swift:X:Y: Expectation failed: (lhs → "abc123").contains(rhs → "d")
<X> Test foo() failed after 0.003 seconds with 1 issue.
```

This can be hard to read for more complex expressions as the real values are nested inside a single line. With this change, we instead output:

```
< > Test foo() started.
<X> Test foo() recorded an issue at FileName.swift:X:Y: Expectation failed: lhs.contains(rhs)
--> lhs.contains(rhs) → false
-->   lhs → "abc123"
-->   rhs → "d"
<X> Test foo() failed after 0.003 seconds with 1 issue.
```

This change also adds the subexpression breakdown to our JSON schema as an experimental field of the "issue" event record:

```json
{
   "kind" : "event",
   "payload" : {
      /* ... */
      "issue" : {
         /* ... */
         "_expectation" : {
            "_expression" : {
               "children" : [
                  {
                     "runtimeTypeName" : "Swift.String",
                     "runtimeValue" : "\"abc123\"",
                     "sourceCode" : "lhs"
                  },
                  {
                     "runtimeTypeName" : "Swift.String",
                     "runtimeValue" : "\"d\"",
                     "sourceCode" : "rhs"
                  }
               ],
               "runtimeTypeName" : "Swift.Bool",
               "runtimeValue" : "false",
               "sourceCode" : "lhs.contains(rhs)"
            }
         },
         /* ... */
      },
   },
   "version" : "99.0.0"
}
```

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -292,7 +292,7 @@ extension Issue.Kind: CustomStringConvertible {
       } else if let mismatchedExitConditionDescription = expectation.mismatchedExitConditionDescription {
         "Expectation failed: \(mismatchedExitConditionDescription)"
       } else {
-        "Expectation failed: \(expectation.evaluatedExpression.expandedDescription())"
+        "Expectation failed: \(expectation.evaluatedExpression.sourceCode)"
       }
     case let .confirmationMiscounted(actual: actual, expected: expected):
       if let expected = expected as? any _RangeExpressionOverIntValues {
@@ -632,7 +632,7 @@ extension Issue.Kind.Snapshot: CustomStringConvertible {
       if let mismatchedErrorDescription = expectation.mismatchedErrorDescription {
         "Expectation failed: \(mismatchedErrorDescription)"
       } else {
-        "Expectation failed: \(expectation.evaluatedExpression.expandedDescription())"
+        "Expectation failed: \(expectation.evaluatedExpression.sourceCode)"
       }
     case let .confirmationMiscounted(actual: actual, expected: expected):
       "Confirmation was confirmed \(actual.counting("time")), but expected to be confirmed \(expected.counting("time"))"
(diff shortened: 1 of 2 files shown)


Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -102,34 +44,15 @@ public struct __Expression: Sendable {
...
 
+  /// Whether or not this instance represents a negated expression (`!foo`).
+  var isNegated = false
+
   /// The source code of the original captured expression.
   @_spi(ForToolsIntegrationOnly)
   public var sourceCode: String {
     switch kind {
-    case let .generic(sourceCode), let .stringLiteral(sourceCode, _):
+    case let .generic(sourceCode):
```
harlanhaskins, line 56: Since `Kind` is internal, do we expect for it to grow new cases, or could it be fully collapsed into `public var sourceCode: String`?

Sources/Testing/SourceAttribution/Expression.swift
```diff
@@ -35,64 +35,6 @@ public struct __Expression: Sendable {
     /// - Parameters:
     ///   - sourceCode: The source code of the represented expression.
     case generic(_ sourceCode: String)
-
```
harlanhaskins, line 38: This is an _awesome_ simplification!

grynspan: 🫡

Sources/Testing/SourceAttribution/Expression.swift
grynspan: It'll _eventually_ be collapsed into `public var sourceCode: String`, yeah. But see my upcoming change where I lazily reconstruct at least some of these strings.

grynspan: Actually, never mind: lazy evaluation didn't save me anything. Yeah I'll just rip this thing out.

PR #1582, opened 2026-02-21
Enhance closure-based expectation failures.

grynspan:
This PR is a follow-up of #1572 and #1581. It cleans up the output produced by the `#expect(throws:)` and `#expect(processExitsWith:)` macros to more closely match what we now output for boolean and optional expectations.

For instance, the following test:

```swift
@Test func foo() {
  #expect(throws: MyError()) {
    throw MyParameterizedError(index: 0)
  }
}
```

Previously output:

```
◇ Test foo() started.
✘ Test foo() recorded an issue at [...]: Expectation failed: expected error "MyError()", but "MyParameterizedError(index: 0)" was thrown instead
↳ throw MyParameterizedError(index: 0) → MyParameterizedError(index: 0)
✘ Test foo() failed after 0.001 seconds with 1 issue.
```

But will now output:

```
◇ Test foo() started.
✘ Test foo() recorded an issue at [...]: Expectation failed: expected error "MyError()", but "MyParameterizedError(index: 0)" was thrown instead
↳ MyError() → MyParameterizedError(index: 0)
↳   index → 0
✘ Test foo() failed after 0.001 seconds with 1 issue.
```

And the following exit test:

```swift
@Test func bar() async {
  await #expect(processExitsWith: .exitCode(EX_USAGE)) {
    exit(EX_IOERR)
  }
}
```

Previously output:

```
◇ Test bar() started.
✘ Test bar() recorded an issue at [...]: Expectation failed: .exitCode(EX_USAGE → 64)
↳ .exitCode(EX_USAGE → 64) → .exitCode(EX_IOERR → 74)
✘ Test bar() failed after 0.040 seconds with 1 issue.
```

But will now output:

```
◇ Test bar() started.
✘ Test bar() recorded an issue at [...]: Expectation failed: expected exit status ".exitCode(EX_USAGE)", but ".exitCode(EX_IOERR)" was reported instead
↳ .exitCode(EX_USAGE) → .exitCode(EX_IOERR)
↳   EX_USAGE → 64
↳   EX_IOERR → 74
✘ Test bar() failed after 0.038 seconds with 1 issue.
```

We already have code in Swift Testing that uses reflection to extract information about fields of captured values, and Xcode uses this code, but it's effectively dead when using `swift test`. With this change in place, we'll output those subexpressions recursively. For example, the following test:

```swift
struct S: Equatable, CustomStringConvertible {
  var x: Int
  var description: String { "DESC" }
}

@Test func quux() {
  let s1 = S(x: 1)
  let s2 = S(x: 2)
  #expect(s1 == s2)
}
```

Would previously output:

```
◇ Test quux() started.
✘ Test quux() recorded an issue at [...]: Expectation failed: s1 == s2
↳ s1 == s2 → false
↳   s1 → DESC
↳   s2 → DESC
✘ Test quux() failed after 0.001 seconds with 1 issue.
```

And will now output:

```
◇ Test quux() started.
✘ Test quux() recorded an issue at [...]: Expectation failed: s1 == s2
↳ s1 == s2 → false
↳   s1 → DESC
↳     x → 1
↳   s2 → DESC
↳     x → 2
✘ Test quux() failed after 0.001 seconds with 1 issue.
```

### Checklist:

### DIFF
--- a/Sources/Testing/ExitTests/ExitTest.swift
+++ b/Sources/Testing/ExitTests/ExitTest.swift
@@ -527,28 +527,26 @@ func callExitTest(
   }
 
   // Plumb the exit test's result through the general expectation machinery.
-  var expression = expression
-  if let expectedExitStatus = expectedExitCondition.exitStatus,
-     let expectedExitStatusName = expectedExitStatus.name {
-    expression.subexpressions.append(
-      __Expression(
-        expectedExitStatusName,
-        runtimeValue: __Expression.Value(describing: expectedExitStatus.code)
-      )
-    )
-  }
-  if let statusName = result.exitStatus.name {
-    expression.subexpressions.append(
-      __Expression(
-        statusName,
-        runtimeValue: __Expression.Value(describing: result.exitStatus.code)
-      )
-    )
+  func expressionWithCapturedRuntimeValues() -> __Expression {
+    var expression = expression.capturingRuntimeValues(result.exitStatus)
+
+    expression.subexpressions = [expectedExitCondition.exitStatus, result.exitStatus]
+      .compactMap { exitStatus in
+        guard let exitStatus, let exitStatusName = exitStatus.name else {
+          return nil
+        }
+        return __Expression(
+          exitStatusName,
+          runtimeValue: __Expression.Value(describing: exitStatus.code)
+        )
+      }
+
+    return expression
   }
   return __checkValue(
     expectedExitCondition.isApproximatelyEqual(to: result.exitStatus),
     expression: expression,
-    expressionWithCapturedRuntimeValues: expression.capturingRuntimeValues(result.exitStatus),
+    expressionWithCapturedRuntimeValues: expressionWithCapturedRuntimeValues(),
     mismatchedExitConditionDescription: #"expected exit status "\#(expectedExitCondition)", but "\#(result.exitStatus)" was reported instead"#,
     comments: comments(),
     isRequired: isRequired,


Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -527,11 +527,29 @@ func callExitTest(
...
+        statusName,
+        runtimeValue: __Expression.Value(describing: result.exitStatus.code)
+      )
+    )
+  }
   return __checkValue(
     expectedExitCondition.isApproximatelyEqual(to: result.exitStatus),
     expression: expression,
     expressionWithCapturedRuntimeValues: expression.capturingRuntimeValues(result.exitStatus),
+    mismatchedExitConditionDescription: #"expected exit status "\#(expectedExitCondition)", but "\#(result.exitStatus)" was reported instead"#,
```
grynspan, line 550: This string matches what we do for `#expect(throws:)`, but we can bikeshed both.

Sources/Testing/ExitTests/ExitTest.swift
```diff
@@ -527,11 +527,29 @@ func callExitTest(
   }
 
   // Plumb the exit test's result through the general expectation machinery.
-  let expression = __Expression(String(describingForTest: expectedExitCondition))
+  var expression = expression
+  if let expectedExitStatus = expectedExitCondition.exitStatus,
```
harlanhaskins: Should we add a fallback for if it exits with a status whose name we don't know? Is that gonna be captured here? (I'm not sure what happens to `expressionWithCapturedRuntimeValues`, so if that covers this then that's fine)

harlanhaskins (approved):
Nice cleanup, appreciate the detail in cleaning `.exitCode(EXIT_SUCCESS -> 0)` to just `.exitCode(EXIT_SUCCESS)`. Left one small comment, but could just be my own unfamiliarity with the runtime value capture system.

Sources/Testing/ExitTests/ExitTest.swift
grynspan: In that case it would be captured as just a plain integer value. When we go to output it to `stderr`, we'll see the stringification is the same and won't bother listing it as a separate line item. The effect is then the same as not including the subexpression at all.

PR #1584, opened 2026-02-23
Use one allocation for the bespoke `Mutex`'s storage.

grynspan:
We currently use two allocations to store the contents of our bespoke Apple-only `Mutex` reimplementation: one for the `os_unfair_lock` and the other for the stored value. We can instead use a single allocation with a bit of careful pointer arithmetic to ensure there's enough storage for both and that it's well-aligned.

This change has the effect of reducing the size of `Mutex` by a word. It also reduces the amount of memory allocated by some vague amount because on Darwin, the minimum allocation size is 16 but the size of `os_unfair_lock` is 4. I place the lock _after_ the value because its alignment will often be smaller, so fewer padding bytes will be needed between the value and the lock than between the lock and the value.

At compile time in release mode, all specializations are known and the extra math is optimized away.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/MutexAdditions.swift
+++ b/Sources/Testing/Support/Additions/MutexAdditions.swift
@@ -40,32 +40,48 @@ struct Mutex<Value>: Sendable, ~Copyable where Value: ~Copyable {
   private typealias _Lock = pthread_mutex_t
 #endif
 
+  /// Storage for both the lock and value.
+  private nonisolated(unsafe) let _baseAddress: UnsafeMutableRawPointer
+
+  /// The offset of `_lockAddress` from `_baseAddress`.
+  private static var _lockOffset: Int {
+    let p = UnsafeRawPointer(bitPattern: MemoryLayout<Value>.stride)!
+    return Int(bitPattern: p.alignedUp(for: _Lock.self))
+  }
+
   /// Storage for the underlying lock.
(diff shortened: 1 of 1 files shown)


Sources/Testing/Support/Additions/MutexAdditions.swift
```diff
@@ -40,32 +40,48 @@ struct Mutex<Value>: Sendable, ~Copyable where Value: ~Copyable {
   private typealias _Lock = pthread_mutex_t
 #endif
 
+  /// Storage for both the lock and value.
+  private nonisolated(unsafe) let _baseAddress: UnsafeMutableRawPointer
+
+  /// The offset of `_lockAddress` from `_baseAddress`.
+  private static var _lockOffset: Int {
+    let p = UnsafeRawPointer(bitPattern: MemoryLayout<Value>.stride)!
```
harlanhaskins, line 48: Would it be easier and avoid this math by storying an `UnsafeMutablePointer<(Value, os_unfair_lock)>`? I think you can still get `_lockAddress` from `_baseAddress.pointer(to: \.1)` and tuples are guaranteed to be laid out contiguously (modulo alignment requirements)

grynspan: Nope. Try it locally and see why: can't create tuples with move-only values, nor can you use `pointer(to:)`.

PR #1585, opened 2026-02-24
Use a different symbol for failure with known issues.

### LINKED ISSUE #1583: Avoid using cross symbol to denote "success but with known issues"
If a test suite passes but has known issues currently swift-testing uses either `xmark.diamond` on macOS or otherwise `HEAVY BALLOT X` to denote the status. This is confusing since it leads to log output that looks contradictory:

```
✘ Test run with 5 tests in 2 suites passed after 3.235 seconds with 1 known issue
```

IMO the fact that there are known issues is useful additional information, but the overall status is still success so it should just be a tick, same as if there were no known issues (or maybe there could be an alternative tick icon if we really care about the distinction)

grynspan:
This PR switches us to a dash character for failure-with-known-issues instead of an X which is the same symbol used for full failure. As well, I've removed the separate (and never-made-not-experimental) `passWithWarnings` symbol and replaced it with the existing `warning` symbol. I had originally intended for `warning` to represent some sort of runtime condition, but we didn't end up building any such things, so let's use the standard warning glyph where users might actually encounter it.

Before (captured on macOS):

Screenshot 2026-02-23 at 7 13 36 PM

After:

Screenshot 2026-02-23 at 7 14 30 PM

Resolves #1583.

### Checklist:

### DIFF
--- a/Sources/SymbolShowcase/SymbolShowcaseMain.swift
+++ b/Sources/SymbolShowcase/SymbolShowcaseMain.swift
@@ -43,7 +43,6 @@ import Foundation
       "Default": .default,
       "Pass": .pass(knownIssueCount: 0),
       "Pass w/known issues": .pass(knownIssueCount: 1),
-      "Pass with warnings": .passWithWarnings,
       "Skip": .skip,
       "Fail": .fail,
       "Difference": .difference,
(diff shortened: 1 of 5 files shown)


grynspan:
Should warnings be purple to match what Xcode does? 🤔

grynspan:
(We'll table that question for now.)

PR #1588, opened 2026-02-24
Expose `ABI.EncodedInstant` as tools SPI.

grynspan:
Follow-up to #1587.

Exposes `ABI.EncodedInstant` as SPI along with some conversion helpers to get a Foundation `Date` or suspending clock instant from an instance thereof. A value of this type is associated with each event in the event stream.

### Checklist:


Sources/Overlays/_Testing_Foundation/Events/Clock+Date.swift
```diff
@@ -12,19 +12,32 @@
...
+    self.init(timeIntervalSince1970: instant.durationSince1970 / .seconds(1))
+  }
+
+  /// Initialize this date to equal an instant from the testing library's event
+  /// stream.
+  ///
+  /// - Parameters:
+  ///   - instant: The encoded instant to initialize this instance from.
+  ///
+  /// The resulting instance is equivalent to the wall-clock time represented by
```
stmontgomery, line 36:
```swift
- Returns: `nil` ...
```

grynspan: We don't do `- Returns` on initializers.

Sources/Testing/ABI/Encoded/ABI.EncodedInstant.swift
```diff
@@ -12,29 +12,54 @@ extension ABI {
...
+extension SuspendingClock.Instant {
+  /// Initialize this instant to equal an instant from the testing library's
+  /// event stream.
+  ///
+  /// - Parameters:
+  ///   - instant: The encoded instant to initialize this instance from.
+  ///
+  /// The resulting instance is equivalent to the suspending-clock time
+  /// represented by `instant`.
+  public init?<V>(decoding instant: ABI.EncodedInstant<V>) {
```
stmontgomery, line 56: Is this one intentionally Optional?

Sources/Overlays/_Testing_Foundation/Events/Clock+Date.swift
grynspan: Additionally, this initializer isn't actually failable as of right now. However, I've marked it failable (`?`) so that we have the option to fail in the future if we change the encoding of these values in a way that's not compatible with these types (e.g. if we were to drop wall-clock time from the schema).

Sources/Testing/ABI/Encoded/ABI.EncodedInstant.swift
grynspan: Yes, see my reply to your other comment :)

PR #1589, opened 2026-02-25
Expose `ABI.EncodedSourceLocation` as tools SPI.

grynspan:
Follow-up to #1587.

Exposes `ABI.EncodedSourceLocation` as SPI along with some conversion helpers to go to/from a `SourceLocation` instance. A value of this type is associated with many events in the event stream.

### Checklist:

### DIFF
--- a/Tests/TestingTests/SourceLocationTests.swift
+++ b/Tests/TestingTests/SourceLocationTests.swift
@@ -86,6 +86,20 @@ struct SourceLocationTests {
     #expect(sourceLocation.moduleName == SourceLocation.synthesizedModuleName)
     #expect(sourceLocation.fileName == "FileName.swift")
   }
+
+  @Test("SourceLocation does not accept a bad file ID from an EncodedSourceLocation")
+  func badFileIDInEncodedSourceLocation() throws {
+#if os(Windows)
+    var json = #"{"filePath": "C:\fake/dir/FileName.swift/", "fileID": "bad", "line": 1, "column": 1}"#
+#else
+    var json = #"{"filePath": "/fake/dir/FileName.swift/", "fileID": "bad", "line": 1, "column": 1}"#
+#endif
+    let esl = try json.withUTF8 { json in
+      try JSON.decode(ABI.EncodedSourceLocation<ABI.v6_3>.self, from: UnsafeRawBufferPointer(json))
+    }
+    let sourceLocation = SourceLocation(decoding: esl)
+    #expect(sourceLocation == nil)
+  }
 #endif
 
   @Test("SourceLocation.line and .column properties")


Sources/Testing/ABI/Encoded/ABI.EncodedSourceLocation.swift
```diff
@@ -12,56 +12,107 @@ extension ABI {
...
+  public init?<V>(decoding sourceLocation: ABI.EncodedSourceLocation<V>) {
     let fileID = sourceLocation.fileID
-    guard let filePath = sourceLocation.filePath ?? sourceLocation._filePath else {
-      return nil
-    }
+    let filePath = sourceLocation.filePath
     let line = max(1, sourceLocation.line)
     let column = max(1, sourceLocation.column)
 
+    if let fileID, !fileID.utf8.contains(UInt8(ascii: "/")) {
```
harlanhaskins, line 109: Was this added just because this method is now `public` and therefore may be misused?

grynspan: The strings this type needs to handle are arbitrary, so we should validate inputs before passing them to an initializer that may abort on bad inputs.

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -96,7 +96,8 @@ extension ABI {
     ///
     /// - Warning: Source locations at this level of the JSON schema are not yet
     ///   part of said JSON schema.
-    var _sourceLocation: EncodedSourceLocation<V>?
+    @_spi(Experimental)
```
jerryjrchen, line 99: Should this be `ForToolsIntegrationOnly`?

Sources/Testing/ABI/Encoded/ABI.EncodedSourceLocation.swift
jerryjrchen: Perhaps overly pedantic, but should we check that the `/` separator isn't the very beginning or end of the fileID as well? e.g. excluding `/foo.swift` and `Foo/`

Tests/TestingTests/SourceLocationTests.swift
jerryjrchen, line 76: Would it be worthwhile to add some tests here for the `fileId` validation?

jerryjrchen (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedSourceLocation.swift
grynspan: Strictly speaking, that's allowed by the definition of `#fileID` and just means one or the other component is empty.

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
grynspan: The whole type already is.

Tests/TestingTests/SourceLocationTests.swift
grynspan: Sure.

PR #1590, opened 2026-02-25
Expose `ABI.EncodedAttachment` as tools SPI.

grynspan:
Follow-up to #1587.

Exposes `ABI.EncodedAttachment` as SPI. A value of this type is associated with the `valueAttached` event kind. The type already conforms to `Attachable` so it can be directly plumbed back into Swift Testing, and XCTest internally knows how to convert an `Attachable` value to an `XCTAttachment`.

### Checklist:

### DIFF
--- a/Tests/TestingTests/AttachmentTests.swift
+++ b/Tests/TestingTests/AttachmentTests.swift
@@ -957,8 +957,10 @@ extension AttachmentTests {
       let url = URL(fileURLWithPath: path, isDirectory: false)
       try Data(expectedBytes).write(to: url)
 
+#if os(Windows)
+      let path = path.replacing(#"\"#, with: #"\\"#)
+#endif
       var json = #"{"path": "\#(path)"}"#
-      print("ATTACHMENT JSON:", json)
       let eattachment = try json.withUTF8 { json in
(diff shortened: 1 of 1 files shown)


Sources/Testing/ABI/Encoded/ABI.EncodedAttachment.swift
```diff
@@ -16,122 +16,203 @@ extension ABI {
...
-      let data = Data(bytesNoCopy: .init(mutating: rawValue.baseAddress!), count: rawValue.count, deallocator: .none)
-      var container = encoder.singleValueContainer()
-      try container.encode(data)
-    }
-#else
-    // Otherwise, it's an array of integers.
-    var container = encoder.singleValueContainer()
-    try container.encode(rawValue)
+        // If possible, decode a whole Foundation Data object.
+        if let data = try? container.decodeIfPresent(Data.self, forKey: .bytes) {
```
harlanhaskins, line 129: How does this encoding differ from `[UInt8]`?

grynspan: `Data` encodes as base64 by default.

Issue #480, opened 2024-06-13
Using recording an issue inside of a detached task causes a crash.


younata:
### Description

This is something I learned while researching potential ways to address #475. Which, at the time I wrote that issue, I thought would have just resulted in the failure being silently dropped.

The current way of determining what the current reporting tests natively in Swift-Testing is broken when you run an expect inside of a detached task. Technically, you get a failing test result, but it's because the test runner crashed. I haven't yet spelunked enough through the codebase to figure out where that crash is.

This is because `Test.current` is stored as a task-local value. Which means that when you make an expect inside of a detached task, Swift Testing has lost that context, which means that `Test.current` is incorrectly (from the semantics of the current running test) as nil.

I don't know the concurrency system well enough to be able to suggest a good solution. The issue is that, because Swift Testing supports task-based parallelization, the value of `Test.current` strongly depends on the task it's running on. Which makes having a global `Test.current` like what's in swift-corelibs-xctest the wrong solution.

### Steps to reproduce

This test should pass. The test runner crashes with a non-zero error instead.

```swift
import Testing

@Test func CheckingExpectWithDetachedTasks() async {
    await Task.detached {
        #expect(Bool(false))
    }.value
}
```

![Screenshot 2024-06-13 at 6 41 15 AM](https://github.com/apple/swift-testing/assets/285321/0637fa14-71f5-46e2-a2b6-7fe000005b88)

### swift-testing version/commit hash

Testing Library Version: 75.7 (arm64e-apple-macos13.0)

stmontgomery:
The crash here is a known issue in Xcode 16 Beta, tracked by rdar://122190668. To resolve it, some work will be needed on the Xcode side, but it's also likely some additional action will be needed here in the swift-testing codebase, so I will leave this issue open to track that side of things.

younata:
Having not tried Swift Testing on Xcode 15, is the behavior that it silently drops the issue?

stmontgomery:
> (To clarify: the crash is specific to Xcode 16.)

Right. But that said, other tools (such as VS Code or `swift test` CLI) _may_ not be handling this situation very clearly in the reporting either, despite not crashing. Because of parallel execution, it's not always obvious which test a particular `Issue` is associated with, so one idea we've had is to keep track of all the currently-running tests and include them as "potentially associated" tests in the reporting, so that at least you can narrow the scope down when troubleshooting.

grynspan:
[...]

PR #1591, opened 2026-02-25
Add diagnostics for issues recorded after their associated test has finished

### LINKED ISSUE #1283: Diagnose issues recorded "late", after their associated test has finished
### Description

It is currently possible for an Issue in a Swift Testing test to be recorded “late”, after the test it's associated with has officially finished. This issue tracks an enhancement which would explain this situation clearly when it happens, to help users understand and modify their test to avoid it.

One common way this can occur is when a test uses an unstructured `Task` and does not `await` its value to ensure it finishes before the `@Test` function returns. Here's an example:

```swift
@Test func example() async throws {
  Task {
    try await Task.sleep(for: .seconds(1))
    Issue.record("Late")
  }
  // Didn't 'await' that Task!
}

@Test func longRunning() async throws {
  try await Task.sleep(for: .seconds(5))
}
```

This outputs the following:

```
◇ Test longRunning() started.
◇ Test example() started.
✔ Test example() passed after 0.001 seconds.
✘ Test example() recorded an issue at Example.swift:6:17: Issue recorded
↳ Late
✔ Test longRunning() passed after 5.194 seconds.
✘ Test run with 2 tests in 0 suites failed after 5.194 seconds with 1 issue.
```

This can be a confusing experience: it says `example()` passed, and yet later it claims `example()` recorded an issue. This could be resolved by modifying `example()` as follows:

```diff
 @Test func example() async throws {
-  Task {
+  await Task {
     try await Task.sleep(for: .seconds(1))
     Issue.record("Late")
-  }
+  }.value
 }
```

This situation is ultimately the responsibility of the test author to resolve, generally by ensuring all issues are recorded before the test function returns. But when this problem happens, the testing library could help clarify what happened by diagnosing this with an extra note in the console message.

### Additional information

_No response_

wrdowney:
Adds a warning diagnostic message for issues that are recorded after their associated test has finished. 

Resolves swiftlang#1283

### Motivation:

Currently, Issues can be recorded after a test has finished. This commonly occurs when asynchronous work completes after the test its associated test has ended. While responsibility falls on the user in this scenario, clear messaging can point them in the correct direction.

From the issue:
> One common way this can occur is when a test uses an unstructured Task and does not await its value to ensure it finishes before the Test function returns. Here's an example:
```swift
@Test func example() async throws {
  Task {
    try await Task.sleep(for: .seconds(1))
    Issue.record("Late")
  }
  // Didn't 'await' that Task!
}

@Test func longRunning() async throws {
  try await Task.sleep(for: .seconds(5))
}
```
> This outputs the following:
```
◇ Test longRunning() started.
◇ Test example() started.
✔ Test example() passed after 0.001 seconds.
✘ Test example() recorded an issue at Example.swift:6:17: Issue recorded
↳ Late
✔ Test longRunning() passed after 5.194 seconds.
✘ Test run with 2 tests in 0 suites failed after 5.194 seconds with 1 issue.
```

### Modifications:
- Adds a boolean flag `hasFinished` to `Test.Case` along with a corresponding backing class that stores whether the test case has finished executing. This flag defaults to false
- When a test case finishes, the `hasFinished` flag is set to true
- Added a check when events are posted. If the event is has kind issueRecorded and the testCase has finished, we emit another issueRecorded event. 
- Added a test case where a task records an issue after it's corresponding test case has finished, validating that both the original and late issue are recorded.

### Checklist:


grynspan:
I'm not sure this is the right abstraction. A problem of this form should probably itself be surfaced as an issue (of kind `.apiMisused` maybe? Or a new kind if needed, although those are more of a pain to add.)

Just adding it to the console output means that, for instance, somebody using Xcode or VS Code won't see anything about the problem unless they go digging in the log file.

wrdowney:
Thanks for the feedback! I agree about surfacing through issues, thanks for pointing that out. I'm working on implementing that now and looking for a little guidance on the following questions:
1. Are there existing APIs that I might be able to leverage to determine if a test case is current executing? I've thought about using Test.current but I believe that will incorrectly report issues emitted from detached tasks as "late". I want to avoid having to store a set of currently executing test cases.
2. Right now I'm looking at posting a second issueRecorded event from Event.post() for an apiMisuse Issue when an Issue is recorded after test case completion, are there concerns with posting two issues from one call to Event.post()?

grynspan:
> 1. Are there existing APIs that I might be able to leverage to determine if a test case is current executing? I've thought about using Test.current but I believe that will incorrectly report issues emitted from detached tasks as "late". I want to avoid having to store a set of currently executing test cases.

If an issue is recorded from within a detached task that was originated by a specific test, we would expect that issue to be associated with that test (if we can figure out that it came from that test, which is hard.)

You shouldn't need to ever store a list of "current" tests anywhere.

> 2. Right now I'm looking at posting a second issueRecorded event from Event.post() for an apiMisuse Issue when an Issue is recorded after test case completion, are there concerns with posting two issues from one call to Event.post()?

The only immediate concern(s) I have are that:

a. We would need to guard against accidentally recursing, i.e. an issue of this specific flavour, not necessarily all `.apiMisused` issues, should not trigger its own "late" issue; and
b. We don't want to go off and try to recompute a bunch of state we computed for the first issue (e.g. source location info, current configuration) if we can avoid it.

For a., an internal-only `isLate: Bool` property would be sufficient, I think.

wrdowney:
I have a working (albeit a little rough) solution for emitting a second recordIssue event when an issue is recorded after test completion. But I'm now running into the following issue in Xcode: recording an issue after test completion results in an internal inconsistency error . 

```swift
@Suite struct ExampleTests {
  @Test func example() async throws {
    Task {
      try await Task.sleep(for: .seconds(1))
      Issue.record("Late")
    }
  }
  @Test func longRunning() async throws {
    try await Task.sleep(for: .seconds(5))
  }
}
```

```
Task 2: Fatal error: Internal inconsistency: No test reporter for test TestingTests.ExampleTests/example()/IssueTests.swift:1840:4 and test case Non-parameterized test case ID. Issue: Issue recorded (error): Late (at IssueTests.swift:1843:19)
```
This is run on the current main using Xcode 26.3

Unfortunately the error is thrown after the initial issue is recorded, thus any subsequent issues are neither logged to the console or recorded in editor.

I see #480 which fixes a similar issue for detached tasks. But those fixes don't seem to resolve this case. This seems to be an Xcode issue (I have confirmed running via CLI and VSCode extension does not throw an error). Although I do recognize this is quite the _unique_ use case.

Assuming I'm correct here, do we want to wait on a potential Xcode update prior to this work or am I able to continue here and for now the second issue(regarding the late issue recording) will only be logged when running in environments outside of Xcode?

I wanted to validate I'm not missing anything here before filing a feedback.

grynspan:
The crash you're seeing should be fixed in the Xcode 26.4 Beta. Our main branch _de facto_ requires Xcode 26.4 (our release/6.3 branch works with earlier Xcode 26.x releases).

wrdowney:
I ended up going with a slightly different approach by setting a flag on the test case. My main concern is that this approach might be too heavy but it's the simplest logic I could think of. I also call the private post method when emitting the second event so that should prevent any issues with accidental recursion or recomputing state.

grynspan:
So far we've avoided adding mutable state to `Test` or `Test.Case`. They get copied around a lot and having each instance carry a heap-allocated mutex with it is a lot of potential overhead.

Something you might want to try instead… install an event handler somewhere under `Runner.run()` that listens for `.testEnded` and `.testCaseEnded` events and tracks the set of ended tests/cases. If it then gets `.issueRecorded` for a test/case that was already reported as ended, it can raise a secondary issue at that point. This then only requires a single lock-guarded `Set`.

Something like:

```swift
enum FinishedItem: Sendable, Equatable, Hashable {
  case test(Test.ID)
  case testCase(Test.Case.ID)
}

let finishedItems = Allocated(Mutex<Set<FinishedItem>>([]))
configuration.eventHandler = { [oldEventHandler = configuration.eventHandler] event, eventContext in
  switch event.kind {
  case .testEnded:
    finishedItems.value.withLock { finishedItems in
      _ = finishedItems.insert(.test(event.testID!))
    }
  case .testCaseEnded:
    finishedItems.value.withLock { finishedItems in
      _ = finishedItems.insert(.testCase(event.testCaseID!))
    }
  case .issueRecorded:
    let finishedItems = finishedItems.value.rawValue
    if let testCaseID = event.testCaseID, finishedItems.contains(.testCase(testCaseID)) {
      // Bad test case
    } else if let testID = event.testID, finishedItems.contains(.test(testID)) {
      // Bad test
    }
  default:
    break
  }
  oldEventHandler(event, eventContext)
}
```

Note I haven't tried to optimize this code or clean up its structure, and I'm not saying you must implement exactly what I've typed here. This is just an idea about how to shoehorn this functionality into the existing architecture.

wrdowney:
Apologies for the long delay between updates. I have made some changes that are similar to your suggestion. Installing a new event handler which can respond to test(Case)Ended event and track which items have finished. Additionally, I needed to store both test and test case IDs due to conflicting test case IDs across different test instances. 

> Instances of this type are considered unique within the scope of a given parameterized test function. They are not necessarily unique across two different ``Test`` instances.

I hope this is a good overall approach although I assume it might need some cleaning up. Let me know if you have any questions/suggestions. Thanks again for all the feedback!

PR #1596, opened 2026-02-26
Add a C-callable "gestalt" function for querying some basic info.

grynspan:
This PR adds an exported C function that lets callers query some info that may be necessary to gather before those callers can reliably invoke the testing library's ABI-stable entry point function.

This function is _not_ meant to be a general-purpose querying function, nor is it intended for use by tools authors. Rather, it is useful if you're writing code that works with the JSON event stream and needs to know how to communicate with the testing library. In particular, a caller would need to know the range of supported JSON schema versions.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/ABI.swift
+++ b/Sources/Testing/ABI/ABI.swift
@@ -214,7 +214,7 @@ extension ABI {
 
 // MARK: -
 
-/// The set of keys accepted by ``swift_testing_copyMetadataValue(_:_:)``.
+/// The set of keys accepted by `_swift_testing_copyMetadataValue(_:_:)`.
 private enum _MetadataKey: String, Sendable, CaseIterable {
   /// The minimum supported ABI version.
   case minimumSupportedABIVersion = "_minimumSupportedABIVersion"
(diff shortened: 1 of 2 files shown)


Sources/Testing/ABI/ABI.swift
```diff
@@ -209,3 +211,57 @@ extension ABI {
     }
   }
 }
+
+// MARK: -
+
+/// The set of keys accepted by ``swift_testing_copyMetadataValue(_:_:)``.
+private enum _MetadataKey: String, Sendable, CaseIterable {
+  /// The minimum supported ABI version.
```
jerryjrchen, line 219: Nit: maybe include an example of what the value can look like:
```suggestion
  /// The minimum supported ABI version, e.g. "6.3"
```

grynspan: I wouldn't want to document a specific version number (it's `0` anyway!) but we could document that the value can be decoded as an instance of `ABI.VersionNumber`.

stmontgomery: Will the JSON string returned from `_swift_testing_copyMetadataValue` for a single `String` value, such as this one, be wrapped in a JSON object of some kind, or will it be a bare string.

grynspan: It will produce the JSON representation of the returned Swift value. In the case of an `ABI.VersionNumber`, that's a bare integer or string, yes.

PR #1597, opened 2026-02-27
Add a mode flags optionset to `FileHandle`.

grynspan:
This PR adds an `OptionSet`-conforming type to `FileHandle` so that we don't have to hard-code mode strings we pass to `fopen()`.

### Checklist:

### DIFF
--- a/Sources/Testing/Attachments/Attachment.swift
+++ b/Sources/Testing/Attachments/Attachment.swift
@@ -432,10 +432,9 @@ extension Attachable where Self: ~Copyable {
 #endif
 
     try withUnsafeBytes(for: attachment) { buffer in
-      // Note "x" in the mode string which indicates that the file should be
-      // created and opened exclusively. The underlying `fopen()` call will thus
-      // fail with `EEXIST` if a file exists at `filePath`.
-      let file = try FileHandle(atPath: filePath, mode: "wxeb")
+      // The underlying `fopen()` call should fail with `EEXIST` if a file
+      // exists at `filePath`.
+      let file = try FileHandle(atPath: filePath, options: [.writeAccess, .creationRequired])
       try file.write(buffer)
     }
   }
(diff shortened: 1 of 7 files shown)


Sources/Testing/Attachments/Attachment.swift
```diff
@@ -432,10 +432,9 @@ extension Attachable where Self: ~Copyable {
 #endif
 
     try withUnsafeBytes(for: attachment) { buffer in
-      // Note "x" in the mode string which indicates that the file should be
-      // created and opened exclusively. The underlying `fopen()` call will thus
-      // fail with `EEXIST` if a file exists at `filePath`.
-      let file = try FileHandle(atPath: filePath, mode: "wxeb")
```
grynspan, line 438: `"e"` (`O_CLOEXEC`) is implied by default with the new optionset; you have to _request_ inheritance explicitly instead (which we don't ever do currently).

Sources/Testing/Support/FileHandle.swift
```diff
@@ -48,15 +48,97 @@ struct FileHandle: ~Copyable, Sendable {
...
+    /// to `EEXIST`.
+    ///
+    /// This option is equivalent to the `"x"` character or `O_CREAT | O_EXCL`.
+    static var creationRequired: Self { .init(rawValue: 1 << 3) }
+
+    /// The string representation of this instance, suitable for passing to
+    /// `fopen()` and related functions.
+    var stringValue: String {
+      var result = ""
+      result.reserveCapacity(8)
```
ankushkushwaha, line 97: Do we really need this `result.reserveCapacity(8)` ?

grynspan: Probably not. :)

Tests/TestingTests/Traits/TagListTests.swift
```diff
@@ -143,7 +143,7 @@ struct TagListTests {
     }
     """
     try jsonContent.withUTF8 { jsonContent in
-      let fileHandle = try FileHandle(forWritingAtPath: jsonPath)
+      let fileHandle = try Testing.FileHandle(forWritingAtPath: jsonPath)
```
stmontgomery, line 146: Is this in anticipation of a conflict?

grynspan: Within the test target it's now ambiguous with Foundation. 🤷🏻‍♂️

PR #1598, opened 2026-02-27
Replace fixture protocol in ConfirmationTests with an existential composition of parameterized protocols

stmontgomery:
This modifies a test fixture in `ConfirmationTests` now that both a Swift language limitation, and an associated runtime bug, have been resolved.

The original issue was that the language did not support writing `any RangeExpression<Int> & Sequence<Int> & Sendable`, but that was resolved in Swift 6.2 in https://github.com/swiftlang/swift/pull/76705. Once that was resolved, it exposed a runtime bug which would crash when forming a collection of such existential types. That was fixed in https://github.com/swiftlang/swift/pull/85346, and landed in Apple's 26.4 OSes (currently in beta).

This PR finally removes the workaround and writes this parameterized test the idiomatic way, using a collection whose element type is a composition of protocols where some are parameterized.

### Checklist:

### DIFF
--- a/Tests/TestingTests/ConfirmationTests.swift
+++ b/Tests/TestingTests/ConfirmationTests.swift
@@ -35,7 +35,7 @@ struct ConfirmationTests {
   func unsuccessfulConfirmations() async {
     // confirmedOutOfRange(_:) is availability-guarded, but if it runs, increase
     // the number of expected miscounts by the number of arguments passed to it.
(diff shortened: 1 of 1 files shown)


grynspan (changes requested):
We only have 26.2 in CI so this change is untestable. Perhaps wait until we can verify it?

PR #1599, opened 2026-03-01
Adopt `Atomic`.

grynspan:
This PR moves us from `Mutex` to `Atomic` for scalar integer values that don't need a full mutex guarding them.

On Darwin, we cannot use types from the `Synchronization` module because we need to back-deploy earlier than they are available, so we simulate a subset of the `Atomic` API using C atomic intrinsics provided by clang and bubbled up through Stubs.h. Unlike the real `Atomic` type, we heap-allocate our storage, but we're heap-allocating the current mutex so that isn't a performance regression.

I have only implemented the set of functionality we're actually using; adding more functionality isn't hard, but is beyond the scope of this PR. In particular, I have not attempted to implement/use any atomic orderings other than `.sequentiallyConsistent`. We can revise orderings later on a case-by-case basis.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/AtomicAdditions.swift
+++ b/Sources/Testing/Support/Additions/AtomicAdditions.swift
@@ -98,7 +98,7 @@ extension Atomic where Value == Int {
     swt_atomicStore(_address, desired)
   }
 
-  func compareExchange(expected: consuming Value, desired: consuming Value) -> (exchanged: Bool, original: Value) {
+  func compareExchange(expected: consuming Value, desired: consuming Value, ordering: Ordering) -> (exchanged: Bool, original: Value) {
     var expected = expected
     let exchanged = swt_atomicCompareExchange(_address, &expected, desired)
     return (exchanged, expected)
@@ -106,8 +106,13 @@ extension Atomic where Value == Int {
 
   @discardableResult
   func add(_ operand: Value, ordering: Ordering) -> (oldValue: Value, newValue: Value) {
-    let newValue = swt_atomicAdd(_address, operand)
-    return (newValue - operand, newValue)
+    while true {
+      let oldValue = load(ordering: ordering)
+      let newValue = oldValue + operand
+      if compareExchange(expected: oldValue, desired: newValue, ordering: ordering).exchanged {
+        return (oldValue, newValue)
+      }
+    }
   }
 
   @discardableResult
--- a/Sources/_TestingInternals/include/Stubs.h
+++ b/Sources/_TestingInternals/include/Stubs.h
@@ -104,23 +104,12 @@ static mach_port_t swt_mach_task_self(void) {
   SWT_SWIFT_NAME(swt_atomicCompareExchange(_:_:_:)) \
   static inline bool SWT_CONCAT(swt_atomicCompareExchange, __COUNTER__)(T *_Nonnull dst, T *_Nonnull expected, T desired) { \
     return __atomic_compare_exchange_n(dst, expected, desired, /*weak: */false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST); \
-  } \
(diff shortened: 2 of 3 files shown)


Sources/_TestingInternals/include/Stubs.h
```diff
@@ -84,6 +84,42 @@ static mach_port_t swt_mach_task_self(void) {
...
+    return __atomic_load_n(src, __ATOMIC_SEQ_CST); \
+  } \
+  \
+  static inline void swt_atomicStore_##T(T *dst, T src) { \
+    __atomic_store_n(dst, src, __ATOMIC_SEQ_CST); \
+  } \
+	static inline bool swt_atomicCompareExchange_##T(T *dst, T *expected, T desired) { \
+		return __atomic_compare_exchange_n(dst, expected, desired, /*weak: */false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST); \
+	} \
+	static inline T swt_atomicAdd_##T(T *dst, T addend) { \
```
jerryjrchen: Nit: indentation on these is inconsistent. And GH code highlighting seems to prefer the four space indentation

Sources/Testing/Support/Additions/AtomicAdditions.swift
```diff
@@ -0,0 +1,118 @@
...
+  }
+}
+
+extension Atomic: Sendable where Value: Sendable {}
+
+// MARK: - Atomic<Bool>
+
+extension Atomic where Value == Bool {
+  func load(ordering: Ordering) -> Value {
+    swt_atomicLoad_bool(_address)
```
jerryjrchen: Nit: maybe `assert` on `ordering == .sequentiallyConsistent`? Would ensure we don't forget to update the implementation if we add a new case to `Ordering`

Sources/Testing/Support/Additions/AtomicAdditions.swift
```diff
@@ -0,0 +1,118 @@
...
+    case sequentiallyConsistent
+  }
+}
+
+extension Atomic: Sendable where Value: Sendable {}
+
+// MARK: - Atomic<Bool>
+
+extension Atomic where Value == Bool {
+  func load(ordering: Ordering) -> Value {
```
jerryjrchen, line 57: Nit: maybe assert on ordering == .sequentiallyConsistent? Would ensure we don't forget to update the implementation if we add a new case to Ordering

Tests/TestingTests/Support/AtomicTests.swift
```diff
@@ -0,0 +1,34 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if SWT_TARGET_OS_APPLE // other platforms use Synchronization
+@testable import Testing
+
+final class `Atomic tests`: Sendable {
```
jerryjrchen, line 14: What the reasoning behind making this a class rather than a struct suite?

Sources/Testing/Support/Additions/AtomicAdditions.swift
grynspan: If we update the implementation to add a new case, it implies we need to audit the entire file anyway since it _all_ assumes seqcst.

Tests/TestingTests/Support/AtomicTests.swift
grynspan: Needs to hold an `Atomic` which is `~Copyable` while also allowing itself to be used concurrently.

Sources/_TestingInternals/include/Stubs.h
grynspan: Fixed.

PR #1600, opened 2026-03-02
Adjust the "Q&A" documentation for `Allocated` so it's not `Mutex`-specific.

grynspan:
Cleanup of internal docs for `Allocated`.

### Checklist:


grynspan:
Docs change only. Self-reviewing.

PR #1606, opened 2026-03-04
Remove some legacy Darwin-only code from `Test.Clock`.

grynspan:
This PR removes some more Darwin-only legacy code from the clock implementation. Our minimum deployment target is now high enough that this code is no longer needed.

### Checklist:


harlanhaskins (approved):
Hooray!

stmontgomery (approved):

PR #1608, opened 2026-03-05
Add support for `.expectationFailed` in interop and exit tests.

grynspan:
This PR enhances the JSON event stream to allow us to recognize and recover issues of kind `.expectationFailed`. This then improves the fidelity of these issues when passed through XCTest interop and exit test process boundaries.

For example, given the following exit test:

```swift
@Test func foo() async {
  await #expect(processExitsWith: .success) {
    struct S: Equatable {
      var x: Int
      var y: String
    }
    let lhs = S(x: 1, y: "abc")
    let rhs = S(x: 2, y: "def")
    #expect(lhs == rhs)
  }
}
```

We'll now get a breakdown of the failed expectation (note I recorded this example with `--verbose` so type info would be visible too):

```
<X>  Test foo() recorded an issue at [...]: Expectation failed: lhs == rhs
`->  lhs == rhs: Swift.Bool → false
`->    lhs: ModuleName.S → S(x: 1, y: "abc")
`->      x: Swift.Int → 1
`->      y: Swift.String → "abc"
`->    rhs: ModuleName.S → S(x: 2, y: "def")
`->      x: Swift.Int → 2
`->      y: Swift.String → "def"
```

In the process of implementing this change, I was able to reconcile the issue-decoding logic used in interop and exit tests so they share an implementation.

This PR also resolves an issue running interop tests inside Xcode 26.4 where the tests could spuriously fail.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
@@ -163,8 +163,7 @@ extension Issue {
       sourceContext: sourceContext
     )
     if issue.isKnown {
-      // The known issue comment, if there was one, is already included in
-      // the `comments` array above.
+      // FIXME: The known issue comment is not currently encoded.
       self.knownIssueContext = Issue.KnownIssueContext()
     }
   }


Sources/Testing/ABI/Encoded/ABI.EncodedExpression.swift
```diff
@@ -51,3 +51,26 @@ extension ABI {
...
+
+// MARK: - Conversion to/from library types
+
+@_spi(ForToolsIntegrationOnly)
+extension __Expression {
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// - Parameters:
+  ///   - expression: The encoded expression to initialize this instance from.
+  init?<V>(decoding expression: ABI.EncodedExpression<V>) {
```
grynspan, line 62: Not ready to make this public/SPI yet.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -99,3 +99,74 @@ extension ABI {
...
+@_spi(ForToolsIntegrationOnly)
+extension Issue {
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// - Parameters:
+  ///   - event: The encoded event to initialize this instance from.
+  ///
+  /// If `event` does not represent a recorded issue, the initializer returns
+  /// `nil`.
+  init?<V>(decoding event: ABI.EncodedEvent<V>) {
```
grynspan, line 113: Not ready to make these public/SPI yet.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -99,3 +99,73 @@ extension ABI {
...
+    self.init(
+      kind: issueKind,
+      severity: severity,
+      comments: [],
+      sourceContext: sourceContext
+    )
+    if issue.isKnown {
+      // The known issue comment, if there was one, is already included in
+      // the `comments` array above.
+      self.knownIssueContext = Issue.KnownIssueContext()
```
stmontgomery: I know we separated the comment on `withKnownIssue` into a discrete field intentionally, so are we potentially conflating them by having them combined into a single field? (That might be a preexisting problem)

grynspan: As of right now, the known-issue comment isn't in the encoded form. It's something we can certainly look into as a separate PR.

grynspan: (The source comment itself is not new, but is out of date.)

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -99,3 +99,72 @@ extension ABI {
...
+    }
+  }
+
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// - Parameters:
+  ///   - issue: The encoded issue to initialize this instance from.
+  ///
+  /// - Note: For higher fidelity, initialize the issue with an encoded event
+  ///   representing a recorded issue rather than just the encoded issue.
```
jerryjrchen, line 129: > For higher fidelity, initialize the issue with an encoded event representing a recorded issue rather than just the encoded issue

I think this could be more explicit about what's expected here. e.g. state that `_expectation` or `_error` should be set to indicate specific issue kinds.

jerryjrchen (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
grynspan: Those are experimental fields, so we don't want to direct people to them unless/until they're an approved part of the schema. Reliance on these fields outside our immediate project(s) would force us into the dark corners of Hyrum's Law.

PR #1609, opened 2026-03-05
Don't use `#_sourceLocation` in the fallback event handler.

grynspan:
We can't use our own macros in the Testing target because they aren't available when we build the toolchain.

Update the expansion of `#_sourceLocation` to detect it's being used in one of our own library targets and emit an error.

Resolves rdar://171774246.

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Event+FallbackEventHandler.swift
+++ b/Sources/Testing/Events/Event+FallbackEventHandler.swift
@@ -163,7 +163,7 @@ extension Event {
         kind: .system,
         comments: [errorContext],
         sourceContext: sourceContext
-      ).record()//(errorContext, sourceLocation: sourceLocation)
+      ).record()
     }
   }
 #endif


harlanhaskins (approved):
Clever use of compiler conditions to prevent this bug in the future, nice

Sources/TestingMacros/SourceLocationMacro.swift
```diff
@@ -20,7 +20,8 @@ public struct SourceLocationMacro: ExpressionMacro, Sendable {
     of macro: some FreestandingMacroExpansionSyntax,
     in context: some MacroExpansionContext
   ) throws -> ExprSyntax {
-    createSourceLocationExpr(of: macro, context: context)
+    diagnoseExpansionInLibraryTarget(of: macro, in: context)
```
stmontgomery, line 23: Doesn't have to be done now, in this PR, but this diagnostic could be applied to every one of our macros, right? E.g. other things like `#__testing(...)`

grynspan: Yes, with some tweaking as this takes an expression macro specifically (`__testing` is an attribute). It was out of scope for this PR.

stmontgomery:
(For tracking purposes) Fallout from #1523

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -108,22 +108,41 @@ extension Event {
...
+  private static func _bestAvailableSourceLocation(forInvalidRecordJSON recordJSON: UnsafeRawBufferPointer) -> SourceLocation {
+    // TODO: try to actually extract a source location from arbitrary JSON?
+
+    // If there's a test associated with the current task, it should have a
+    // source location associated with it.
+    if let test = Test.current {
+      return test.sourceLocation
+    }
+
+    return SourceLocation(fileID: "<unknown>/<unknown>", filePath: "<unknown>", line: 1, column: 1)
```
stmontgomery, line 128: Generally suggest 0 for line and column for unknown locations, I think that's what the compiler does too
```suggestion
    return SourceLocation(fileID: "<unknown>/<unknown>", filePath: "<unknown>", line: 0, column: 0)
```

grynspan: We explicitly document that both must be at least `1`, and we assert if they're not.

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -132,8 +151,19 @@ extension Event {
...
+
+      // Try to figure out a reasonable source context for this issue.
+      let sourceContext = SourceContext(
+        backtrace: .current(),
+        sourceLocation: _bestAvailableSourceLocation(forInvalidRecordJSON: recordJSON)
+      )
+
+      // Record the issue.
+      Issue(
+        kind: .apiMisused,
```
stmontgomery: Since you're modifying it, I think `.system` might be even more appropriate for this one

finagolfin:
I just ran into this build error when trying to build latest trunk locally on macOS, thought maybe I did something wrong. 😄 May help to set up some kind of nightly snapshot CI using `build-script` to build this repo, as [that would likely have turned up this bug](https://ci.swift.org/job/oss-swift-package-macos/4813/console) during PR testing.

grynspan:
> I just ran into this build error when trying to build latest trunk locally on macOS, thought maybe I did something wrong. 😄 May help to set up some kind of nightly snapshot CI using `build-script` to build this repo, as [that would likely have turned up this bug](https://ci.swift.org/job/oss-swift-package-macos/4813/console) during PR testing.

I'd have to defer to our CI wranglers on that topic.

PR #1607, opened 2026-03-04
Restore inadvertently reverted fallback event handler optimization.

grynspan:
When I reverted our `RawSpan` adoption in #1541, I inadvertently reverted the optimizations I'd made to Jerry's interop work that reduce the overhead of calling the fallback event handler more than once. This PR restores those optimizations.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1610, opened 2026-03-05
Guard `_postToFallbackEventHandler` with an `SWT_NO_INTEROP` check.

grynspan:
Missing a check for `SWT_NO_INTEROP`. Regressed in #1607.

### Checklist:


grynspan:
Self-reviewing as it's an obvious change.

PR #1615, opened 2026-03-07
Configurable interop modes

jerryjrchen:
Support none/limited/complete/strict interop modes with varying degrees of noisiness when handling interop issues.

### Motivation:

Allow users to configure the interop mode for their tests.

### Modifications:

* Modes will only apply if `SWT_EXPERIMENTAL_INTEROP_ENABLED` is enabled.

* limited: warnings only, complete: report issues as errors, strict: fatalError when handling interop issues

* Default mode is "complete" if opted-in to interop. If interop becomes enabled by default in the future, this will be updated.

Resolves rdar://171907588
### Checklist:

### DIFF
--- a/Sources/Testing/Events/Event+FallbackEventHandler.swift
+++ b/Sources/Testing/Events/Event+FallbackEventHandler.swift
@@ -14,8 +14,7 @@ private import _TestingInternals
 public enum Interop: Sendable {}
 
 extension Interop {
-  @_spi(Experimental) @_spi(ForToolsIntegrationOnly)
-  public enum Mode: String, Sendable, CaseIterable {
+  public enum Mode: String, Sendable, Codable, CaseIterable {
     /// The interop feature is not active.
     case none
 
@@ -43,22 +42,29 @@ extension Interop {
   }
 }
 
+extension Interop {
+  /// Name of the environment variable flag to set when opting-in to the
+  /// experimental interop feature.
+  static let experimentalOptInKey = "SWT_EXPERIMENTAL_INTEROP_ENABLED"
+}
+
 extension Interop.Mode {
+  /// The name for the environment variable which if set, overrides the default
+  /// interop mode.
   static let interopModeEnvKey = "SWIFT_TESTING_XCTEST_INTEROP_MODE"
 
+  /// Whether this interop mode causes Swift Testing to install a fallback event
+  /// handler ahead of running tests.
   var requiresInstallation: Bool {
-    self != .none
+    Environment.flag(named: Interop.experimentalOptInKey) == true && self != .none
   }
 
   /// Current interop mode, which should not be changed after tests start
   /// running.
   @_spi(Experimental) @_spi(ForToolsIntegrationOnly)
-  public static let current: Self = _currentImpl()
-
-  /// Exposed for testing only.
-  static func _currentImpl() -> Self {
-    Environment.variable(named: interopModeEnvKey).flatMap(Interop.Mode.init) ?? .complete
-  }
+  public static let current: Self = {
+    Environment.variable(named: interopModeEnvKey).flatMap(Interop.Mode.init) ?? .limited
+  }()
 }
 
 extension Event {
@@ -84,12 +90,15 @@ extension Event {
       return
     }
 
+    let xctestWarningMessage =
+      "XCTest API was used in a Swift Testing test. Adopt Swift Testing primitives, such as #expect, instead."
+
     // For the time being, assume that foreign test events originate from XCTest
-    let warnForXCTestUsageIssue =
+    lazy var warnForXCTestUsageIssue =
       Issue(
         kind: .apiMisused, severity: .warning,
         comments: [
-          "XCTest API was used in a Swift Testing test. Adopt Swift Testing primitives, such as #expect, instead."
+          .init(rawValue: xctestWarningMessage)
         ], sourceContext: issue.sourceContext)
 
     switch Interop.Mode.current {
@@ -106,7 +115,7 @@ extension Event {
       issue.severity = .error
       issue.record()
       fatalError(
-        "XCTest API was used in a Swift Testing test. Adopt Swift Testing primitives, such as #expect, instead. This is a fatal error because strict interop mode is active (\(Interop.Mode.interopModeEnvKey)=strict)",
+        "\(xctestWarningMessage) This is a fatal error because strict interop mode is active (\(Interop.Mode.interopModeEnvKey)=strict)",
       )
     }
   }
@@ -119,7 +128,9 @@ extension Event {
   ///
   /// - Returns: A source location to use when reporting an issue about
   ///   `recordJSON`.
-  private static func _bestAvailableSourceLocation(forInvalidRecordJSON recordJSON: UnsafeRawBufferPointer) -> SourceLocation {
+  private static func _bestAvailableSourceLocation(
+    forInvalidRecordJSON recordJSON: UnsafeRawBufferPointer
+  ) -> SourceLocation {
     // TODO: try to actually extract a source location from arbitrary JSON?
 
     // If there's a test associated with the current task, it should have a
@@ -131,12 +142,13 @@ extension Event {
     return .unknown
   }
 
-#if !SWT_NO_INTEROP
+  #if !SWT_NO_INTEROP
   /// The fallback event handler to install when Swift Testing is the active
   /// testing library.
   private static let _ourFallbackEventHandler: SWTFallbackEventHandler = {
     recordJSONSchemaVersionNumber, recordJSONBaseAddress, recordJSONByteCount, _ in
-    let version = String(validatingCString: recordJSONSchemaVersionNumber)
+    let version =
+      String(validatingCString: recordJSONSchemaVersionNumber)
       .flatMap(VersionNumber.init)
       .flatMap { ABI.version(forVersionNumber: $0) } ?? ABI.v0.self
     let recordJSON = UnsafeRawBufferPointer(
@@ -169,17 +181,15 @@ extension Event {
       ).record()
     }
   }
-#endif
+  #endif
 
   /// The implementation of ``installFallbackEventHandler()``.
   private static let _installFallbackEventHandler: Bool = {
-#if !SWT_NO_INTEROP
-    if Environment.flag(named: "SWT_EXPERIMENTAL_INTEROP_ENABLED") == true
-      && Interop.Mode.current.requiresInstallation
-      {
+    #if !SWT_NO_INTEROP
+    if Interop.Mode.current.requiresInstallation {
       return _swift_testing_installFallbackEventHandler(Self._ourFallbackEventHandler)
     }
-#endif
+    #endif
     return false
   }()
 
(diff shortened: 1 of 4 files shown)


Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -10,6 +10,57 @@
 
 private import _TestingInternals
 
+@_spi(Experimental) @_spi(ForToolsIntegrationOnly)
+public enum Interop: Sendable {}
+
+extension Interop {
+  @_spi(Experimental) @_spi(ForToolsIntegrationOnly)
```
stmontgomery: These `@_spi` groups are implied by the enclosing type, so you don't have to repeat them here

Tests/TestingTests/InteropModeTests.swift
```diff
@@ -0,0 +1,68 @@
...
+  @Test func `Default interop mode`() async {
+    #expect(Interop.Mode.current == .complete)
+  }
+
+  /// Run this test case in a separate process via exit tests since we will
+  /// be modifying the environment during the test.
+  ///
+  /// Ideally we'd run each value in a separate exit test, but that requires
+  /// combining parameterized tests and exit tests which is currently unsupported.
+  @Test func `Read interop modes from environment`() async {
```
stmontgomery: This is the only test function in this suite which uses exit tests, right? Could the `#if !SWT_NO_EXIT_TESTS` guard be more narrowly-scoped so that it only surrounds this one test function?

Tests/TestingTests/InteropModeTests.swift
```diff
@@ -0,0 +1,68 @@
...
+
+  @Test func `Default interop mode`() async {
+    #expect(Interop.Mode.current == .complete)
+  }
+
+  /// Run this test case in a separate process via exit tests since we will
+  /// be modifying the environment during the test.
+  ///
+  /// Ideally we'd run each value in a separate exit test, but that requires
+  /// combining parameterized tests and exit tests which is currently unsupported.
```
stmontgomery: You should be able to use exit testing in a parameterized test function as long as the parameter is `Codable & Sendable` and (importantly) you use an explicit capture list in the exit test closure. See [ST-0012: Capturing values in exit tests](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0012-exit-test-value-capturing.md).

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -10,6 +10,57 @@
...
+    /// `fatalError` upon any XCTest assertion failures when running a
+    /// Swift Testing test.
+    case strict
+  }
+}
+
+extension Interop.Mode {
+  static let interopModeEnvKey = "SWIFT_TESTING_XCTEST_INTEROP_MODE"
+
+  var requiresInstallation: Bool {
```
stmontgomery, line 58: These could use documentation

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -28,22 +79,36 @@ extension Event {
...
-          let issue = Issue(decoding: event) else {
+      var issue = Issue(decoding: event)
+    else {
       return
     }
 
     // For the time being, assume that foreign test events originate from XCTest
-    let warnForXCTestUsageIssue = {
-      return Issue(
+    let warnForXCTestUsageIssue =
```
stmontgomery: How about we make this `lazy var` instead of `let` so that it's only created lazily on-demand if it's needed. That'll avoid creating issues when mode is `.none`

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -28,22 +79,36 @@ extension Event {
...
+      warnForXCTestUsageIssue.record()
+    case .complete:
+      issue.severity = .error
+      issue.record()
+      warnForXCTestUsageIssue.record()
+    case .strict:
+      issue.severity = .error
+      issue.record()
+      fatalError(
+        "XCTest API was used in a Swift Testing test. Adopt Swift Testing primitives, such as #expect, instead. This is a fatal error because strict interop mode is active (\(Interop.Mode.interopModeEnvKey)=strict)",
```
stmontgomery: The first part of this string literal, 

```swift
"XCTest API was used in a Swift Testing test. Adopt Swift Testing primitives, such as #expect, instead."
```

is copied above in the `warnForXCTestUsageIssue` comment. Could that substring be hoisted out as a local `let` constant and shared in both places?

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -109,7 +174,9 @@ extension Event {
   /// The implementation of ``installFallbackEventHandler()``.
   private static let _installFallbackEventHandler: Bool = {
 #if !SWT_NO_INTEROP
-    if Environment.flag(named: "SWT_EXPERIMENTAL_INTEROP_ENABLED") == true {
+    if Environment.flag(named: "SWT_EXPERIMENTAL_INTEROP_ENABLED") == true
```
stmontgomery: Now that you've added a top-level `Interop` namespace type in the module, it might be nice to move the definition of this `SWT_EXPERIMENTAL_INTEROP_ENABLED` env var string there (similar to how you defined the other env var as a constant in this PR). There's at least one place where tests hardcode it

Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -109,7 +174,9 @@ extension Event {
...
+      && Interop.Mode.current.requiresInstallation
```
stmontgomery: Can the `requiresInstallation` computed property be the place where the `Environment.flag(named: "SWT_EXPERIMENTAL_INTEROP_ENABLED")` check is performed, to centralize that? I.e.

```swift
  var requiresInstallation: Bool {
    Environment.flag(named: "SWT_EXPERIMENTAL_INTEROP_ENABLED") == true && self != .none
  }
```

Tests/TestingTests/EventHandlingInteropTests.swift
```diff
@@ -148,81 +155,133 @@ struct EventHandlingInteropTests {
     }
   }
 
+  // MARK: - End to end handling of an issue with interop
+
+  /// Simulates the behaviour of XCTFail when called in a Swift Testing test.
+  /// This always forwards a test failure through the fallback event handler if it can find one.
+  /// It is an error to call this when a handler hasn't been installed yet.
+  /// - Parameter payload: Optional payload to use instead of generating a standard one.
+  private static func FakeXCTFail(payload: Data? = nil) throws {
```
stmontgomery: style nit (for private funcs)
```suggestion
  private static func _fakeXCTFail(payload: Data? = nil) throws {
```

Tests/TestingTests/EventHandlingInteropTests.swift
```diff
@@ -148,81 +155,133 @@ struct EventHandlingInteropTests {
...
+    let currentHandler = try #require(
+      _swift_testing_getFallbackEventHandler(),
+      "A fallback event handler must be installed ahead of time")
```
stmontgomery: Totally optional, but FYI: When a comment passed to `#expect` is just a constant string (without any interpolation segments), it can be a `//`-style code comment on the previous line, and magically(!) the macro will see it and include it in the results as if you'd written it the same way you have here. So this is exactly equivalent:
```suggestion
    // A fallback event handler must be installed ahead of time
    let currentHandler = try #require(_swift_testing_getFallbackEventHandler())
```

Tests/TestingTests/EventHandlingInteropTests.swift
stmontgomery: Separate from its name, you could make this a global function (within this file) rather than a `static func`, so that it more closely resembles `XCTFail()`. Very stylistic and optional suggestion though

Tests/TestingTests/EventHandlingInteropTests.swift
```diff
@@ -231,3 +290,19 @@ struct EventHandlingInteropTests {
   }
 }
 #endif
+
+extension Test {
+  func runCapturingIssues() async -> [Issue] {
```
stmontgomery: Can this be `fileprivate`?

Alternatively: If you foresee this utility being useful to other tests in the future (and I could definitely imagine that), it could be moved to `TestingAdditions.swift` and left (implicitly) `internal`

Tests/TestingTests/EventHandlingInteropTests.swift
```diff
@@ -148,81 +155,133 @@ struct EventHandlingInteropTests {
...
+      try #require(Event.installFallbackEventHandler())
+
+      // Run the test, which should record two issues in response to the interop one
+      let issues = await Test {
+        try Self.FakeXCTFail()
+      }.runCapturingIssues()
+
+      // Generate the interop test failure and warning about XCTest API usage
+      #expect(
+        issues.map { $0.severity } == [.error, .warning]
```
stmontgomery: Are you intentionally trying to guarantee that the relative order of the two issues (in `complete` mode) is that the "assertion failure" one is always first (as error severity), followed by the runtime warning? Not objecting necessarily, but observing that this test requires that order to be maintained.

Tests/TestingTests/EventHandlingInteropTests.swift
jerryjrchen: Hmm, I was trying to make this look like `XCTFail()` but admittedly I'm already making compromises there by making it `throws` so I might as well update the naming

Tests/TestingTests/InteropModeTests.swift
jerryjrchen: Very neat, glad I wrote the comment then :P

Tests/TestingTests/EventHandlingInteropTests.swift
jerryjrchen: Makes sense here since I'm not creating a formatted string or anything like that

Tests/TestingTests/EventHandlingInteropTests.swift
jerryjrchen: Moved to TestingAdditions

Tests/TestingTests/EventHandlingInteropTests.swift
jerryjrchen: The order doesn't matter, so I'll sort these for now

Tests/TestingTests/EventHandlingInteropTests.swift
jerryjrchen: Yeah, makes sense here since I'm not doing any string interpolation

jerryjrchen:
```
2026-03-11T22:44:35.4518324Z ✘ Test "Strict interop mode causes a process exit" recorded an issue at EventHandlingInteropTests.swift:235:11: Expectation failed: expected exit status ".signal(SIGTRAP)", but ".signal(SIGILL)" was reported instead
2026-03-11T22:44:35.4520086Z ↳ .signal(SIGTRAP) → .signal(SIGILL)
2026-03-11T22:44:35.4520647Z ↳   SIGTRAP → 5
2026-03-11T22:44:35.4520955Z ↳   SIGILL → 4
```

Hmm I think this is x86_64 behaviour 😓 : https://developer.apple.com/documentation/xcode/sigtrap_sigill

PR #1617, opened 2026-03-09
Expose SPI to cast `ABI.EncodedTest` to/from `Test`.

grynspan:
Follow-up to #1587.

Exposes SPI initializers on `ABI.EncodedTest` and `Test` to convert to/from each other. The initializers are inherently lossy (after all, you can't encode a test body in JSON).

### Checklist:

### DIFF
--- a/Tests/TestingTests/ABIEntryPointTests.swift
+++ b/Tests/TestingTests/ABIEntryPointTests.swift
@@ -10,11 +10,15 @@
 
 #if canImport(Foundation) && !SWT_NO_ABI_ENTRY_POINT
 @testable @_spi(Experimental) @_spi(ForToolsIntegrationOnly) import Testing
+private import _TestingInternals
 
 #if canImport(Foundation)
 private import Foundation
 #endif
-private import _TestingInternals
+
+#if !SWT_TARGET_OS_APPLE && canImport(Synchronization)
+import Synchronization
+#endif
 
 @Suite("ABI entry point tests")
 struct ABIEntryPointTests {


Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -38,7 +38,7 @@ extension ABI {
     var displayName: String?
 
     /// The source location of this test.
-    public var sourceLocation: EncodedSourceLocation<V>
+    var sourceLocation: EncodedSourceLocation<V>
```
jerryjrchen, line 41: Is this change intentional? My understanding is that this is breaking, although it's only been public for like a few days so unlikely anyone has used it. Would it be worth doing a toolchain build just to be extra sure?

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -172,3 +153,168 @@ extension ABI {
...
+  /// - Parameters:
+  ///   - test: The encoded test to initialize this instance from.
+  ///
+  /// The resulting instance of ``Test`` cannot be run; attempting to do so will
+  /// throw an error.
+  public init?<V>(decoding test: ABI.EncodedTest<V>) {
+    let sourceLocation = SourceLocation(decoding: test.sourceLocation) ?? .unknown
+    let typeInfo = Self._makeTypeInfo(for: test)
+
+    // Construct the (partial) list of traits available in the encoded test.
```
jerryjrchen, line 268: Is there any risk here to not including the full list of traits, or at least hinting there are more traits? I'll mention that I'm in favour of this approach where we do encode/decode whatever traits we can.

If we drop traits when going from one form to another, I think it might be confusing/surprising to an end user.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -115,7 +115,7 @@ func entryPoint(passing args: __CommandLineArguments_v0?, eventHandler: Event.Ha
 
       // Post an event for every discovered test. These events are turned into
       // JSON objects if JSON output is enabled.
-      for test in tests {
+      for test in tests where !test.isHidden {
```
jerryjrchen, line 118: Is this change related?

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -172,3 +153,168 @@ extension ABI {
...
+      // The string wasn't structured as expected for a Swift Testing or XCTest
+      // test ID.
+      return nil
+    }
+
+    // Replace the first component string, which is currently shaped like
+    // "ModuleName.TypeName", with ["ModuleName", "TypeName"]
+    let secondTestIDComponent = testID[moduleName.endIndex ..< firstComponent.endIndex].dropFirst()
+    testIDComponents[0] = moduleName
+    testIDComponents.insert(secondTestIDComponent, at: 1)
```
jerryjrchen, line 235: Could we get some test coverage for different testID formats? It would be helpful to see some examples of how the `/` and `.` separators can mix together.

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
grynspan: Nobody's using it. I lowered it because my thinking is now that, outside Swift Testing itself, if you want to decode an instance of Test, you presumably want to decode the whole thing.

(We can open it up again later if I'm wrong!)

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
grynspan: It's impossible to encode/decode all traits. The decoded form is only an approximation of the original data.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: Yes, unborks the test I wrote.

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
grynspan: There are no other test ID formats.

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -172,3 +153,168 @@ extension ABI {
...
+  /// The resulting instance of ``Test`` cannot be run; attempting to do so will
+  /// throw an error.
+  public init?<V>(decoding test: ABI.EncodedTest<V>) {
+    let sourceLocation = SourceLocation(decoding: test.sourceLocation) ?? .unknown
+    let typeInfo = Self._makeTypeInfo(for: test)
+
+    // Construct the (partial) list of traits available in the encoded test.
+    // Note we do not try to encode _all_ traits because many trait types simply
+    // cannot be represented as JSON.
+    var traits = [any Trait]()
```
harlanhaskins, line 271: Do we have some larger grand design for formalizing this with a protocol? I mean I suppose `Encodable & Trait` is enough

grynspan: Not as such. We can _encode_ arbitrary traits, but cannot _decode_ them, and traits being a Swift-Testing-specific feature means they don't translate well to other testing libraries that might e.g. track "tags" but not "traits".

PR #1621, opened 2026-03-10
Fix incorrect type when trying to encode an attachment as JSON.

grynspan:
This fixes us using the wrong type when encoding an attachment as JSON if Foundation is not available. (This is not expected to be a fast path, so the array copy is fine).

### Checklist:


grynspan:
Self-reviewing as it's a tiny change.

PR #1622, opened 2026-03-10
Reimplement `_TestingInterop` in C++.

grynspan:
Normally we try to _reduce_ our C++ content, but the complexity's much lower here since we can't use Swift's atomics properly on Darwin yet and since we don't need to juggle Swift module visibility.

### Checklist:


Sources/_TestingInterop/include/Empty.h
jerryjrchen, line 1: Empty file here is just to keep the compiler happy?

Sources/_TestingInterop/CMakeLists.txt
```diff
@@ -8,10 +8,19 @@
 
 include(ModuleABIName)
 add_library(_TestingInterop
-  FallbackEventHandler.swift)
+  FallbackEventHandler.cpp)
+if("${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}" STREQUAL "MSVC" OR
+   "${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+  target_compile_options(_TestingInternals PRIVATE
```
jerryjrchen: Is this supposed to be compile options for `_TestingInterop` instead?

jerryjrchen (approved):

Sources/_TestingInterop/CMakeLists.txt
grynspan: Probably!

Sources/_TestingInterop/include/Empty.h
grynspan: Yarp.

PR #1626, opened 2026-03-13
Fix unintentionally-quadratic COW copies on insertion

harlanhaskins:
This fixes an accidentally-quadratic COW copy of a subgraph whenever an item is inserted.

### Motivation:

For large Swift Testing test suites with many parameterized test cases, we accidentally tripped this specific pathological failure mode of the graph insertion code; when we insert a new child, we mutate each graph entry along the ancestor chain. If we mutate our local copy while leaving the existing copy in the tree, it causes a COW copy of the whole tree on every insertion.

For the attached stress test, without the `.take()` it takes *2 seconds* on my MacBook Pro, and with the `.take()` it takes 40 milliseconds.

### Modifications:

The bulk of this fix is to `.take()` the child from the dictionary (`Optional.take()` sets the receiver to nil and returns the previous wrapped value).

This is a very similar problem and fix to this PR in `swift-driver`: https://github.com/swiftlang/swift-driver/pull/654

### Checklist:

### DIFF
--- a/Tests/TestingTests/Support/GraphTests.swift
+++ b/Tests/TestingTests/Support/GraphTests.swift
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-@_spi(Experimental) @_spi(ForToolsIntegrationOnly) @testable import Testing
+@_spi(Experimental) @testable import Testing
 
 @Suite("Graph<K, V> Tests")
 struct GraphTests {
@@ -550,7 +550,7 @@ struct GraphTests {
     #expect(Set(values) == [123, 124, 456, 457, 13579, 13580, 789, 790, 2468, 2469])
   }
 
-  @Test
+  @Test(.enabled(if: performanceTestsEnabled))
   func `inserting many items with the same prefix is not quadratic`() {
     // This test verifies that inserting various numbers of elements that
     // all share a common prefix scales ~linearly.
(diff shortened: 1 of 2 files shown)


Tests/TestingTests/Support/GraphTests.swift
```diff
@@ -549,4 +549,37 @@ struct GraphTests {
...
+    let totalTime = ContinuousClock.now - startTime
+    let avgTime = totalTime / Double(itemCount)
+
+    print("  \(itemCount) insertions: \(totalTime) total")
+    print("  Average per insert: \(avgTime)")
+
+    #expect(graph[sharedPrefix + ["Argument 0"]] == 0)
+    #expect(graph[sharedPrefix + ["Argument \(itemCount - 1)"]] == itemCount - 1)
+
+    #expect(totalTime.components.seconds < 1)
```
stmontgomery: I worry this may be flaky, since this test may run in constrained environments. I realize we don't have a good benchmarking solution currently

harlanhaskins: Yeah, I don't like this method of measuring. I wonder if we can reduce the flake by running 3 different scales and making sure the time is not quadratic?

stmontgomery: Yeah, that'd be better, assuming there's some tolerance for variability

grynspan:
Nice. Please give me a chance to review on Monday before merging!

Tests/TestingTests/Support/GraphTests.swift
harlanhaskins: Updated it to a scale test; the scale test just verifies that at various different scales, no item count takes more than 30% of the average time per-item

Tests/TestingTests/Support/GraphTests.swift
```diff
@@ -8,7 +8,7 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-@testable import Testing
+@_spi(Experimental) @_spi(ForToolsIntegrationOnly) @testable import Testing
```
grynspan: Nitpick: anything experimental actually being used here?

harlanhaskins: Just `Test.Clock`, but also that doesn't need `ForToolsIntegrationOnly`

grynspan: I think you have that backwards?

harlanhaskins: Ah, turns out you only need one of the SPI tags to get the declaration. So I could pick either or leave both

PR #1627, opened 2026-03-15
Preserve contextual type for literal `arguments:` expressions in `@Test`

ojun9:
Preserve contextual type for literal `arguments:` expressions in `@Test`

### Motivation:

Parameterized tests using `@Test(arguments:)` may fail to compile when the collection is written as an array literal containing values that rely on contextual type inference.

For example:

```swift
@Test<[(String?, Int?)]>(arguments: [
  (nil, 2),
  ("c", nil),
  ("d", nil)
])
func f(s: String?, i: Int?) {}
```

During macro expansion, the `arguments:` expression is wrapped in a closure so the collection can be evaluated lazily:

```swift
arguments: { [...] }
```

When the array literal is evaluated inside this closure, the contextual type that would normally be provided by the test function parameters is no longer available. As a result, values such as `nil` cannot be inferred and the code fails to type-check.

The same code works if the array literal is explicitly cast:

```swift
@Test(arguments: [
  (nil, 1),
  ("a", nil)
] as [(String?, Int?)])
func f(s: String?, i: Int?) {}
```

In this case the generated code becomes:

```swift
arguments: { [...] as [(String?, Int?)] }
```

The explicit cast preserves the contextual type inside the closure and allows the literals to type-check.

### Modifications:

When `arguments:` is supplied as a single array literal, the macro derives the expected array type from the test function parameter list and applies it as an explicit cast inside the generated closure.

For example, the generated expression becomes:

```swift
arguments: { [...] as [(String?, Int?)] }
```

This preserves the contextual type required for literal inference while keeping the existing lazy evaluation behavior.

Macro expansion only has access to source syntax and does not observe the inferred collection type directly. For this reason, the array type is derived from the test function parameter list. The change is intentionally limited to the case where `arguments:` is a single array literal so that other overloads remain unaffected.

Additional tests verify that contextual types are preserved for empty array literals and tuple literals containing optionals.

### Checklist:


grynspan:
Type checking of arguments to the `@Test` macro occurs _before_ the expansion itself. Did you test this change to confirm it actually works? The added test only checks the macro expansion (without any type checking).

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +185,38 @@ struct AttributeInfo {
...
+    guard let functionDecl = declaration.as(FunctionDeclSyntax.self) else {
+      return nil
+    }
+
+    let parameters = functionDecl.signature.parameterClause.parameters
+    guard !parameters.isEmpty else {
+      return nil
+    }
+
+    if testFunctionArguments.count == 1, expression.is(ArrayExprSyntax.self) {
```
grynspan: It should be safe to generalize this so that the argument count only needs to match the parameter count.

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +185,38 @@ struct AttributeInfo {
...
+  private func _contextualTypeForLiteralArgument(
+    for expression: ExprSyntax,
+    among testFunctionArguments: [Argument]
+  ) -> String? {
+    guard let functionDecl = declaration.as(FunctionDeclSyntax.self) else {
+      return nil
+    }
+
+    let parameters = functionDecl.signature.parameterClause.parameters
+    guard !parameters.isEmpty else {
```
grynspan:
```suggestion
    if parameters.isEmpty {
```

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +185,38 @@ struct AttributeInfo {
...
+  ///
+  /// Parameterized `@Test` declarations are modeled in terms of the collection
+  /// type supplied to the macro, but macro expansion only sees source syntax.
+  /// When the `arguments:` parameter is supplied as a single array literal,
+  /// reconstruct the array type from the test function's parameters so the
+  /// literal retains enough contextual type information after lazy wrapping.
+  private func _contextualTypeForLiteralArgument(
+    for expression: ExprSyntax,
+    among testFunctionArguments: [Argument]
+  ) -> String? {
```
grynspan:
```suggestion
  ) -> TypeSyntax? {
```

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +185,38 @@ struct AttributeInfo {
...
+      return nil
+    }
+
+    if testFunctionArguments.count == 1, expression.is(ArrayExprSyntax.self) {
+      if parameters.count == 1, let parameter = parameters.first {
+        // A single-parameter test expects collection elements of the parameter
+        // type itself, not tuple-shaped elements.
+        return "[\(parameter.baseTypeName)]"
+      }
+      let elementType = parameters.map(\.baseTypeName).joined(separator: ", ")
```
grynspan: Can we construct an `ArrayTypeSyntax` here instead and leave the string interpolation to swift-syntax?

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +185,38 @@ struct AttributeInfo {
...
+    let parameters = functionDecl.signature.parameterClause.parameters
+    guard !parameters.isEmpty else {
+      return nil
+    }
+
+    if testFunctionArguments.count == 1, expression.is(ArrayExprSyntax.self) {
+      if parameters.count == 1, let parameter = parameters.first {
+        // A single-parameter test expects collection elements of the parameter
+        // type itself, not tuple-shaped elements.
+        return "[\(parameter.baseTypeName)]"
```
grynspan: Tokens from the original source need to be trimmed.

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +185,38 @@ struct AttributeInfo {
 
     return LabeledExprListSyntax(arguments)
   }
+
+  /// The contextual type to explicitly apply to a literal `arguments:`
+  /// expression after it is wrapped in a closure for lazy evaluation.
+  ///
```
grynspan, line 200: Need parameter and return callouts.

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -171,7 +171,12 @@ struct AttributeInfo {
     if let testFunctionArguments {
       arguments += testFunctionArguments.map { argument in
         var copy = argument
-        copy.expression = .init(ClosureExprSyntax { argument.expression.trimmed })
+        let argumentExpr = argument.expression.trimmed
```
grynspan: We can simplify this part of the diff by modifying `argumentExpr` before creating the closure wrapper.

grynspan (changes requested):
Additional test coverage is needed too.

ojun9:
Thanks for the question.

You're right that the existing test only verifies the macro expansion.
The change is not intended to make the following case compile:

```swift
@Test(arguments: [
  (nil, 1),
  ("a", nil)
])
func f(s: String?, i: Int?) {}
```

This case fails with `Type of expression is ambiguous without a type annotation` even without the macro, because the array literal lacks a contextual type.

The change addresses the situation where the collection type is already provided, for example via a generic parameter:

```swift
@Test<[(String?, Int?)]>(arguments: [
  (nil, 2),
  ("c", nil),
  ("d", nil)
])
func f(s: String?, i: Int?) {}
```

In this situation the collection element type is known, but wrapping the literal inside the generated closure removes the contextual type required for literal inference.

The added cast restores that contextual type and allows the code to compile.
I confirmed that this generic-parameter case compiles correctly with the change.

grynspan:
I haven't forgotten about this PR, but I have been tied up with other work.

We should make sure to handle dictionary literals correctly as well, and can improve test efficiency by inferring them as instances of `KeyValuePairs` rather than `Dictionary`. Note that tuple desugaring on dictionary elements is always to exactly two test function arguments and does not occur if there are multiple argument collections in the `@Test` macro.

Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
```diff
@@ -174,12 +174,18 @@ extension FunctionParameterSyntax {
...
-  /// The base type name of this parameter.
-  var baseTypeName: String {
+  /// The underlying type of this parameter with any attributed type wrappers
+  /// (such as `inout`) removed.
+  var baseType: TypeSyntax {
     // Discard any specifiers such as `inout` or `borrowing`, since we're only
     // trying to obtain the base type to reference it in an expression.
     let baseType = type.as(AttributedTypeSyntax.self)?.baseType ?? type
-    return baseType.trimmedDescription
+    return baseType.trimmed
```
grynspan: Avoid trimming both here and in the  call sites as it's somewhat expensive. I would leave it untrimmed here so that it can still be used for diagnostic attribution, but either choice is fine.

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -169,9 +169,24 @@ struct AttributeInfo {
...
+        ) {
+          expr = ExprSyntax(
+            AsExprSyntax(
+              expression: expr,
+              asKeyword: .keyword(.as, leadingTrivia: .space, trailingTrivia: .space),
+              type: contextualType.trimmed
+            )
+          )
+        }
+        let lazyExpression = expr.trimmed
```
grynspan: Already trimmed by this point. (New nodes created programmatically are _de facto_ trimmed if you don't explicitly specify trivia).

grynspan (commented):
We still need some fixture tests in the TestingTests target. See `NonSendableTests` for an example of how to write one (empty test that triggers this macro expansion code when we compile our own tests so that if the code isn't working, we fail to build). Those tests should go in the TestingTests target.

Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
```diff
@@ -174,12 +174,18 @@ extension FunctionParameterSyntax {
...
+  }
+
+  /// The base type name of this parameter.
+  var baseTypeName: String {
```
grynspan, line 186: Is anything still using this that can't accept a `TypeSyntax` instead? We should (eventually) replace any such code, but it isn't _critical_ for this PR. 🙂

ojun9: I don't think there's anything in this PR that needs to change for that, but I agree we should eventually migrate any remaining String-based uses to TypeSyntax where possible.

grynspan: Sounds good, thanks.

Sources/TestingMacros/Support/Additions/FunctionDeclSyntaxAdditions.swift
```diff
@@ -174,12 +174,18 @@ extension FunctionParameterSyntax {
...
-  /// The base type name of this parameter.
-  var baseTypeName: String {
+  /// The underlying type of this parameter with any attributed type wrappers
+  /// (such as `inout`) removed.
+  var baseType: TypeSyntax {
     // Discard any specifiers such as `inout` or `borrowing`, since we're only
     // trying to obtain the base type to reference it in an expression.
     let baseType = type.as(AttributedTypeSyntax.self)?.baseType ?? type
-    return baseType.trimmedDescription
+    return baseType
```
grynspan:
```suggestion
    type.as(AttributedTypeSyntax.self)?.baseType ?? type
```

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +194,84 @@ struct AttributeInfo {
...
+            for parameter in parameters {
+              TupleTypeElementSyntax(type: parameter.baseType.trimmed)
+            }
+          })
+        )
+        return TypeSyntax(ArrayTypeSyntax(element: elementType))
+      }
+    }
+
+    if expression.is(DictionaryExprSyntax.self) {
```
grynspan:
```suggestion
    } else if expression.is(DictionaryExprSyntax.self) {
```

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +194,84 @@ struct AttributeInfo {
...
+      if testFunctionArguments.count == 1, parameters.count == 2 {
+        return TypeSyntax(
+          IdentifierTypeSyntax(
+            name: .identifier("KeyValuePairs"),
```
grynspan, line 262: We need to always explicitly specify the module name in our macro expansions (`"Swift"` in this case). So you need to use a `MemberTypeSyntax` here rather than a bare `IndentifierTypeSyntax`.

Tests/TestingTests/Test.Case.ArgumentTests.swift
```diff
@@ -165,11 +165,24 @@ struct ParameterizedTests {
   @Test(.hidden, arguments: [("value", 123)])
   func one2TupleParameter(x: (String, Int)) {}
 
+  @Test<[(String?, Int?)]>(.hidden, arguments: [
```
grynspan:
```suggestion
  @Test(.hidden, arguments: [
```

Generic parameters on `@Test` are not supported and will diagnose if used.

Tests/TestingTests/Test.Case.ArgumentTests.swift
```diff
@@ -165,11 +165,24 @@ struct ParameterizedTests {
...
+  ])
+  func contextualArrayLiteral(x: String?, y: Int?) {}
+
   @Test(.hidden, arguments: ["value": 123])
   func twoDictionaryElementParameters(x: String, y: Int) {}
 
   @Test(.hidden, arguments: ["value": 123])
   func oneDictionaryElementTupleParameter(x: (key: String, value: Int)) {}
 
+  @Test<KeyValuePairs<String, Int?>>(.hidden, arguments: [
```
grynspan:
```suggestion
  @Test(.hidden, arguments: [
```

Generic parameters on `@Test` are not supported and will diagnose if used.

grynspan:
Sorry for the delay! I was out sick.

grynspan:
Let me know when you're ready for another review.

Tests/TestingTests/Test.Case.ArgumentTests.swift
ojun9: I removed the unsupported generic argument clause from @Test.

The remaining type annotation is expressed as an explicit cast on the arguments: expression.
This preserves contextual type information (particularly for nil literals) and is consistent with the macro’s supported expansion semantics.

ojun9:
@grynspan 
No worries, hope you're feeling better!
Ready for another review!

grynspan (approved):

grynspan:
If you see the "merge" button, please make sure to squash. Otherwise ping me and I'll click the button.

Sources/TestingMacros/Support/AttributeDiscovery.swift
```diff
@@ -180,4 +194,83 @@ struct AttributeInfo {
...
+          })
+        )
+        return TypeSyntax(ArrayTypeSyntax(element: elementType))
+      }
+    } else if expression.is(DictionaryExprSyntax.self) {
+      if testFunctionArguments.count == 1, parameters.count == 2 {
+        return TypeSyntax(
+          MemberTypeSyntax(
+            baseType: IdentifierTypeSyntax(name: .identifier("Swift")),
+            name: .identifier("KeyValuePairs"),
```
stmontgomery, line 262: I'm surprised to see `DictionaryExprSyntax` being interpreted as `KeyValuePairs`. I see [this comment](https://github.com/swiftlang/swift-testing/pull/1627#issuecomment-4126185940) which I'm guessing is what led to that:

> We should make sure to handle dictionary literals correctly as well, and can improve test efficiency by inferring them as instances of `KeyValuePairs` rather than `Dictionary`.

That changes the semantics some, though: elements with equal keys will run the test multiple times, and the order we dispatch test cases in will be stable, whereas with a dictionary the order is nondeterministic due to hashing.

Is switching to `KeyValuePairs` _necessary_ for this overall PR, or is it more of an opportunistic performance improvement? It feels like a potentially user-observable behavior change.

grynspan: This was my suggestion, yes.

> That changes the semantics some, though: elements with equal keys will run the test multiple times

Equal keys crash outright with a dictionary when it is initialized.

> and the order we dispatch test cases in will be stable, whereas with a dictionary the order is nondeterministic due to hashing.

This change is more consistent with how we treat arrays, yes.

> Is switching to `KeyValuePairs` _necessary_ for this overall PR, or is it more of an opportunistic performance improvement? It feels like a potentially user-observable behavior change.

It isn't _strictly_ necessary, but the only user-observable change is that the random order of elements is no longer random (which we never really intended it to be anyway).

Tests/TestingTests/Test.Case.ArgumentTests.swift
```diff
@@ -165,11 +165,24 @@ struct ParameterizedTests {
   @Test(.hidden, arguments: [("value", 123)])
   func one2TupleParameter(x: (String, Int)) {}
 
+  @Test(.hidden, arguments: [
+    (nil, 123),
+    ("value1", nil),
+    ("value2", nil),
+  ] as [(String?, Int?)])
```
stmontgomery, line 172: Why does this test, and the other new one below, include an explicit `as` cast on the passed-in collection? I thought this PR was intended to remove the need to specify that at the point where `@Test` is used?

I just pulled the branch locally, and when I try this example mentioned in the PR description _without_ an `as` cast **and** without `<...>` generic type parameters on `@Test`, it's still emitting an error:

```swift
@Test(arguments: [
  (nil, 2), // ❌ error: 'nil' cannot initialize specified type 'String'
  ("c", nil),
  ("d", nil)
])
func f(s: String?, i: Int?) {}
```

If the outcome of this PR is that a user still needs to either (a) use an explicit `as` cast or (b) include generic type parameters on `@Test`, then I'm not sure what the benefit is. Users can already do (a), and we recently began emitting a warning diagnostic for (b), and it's not meaningfully simpler than option (a) IMO anyway.

stmontgomery (changes requested):
I appreciate your effort on this @ojun9 — I really do _want_ to see this problem resolved so that test authors don't need to provide type information for test arguments explicitly. In fact, I took a stab at it myself a while back in [this PR](https://github.com/swiftlang/swift-testing/pull/808).

I'm [concerned](https://github.com/swiftlang/swift-testing/pull/1627#discussion_r3175352330) though that as written, it doesn't really help much in practice towards that goal. Unless I'm missing something, it _seems_ like it still requires specifying the type of the argument collection explicitly.

If that's not accurate, or if there are scenarios where this is helpful _without_ explicit type info that I'm overlooking, could you please give an example of that?

ojun9:
@stmontgomery 

Sorry for the very delayed reply.

Thanks, I see your point.

The case this PR helps with is one where the type is already provided somehow before expansion, for example:

```swift
@Test<[(String?, Int?)]>(arguments: [
(nil, 2),
("c", nil),
("d", nil),
])
func f(s: String?, i: Int?) {}
```

In this case, the original arguments: expression has enough context to type-check, but the generated closure loses that context unless the expansion adds as `[(String?, Int?)]`.

That said, I agree this is still explicit type information at the use site. I don’t currently know of an example where this PR makes the literal work with neither an as cast nor a generic argument clause.

PR #1628, opened 2026-03-16
Add `wasm32-unknown-emscripten` platform support

MaxDesiatov:
Now that both `swiftc` and `swift-frontend` support Emscripten target triple we can follow up on [the previously pitched Emscripte support](https://forums.swift.org/t/pitch-emscripten-target-support-for-swift/85310) in libraries like `swift-foundation`.


grynspan:
I would expect this to be tracked by `targetEnvironment()` rather than `os()` but okay.

If it is the intent of the Platform Steering Group to support Emscripten, we need CI support for it as well and some way to test changes.

Sources/Testing/CMakeLists.txt
```diff
@@ -132,12 +133,14 @@ target_link_libraries(Testing PRIVATE
...
 if(NOT APPLE)
-  if(NOT CMAKE_SYSTEM_NAME STREQUAL WASI)
+  if(NOT CMAKE_SYSTEM_NAME STREQUAL WASI AND NOT CMAKE_SYSTEM_NAME STREQUAL Emscripten)
     target_link_libraries(Testing PUBLIC
       dispatch)
   endif()
-  target_link_libraries(Testing PUBLIC
-    Foundation)
+  if(NOT CMAKE_SYSTEM_NAME STREQUAL Emscripten)
+    target_link_libraries(Testing PUBLIC
```
grynspan, line 144: Is Foundation not available in this environment? Is this a temporary constraint?

MaxDesiatov: This PR is marked as draft, as it's not ready for review yet.

PR #1635, opened 2026-03-19
Factor out bug report URL boilerplate into a utility function

stmontgomery:
Many places in the codebase terminate the program (via `fatalError`, `precondition`, or `assert`) or throw `SystemError` and include a message asking the developer to [file a bug report](https://github.com/swiftlang/swift-testing/issues/new). The URL and surrounding boilerplate text for this request are duplicated verbatim across 18 call sites. 

This PR factors this into a utility function to reduce repetition and make it easier to include these helpful messages consistently when adding new assertion sites in the future.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/BugReporting.swift
+++ b/Sources/TestingMacros/Support/BugReporting.swift
@@ -10,6 +10,4 @@
 
 /// A standardized message inviting a user to file a bug report at the provided
 /// URL.
-var fileABugMessage: String {
-  "Please file a bug report at https://github.com/swiftlang/swift-testing/issues/new"
-}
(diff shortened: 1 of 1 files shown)


grynspan:
Does it need to be a function call? Can't it just be an internal global and we interpolate it in?

grynspan:
Or even better, just have `SystemError` and/or `.system` synthesize it?

stmontgomery:
`SystemError`/`.system` is just one of the several ways the URL string is used. And the helper function also formats the _"and include this information: ..."_ suffix when relevant, so I find that valuable.

grynspan:
Okay, I don't morally object or anything. (We might want to noodle on it a bit before you merge it.)

PR #1642, opened 2026-03-26
Change Argument.ID.bytes to a SHA256 hash

harlanhaskins:
It's wasteful to send the entire JSON content of an argument as the ID for an argument. Use the SHA256 instead.

### Motivation:

The JSON string for each parameterized test argument can be very large; for one sample project, each argument was half a kilobyte, multiplied by ~10,000 cases and it starts to add up.

This PR standardizes argument IDs as SHA256 hashes. I considered doing this only when it would be profitable to do so (e.g. if the JSON encoding is less than 256 bits), but this makes collisions more likely since we wouldn't be able to guarantee the distribution of bytes that SHA256 gives us. But if that's not really a concern, it's easy enough to only perform this transformation when the encoded input is >32 bytes.

Fixes rdar://118992957

### Modifications:

- Adds a `_TestingUtilities` module and moves SHA256.swift into it
- Imports that from both TestingMacros and Testing
  - This required rebuilding it a 2nd time while building TestingMacros in CMake, which I think is acceptable since it's just one file for now.

### Checklist:

### DIFF
--- a/Tests/TestingTests/SHA256Tests.swift
+++ b/Tests/TestingTests/SHA256Tests.swift
@@ -11,7 +11,7 @@
 #if canImport(CryptoKit)
 
 import CryptoKit
-import Testing
+@testable import Testing
 
 @Suite
 struct SHA256Tests {


Tests/TestingTests/Test.Case.Argument.IDTests.swift
```diff
@@ -23,13 +23,14 @@ struct Test_Case_Argument_IDTests {
     let arguments = try #require(testCase.arguments)
     #expect(arguments.count == 1)
     let argument = try #require(arguments.first)
-    #expect(String(decoding: argument.id.bytes, as: UTF8.self) == "123")
+    #expect(argument.id.bytes == SHA256.hash("123".utf8))
```
jerryjrchen, line 26: Nit: maybe these should be SHA256 from CryptoKit as well kinda like the checks in `SHA256Tests.swift`?

grynspan: Then the tests would only run on Darwin. Gotta limit the impact of that dependency. 😬

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
```diff
@@ -110,13 +110,16 @@ extension Test.Case.Argument.ID {
   ///
   /// - Throws: Any error encountered during encoding.
   private static func _encode(_ value: some Encodable, parameter: Test.Parameter) throws -> [UInt8] {
-    try JSON.withEncoding(of: value, userInfo: [._testParameterUserInfoKey: parameter], Array.init)
+    /// The encoded representation of an argument is the SHA256 hash of its Codable JSON representation.
+    try JSON.withEncoding(of: value, userInfo: [._testParameterUserInfoKey: parameter]) { buffer in
+      SHA256.hash(buffer)
```
grynspan, line 115: Food for thought: is this the right layer at which to perform the hash? Or does it belong higher/lower on the food chain? (I don't know, just asking.)

harlanhaskins: My thought was that `_encode` is the place that makes the decision of what the bytes are for a given argument value, and that that encoding is an implementation detail. We could eventually build a custom Coder or whatever that builds an incremental hash or something and other clients wouldn't care, since these identifiers are only valid for a single test run (right? oh god right?)

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
```diff
@@ -110,13 +110,16 @@ extension Test.Case.Argument.ID {
...
+    }
   }
 #endif
 }
 
 /// A encodable type which wraps a ``CustomTestArgumentEncodable`` value.
-private struct _CustomArgumentWrapper<T>: RawRepresentable, Encodable where T: CustomTestArgumentEncodable {
+struct _CustomArgumentWrapper<T>: RawRepresentable, Encodable where T: CustomTestArgumentEncodable {
```
grynspan: Style: We only place these underscores on `private` symbols*, so if this is now `internal` you should remove the underscore.

(If this is a useful utility type you want to use elsewhere, consider moving it to its own file in the `Support` folder.)

\* We also add single leading underscores for symbols that provide backing storage for higher-visibility symbols, e.g.:

```swift
struct S {
  /// Storage for ``foo``.
  private var _foo: Foo?

  /// Blah blah.
  public var foo: Foo {
    get { ... }
    set { ... }
  }
}
```

harlanhaskins: Ah, good callout!

Sources/Testing/Support/SHA256.swift
```diff
@@ -0,0 +1,184 @@
+//
```
grynspan: Did GitHub actually duplicate this file?

harlanhaskins: It was a move; I figured it made more sense to be in Testing and referenced from TestingMacros than the other way around, but I can put it back if the symlink should point the other direction

Tests/TestingTests/SHA256Tests.swift
```diff
@@ -0,0 +1,32 @@
...
+import CryptoKit
+import Testing
+
+@Suite
+struct SHA256Tests {
+  @Test(arguments: [
+    [],
+    Array(0..<20),
+    Array("Hello, world".utf8),
+    Array(#"{"key": "value", "key2": 123, "key3": null}"#.utf8)
```
grynspan: Can we add:

- A random `UInt64`; and
- An exceptionally long (> 1KB), ideally RNG-generated, sequence of bytes?

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
grynspan: These identifiers need to ultimately be reusable so you can re-run individual test cases.

Sources/Testing/Support/SHA256.swift
grynspan: Well, it makes the diff bigger because it doesn't look like git understands the file _moved_, so for the sake of preserving git history I'd point the symlink the other way, yeah.

Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
harlanhaskins: Right, but if we change the encoding between releases of Swift Testing, that should be fine for that use case

grynspan: Maybe? We haven't specced it out either way.

Sources/TestingMacros/Support/SHA256.swift
```diff
@@ -13,7 +13,7 @@
 
 /// SHA-256 implementation from Secure Hash Algorithm 2 (SHA-2) set of
 /// cryptographic hash functions (FIPS PUB 180-2).
-enum SHA256 {
+package enum SHA256 {
```
grynspan: Does this need to be `package` if we emit the source into each target?

PR #1644, opened 2026-03-30
Avoid hanging trying to diff `Data` and similar collections.

grynspan:
This PR suppresses our diffing logic for collections with element types `Int8` or `UInt8`. The output is not currently useful for these types.

Resolves rdar://173002947.

### Checklist:


harlanhaskins (approved):
Makes sense, and we can always tweak this heuristic in the future.

stmontgomery (approved):

PR #1649, opened 2026-03-31
Add support for module selectors in types during macro expansion.

grynspan:
This PR modifies our code that detects types so that it correctly handles the module selector if present. For example, we currently misfire on a pattern like this:

```swift
try #require(throws: Swift::Never.self) {
  ...
}
```

We would expect to diagnose with a warning of the form:

> ⚠️ Passing 'Never.self' to 'require(\_:\_:)' is redundant; invoke non-throwing test code directly instead

But we don't because we don't recognize that the listed error type is equivalent to `Never`. This PR fixes that, as well as similar problems with `Tag`, `Tag.List`, `ParallelizationTrait`, and the use of `as` in an expectation expression.

Migrating macro expansion code of the form `Testing.T` to `Testing::T` is a future direction.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
+++ b/Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
@@ -26,9 +26,9 @@ extension AttributeInfo {
          let calledExpr = functionCallExpr.calledExpression.as(MemberAccessExprSyntax.self) {
         // Check for .tags() traits.
         switch calledExpr.tokens(viewMode: .fixedUp).map(\.textWithoutBackticks).joined() {
-        case ".tags", "Tag.List.tags", "Testing.Tag.List.tags":
+        case ".tags", "Tag.List.tags", "Testing.Tag.List.tags", "Testing::Tag.List.tags", "Testing::Testing.Tag.List.tags":
           _diagnoseIssuesWithTagsTrait(functionCallExpr, addedTo: self, in: context)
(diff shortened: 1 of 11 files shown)


Sources/TestingMacros/Support/DiagnosticMessage+Diagnosing.swift
```diff
@@ -338,7 +338,8 @@ func diagnoseExpansionInLibraryTarget(of macro: some FreestandingMacroExpansionS
   }
 
   var targetName = "<unknown>"
-  if let fileID = context.location(of: macro, at: .afterLeadingTrivia, filePathMode: .fileID)?.file.trimmedDescription,
+  if let location = context.location(of: macro, at: .afterLeadingTrivia, filePathMode: .fileID),
```
grynspan, line 341: Drive-by fix, unrelated.

harlanhaskins:
It seems like a compiler bug that `Swift::Swift.XYZ` is parsed; it doesn't make sense, but I tried it locally and it parses...

grynspan:
It's definitely weird, but since it's valid, we gotta accept it.

grynspan:
> Swift 6.3 also enables using the Swift module name to access concurrency and String processing library APIs:

```swift
let task = Swift::Task {
  // async work
}

```

Gotta account for that too.

harlanhaskins (approved):

stmontgomery (approved):

PR #1650, opened 2026-03-31
Diagnose the use of a generic clause on `@Test`, `@Suite`, and `@Tag`.

grynspan:
For example:

```swift
@Test<[Int]>(arguments: [1, 2, 3]) func f(i: Int) { ... }
```

We never intended for this to be valid syntax. It appears some folks _are_ using it, so deprecate for now and make it an error in the next language mode. The diagnostic looks like this in the Swift 6 language mode:

> ⚠️ Generic argument clause of attribute 'Test' is unsupported; this is an error in the Swift 7 language mode

And in the Swift 7 language mode, it's an error. (In the future, if we find that this syntax is useful and want to build out some sort of support for it, we can of course remove the diagnostic.)

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/DiagnosticMessage.swift
+++ b/Sources/TestingMacros/Support/DiagnosticMessage.swift
@@ -9,6 +9,7 @@
 //
 
 import SwiftDiagnostics
+import SwiftIfConfig
 import SwiftParser
 import SwiftSyntax
 import SwiftSyntaxBuilder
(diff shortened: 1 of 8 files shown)


Sources/TestingMacros/Support/DiagnosticMessage.swift
```diff
@@ -197,6 +209,39 @@ struct DiagnosticMessage: SwiftDiagnostics.DiagnosticMessage {
...
+  /// generic argument clause (e.g. `@Test<T>`).
+  ///
+  /// - Parameters:
+  ///   - attribute: The `@Test` or `@Suite` attribute.
+  ///   - decl: The generic declaration in question.
+  ///   - genericClause: The child node on `attribute` that makes it generic.
+  ///
+  /// - Returns: A diagnostic message.
+  static func genericAttributeNotSupported(_ attribute: AttributeSyntax, on decl: some SyntaxProtocol, becauseOf genericClause: some SyntaxProtocol, languageMode: VersionTuple?) -> Self {
+    let fixIts: [FixIt] = [
```
harlanhaskins, line 222: Thanks for adding a FixIt!

Tests/TestingTests/EventIterationTests.swift
```diff
@@ -51,9 +51,9 @@ struct EventIterationTests {
   private static func matchesTestLifetimeEventKind(_ expected: Event.Kind, _ actual: Event.Kind) -> Bool {
     switch (expected, actual) {
     case (.testStarted, .testStarted),
-         (.testEnded, .testEnded),
-         (.testCaseStarted, .testCaseStarted),
-         (.testCaseEnded, .testCaseEnded):
+      (.testEnded, .testEnded),
```
harlanhaskins: Unrelated? Or is this the stylistic preference of the project?

grynspan: Unrelated, looks like Xcode did it automatically when I wasn't looking.

harlanhaskins (approved):

stmontgomery (approved):

PR #1651, opened 2026-03-31
Respect `Configuration` when snapshotting a runner plan

harlanhaskins:
This introduces another initializer of `Runner.Plan.Snapshot` that overrides `Configuration.current`

### Motivation:

Currently, Runner.Plan.Snapshot doesn't respect settings in `Configuration`s like `valueReflectionOptions` because those read from the Task-local static properties.

### Modifications:

Adds one initializer and wraps the existing initializer body with `Configuration.withCurrent(...) { ... }`, and then defers to the more specific one.

Fixes rdar://173801709

### Checklist:


harlanhaskins:
Not needed

PR #1659, opened 2026-04-02
Export SwiftTesting_{BUILD_DIR,MACRO_PATH} in CMake

jerryjrchen:
### Motivation:

These are consumed by XCTest when building lit tests in
https://github.com/swiftlang/swift-corelibs-xctest/pull/538

### Modifications:

* Export SwiftTesting_BUILD_DIR in SwiftTestingConfig

* Export SwiftTesting_MACRO_PATH in SwiftTestingMacrosConfig

  SwiftTestingMacrosConfig is a new module exported in this change.

  Note we can't export this as part of SwiftTestingConfig, because we
  set SwiftTesting_MACRO=NO when building the Testing project for Linux,
  so there is no macro path available.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/CMakeLists.txt
+++ b/Sources/TestingMacros/CMakeLists.txt
@@ -127,3 +127,8 @@ if(SwiftTesting_BuildMacrosAsExecutables)
   target_link_libraries(TestingMacros PRIVATE
     SwiftSyntax::SwiftCompilerPlugin)
 endif()
+
+# Makes SwiftTesting_MACRO_DIR available after find_package(SwiftTestingMacros)
+file(GENERATE
+  OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/cmake/modules/SwiftTestingMacrosConfig.cmake"
+  CONTENT "get_filename_component(SwiftTesting_MACRO_DIR \"$<TARGET_FILE:TestingMacros>\" DIRECTORY)\n")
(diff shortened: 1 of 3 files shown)


jerryjrchen:
Ran a toolchain build and it doesn't work: [consoleText.txt](https://github.com/user-attachments/files/26451631/consoleText.txt)

We set `SwiftTesting_MACRO=NO` when building the toolchain and that empties out the value for `SwiftTesting_MACRO_PATH`

jerryjrchen:
Trying another toolchain build after exporting SwiftTestingMacros as a separate module

Sources/TestingMacros/CMakeLists.txt
```diff
@@ -127,3 +127,8 @@ if(SwiftTesting_BuildMacrosAsExecutables)
   target_link_libraries(TestingMacros PRIVATE
     SwiftSyntax::SwiftCompilerPlugin)
 endif()
+
+# Makes SwiftTesting_MACRO_PATH available after find_package(SwiftTestingMacros)
+file(GENERATE
+  OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/cmake/modules/SwiftTestingMacrosConfig.cmake"
+  CONTENT "set(SwiftTesting_MACRO_PATH \"$<TARGET_FILE:TestingMacros>\")\n")
```
briancroom: Is the `DIR` vs. `PATH` difference (`SwiftTesting_BUILD_DIR` vs. `SwiftTesting_MACRO_PATH`) meaningful/intentional?

briancroom (approved):

Sources/TestingMacros/CMakeLists.txt
jerryjrchen: - [x] Used PATH here only because it was easier to get as a generator expression, but makes sense to standardise on dir

jerryjrchen:
Test failures are unrelated.

PR #1671, opened 2026-04-13
Avoid unknown platform warning when setting Embedded-related defines which lack a non-Embedded fallback condition

stmontgomery:
This modifies the technique we use in `Package.swift` to set Embedded-related define (`-D`) flags so that we no longer need to reference a fake/non-existent platform `.custom("DoesNotExist")`. When building the package with Swift Build, this now produces the following warning:

```
warning: PIF: Ignoring settings assignments for unknown platform 'DoesNotExist'
```

To resolve this, this PR avoids adding the relevant defines when there is no fallback condition and we're not building for Embedded mode. As a bonus, this leverages a helper protocol to consolidate the Swift and C++ target styles and consolidates the (identical) set of defines into a single place for easier maintainability.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -319,30 +319,6 @@ package.targets.append(contentsOf: [
 #endif
 
 extension BuildSettingCondition {
-  /// Creates a build setting condition that evaluates to `true` for Embedded
-  /// Swift.
-  ///
-  /// - Parameters:
-  ///   - nonEmbeddedCondition: The value to return if the target is not
-  ///     Embedded Swift. If `nil`, the build condition evaluates to `false`.
-  ///
-  /// - Returns: A build setting condition that evaluates to `true` for Embedded
-  ///   Swift or is equal to `nonEmbeddedCondition` for non-Embedded Swift.
-  static func whenEmbedded(or nonEmbeddedCondition: @autoclosure () -> Self? = nil) -> Self? {
-    if !buildingForEmbedded {
-      if let nonEmbeddedCondition = nonEmbeddedCondition() {
-        nonEmbeddedCondition
-      } else {
-        // The caller did not supply a fallback. Specify a non-existent platform
-        // to ensure this condition never matches.
-        .when(platforms: [.custom("DoesNotExist")])
-      }
-    } else {
-      // Enable unconditionally because the target is Embedded Swift.
-      nil
-    }
-  }
-
   /// A build setting condition representing all Apple or non-Apple platforms.
   ///
   /// - Parameters:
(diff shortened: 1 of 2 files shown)


Package.swift
```diff
@@ -522,3 +476,60 @@ extension Array where Element == PackageDescription.CXXSetting {
...
+      "SWT_NO_EXIT_TESTS": .when(platforms: [.iOS, .watchOS, .tvOS, .visionOS, .wasi, .android]),
+      "SWT_NO_PROCESS_SPAWNING": .when(platforms: [.iOS, .watchOS, .tvOS, .visionOS, .wasi, .android]),
+      "SWT_NO_SNAPSHOT_TYPES": .whenApple(false),
+      "SWT_NO_DYNAMIC_LINKING": .when(platforms: [.wasi]),
+      "SWT_NO_PIPES": .when(platforms: [.wasi]),
+      "SWT_NO_FOUNDATION_FILE_COORDINATION": .whenApple(false),
+      "SWT_NO_IMAGE_ATTACHMENTS": .when(platforms: [.linux, .custom("freebsd"), .openbsd, .wasi, .android]),
+      "SWT_NO_FILE_CLONING": .when(platforms: [.openbsd, .wasi, .android]),
+
+      "SWT_NO_LEGACY_TEST_DISCOVERY": nil,
```
grynspan: Perhaps we just put these in a separate "unconditional" array for clarity's sake?

Package.swift
```diff
@@ -522,3 +476,60 @@ extension Array where Element == PackageDescription.CXXSetting {
...
+      if !buildingForEmbedded {
+        if let condition {
+          append(.define(name, condition))
+        } else {
+          // Since there was no condition and we're not building for Embedded,
+          // the intention was to never match so don't append this define.
+        }
+      } else {
+        // Since we're building for Embedded, append this define unconditionally.
+        append(.define(name, nil))
```
grynspan, line 519: This seems wrong? We don't necessarily want to define all of these conditions in Embedded Swift.

Package.swift
grynspan: If we do that, the loop below can be significantly simplified into something like:

```swift
if buildingForEmbedded {
  self += embeddedOnlyDefines
}
self += conditionalDefines.map { .define($0, $1) }
```

Or something like that.

Package.swift
```diff
@@ -522,3 +476,60 @@ extension Array where Element == PackageDescription.CXXSetting {
     return result
   }
 }
+
+extension Array where Element: _LanguageBuildSetting {
+  /// Append defines which are conditionalized for the Embedded language mode.
+  mutating func appendEmbeddedRelatedDefines() {
```
grynspan:
```suggestion
  mutating func appendFeatureFlags() {
```

Or something like that?

grynspan (changes requested):
I think you may have reversed the sense of some conditions here.

Package.swift
stmontgomery: That was the prior behavior
```swift
      // Enable unconditionally because the target is Embedded Swift.
      nil
```

grynspan: Because the previous function was "when Embedded or any of these platforms, set this condition". :)

stmontgomery: I reviewed this again and believe the behavior is semantically equivalent to what we had before. In Embedded mode, _all_ of these defines were already being set unconditionally — that's what the old `whenEmbedded()` helper did (it returned `nil`, meaning "unconditional"). The new code just makes that explicit in the `else` branch.

Here's a table tracing through two representative defines — one **with** a fallback condition and one **without** — under both the old and new logic:

| Define | Fallback condition | Mode | Old behavior (`whenEmbedded(or:)`) | New behavior (`appendEmbeddedRelatedDefines`) |
|---|---|---|---|---|
| `SWT_NO_EXIT_TESTS` | `.when(platforms: [.iOS, …])` | Non-Embedded | `whenEmbedded(or: .when(…))` → returns `.when(…)` → appends `.define(…, .when(…))` | `condition` is non-nil → appends `.define(…, .when(…))` |
| `SWT_NO_EXIT_TESTS` | `.when(platforms: [.iOS, …])` | Embedded | `whenEmbedded(or: .when(…))` → returns `nil` → appends `.define(…, nil)` (unconditional) | Embedded path → appends `.define(…, nil)` (unconditional) |
| `SWT_NO_LEGACY_TEST_DISCOVERY` | _(none)_ | Non-Embedded | `whenEmbedded()` → no fallback → returns `.when(platforms: [.custom("DoesNotExist")])` → appends a **never-matching** define | `condition` is `nil` → **skips entirely** (doesn't append) |
| `SWT_NO_LEGACY_TEST_DISCOVERY` | _(none)_ | Embedded | `whenEmbedded()` → returns `nil` → appends `.define(…, nil)` (unconditional) | Embedded path → appends `.define(…, nil)` (unconditional) |

The only behavioral difference is row 3: instead of emitting a define with `.custom("DoesNotExist")` that never matches any real platform (and triggers the Swift Build warning this PR fixes), we simply omit the define entirely. A define conditioned on a platform that doesn't exist vs. a define that isn't present at all are semantically identical — neither results in the flag being passed to the compiler.

Package.swift
stmontgomery: Changed style slightly and took this suggestion

Package.swift
```diff
@@ -319,42 +319,30 @@ package.targets.append(contentsOf: [
...
+  /// An array representing all Apple or non-Apple platforms.
   ///
   /// - Parameters:
   ///   - isApple: Whether or not the result represents Apple platforms.
   ///
-  /// - Returns: A build setting condition that evaluates to `isApple` for Apple
-  ///   platforms.
-  static func whenApple(_ isApple: Bool = true) -> Self {
+  /// - Returns: An array containing the requested platforms.
+  static func applePlatforms(_ isApple: Bool = true) -> Self {
```
grynspan: I think I'd just make this `static let applePlatforms: Self` and `static let nonApplePlatforms: Self`.

grynspan (approved):

grynspan:
Please also update Porting.md!

stmontgomery:
CI failures are unrelated

PR #1672, opened 2026-04-13
Share `TestContentKind` definition/implementation between targets.

grynspan:
This PR ensures `TestContentKind` is defined and available in the macro target so that we don't have to redeclare it there with slightly different semantics.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/TestContentGeneration.swift
+++ b/Sources/TestingMacros/Support/TestContentGeneration.swift
@@ -13,28 +13,14 @@ import SwiftSyntax
 import SwiftSyntaxBuilder
 import SwiftSyntaxMacros
 
-/// An enumeration representing the different kinds of test content known to the
-/// testing library.
-///
-/// When adding cases to this enumeration, be sure to also update the
-/// corresponding enumeration in TestContent.md.
-///
-/// - Bug: This type should be imported directly from `_TestDiscovery` instead
-///   of being redefined (differently) here.
-enum TestContentKind: UInt32 {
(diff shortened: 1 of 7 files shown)


Sources/TestingMacros/Support/TestContentGeneration.swift
```diff
@@ -13,28 +13,14 @@ import SwiftSyntax
...
-enum TestContentKind: UInt32 {
-  /// A test or suite declaration.
-  case testDeclaration = 0x74657374
-
-  /// An exit test.
-  case exitTest = 0x65786974
-
+extension TestContentKind {
   /// This kind value as a comment (`/* 'abcd' */`) if it looks like it might be
   /// a [FourCC](https://en.wikipedia.org/wiki/FourCC) value, or `nil` if not.
```
jerryjrchen: > or `nil` if not.

Is `spaces(0)` the same as `nil`?

jerryjrchen (approved):

Sources/TestingMacros/Support/TestContentGeneration.swift
grynspan: Now I'm not sure why I changed it.

stmontgomery (approved):

grynspan:
CI failures are unrelated.

Sources/TestingMacros/Support/TestContentGeneration.swift
grynspan: Oh, the comment is just wrong. :)

PR #1675, opened 2026-04-14
Add an environment variable to allow suspending child processes.

### LINKED ISSUE #1670: Environment variable for starting exit test processes in debuggable state
### Motivation

Right now, swift-testing will begin an exit test sub process in a stopped state when built in `DEBUG` mode to allow debuggers to attach before the process begins. This is helpful when debugging the behavior of exit test code. However, it currently requires building swift-testing from source and stepping through its code in the debugger to get the child task PID.

### Proposed solution

It would be great if swift-testing could provide an environment variable that allows for more ease of debugging child exit test processes. For example, the environment variable could start child processes in the stopped state (even when not building swift-testing from source) and could print the PID/name of the child process to stdout for easy attaching via the debugger.

### Alternatives considered

_No response_

### Additional information

_No response_

grynspan:
This PR adds an environment variable check in our `posix_spawn()` wrapper that causes child processes to start in a suspended state on Darwin. The equivalent logic is implemented on Windows too, but other platforms don't support this functionality.

This change gives developers a chance to manually attach a debugger to the spawned child process.

This environment variable, as with all others we use, is not stable API and is subject to change at any time.

Resolves #1670.

### Checklist:

### DIFF
--- a/Documentation/EnvironmentVariables.md
+++ b/Documentation/EnvironmentVariables.md
@@ -53,7 +53,7 @@ names prefixed with `SWT_`.
 | `SWT_CAPTURED_VALUES` | `CInt`/`HANDLE` | A file descriptor (handle on Windows) containing captured values passed to the exit test. |
 | `SWT_CLOSEFROM` | `CInt` | Used on OpenBSD to emulate `posix_spawn_file_actions_addclosefrom_np()`. |
 | `SWT_EXIT_TEST_ID` | `String` (JSON) | Specifies which exit test to run. |
-| `SWT_START_CHILDREN_SUSPENDED` | `Bool` | Start child processes (such as those for exit tests) in a suspended state where supported (Apple platforms and Windows). |
+| `SWT_START_CHILD_PROCESSES_SUSPENDED` | `Bool` | Start child processes (such as those for exit tests) in a suspended state where supported (Apple platforms and Windows). |
 | `XCTestBundlePath`\* | `String` | Used on Apple platforms to determine if Xcode is hosting the test run. |
 
 ## Miscellaneous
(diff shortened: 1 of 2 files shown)


Documentation/EnvironmentVariables.md
```diff
@@ -53,6 +53,7 @@ names prefixed with `SWT_`.
 | `SWT_CAPTURED_VALUES` | `CInt`/`HANDLE` | A file descriptor (handle on Windows) containing captured values passed to the exit test. |
 | `SWT_CLOSEFROM` | `CInt` | Used on OpenBSD to emulate `posix_spawn_file_actions_addclosefrom_np()`. |
 | `SWT_EXIT_TEST_ID` | `String` (JSON) | Specifies which exit test to run. |
+| `SWT_START_CHILDREN_SUSPENDED` | `Bool` | Start child processes (such as those for exit tests) in a suspended state where supported (Apple platforms and Windows). |
```
stmontgomery: Instead of "children" could the env var reference exit tests by name?

```suggestion
| `SWT_START_EXIT_TESTS_SUSPENDED` | `Bool` | Start child processes (such as those for exit tests) in a suspended state where supported (Apple platforms and Windows). |
```

grynspan: This change is not specific to exit tests. In particular, it also affects directory attachments.

stmontgomery: Ok, given that I'd suggest "child processes" to avoid confusion with other notions of children, such as child tests contained in a suite
```suggestion
| `SWT_START_CHILD_PROCESSES_SUSPENDED` | `Bool` | Start child processes (such as those for exit tests) in a suspended state where supported (Apple platforms and Windows). |
```

stmontgomery (approved):

PR #1677, opened 2026-04-15
Preserve issue severity = warning across interop boundary

jerryjrchen:
### Motivation:

If XCTest records a warning severity issue, it should stay as a warning severity issue. This is because the intention on the XCTest side was not to create a "test failing event".

Prior to this change, we were always recording such issues with *error* severity in the complete interop mode.

### Modifications:

* For interop issues, only "promote" severity to error if it was reported as an error in the first place

* Add regression tests for this behaviour

### Checklist:

Resolves rdar://174841005


stmontgomery (approved):

jerryjrchen:
CI failures are unrelated

PR #1680, opened 2026-04-17
Decode EncodedEvent -> SkipInfo

jerryjrchen:
### Motivation:

Required for ultimately decoding EncodedEvent -> Event

Resolves rdar://174974921

### Modifications:

Decode EncodedEvent -> SkipInfo

### Checklist:

### DIFF
--- a/Sources/Testing/Running/SkipInfo.swift
+++ b/Sources/Testing/Running/SkipInfo.swift
@@ -78,6 +78,35 @@ extension SkipInfo {
   }
 }
 
+// MARK: - Conversion to/from ABI types
+
+extension SkipInfo {
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// SkipInfo is only non-nil for the skip/cancel event kinds.
+  ///
+  /// - Parameters:
+  ///   - event: The encoded event to initialize this instance from.
+  ///
+  /// Reconstructs ``SkipInfo`` from the comments and
+  /// source location stored in the encoded event.
+  init?<V>(decoding event: ABI.EncodedEvent<V>) {
+    // Only skip/cancel event kinds can decode SkipInfo.
+    switch event.kind {
+    case .testCancelled, .testCaseCancelled, .testSkipped:
+      break
+    default:
+      return nil
+    }
+
+    // Typically only a single comment is expected for SkipInfo.
+    let comment = event._comments?.first.map(Comment.init(rawValue:))
+    let sourceLocation = event._sourceLocation.flatMap(SourceLocation.init(decoding:))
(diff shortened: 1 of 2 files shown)


Sources/Testing/Running/SkipInfo.swift
```diff
@@ -78,6 +78,25 @@ extension SkipInfo {
...
+
+extension SkipInfo {
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// - Parameters:
+  ///   - event: The encoded event to initialize this instance from.
+  ///
+  /// Reconstructs ``SkipInfo`` from the comments and
+  /// source location stored in the encoded event.
+  init<V>(decoding event: ABI.EncodedEvent<V>) {
```
grynspan: As with the other `init(decoding:)` initializers, I recommend making this optional to give us an escape hatch for future incompatible JSON changes.

grynspan:
An encoded event doesn't really naturally map to a `SkipInfo` instance as the latter is just a bookkeeping error type. Can you give some more background as to how this will be useful (since nothing is adopting it in this PR)?

jerryjrchen:
> An encoded event doesn't really naturally map to a `SkipInfo` instance as the latter is just a bookkeeping error type. Can you give some more background as to how this will be useful (since nothing is adopting it in this PR)?

For the EncodedEvent -> testSkipped/testCancelled cases, I was hoping to do something like this:
```swift
public init?<V>(decoding event: ABI.EncodedEvent<V>) {
  let kind: Kind
  switch event.kind {
  // ...
  case .testSkipped:
    kind = .testSkipped(SkipInfo(decoding: event))
  case .testCancelled:
    kind = .testCancelled(SkipInfo(decoding: event))
```

Sources/Testing/Running/SkipInfo.swift
```diff
@@ -78,6 +78,25 @@ extension SkipInfo {
...
+
+extension SkipInfo {
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// - Parameters:
+  ///   - event: The encoded event to initialize this instance from.
+  ///
+  /// Reconstructs ``SkipInfo`` from the comments and
+  /// source location stored in the encoded event.
+  init?<V>(decoding event: ABI.EncodedEvent<V>) {
```
stmontgomery, line 93: If the passed-in `event` has a `kind` value which is something other than `.testCancelled`, should this return `nil`? In other words should it only decode to a valid instance for certain kinds of events? (And if you make this change, I recommend adding a unit test.)

jerryjrchen: Done and tested

stmontgomery (approved):

PR #1681, opened 2026-04-18
Decode EncodedInstant -> Test.Clock.Instant

jerryjrchen:
### Motivation:

Required for ultimately decoding EncodedEvent -> Event

Resolves rdar://175082804

### Modifications:

- Decode `EncodedInstant` to `Test.Clock.Instant`

- Test round trip encode/decode

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Clock.swift
+++ b/Sources/Testing/Events/Clock.swift
@@ -24,7 +24,7 @@ extension Test {
       fileprivate(set) var suspending = TimeValue(rawValue: SuspendingClock().systemEpoch.duration(to: .now))
 
 #if !SWT_NO_UTC_CLOCK
-      /// The wall-clock time corresponding to this instant.
+      /// The wall-clock time since 1970 corresponding to this instant.
       fileprivate(set) var wall: TimeValue = {
 #if !SWT_NO_TIMESPEC
         var wall = timespec()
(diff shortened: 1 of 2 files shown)


Sources/Testing/Events/Clock.swift
```diff
@@ -67,6 +67,25 @@ extension SuspendingClock.Instant {
...
+extension Test.Clock.Instant {
+  /// Initialize this instant to be exactly equal to an instant from the testing
+  /// library's event stream.
+  ///
+  /// Note that when the original instant is encoded to the event stream,
+  /// it loses some precision.
+  ///
+  /// - Parameters:
+  ///   - instant: The encoded instant to initialize this instant from.
+  public init<V>(decoding instant: ABI.EncodedInstant<V>) {
```
grynspan: Should be optional like the other decoding inits.

Sources/Testing/Events/Clock.swift
```diff
@@ -67,6 +67,25 @@ extension SuspendingClock.Instant {
...
+    self.init()
```
grynspan: This line isn't necessary.

Sources/Testing/Events/Clock.swift
```diff
@@ -24,7 +24,7 @@ extension Test {
       fileprivate(set) var suspending = TimeValue(rawValue: SuspendingClock().systemEpoch.duration(to: .now))
 
 #if !SWT_NO_UTC_CLOCK
-      /// The wall-clock time corresponding to this instant.
+      /// The wall-clock time since the Unix epoch corresponding to this instant.
```
grynspan: Nitpick: I think we just say "since 1970" elsewhere. The UNIX epoch does not technically apply to all systems.

grynspan (dismissed):

Sources/Testing/Events/Clock.swift
jerryjrchen: Just curious, is this a stylistic choice since some of the decoding inits do actually need to be option? Or are there other reasons we have that pattern?

Sources/Testing/Events/Clock.swift
```diff
@@ -67,6 +67,24 @@ extension SuspendingClock.Instant {
   }
 }
 
+@_spi(ForToolsIntegrationOnly)
+extension Test.Clock.Instant {
+  /// Initialize this instant to be exactly equal to an instant from the testing
+  /// library's event stream.
+  ///
+  /// Note that when the original instant is encoded to the event stream,
```
grynspan:
```suggestion
  /// - Note: that when the original instant is encoded to the event stream,
```

grynspan:
```suggestion
  /// - Note: that when the original instant is encoded to the event stream,
```

grynspan (approved):

PR #1684, opened 2026-04-21
Decode EncodedEvent -> Event

jerryjrchen:
### Motivation:

Enables re-use of EncodedEvent decoding logic for interop and exit tests

The decoding is lossy, i.e. you can't recover all the original Event
details. For example, testID is left as nil, which we may want to
address in the future.

The current decoding supports these use cases:

- Corelibs XCTest manually re-defined the record type to decode events
  from the fallback event handler.

  Now, it can use this decoding init directly.

- Exit tests only decoded a subset of event kinds.

  Now, they still only process a limited set of event kinds. But, the
  implementation is simplified and we can easily extend it to support
  future event kinds if so desired.

Resolves rdar://175287183

### Modifications:

* Implement decoding EncodedEvent -> Event

* Adopt for exit tests

* Test common decoding scenarios, included decoding issueRecorded and valueAttached event kinds

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
@@ -180,3 +180,65 @@ extension ABI {
 
 extension ABI.EncodedEvent: Codable {}
 extension ABI.EncodedEvent.Kind: Codable {}
+
+// MARK: - Conversion to/from library types
+
+@_spi(ForToolsIntegrationOnly)
+extension Event {
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// - Parameters:
+  ///   - event: The encoded event to initialize this instance from.
+  ///
+  /// ``testID`` and ``testCaseID`` are always `nil` because we need information
+  /// from the associated `ABI.EncodedTest` to properly decode those values.
+  public init?<V>(decoding event: ABI.EncodedEvent<V>) {
+    // SkipInfo will only be decoded for skip/cancel event kinds
+    lazy var skipInfo = SkipInfo(decoding: event)
+
+    let kind: Kind
+    switch event.kind {
+    case .runStarted:
+      kind = .runStarted
+    case .testStarted:
+      kind = .testStarted
+    case .testCaseStarted:
+      kind = .testCaseStarted
+    case .issueRecorded:
+      guard let issue = Issue(decoding: event) else {
+        return nil
+      }
+      kind = .issueRecorded(issue)
+    case .valueAttached:
+      guard let attachment = Attachment<AnyAttachable>(decoding: event) else {
+        return nil
+      }
+      kind = .valueAttached(attachment)
+    case .testCaseEnded:
+      kind = .testCaseEnded
+    case .testCaseCancelled:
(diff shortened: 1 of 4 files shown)


Tests/TestingTests/ABI.EncodedEventTests.swift
```diff
@@ -0,0 +1,122 @@
...
+      {
+        "kind": "issueRecorded",
+        "instant": {"absolute": 0, "since1970": 0},
+        "messages": [],
+        "issue": {"isKnown": true}
+      }
+      """)
+    let decoded = try #require(Event(decoding: event))
+
+    guard case .issueRecorded(let issue) = decoded.kind else {
```
jerryjrchen, line 97: I wasn't sure if there was a nicer way of checking the enum case for `kind` since this enum has associated values

grynspan: This is how we do it for the moment.

Tests/TestingTests/SkipInfoTests.swift
```diff
@@ -9,9 +9,6 @@
 //
 
 @testable @_spi(ForToolsIntegrationOnly) import Testing
-#if canImport(Foundation)
```
grynspan, line 12: Drive-by change?

jerryjrchen: Yeah, realised I didn't need the import in my earlier change after writing the tests in this one. I can leave it out though

grynspan: It's fine, just helps to explain this sort of thing in the PR. :)

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -180,3 +180,65 @@ extension ABI {
...
+
+@_spi(ForToolsIntegrationOnly)
+extension Event {
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// - Parameters:
+  ///   - event: The encoded event to initialize this instance from.
+  ///
+  /// ``testID`` and ``testCaseID`` are always `nil` because we need information
+  /// from the associated `ABI.EncodedTest` to properly decode those values.
```
stmontgomery, line 194: So it'll just be follow-on work to fill in these missing property values, I take it?

stmontgomery (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
grynspan: `Test.Case.ID` isn't stable yet, so we can't really decode it right now. `Test.ID` may need to be modified to allow it to take a flat string instead of a `SourceLocation`, but I think it's feasible.

grynspan (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
jerryjrchen: It's ok to have as nil for the time being because interop does not look at this field.

Issue #1366, opened 2025-10-15
Add support for Android


grynspan:
Add support for Android:

- [x] Audit our platform-specific code and add branches for Android (mostly, but not entirely, the same as Linux)
- [ ] Set up team workstations or VMs
- [ ] Set up CI jobs

finagolfin:
Note that while cross-compiling some proportion of Testing tests has worked for a while now, long-standing SwiftPM bugs like swiftlang/swift-package-manager#8094 have broken cross-compiling this repo's tests to _any_ platform, so the remaining steps here are blocked on those underlying bugs, which I'm told the new swift-build backend may soon fix. Instead, I have been occasionally natively building the trunk toolchain on Android in the Termux app then testing this repo, and with #1373 or an earlier patch, all tests passed.

grynspan:
@finagolfin @compnerd Any thoughts on how we can run our test suite automatically on Android as part of our GHA workflow suite? If we can do that and we can sort out instructions for at-desk testing, we can mark Android as fully supported I believe.

compnerd:
Yes, that is possible; https://github.com/compnerd/ds2 was running the tests on Android, so it is definitely possible to do in GHA.

finagolfin:
There are potentially two separate issues with cross-compiling Testing tests for Android:

1. We need to check if the recent `swift-build` switch in trunk SwiftPM fixed the problems with using macros, or if more work is needed in SwiftPM.
2. Once a test runner is successfully produced, it _should_ just work with the current official workflows, which @marcprux already has [working with `Testing`](https://github.com/swiftlang/github-workflows/blob/1ffcd38c2569b09b1641b1f0d98b2a2773eb5f9a/.github/workflows/scripts/android/android-emulator-tests.sh#L153), provided packages don't define their own macros and avoid that long-standing SwiftPM bug.

We won't know if 2. works for sure till we fix 1., but the emulator testing probably works fine already.

marcprux:
We still can't cross-compile the `Testing` tests due to https://github.com/swiftlang/swift-package-manager/issues/8094. I just tried again with 6.3.1 and got the expected failure:

```
zap swift-android-sdk/swift-testing [android-checks] % ANDROID_SERIAL=emulator-5554 skip android test

[0/1] Planning build
Building for debugging...
[0/21] Write sources
[37/65] Write swift-version-27A8E6E6C94AF7AD.txt
[38/65] Write sources
[...]

PR #1685, opened 2026-04-22
Update the Swift versions built on CI

finagolfin:
Update the CI built using the release compilers to build using the latest release also, not just the release snapshots.

Don't update linux because that container is currently missing the `zip` package and fails a test that requires it.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -25,14 +25,14 @@ jobs:
     with:
       linux_swift_versions: '["nightly-main", "nightly-6.3"]'
       linux_os_versions: '["amazonlinux2", "jammy"]'
-      linux_static_sdk_versions: '["nightly-main"]'
-      windows_swift_versions: '["nightly-main", "nightly-6.3"]'
+      linux_static_sdk_versions: '["6.3", "nightly-main", "nightly-6.3"]'
+      windows_swift_versions: '["6.3", "nightly-main", "nightly-6.3"]'
       enable_macos_checks: true
       macos_xcode_versions: '["swift_6.3"]'
       enable_ios_checks: true
       ios_host_xcode_versions: '["swift_6.3"]'
       enable_wasm_sdk_build: true
-      wasm_sdk_versions: '["nightly-main", "nightly-6.3"]'
+      wasm_sdk_versions: '["6.3", "nightly-main", "nightly-6.3"]'
       enable_android_sdk_build: true
       android_ndk_versions: '["r28c", "r29"]'
       android_sdk_triples: '["aarch64-unknown-linux-android33", "x86_64-unknown-linux-android28"]'
--- a/.github/workflows/main_using_release.yml
+++ b/.github/workflows/main_using_release.yml
@@ -19,12 +19,12 @@ jobs:
     with:
       linux_swift_versions: '["nightly-6.3"]'
       linux_os_versions: '["amazonlinux2", "jammy"]'
-      linux_static_sdk_versions: '["nightly-6.3"]'
-      windows_swift_versions: '["nightly-6.3"]'
+      linux_static_sdk_versions: '["6.3", "nightly-6.3"]'
+      windows_swift_versions: '["6.3", "nightly-6.3"]'
       enable_macos_checks: true
       macos_xcode_versions: '["swift_6.3"]'
       enable_wasm_sdk_build: true
-      wasm_sdk_versions: '["nightly-6.3"]'
+      wasm_sdk_versions: '["6.3", "nightly-6.3"]'
       enable_android_sdk_build: true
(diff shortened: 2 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -23,18 +23,16 @@ jobs:
...
       linux_os_versions: '["amazonlinux2", "jammy"]'
-      linux_static_sdk_versions: '["nightly-main"]'
-      windows_swift_versions: '["nightly-main", "nightly-6.3"]'
+      windows_swift_versions: '["6.3", "nightly-main", "nightly-6.3"]'
       enable_macos_checks: true
       macos_xcode_versions: '["swift_6.3"]'
       enable_ios_checks: true
       ios_host_xcode_versions: '["swift_6.3"]'
       enable_wasm_sdk_build: true
-      wasm_sdk_versions: '["nightly-main", "nightly-6.3"]'
```
finagolfin, line 35: Removing these versions simply resets to the default of building with all the three actively developed compilers, ie simply adding the 6.3 releases here, same for the static SDK above.

This can't be done for linux and Windows, as they build using the Swift 5 compilers also by default.

finagolfin:
The 6.3 release linux compiler has one attachment test failing when building and testing main, `Test attachContentsOfDirectoryURL() recorded an issue at AttachmentTests.swift:370:18: Caught error: Error Domain=NSCocoaErrorDomain Code=512 "(null)"UserInfo={NSLocalizedDescription=The directory at '/tmp/A7D4DEEA-C657-40BC-8D06-DF200ABE6829' could not be compressed (.exitCode(127)).}`

grynspan:
> The 6.3 release linux compiler has one attachment test failing when building and testing main, `Test attachContentsOfDirectoryURL() recorded an issue at AttachmentTests.swift:370:18: Caught error: Error Domain=NSCocoaErrorDomain Code=512 "(null)"UserInfo={NSLocalizedDescription=The directory at '/tmp/A7D4DEEA-C657-40BC-8D06-DF200ABE6829' could not be compressed (.exitCode(127)).}`

I believe that's expected as we don't know where the `zip` or `tar` tool might be in the Android filesystem. We could use your help finding it, if it's present.

.github/workflows/pull_request.yml
grynspan: We'd like to keep them regardless as it acts as a breadcrumb when we move to the next Swift version.

finagolfin:
Android builds fine: this is [a linux-only issue, only with the 6.3 release compiler](https://github.com/swiftlang/swift-testing/actions/runs/24766869021/job/72463147659?pr=1685). I will rebase and see if it hits the new 6.3 snapshot also.

.github/workflows/pull_request.yml
finagolfin: OK, will put them back and add to them instead.

grynspan:
Oh, sorry, I misunderstood. Looks like we got exit code 127 which means "command not found" i.e. the Linux the test is running on does not have `/usr/bin/zip`. `zip` is one of the Swift toolchain's dependencies on Linux, so it ought to be present.

grynspan:
I pushed a commit to your branch to use the same logic as on ~~FreeBSD~~ OpenBSD for checking if `zip` is present before running it, so we'll get a clearer diagnostic when it's missing.

grynspan:
>✘ Test attachContentsOfDirectoryURL() recorded an issue at AttachmentTests.swift:370:18: Caught error: Error Domain=NSCocoaErrorDomain Code=4 "The file doesn’t exist."UserInfo={NSFilePath=/usr/bin/zip, NSLocalizedDescription=The 'zip' package is not installed.}

Let's just leave "6.3" out of the YAML files for now. I'll reach out to @shahmishal about the docker (?) configuration for 6.3 and make sure it contains the zip package, then we can revisit adding it.

grynspan:
Also, if this PR is enabling _testing_ on Android, then does it resolve #1366?

finagolfin:
Sounds good, strange that the 6.3 snapshot container for linux has it but the 6.3 release container does not. 🤷‍♂️

> Also, if this PR is enabling testing on Android, then does it resolve https://github.com/swiftlang/swift-testing/issues/1366?

This pull doesn't touch Android anymore, and that issue is still blocked on the linked SwiftPM inability to cross-compile when new macros are present. 🫤

.github/workflows/pull_request.yml
finagolfin: Done

finagolfin:
Alright, passed all CI now 👍

grynspan:
Please squash-merge when ready!

grynspan (approved):

stmontgomery (approved):

PR #1686, opened 2026-04-22
Promote `CustomTestReflectable` to API.

grynspan:
Per [ST-0022](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0022-customtestreflectable.md), promote this interface to API.

### Checklist:

### DIFF
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -91,7 +91,6 @@ the test when the code doesn't satisfy a requirement, use
 
 - ``Expectation``
 - ``ExpectationFailedError``
-- ``CustomTestStringConvertible``
 
 ### Representing source locations
 
--- a/Sources/Testing/Testing.docc/describing-values.md
+++ b/Sources/Testing/Testing.docc/describing-values.md
@@ -1,0 +1,170 @@
+# Describing and reflecting values
+
+<!--
+This source file is part of the Swift.org open source project
+
+Copyright (c) 2024–2026 Apple Inc. and the Swift project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+-->
+
+Add custom descriptions and mirrors to values you use in your tests.
+
+## Overview
+
+The testing library provides two protocols, ``CustomTestStringConvertible`` and
+``CustomTestReflectable``, that you can use to customize the appearance of
+values in Swift. The testing library uses these protocols to describe
+parameterized test arguments and, if a call to ``expect(_:_:sourceLocation:)``
+or ``require(_:_:sourceLocation:)-5l63q`` fails, to describe any values you pass
+to them.
+
+## Customize the description of a value
+
+You use the ``CustomTestStringConvertible`` protocol when you want to customize
+the description of a value _during testing only_. Values whose types conform to
+this protocol use it to describe themselves when the testing library presents
+them as part of the output of a test. For example, this protocol affects the
+display of values you pass as arguments to test functions or that are elements
+of an expectation failure.
+
+By default, the testing library converts values to strings using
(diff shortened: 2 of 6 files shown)


harlanhaskins (approved):

Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -91,7 +91,6 @@ the test when the code doesn't satisfy a requirement, use
 
 - ``Expectation``
 - ``ExpectationFailedError``
-- ``CustomTestStringConvertible``
```
stmontgomery, line 94: Are we unable to maintain this symbol reference here too? Ideally I think I'd like to be able to link to both this existing protocol and the new `CustomTestReflectable` from multiple places, but does DocC disallow that?

Sources/Testing/Testing.docc/describing-values.md
```diff
@@ -0,0 +1,170 @@
...
+or ``require(_:_:sourceLocation:)-5l63q`` fails, to describe any values you pass
+to them.
+
+## Customize the description of a value
+
+You use the ``CustomTestStringConvertible`` protocol when you want to customize
+the description of a value _during testing only_. Values whose types conform to
+this protocol use it to describe themselves when the testing library presents
+them as part of the output of a test. For example, this protocol affects the
+display of values you pass as arguments to test functions or that are elements
```
stmontgomery, line 30: "elements" feels like a confusing word in this position. "…or that are **sub-expressions** of…"  perhaps?

Sources/Testing/Testing.docc/describing-values.md
```diff
@@ -0,0 +1,170 @@
...
+extension Food {
+  package var testDescription: String { ... }
+  package var customTestMirror: Mirror { ... }
+}
+  
+// In your test target:
+
+import FoodTruck
+
+extension Food: CustomTestStringConvertible, CustomTestReflectable {}
```
stmontgomery, line 166: Would these typically need `@retroactive` too?

grynspan: No because they'll be in the same package/project.

Sources/Testing/Testing.docc/describing-values.md
grynspan: Neither's a great choice but I'm not sure what exactly to put.

Sources/Testing/Testing.docc/Expectations.md
grynspan: We can but I think we're only supposed to put it under one heading?

Sources/Testing/Testing.docc/describing-values.md
stmontgomery: Fair. Although it's not always the case that a test target is within a project which _specifies_ a package name. For example, in Xcode projects you can specify a package identifier, but not all projects have adopted that feature. Since you have a `- Note:` callout at the end, consider mentioning that possibility there

stmontgomery (approved):

Sources/Testing/Testing.docc/describing-values.md
stmontgomery: "…or that are **included in an expectation**." sounds good to me

grynspan: That would be the passive voice.

Sources/Testing/Testing.docc/describing-values.md
grynspan: I'm not sure how to phrase it exactly. "If that doesn't work, try `internal`, `@testable`, and `@retroactive`" sounds a bit hammer-headed.

Sources/Testing/Testing.docc/describing-values.md
stmontgomery: "…or that **you include in an expectation.**" ?

grynspan:
Let's plan to revisit the documentation tweaks with our DevPubs friends.

PR #1689, opened 2026-04-25
@_spi(ForToolsIntegrationOnly) viz for ABI.EncodedEvent.messages

jerryjrchen:
### Motivation:

Required for interop test libraries to read issue messages when handling EncodedEvent. This message content is not available in the Issue comments after decoding.

### Modifications:

Affected types are already under the @_spi umbrella of the ABI type.

- Add public viz to ABI.EncodedEvent.messages

- Add public viz to ABI.EncodedMessage.text

Resolves rdar://175600485

### Checklist:


stmontgomery (approved):

grynspan:
Based on private discussion, I don't think this is the right way to get the metadata you're looking for.

PR #1692, opened 2026-04-28
Explicitly import '_Concurrency' module when building for Embedded

stmontgomery:
This explicitly imports the `_Concurrency` module when building the package for [Embedded Swift](https://docs.swift.org/embedded/documentation/embedded/), since in that context it's not implicitly imported.

### Motivation:

This resolves various errors when attempting to build for Embedded, such as:

```
Test.swift:345:39: error: '_Concurrency' module not imported, required for async
```

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -364,6 +364,11 @@ extension Array where Element == PackageDescription.SwiftSetting {
 
     if buildingForEmbedded {
       result.append(.enableExperimentalFeature("Embedded"))
+
+      // Swift's concurrency module is not implicitly imported when building for
+      // Embedded, so we must explicitly import it since this project uses
+      // async and other concurrency-related language features.
+      result.append(.unsafeFlags(["-Xfrontend", "-import-module", "-Xfrontend", "_Concurrency"]))
     }
 
     // Define a compiler condition so we can discover at macro expansion time if


harlanhaskins (approved):
Approving in theory, but do we hit any issues with package dependents by our usage of unsafeFlags? IIRC unsafeFlags means clients can't use our package as a dependency

stmontgomery:
> Approving in theory, but do we hit any issues with package dependents by our usage of unsafeFlags? IIRC unsafeFlags means clients can't use our package as a dependency

That was true for a long time, but the prohibition against using unsafeFlags in dependencies was lifted in Swift 6.2 in https://github.com/swiftlang/swift-package-manager/pull/8914. We've been using unsafeFlags in our package since 6.3.

harlanhaskins:
Oh, awesome! My GameBoy Advance experiments just became a lot more reusable!

grynspan:
Seems to me this ought to happen automatically once there's a minimal implementation for Embedded?

stmontgomery:
> Seems to me this ought to happen automatically once there's a minimal implementation for Embedded?

My understanding is that there is already a minimal implementation for Embedded, but when building in that mode you must explicitly opt-in to using it. And this is one way to do so.

grynspan:
I'd like to get clarification on the "right" way to do so before merging.

PR #1697, opened 2026-04-29
A cornucopia of Embedded Swift fixes.

grynspan:
Among them:

- Stub out all code related to the "ABI JSON schema".
- Mark some unsupported features as unsupported in the package manifest.
- Remove dependencies on the suspending clock (TODO: provide an alternative when an Embedded Swift target does support the functionality therein).
- Fall back to the wall clock for timestamp tracking if it is available but the suspending clock is not.
- Stub out `CommandLine.arguments` (TODO: provide an alternate mechanism for configuring a test run).
- Conditionalize remaining link-time dependencies on Foundation.
- Conditionalize underscored `Error` protocol requirement usages.
- Provide a stub implementation of `String.init(describing:)` for types we know conform to `CustomStringConvertible` (see https://github.com/swiftlang/swift/pull/88738)
- Avoid using `withVaList()` which is unavailable in Embedded Swift.

### Checklist:


Package.swift
```diff
@@ -137,8 +137,9 @@ let package = Package(
       dependencies: [
         "_TestDiscovery",
         "_TestingInternals",
-        "TestingMacros",
-      ],
+      ] + {
+        buildingForEmbedded ? [] : ["TestingMacros"]
```
stmontgomery, line 142: Could we leave a comment indicating this is meant as a workaround for a Swift package manager and/or build system problem? It'll be another TODO we need to follow up about and resolve

Sources/Testing/Running/Runner.swift
```diff
@@ -277,8 +277,10 @@ extension Runner {
   ///
   /// - ``Runner/run()``
   private static func _runStep(atRootOf stepGraph: Graph<String, Plan.Step?>, context: _Context) async throws {
+#if !hasFeature(Embedded)
```
stmontgomery, line 280: Would these be worth a SWT_NO_TASK_CANCELLATION guard?

Sources/Testing/Expectations/ExpectationChecking+Macro.swift
```diff
@@ -1217,10 +1217,10 @@ public func __checkClosureCall<each T>(
 ///
 /// - Returns: A string equivalent to `String(describingForTest: error)` with
 ///   information about its type added if not already present.
-private func _description(of error: some Error) -> String {
+private func _description(of error: any Error) -> String {
```
stmontgomery, line 1220: Can you say why the switch from some to any Error here and a few other places? Apologies if I missed it but didn't see discussion of that in the PR description. I thought for Embedded, if anything we'd move in the opposite direction

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -60,6 +60,14 @@ static int swt_errno(void) {
   return errno;
 }
 
+/// Print a description of the given time value.
+///
+/// This function is provided because Embedded Swift does not support creating
+/// `va_list` arguments to pass to variadic C functions.
+static void swt_printTimeValue(char *buffer, size_t count, long long seconds, int milliseconds) {
+  snprintf(buffer, count, "%lld.%03d seconds", seconds, milliseconds);
```
harlanhaskins: One of these days I would like to rework this to our reporting to show ns/us/ms (whichever is most appropriate for the scale).

Sources/Testing/Expectations/ExpectationChecking+Macro.swift
harlanhaskins: Is there some interesting interplay with the new Embedded support for error existentials? Does the special `any Error` -> `some Error` existential opening not work on Embedded?

Sources/Testing/Attachments/Attachment.swift
```diff
@@ -522,13 +522,22 @@ extension Attachment where AttachableValue: ~Copyable {
         let preferredName = preferredNameComponents.joined(separator: ".")
         let preferredPath = appendPathComponent(preferredName, to: directoryPath)
 
+        func isEEXIST(_ error: any Error) -> Bool {
+#if !hasFeature(Embedded)
+          error._code == swt_EEXIST() && error._domain == "NSPOSIXErrorDomain"
+#else
+          // TODO: detect EEXIST without using _code/_domain?
```
harlanhaskins, line 529: Is the `System` package gonna be usable on Embedded? At the very least we could probably use its errno wrapper (once that's shipped in the toolchain).

Sources/Testing/Expectations/ExpectationChecking+Macro.swift
grynspan: The error passed to this function is already an `any Error` existential, and we cannot open an existential (at least not yet) so the signature of this function must also take `any Error`. If it takes `some Error`, that requires existential opening.

Sources/Testing/Attachments/Attachment.swift
grynspan: We cannot have a dependency on components not in the toolchain, but further any component we are dependent on needs to jump through additional hoops to use Swift Testing.

In any event, we already have our own C error wrapper type. This bit of the PR is likely obsoleted by https://github.com/swiftlang/swift/pull/88954.

Sources/_TestingInternals/include/Stubs.h
grynspan: We could look into it, but in practical terms any granularity below milliseconds is likely to be noise.

Sources/Testing/Running/Runner.swift
grynspan: I'm open to it if you think it's important, but I'm hoping we get a stub implementation of task cancellation that would moot it.

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -60,6 +60,14 @@ static int swt_errno(void) {
   return errno;
 }
 
+/// Print a description of the given time value.
+///
+/// This function is provided because Embedded Swift does not support creating
+/// `va_list` arguments to pass to variadic C functions.
+static void swt_printTimeValue(char *buffer, size_t count, long long seconds, int milliseconds) {
```
grynspan: This change is probably obsoleted by https://github.com/swiftlang/swift/pull/88738.

stmontgomery (approved):
Thanks! Lot of helpful things here

I'd love if you'd add that little TODO [comment](https://github.com/swiftlang/swift-testing/pull/1697#discussion_r3209089840) to `Package.swift` before landing, just since the omission of `TestingMacros` might be confusing. Otherwise everything LGTM!

harlanhaskins (approved):

PR #1701, opened 2026-05-05
Get the names of signals like `SIGABRT` on Windows.

grynspan:
This change enhances the stringification of `ExitStatus.signal(_:)` on Windows to include the name of a signal as it does on Darwin, Linux, and others. For example, `.signal(SIGABRT)` is now stringified as `".signal(SIGABRT)"` instead of `".signal(22)"`.

On Windows, the set of named signals is small, and we already need to enumerate it anyway during exit test preflighting, so since we have a hard-coded list of them we might as well name them too.

I didn't attempt to do something similar for WASI because WASI doesn't support exit tests _anyway_ (so it would be dead code) and because the list of nominally supported signals on WASI (inherited from Musl) is quite a bit longer than on Windows.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/WinSDKAdditions.swift
+++ b/Sources/Testing/Support/Additions/WinSDKAdditions.swift
@@ -50,4 +50,29 @@ let STATUS_SIGNAL_CAUGHT_BITS = {
 
   return result
 }()
+
+/// The names of all signals supported on Windows (to the degree that Windows
+/// supports signals).
+///
+/// Most other platforms provide a C API for getting the name of a signal code.
+/// Windows doesn't have such an API, but it does have a short [list](https://learn.microsoft.com/en-us/cpp/c-runtime-library/signal-constants?view=msvc-170)
+/// of supported signals that we can just hard-code here.
+let windowsSignals: [CInt: String] = {
+  var result = [
+    SIGINT: "SIGINT",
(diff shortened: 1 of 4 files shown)


compnerd (approved):

Sources/Testing/Support/Additions/WinSDKAdditions.swift
```diff
@@ -50,4 +50,29 @@ let STATUS_SIGNAL_CAUGHT_BITS = {
 
   return result
 }()
+
+/// The names of all signals supported on Windows (to the degree that Windows
+/// supports signals).
+///
+/// Most other platforms provide a C API for getting the name of a signal code.
+/// Windows doesn't have such an API, but it does have a short [list](https://learn.microsoft.com/en-us/cpp/c-runtime-library/signal-constants?view=msvc-170)
```
harlanhaskins, line 58: I was a bit concerned originally, but the page clearly says `must be`, so...

harlanhaskins (approved):

Sources/Testing/Support/Additions/WinSDKAdditions.swift
grynspan: Ironically the list on that page is incomplete.

stmontgomery (approved):

PR #1704, opened 2026-05-06
[windows] add PDB files to install target

charles-zablit:
This is part of https://github.com/swiftlang/swift/issues/88869.

### DIFF
--- a/Sources/TestingMacros/CMakeLists.txt
+++ b/Sources/TestingMacros/CMakeLists.txt
@@ -60,6 +60,11 @@ if(SwiftTesting_BuildMacrosAsExecutables)
 
   install(TARGETS TestingMacros
     RUNTIME DESTINATION bin)
+  if(MSVC)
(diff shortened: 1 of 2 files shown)


stmontgomery:
I can understand why we would want symbol files for the testing library itself, i.e. the `Testing` module, since that's linked in to test processes that Swift users may need to debug through. But I'm unclear why we need to include one for the macro plugin (`TestingMacros`) since it's used at build time and user code isn't loaded into that process. Can you elaborate there?

grynspan:
@compnerd Please advise!

charles-zablit:
> I can understand why we would want symbol files for the testing library itself, i.e. the `Testing` module, since that's linked in to test processes that Swift users may need to debug through. But I'm unclear why we need to include one for the macro plugin (`TestingMacros`) since it's used at build time and user code isn't loaded into that process. Can you elaborate there?

I'm not familiar with this part of the toolchain. The overall goal of the debug symbol server is to provide debug symbols for users of the toolchain as well as maintainers. We will also provide debug symbols for lldb and swiftc for instance.

PR #1710, opened 2026-05-12
Remove temporary '_Testing_ExperimentalImageAttachments' library product

stmontgomery:
This removes the `_Testing_ExperimentalImageAttachments` library product from the Swift package. This was added to ease the "Trying it out" instructions for [ST-0014](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0014-image-attachments-in-swift-testing-apple-platforms.md), [ST-0015](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0015-image-attachments-in-swift-testing-windows.md), and [ST-0017](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0017-image-attachment-consolidation.md), but it's no longer necessary since those proposals' reviews have concluded.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -84,19 +84,6 @@ let package = Package(
     )
 #endif
 
-    result += [
-      .library(
-        name: "_Testing_ExperimentalImageAttachments",
-        targets: [
-          "_Testing_AppKit",
-          "_Testing_CoreGraphics",
-          "_Testing_CoreImage",
-          "_Testing_UIKit",
-          "_Testing_WinSDK",
-        ]
-      )
-    ]
-
     result.append(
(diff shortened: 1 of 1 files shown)


grynspan:
_Well akshully_… since SwiftPM does not have support for building cross-import overlays, this is still a convenient way to ensure they're all built when using Swift Testing as a package. But we can also structure the package so that the overlays are just built directly into the main library target when building as a package.

grynspan (approved):
We might want to consider adding a dedicated target for all overlays for other reasons in the future, but it's not a blocker for this PR.

PR #1712, opened 2026-05-13
The subprocess wait thread's name shouldn't be exit-test-specific.

grynspan:
We name the thread we create on Linux/BSD to indicate we're using it to monitor for subprocess termination, but it says "exit test" instead of "subprocess" and we already use this code for other purposes (namely spawning zip/tar). Rename the thread.

### Checklist:


Sources/Testing/ExitTests/WaitFor.swift
harlanhaskins, line 191: What an incredible footgun for cross-platform libraries.

harlanhaskins (approved):

stmontgomery (approved):

Sources/Testing/ExitTests/WaitFor.swift
grynspan: 15 characters for a thread name on Linux. 15!

PR #1716, opened 2026-05-15
Avoid casts to `any ___Trait` in Embedded Swift.

grynspan:
This PR adds helper members on `Trait` to allow casting between the different refining protocols and concrete trait types without relying on `as?` in Embedded Swift.

I have left an `as? TimeLimitTrait` cast unmodified as I expect we won't support time limits in Embedded Swift and will probably just stub out that whole type.

### Checklist:

### DIFF
--- a/Sources/Testing/Traits/Comment.swift
+++ b/Sources/Testing/Traits/Comment.swift
@@ -8,8 +8,6 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-private import _TestingInternals
-
 /// A type that represents a comment related to a test.
 ///
 /// Use this type to provide context or background information about a
--- a/Sources/Testing/Traits/Comment.swift
+++ b/Sources/Testing/Traits/Comment.swift
@@ -8,6 +8,8 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
+private import _TestingInternals
+
 /// A type that represents a comment related to a test.
 ///
 /// Use this type to provide context or background information about a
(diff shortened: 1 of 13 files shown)


Sources/Testing/Traits/Comment.swift
```diff
@@ -175,18 +183,4 @@ extension Test {
...
-
-  /// The complete set of comments about this test from all traits of a certain
-  /// type.
-  ///
-  /// - Parameters:
-  ///   - traitType: The type of ``Trait`` whose comments should be returned.
-  ///
-  /// - Returns: The comments found for the specified test trait type.
-  @_spi(Experimental)
-  public func comments<T>(from traitType: T.Type) -> [Comment] where T: Trait {
```
stmontgomery, line 187: Could we keep this but guard it by `#if !hasFeature(Embedded)`, in case anyone is using it? (Replacing its usage elsewhere in this PR is fine though)

Sources/Testing/Traits/Comment.swift
```diff
@@ -8,6 +8,8 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
+private import _TestingInternals
```
stmontgomery: What necessitates this new import?

Sources/Testing/Traits/Comment.swift
grynspan: It's experimental, so… no. :)

Sources/Testing/Traits/Comment.swift
grynspan: I think I had `swt_unreachable()` in an earlier revision. Will remove.

stmontgomery (approved):

PR #1718, opened 2026-05-16
Opt clients into per-test-case iteration by default

harlanhaskins:
This enables the per-test-case iteration feature by default in Swift 6.4 mode.

Relevant SE proposal: https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0024-per-test-case-repetitions.md

### DIFF
--- a/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
@@ -674,9 +674,6 @@ public func configurationForEntryPoint(from args: __CommandLineArguments_v0) thr
   }
   configuration.repetitionPolicy = repetitionPolicy
 
-  // Opt in to per-test-case repetition
-  configuration.shouldUseLegacyPlanLevelRepetition = false
-
 #if !SWT_NO_EXIT_TESTS
   // Enable exit test handling via __swiftPMEntryPoint().
   configuration.exitTestHandler = ExitTest.handlerForEntryPoint()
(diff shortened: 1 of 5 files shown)


Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -681,7 +675,7 @@ public func configurationForEntryPoint(from args: __CommandLineArguments_v0) thr
   configuration.repetitionPolicy = repetitionPolicy
 
   // Opt in to per-test-case repetition
-  configuration.shouldUseLegacyPlanLevelRepetition = !args.usePerTestCaseRepetition
+  configuration.shouldUseLegacyPlanLevelRepetition = false
```
stmontgomery: Can the default value of the `Configuration.shouldUseLegacyPlanLevelRepetition` property itself become `false` now too, and then we could remove this line outright?

Also, that property has `@_spi(Experimental)` and that could be removed now with this PR

Sources/Testing/Running/Configuration.swift
```diff
@@ -135,7 +135,7 @@ public struct Configuration: Sendable {
   /// Whether to perform test repetition at the plan level or on a per-test-
   /// case basis.
   @_spi(Experimental)
```
stmontgomery: Now this line can go too! (Which will leave _only_ the inherited `@_spi(ForToolsIntegrationOnly)`)
```suggestion
```

stmontgomery (approved):

PR #1722, opened 2026-05-20
Improve stringification support in Embedded Swift.

grynspan:
Add more concrete overloads of `String(describingForTest:)` and `appendInterpolation(_:)` to eliminate more string-related diagnostics. And a few other related changes.

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/CustomTestStringConvertible.swift
+++ b/Sources/Testing/SourceAttribution/CustomTestStringConvertible.swift
@@ -71,26 +71,34 @@ extension String {
   }
 
 #if hasFeature(Embedded)
-  /// Initialize this instance so that it can be presented in a test's output.
-  ///
-  /// - Parameters:
-  ///   - value: The value to describe.
-  ///
-  /// ## See Also
-  ///
-  /// - ``CustomTestStringConvertible``
-  public init(describingForTest value: some CustomTestStringConvertible) {
+  public init(describingForTest value: borrowing some CustomTestStringConvertible) {
     self = value.testDescription
   }
 
-  /// Initialize this instance so that it can be presented in a test's output.
-  ///
-  /// - Parameters:
-  ///   - value: The value to describe.
-  ///
-  /// ## See Also
-  ///
-  /// - ``CustomTestStringConvertible``
+  public init(describingForTest value: borrowing some CustomStringConvertible & CustomTestStringConvertible) {
+    self = value.testDescription
+  }
+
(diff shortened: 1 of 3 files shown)


Sources/Testing/Traits/Comment.swift
```diff
@@ -110,9 +110,53 @@ extension Comment: ExpressibleByStringInterpolation {
       rawValue += literal
     }
 
+#if !hasFeature(Embedded)
     @inlinable public mutating func appendInterpolation(_ value: (some Any)?) {
       rawValue += String(describingForTest: value)
     }
+#else
+    @inlinable public mutating func appendInterpolation(_ value: borrowing some CustomTestStringConvertible) {
```
harlanhaskins, line 118: I'm sure this is fine, but I hope this doesn't cause an issue where a client's wild interpolations explode the overload set. I could see something like

```swift
#expect(calculation.isCorrect(), "Calculation incorrect; should be \(object.x + (object.y - object.z) * object.w + object.a + object.b / object.c")
```

causing a new overload resolution issue...

Sources/Testing/SourceAttribution/CustomTestStringConvertible.swift
```diff
@@ -71,26 +71,34 @@ extension String {
...
 
-  /// Initialize this instance so that it can be presented in a test's output.
-  ///
-  /// - Parameters:
-  ///   - value: The value to describe.
-  ///
-  /// ## See Also
-  ///
-  /// - ``CustomTestStringConvertible``
+  public init(describingForTest value: borrowing some CustomStringConvertible & CustomTestStringConvertible) {
```
harlanhaskins, line 78: Unfortunate that we've lost these comments, can we re-copy them to the overloads?

harlanhaskins: Oh, I guess since this is only enabled for Embedded Swift it doesn't really matter.

Sources/Testing/Traits/Comment.swift
harlanhaskins: I missed that this is only on embedded (same as the previous comment).

harlanhaskins (approved):

Sources/Testing/Traits/Comment.swift
grynspan: They absolutely will explode the type checker in Embedded Swift. So I'm hoping we can discuss with the Embedded Swift stakeholders some way to handle combinatoric conformances that boil down to the same code.

PR #1725, opened 2026-05-22
Reduce key path COW copies in Graph forEach and compactMapValues

inju2403:
This avoids an accidentally-superlinear COW copy of the key path during `Graph.forEach` and `Graph.compactMapValues` traversal.

  ### Motivation:

  While reading [@harlanhaskins](https://github.com/harlanhaskins)' [#1626](https://github.com/swiftlang/swift-testing/pull/1626) (which fixes a quadratic COW issue on `Graph.insertValue`), I noticed that `_forEach` and
  `_compactMapValues` carry a related — though less dramatic — version of the same pattern. Each recursive call builds the descendant's key path with `var childKeyPath = keyPath; childKeyPath.append(key)`, and because
  `childKeyPath` is COW-shared with `keyPath`, the `append` always triggers a copy of the whole key-path array. For a graph with `N` nodes at average depth `d`, this turns traversal into _O(N · d)_ instead of the _O(N)_ it should
  be.

  In isolated benchmarks of the key-path copy pattern at depths 3–6 with branching factor 3–5, the new inout-based version is consistently **3–5× faster**, with the speedup growing with depth — exactly as expected from the _O(N ·
  d) → O(N)_ change:

  | Config | Nodes | Old | New | Speedup |
  | --- | --- | --- | --- | --- |
  | depth 3, branch 3 | 40 | 0.006 ms | 0.002 ms | 3.51× |
  | depth 4, branch 4 | 341 | 0.059 ms | 0.012 ms | 4.84× |
  | depth 4, branch 5 | 781 | 0.138 ms | 0.029 ms | 4.84× |
  | depth 5, branch 4 | 1,365 | 0.242 ms | 0.046 ms | 5.21× |
  | depth 6, branch 4 | 5,461 | 0.830 ms | 0.159 ms | 5.23× |

  ### Modifications:

  The four `_forEach` / `_compactMapValues` overloads (sync + async for each) now thread a single `inout [K]` through the recursion. After invoking the body, the child's key is appended once, recursion descends, and the key is
  popped on the way back up — mutating the key path in place rather than copying it on every descent. The public `forEach(_:)` and `compactMapValues(_:)` wrappers create the initial `var keyPath: [K] = []` and pass it through; the
  public API surface is unchanged.

  All six current `forEach` callsites (two in `Runner.Plan`, four in `AdvancedConsoleOutputRecorder`) consume the key path inline — none of them escape it — so the existing semantics are preserved. The existing `GraphTests` for
  `forEach`, `forEach (async)`, and `compactMapValues` pass unchanged.

  ### Checklist:


Tests/TestingTests/Support/GraphTests.swift
```diff
@@ -588,4 +588,53 @@ struct GraphTests {
     // Standard deviation should be less than 25% of mean
     #expect(standardDeviation < mean * 0.25)
   }
+
```
harlanhaskins: I’ll say I think the pattern in the existing perf test for this is inadequate and doesn’t work reliably, but I haven’t had the time to improve it in a way that’s reliable across CI invocations, so I’m not sure this additional test is gonna be necessary.

harlanhaskins:
The algorithmic improvements look sound here, and assuming it passes the existing test suite I think it’s a good change. Thanks for looking into this!

Tests/TestingTests/Support/GraphTests.swift
inju2403: @harlanhaskins 
Makes sense, thanks for the heads-up! Dropped the perf test in the follow-up commit.

Sources/Testing/Support/Graph.swift
harlanhaskins: Can we clarify these comments with the new `inout` behavior?

grynspan:
Makes sense to me. Perf test aside, do you have any metrics showing an improvement?

Sources/Testing/Support/Graph.swift
inju2403: @harlanhaskins 
Updated the docs for all four overloads (sync + async × forEach + compactMapValues) to describe the reused key path and the COW copy it avoids — see the next commit. Let me know if you'd prefer different wording!

inju2403:
Thanks, @grynspan! There's a benchmark table in the PR description with isolated measurements (3-5× speedup, scaling with depth) — let me copy it here for visibility:

  | Config | Nodes | Old | New | Speedup |
  | --- | --- | --- | --- | --- |
  | depth 3, branch 3 | 40 | 0.006 ms | 0.002 ms | 3.51× |
  | depth 4, branch 4 | 341 | 0.059 ms | 0.012 ms | 4.84× |
  | depth 4, branch 5 | 781 | 0.138 ms | 0.029 ms | 4.84× |
  | depth 5, branch 4 | 1,365 | 0.242 ms | 0.046 ms | 5.21× |
  | depth 6, branch 4 | 5,461 | 0.830 ms | 0.159 ms | 5.23× |

  These are from an isolated micro-benchmark of just the keyPath copy pattern (release build, best of 20 runs, warmup ×5). The speedup grows roughly with depth — consistent with the O(N·d) → O(N) reduction. Happy to put together
  end-to-end numbers against a real test plan if that's more useful.

grynspan:
Thanks, sorry! Looks like it wasn't rendering in the GitHub app.

Sources/Testing/Support/Graph.swift
```diff
@@ -389,8 +389,12 @@ extension Graph {
   /// The recursive implementation of `forEach(_:)`.
   ///
   /// - Parameters:
-  ///   - keyPath: The key path to use for the root node when passing it to
-  ///     `body`.
+  ///   - keyPath: The key path corresponding to the current node. The same
+  ///     array is reused across the entire traversal: each descent appends
+  ///     the child key, and the recursive call pops it on the way back up.
+  ///     This avoids the per-descent COW copy that
```
harlanhaskins: I don't think we need this rationale, just the description of what the keyPath is used for

inju2403: @harlanhaskins 
Good call — simplified to just describe what the keyPath is. The rationale lives better in the commit message anyway.

grynspan (approved):

harlanhaskins (approved):

harlanhaskins:
@inju2403 Thanks! Feel free to merge. It should be set as the default, but if not, make sure it's set to "Squash and merge"

grynspan:
I think you need to do that @harlanhaskins.

harlanhaskins:
Ah, sorry. I figured with approval they'd be able to. But at any rate, thanks @inju2403!

grynspan:

inju2403:

PR #1732, opened 2026-05-27
Add an overload of Issue.record that takes all its parts up-front

harlanhaskins:
This adds an overload of `Issue.record` that takes a `SourceContext` and propagates a client-provided backtrace and source location.

### Motivation:

This is useful for interop to make sure we don't use implicit locations and backtraces and instead use those captured when the issue was recorded.

### Modifications:

Just one additional overload guarded by `@_spi(ForToolsIntegrationOnly)` and a test for it.

### Checklist:

rdar://178080757

### DIFF
--- a/Sources/Testing/Issues/Issue+Recording.swift
+++ b/Sources/Testing/Issues/Issue+Recording.swift
@@ -155,6 +155,27 @@ extension Issue {
     return issue.record()
   }
 
+  /// Records an issue by explicitly providing all information up-front.
+  ///
+  /// - Parameters:
+  ///   - comments: All comments that should be attached to the issue.
+  ///   - kind: The specific kind of issue being recorded.
+  ///   - severity: The severity of the issue being recorded.
+  ///   - sourceContext: The context in which this issue was originally recorded.
+  ///
+  /// - Returns: The issue that was recorded.
+  /// - Note: This is only intended for use in interoperability.
+  @_spi(ForToolsIntegrationOnly)
+  @discardableResult public static func record(
+    comments: [Comment] = [],
+    kind: Issue.Kind,
+    severity: Severity,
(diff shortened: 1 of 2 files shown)


Sources/Testing/Issues/Issue+Recording.swift
```diff
@@ -155,6 +155,27 @@ extension Issue {
...
+  ///   - comments: All comments that should be attached to the issue.
+  ///   - kind: The specific kind of issue being recorded.
+  ///   - severity: The severity of the issue being recorded.
+  ///   - sourceContext: The context in which this issue was originally recorded.
+  ///
+  /// - Returns: The issue that was recorded.
+  /// - Note: This is only intended for use in interoperability.
+  @_spi(ForToolsIntegrationOnly)
+  @discardableResult public static func record(
+    comments: [Comment] = [],
```
stmontgomery: Given that it accepts "all information up-front", should we avoid this default `= []` value?

harlanhaskins: Oh good point

stmontgomery (approved):

Tests/TestingTests/IssueTests.swift
```diff
@@ -1712,6 +1712,23 @@ final class IssueTests: XCTestCase {
     let issue = Issue(kind: .errorCaught(error), severity: .error, comments: [], sourceContext: .init())
     #expect(String(describing: issue).contains("RIGHT"))
   }
+
+  func testExplicitSourceContextWithoutLocation() async {
+    // This just ensures that `Issue.record` with an explicit sourceContext doesn't infer the location via #_sourceLocation.
```
grynspan: Nitpick: wrap comments at 80 columns.

PR #1733, opened 2026-05-28
Expose a non-underscored version of `#_sourceLocation` as SPI.

grynspan:
This PR clones `#_sourceLocation` to `#Testing::sourceLocation` guarded by `@_spi(Experimental)`. Because `#sourceLocation(file:line:)` is a compiler-side statement and not a macro, developers currently can't just write `#sourceLocation` and must specify a module selector. This is still (hopefully) better than the underscored `#_sourceLocation`.

Example usage:

```swift
func foo(sourceLocation: SourceLocation = #Testing::sourceLocation) {
  print("Called from \(sourceLocation.fileID)")
}
```

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
+++ b/Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
@@ -8,16 +8,60 @@
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 //
 
-/// Get the current source location as a compile-time constant.
+/// Get the current source location.
 ///
-/// - Returns: The source location at which this macro is applied.
+/// - Returns: This expression's location in the current Swift source file.
 ///
-/// You can use this macro in place of `#fileID`, `#filePath`, `#line`, and
-/// `#column` as a default argument to a function. It expands to an instance of
-/// ``SourceLocation`` referring to the location of the macro invocation itself
-/// (similar to how `#fileID` expands to the ID of the file containing the
-/// `#fileID` invocation).
+/// At compile time, the testing library expands this macro to an instance of
+/// ``SourceLocation`` referring to the location of the macro invocation itself.
+/// If you want to create an instance of ``SourceLocation`` from specific file
+/// ID, file path, line, and column values, use ``SourceLocation/init(fileID:filePath:line:column:)``
+/// instead.
+///
+/// You can use this expression macro in place of [`#fileID`](https://developer.apple.com/documentation/swift/fileid()),
+/// [`#filePath`](https://developer.apple.com/documentation/swift/filepath()),
+/// [`#line`](https://developer.apple.com/documentation/swift/line()), and
+/// [`#column`](https://developer.apple.com/documentation/swift/column()) as a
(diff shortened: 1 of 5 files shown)


rr-codes (approved):

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
```diff
@@ -8,17 +8,61 @@
...
+/// [`#column`](https://developer.apple.com/documentation/swift/column()) as a
+/// default argument to a function.
+///
+/// ```swift
+/// func cookBurger(sourceLocation: SourceLocation = #Testing::sourceLocation) {
+///   // ...
+/// }
+/// ```
+@_spi(Experimental)
+@freestanding(expression) public macro sourceLocation() -> SourceLocation = #externalMacro(module: "TestingMacros", type: "SourceLocationMacro")
```
stmontgomery, line 64: If we wanted, it could be defined in terms of the original right? (For now, at least, until it becomes fully public, at which point we could reverse their definitions.)
```suggestion
@freestanding(expression) public macro sourceLocation() -> SourceLocation = _sourceLocation()
```

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
```diff
@@ -8,17 +8,61 @@
...
+/// If you want to create an instance of ``SourceLocation`` from specific file
+/// ID, file path, line, and column values, use ``SourceLocation/init(fileID:filePath:line:column:)``
+/// instead.
+///
+/// - Important: You must specify a module selector when you use this expression
+///   macro to avoid conflicting with the Swift compiler's [`#sourceLocation(file:line:)`](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Line-Control-Statement)
+///   statement.
+///
+///   ```swift
+///   let here = #Testing::sourceLocation
```
stmontgomery, line 49: This is perfectly valid, but I feel like most _practical_ usages will be in the context of some helper/utility function. I wonder if giving an example of such a function, with a parameter whose default value is this macro, would be more useful to readers.

stmontgomery: Never mind! I see such an example a few lines below. But that makes me wonder why this example is useful. Could the `- Important` callout be consolidated with the code example below somehow?

stmontgomery (approved):

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
grynspan: I went back and forth about it before posting the PR. This callout comes first because you cannot use the macro at all without the module selector; it then needs an example because module selectors are not a well-known, well-understood corner of the language.

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
grynspan: They're logically equivalent.

PR #1735, opened 2026-05-29
docs: Expand discussion on when to use #expect vs. #require

### LINKED ISSUE #935: Expand discussion on when to use #expect vs. #require
The [Expectations article](https://github.com/swiftlang/swift-testing/blob/main/Sources/Testing/Testing.docc/Expectations.md#validate-your-codes-result) briefly describes the behavioral difference between `#expect` and `#require`, but doesn't elaborate much on how to decide which one to use. I keep seeing people coming around and asking for advice about this, so it would be nice to expand this discussion a bit more.

adityasingh2400:
Expand the Expectations article with guidance on choosing between `#expect` and `#require`.

### Motivation:

The [Expectations article](https://github.com/swiftlang/swift-testing/blob/main/Sources/Testing/Testing.docc/Expectations.md#validate-your-codes-result) briefly describes the behavioral difference between `#expect` and `#require`, but doesn't elaborate on how to decide which one to use. People regularly ask for advice about this, so the article should discuss the trade-off more thoroughly.

Fixes #935.

### Modifications:

- Added a new "Decide between an expectation and a requirement" section to `Sources/Testing/Testing.docc/Expectations.md`.
- The section explains the behavioral difference (`#expect` records an issue and continues; `#require` records an issue and throws, ending the test) and gives a rule of thumb for choosing between them.
- It also covers the optional-unwrapping overload of `#require`, contrasting it with using `#expect` for a non-`nil` check when you don't need the unwrapped value and want the test to keep running.
- All DocC symbol references use the same disambiguated links already used elsewhere in the article (`require(_:_:sourceLocation:)-5l63q` and `require(_:_:sourceLocation:)-6w9oo`).

### Result:

The Expectations article now provides actionable guidance for deciding between `#expect` and `#require`. Documentation-only change; the `Testing` target continues to build cleanly.

### Checklist:

- [x] If public symbols are renamed or modified, DocC references should be updated. (No symbols changed; existing disambiguated DocC references reused.)

### DIFF
--- a/Sources/Testing/Testing.docc/Expectations.md
+++ b/Sources/Testing/Testing.docc/Expectations.md
@@ -54,6 +54,36 @@ the test when the code doesn't satisfy a requirement, use
 ``require(_:_:sourceLocation:)-5l63q`` throws an instance of
 ``ExpectationFailedError`` when your code fails to satisfy the requirement.
 
+### Decide between an expectation and a requirement
+
+``expect(_:_:sourceLocation:)`` and ``require(_:_:sourceLocation:)-5l63q``
+check the same kinds of conditions, but they differ in what happens when a
+check fails:
+
+- ``expect(_:_:sourceLocation:)`` records an issue and then _continues_
+  running the rest of the test. Use it when a single failed check doesn't
+  prevent the remaining checks in the test from producing meaningful results,
+  so that one test run can report several independent failures at once.
+- ``require(_:_:sourceLocation:)-5l63q`` records an issue and then throws an
+  error, which _ends_ the current test. Because it throws, you call it with
+  `try`. Use it when the rest of the test can't run meaningfully unless the
+  condition holds, to avoid reporting a cascade of failures that all stem from
+  the same root cause.
+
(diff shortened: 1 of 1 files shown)


grynspan:
Did you use generative AI to write this documentation?

adityasingh2400:
I wrote it myself, based on my own experience using the two macros and the existing API documentation. The point I was trying to make clearer is the behavioral split: #expect records the failure and lets the test keep running, so it fits independent checks where the later assertions still give useful signal, while #require throws and ends the current test when continuing would be pointless or unsafe, for example right after unwrapping an optional you are about to use. If the framing or any of the wording does not line up with how you would describe it, I am glad to revise or trim it down to whatever fits the documentation best.

grynspan:
No need (I'll review the text later today). Just making sure that we keep track of documentation authorship.

grynspan:
The team discussed this PR privately. We like some of the approach in the PR but at this time, we are going to close it until we can figure out as a team how to approach/next steps. We can always reopen but would rather do this than leave the PR hanging. Feel free to discuss more over in #935.

PR #1736, opened 2026-05-31
Remove unnecessary type casting

compnerd:
The compiler now imports these as their canonical type, allowing us to directly use without the explicit type cast.


grynspan:
Well… clearly not. 😄

Keep in mind we intend for our main branch to compile against the previous release (which is still 6.3 for now).

compnerd:
Clearly it does! You just didn't cherry-pick all the necessary bundled changes :)

We will likely have to split this into two changes - force the types to the correct shapes where we were incorrect, and then we can do the clean up on the explicit types later.

grynspan:
Oh I figured there were upstream changes we'd need. But we won't be able to accept this PR while we still need to build for 6.3.

compnerd:
Ah, that is unfortunate. But, understandable, we can live with this until we can adopt (hopefully) the next release.

PR #1740, opened 2026-06-03
Silence a Windows-only warning about unused `try`.

grynspan:
Silences a warning emitted from a test target on Windows.

### Checklist:


grynspan:
Self-reviewing as this is a tiny change in tests only.

PR #1743, opened 2026-06-05
Don't send fallback events when interop mode is none

jerryjrchen:
### Motivation:

Today, XCTest doesn't have any mode checking logic so it can't actually opt-out of interop.

That's because Swift Testing will happily send fallback events, regardless of the mode, if it detects a fallback event handler is installed.

Note that Swift Testing already refuses to *receive* fallback events when the interop mode is none.

### Modifications:

Add a check for interop mode != none when resolving the _postToFallbackEventHandler closure.

Resolves rdar://178774969

### Checklist:

### DIFF
--- a/Sources/Testing/Events/Event+FallbackEventHandler.swift
+++ b/Sources/Testing/Events/Event+FallbackEventHandler.swift
@@ -234,7 +234,12 @@ extension Event {
   /// handler belongs to the testing library (and so shouldn't be called by us),
   /// the value of this property is `nil`.
   private static let _postToFallbackEventHandler: Event.Handler? = {
-    guard let fallbackEventHandler = _swift_testing_getFallbackEventHandler() else {
+    // If Swift Testing API is called when mode == none, XCTest might still have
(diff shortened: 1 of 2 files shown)


Sources/Testing/Events/Event+FallbackEventHandler.swift
```diff
@@ -234,7 +234,12 @@ extension Event {
   /// handler belongs to the testing library (and so shouldn't be called by us),
   /// the value of this property is `nil`.
   private static let _postToFallbackEventHandler: Event.Handler? = {
-    guard let fallbackEventHandler = _swift_testing_getFallbackEventHandler() else {
+    // If Swift Testing API is called when mode == none, XCTest might still have
+    // installed a fallback event handler. We should refuse to send any events.
+    guard
+      // Interop.Mode.current != .none,
```
jerryjrchen: I was a dummy and forgot to uncomment out the check (I used it to qualify my tests locally :))

stmontgomery (approved):

PR #1745, opened 2026-06-05
Enable Android emulator testing with the snapshot toolchains employing swift-build in SwiftPM

finagolfin:
@marcprux found that [the trunk snapshot toolchain now successfully cross-compiles the tests for Android](https://github.com/swiftlang/swift-package-manager/issues/8094#issuecomment-4616423370), which I was able to reproduce locally with the latest 6.4 snapshot also. With the latest GitHub workflows using `swift-build` again for the 6.4 and trunk snapshots, swiftlang/github-workflows#269, let's see if the trunk tests run in the emulator.

We currently cross-compile this repo to Android using the defaults of 6.3, nightly-6.3, and trunk, and we'll probably add nightly-6.4 soon: can we drop cross-compiling for Android with 6.3 then? That will allow us to run the tests for 6.4 and trunk in the emulator, without any workarounds needed for 6.3 being unable to cross-compile these tests.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -21,7 +21,7 @@ jobs:
             ls -ld /Applications/*Xcode*
   tests:
     name: Test
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.11
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.12
     with:
       linux_swift_versions: '["nightly-main", "nightly-6.3"]'
       linux_os_versions: '["amazonlinux2", "jammy"]'
@@ -35,13 +35,15 @@ jobs:
       wasm_sdk_versions: '["6.3", "nightly-main", "nightly-6.3"]'
       enable_android_sdk_build: true
       android_ndk_versions: '["r28c", "r29"]'
+      android_sdk_versions: '["nightly-main", "nightly-6.4.x"]'
       android_sdk_triples: '["aarch64-unknown-linux-android33", "x86_64-unknown-linux-android28"]'
+      enable_android_sdk_checks: true
       enable_freebsd_checks: true
       freebsd_swift_versions: '["nightly-main"]'
       freebsd_os_versions: '["14.3"]'
   soundness:
     name: Soundness
-    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.11
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.12
     with:
       license_header_check_project_name: "Swift"
       docs_check_enabled: false
--- a/.github/workflows/automerge.yml
+++ b/.github/workflows/automerge.yml
@@ -18,7 +18,7 @@ on:
 jobs:
   create_merge_pr:
     name: Create PR to merge main into release branch
-    uses: swiftlang/github-workflows/.github/workflows/create_automerge_pr.yml@0.0.11
+    uses: swiftlang/github-workflows/.github/workflows/create_automerge_pr.yml@0.0.12
     with:
       head_branch: main
       base_branch: release/6.4.x
--- a/.github/workflows/main_using_main.yml
+++ b/.github/workflows/main_using_main.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:
   tests:
     name: Test
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.11
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.12
     with:
       linux_swift_versions: '["nightly-main"]'
       linux_os_versions: '["amazonlinux2", "jammy"]'
@@ -32,3 +32,4 @@ jobs:
       android_sdk_versions: '["nightly-main"]'
       android_ndk_versions: '["r28c", "r29"]'
       android_sdk_triples: '["aarch64-unknown-linux-android33", "x86_64-unknown-linux-android28"]'
+      enable_android_sdk_checks: true
--- a/.github/workflows/main_using_release.yml
+++ b/.github/workflows/main_using_release.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:
   tests:
     name: Test
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.11
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.12
     with:
       linux_swift_versions: '["nightly-6.3"]'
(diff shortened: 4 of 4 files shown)


.github/workflows/pull_request.yml
```diff
@@ -21,7 +21,7 @@ jobs:
             ls -ld /Applications/*Xcode*
   tests:
     name: Test
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.11
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
```
harlanhaskins: @shahmishal Should we cut a new tag of `swift_package_test`? I would rather we not depend on `main` here, but the last tag was April 14th

marcprux:
Weird, [failing](https://github.com/swiftlang/swift-testing/actions/runs/27017294438/job/79735581384?pr=1745#step:9:609) with:

```
/home/runner/work/swift-testing/swift-testing/Sources/Testing/SourceAttribution/Backtrace.swift:75:27: error: cannot find 'backtrace' in scope
 73 |       if #available(Android 33, *) {
 74 |         initializedCount = addresses.withMemoryRebound(to: UnsafeMutableRawPointer.self) { addresses in
 75 |           .init(clamping: backtrace(addresses.baseAddress!, .init(clamping: addresses.count)))
    |                           `- error: cannot find 'backtrace' in scope
 76 |         }
 77 |       }
```

Why isn't the `#available(Android 33, *)` guard working? Some new regression? Any ideas, @madsodgaard?

finagolfin:
I am trying to reproduce that failure locally with the trunk snapshot toolchain, will let you know.

finagolfin:
Huh, I cannot reproduce that build error in a linux environment running the same commands, implying it is something specific to the CI environment.

finagolfin:
Oh, found it, `swift-build` is using the wrong NDK, not the ones we specified: `-sdk /usr/local/lib/android/sdk/ndk/27.3.13750724/toolchains/llvm/prebuilt/linux-x86_64/sysroot -sysroot /usr/local/lib/android/sdk/ndk/27.3.13750724/toolchains/llvm/prebuilt/linux-x86_64/sysroot`.

.github/workflows/pull_request.yml
finagolfin: Hold on, we're still adjusting the Android workflow to `swift-build`, https://github.com/swiftlang/github-workflows/issues/277 and the NDK bug below, for which this repo is a good test.

finagolfin:
The workflow was updated to add the 6.4 snapshots, will submit a fix for the NDK issue next.

grynspan:
Sounds like this isn't ready to review yet?

finagolfin:
No, the failed tests are an indicator, 😉 but what do you think about disabling 6.3 testing?

grynspan:
> what do you think about disabling 6.3 testing?

Specifically for Android? Sure, I buy it. For all platforms? Nay!

finagolfin:
> Specifically for Android?

Yep, just stop cross-compiling trunk for Android with the 6.3 toolchain, as we're doing with this pull still.

> For all platforms?

Why not? Doesn't much matter to me, but I don't see any value in continuing to build trunk Testing with 6.3, and I think you once said the practice here is only to keep it compiling with the previous release branch, which is now 6.4.

grynspan:
6.4 hasn't been released yet, and we don't have 6.4 toolchains set up in CI for all platforms. So 6.3 needs to stay for now, but since we don't formally support Android with Swift Testing yet (since, hey, no CI tests!) we can probably drop the 6.3 Android jobs without much fuss.

.github/workflows/pull_request.yml
stmontgomery: Just echoing @harlanhaskins: I think we should not switch to `main`, and should instead wait for a new tag of `github-workflows` containing whatever new content you need. We use pinned versions intentionally.

stmontgomery:
Is the problem that the tests _build_ for 6.3 but they encounter failures? Or do the tests (or the product code itself) fail to build outright? If the former, my preference would be to maintain the CI which builds everything for 6.3, but just disable _running_ the tests on that Swift version.

.github/workflows/pull_request.yml
finagolfin: Sure, but I can't test those new workflow changes from upstream without this update of the workflow branch here, while I work on this pull and the workflow changes needed.

stmontgomery: Absolutely, no problem. I only meant to say that I think by the time this lands, we ought to replace this with a new tag. Using it temporarily to iterate makes sense

finagolfin:
@stmontgomery, the latter, ie the tests do not cross-compile for Android with Swift 6.3, whereas they now do with Swift 6.4 and 6.5 using `swift-build`.

Also, the official Swiftlang workflows don't provide ways to only build for one Swift version while testing other versions.

stmontgomery:
> @stmontgomery, the latter, ie the tests do not cross-compile for Android with Swift 6.3, whereas they now do with Swift 6.4 and 6.5 using `swift-build`.

In that case I'm fine with dropping the 6.3 CI build for Android

finagolfin:
Got the upstream workflow to use the latest NDKs with the snapshot toolchains, rebased this pull, and disabled building for 6.3 as discussed. That got this repo to build and run its tests in the Android emulator for the first time. 😸

The only remaining issue is that the upstream workflow doesn't yet support multiple test runners, since `swift-build` now produces a test runner _per_ test target [as I linked earlier](https://github.com/swiftlang/swift-testing/pull/1745#discussion_r3363705647), which means only the macro tests here are run.

@marcprux, do you have that runner fix available locally? We have long done it that way on my Android CI for multiple Swift packages tested together, with such a test script produced by Marc, so we can probably do the same here.

marcprux:
> @marcprux, do you have that runner fix available locally? We have long done it that way on my Android CI for multiple Swift packages tested together, with such a test script produced by Marc, so we can probably do the same here.

Not yet. It's on my TODO list.

finagolfin:
Rebased, updated all official workflows to [the latest tag](https://forums.swift.org/t/swiftlang-github-workflows-tagged-0-0-12/87746), and added Android testing to the trunk commit workflow also. I think we can go ahead and get this in, as it cross-compiles all the trunk tests for Android and runs a subset of them in the Android emulator, with the remaining waiting on swiftlang/github-workflows#277 getting fixed.

.github/workflows/pull_request.yml
finagolfin: Done

finagolfin:
Ping, this has been ready for weeks, just looking for sign off on merging.

grynspan:
Looks like this did make Android builds happy so I'll merge it now.

grynspan:
✅ because GitHub won't let me add it _post facto_.

PR #1749, opened 2026-06-09
Adopt `objc_copyImageHeaders()` for safer image discovery on Darwin.

grynspan:

We must maintain code paths for earlier Apple OSes as well as Apple configurations that disable Objective-C interop. For earlier Apple OSes, we continue to take our existing approach; for no-Objective-C configurations, which are not customer-facing, dyld has lower-level API we can call.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -405,7 +405,7 @@ extension Array where Element == PackageDescription.SwiftSetting {
       .enableExperimentalFeature("AvailabilityMacro=_typedThrowsAPI:macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0"),
       .enableExperimentalFeature("AvailabilityMacro=_transferableAPI:macOS 15.2, iOS 18.2, watchOS 11.2, tvOS 18.2, visionOS 2.2"),
       .enableExperimentalFeature("AvailabilityMacro=_castingWithNonCopyableGenerics:macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0"),
-      .enableExperimentalFeature("AvailabilityMacro=_objcCopyImageHeadersAPI:anyAppleOS 27.0"),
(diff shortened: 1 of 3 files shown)


mikeash (approved):

Package.swift
```diff
@@ -405,6 +405,7 @@ extension Array where Element == PackageDescription.SwiftSetting {
       .enableExperimentalFeature("AvailabilityMacro=_typedThrowsAPI:macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0"),
       .enableExperimentalFeature("AvailabilityMacro=_transferableAPI:macOS 15.2, iOS 18.2, watchOS 11.2, tvOS 18.2, visionOS 2.2"),
       .enableExperimentalFeature("AvailabilityMacro=_castingWithNonCopyableGenerics:macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0"),
+      .enableExperimentalFeature("AvailabilityMacro=_objcCopyImageHeadersAPI:anyAppleOS 27.0"),
```
stmontgomery: `anyAppleOS` isn't supported by 6.3. I know we just turned off auto-merging `main` back to 6.4 branch, so if your intention is to drop support for building `main` with a 6.3 toolchain we could do that, but I think we should first prepare our CI for that.

stmontgomery: Probably easiest to just stick with explicit OS versions in this PR though

grynspan: Yeah I'll tidy it up. The build failed anyway due to the older SDK content.

harlanhaskins (approved):

grynspan:
Linux failures are due to a toolchain bug.

hamishknight:
Seems like this broke arm64 CI: https://ci.swift.org/job/oss-swift-incremental-RA-macos-apple-silicon/10858/

grynspan:
Wonderful.

finagolfin:
It just [broke the macOS x86_64 trunk toolchain CI](https://ci.swift.org/job/oss-swift-package-macos/5023/console) too.

grynspan:
We've reverted the change for now and I'll open a new PR once I sort out the appropriate adjustments.

PR #1750, opened 2026-06-11
Remove legacy plan-level repetition behavior

harlanhaskins:
This removes the legacy plan-level test repetition behavior now that the main branch is no longer targeting 6.4.

### Motivation:

Per the [ST-0024](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0024-per-test-case-repetitions.md) evolution proposal, the legacy behavior was only intended to last for a small transitionary period. Because SwiftPM did not previously support the legacy behavior, there are no backwards compatibility concerns there.

### Modifications:

This removes the behavior from the runner, deprecates the `shouldUseLegacyPlanLevelRepetition` property in `Configuration`, removes all handling for `iterationStarted` and `iterationEnded` events, and removes the existing tests for the legacy behavior. While updating this, I removed `EventIterationTests` as well, since it's redundantly testing the same behavior that's in `TestCaseIterationTests.`

### Checklist:

Fixes rdar://177379218


Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -301,9 +297,7 @@ extension Event.HumanReadableOutputRecorder {
         context.runStartInstant = instant
 
       case .iterationStarted:
```
grynspan: There's a `default` case on this switch you can use for events that don't need to touch any state.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -409,15 +403,9 @@ extension Event.HumanReadableOutputRecorder {
         )
       ) + _formattedComments(comments)
 
-    case let .iterationStarted(index):
```
grynspan, line 412: Nitpick, pair these two events together like `.planStepStarted`/`Ended`?

Sources/Testing/Events/Event.swift
```diff
@@ -440,17 +440,7 @@ extension Event.Kind {
...
-    /// - Parameters:
-    ///   - index: The index of the iteration. The first iteration has an index
-    ///     of `0`.
-    ///
-    /// This event is posted at the start of each test plan iteration.
-    ///
-    /// By default, a test plan runs for one iteration, but the
-    /// ``Configuration/repetitionPolicy-swift.property`` property can be set to
-    /// allow for more iterations.
+    /// Deprecated; this is no longer emitted.
```
grynspan: Use `@available` for this.

Sources/Testing/Events/Event.swift
```diff
@@ -542,17 +532,7 @@ extension Event.Kind {
...
-    /// - Parameters:
-    ///   - index: The index of the iteration. The first iteration has an index
-    ///     of `0`.
-    ///
-    /// This event is posted at the end of each test plan iteration.
-    ///
-    /// By default, a test plan runs for one iteration, but the
-    /// ``Configuration/repetitionPolicy-swift.property`` property can be set to
-    /// allow for more iterations.
+    /// Deprecated; this is no longer emitted.
```
grynspan: `@available` here too.

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
harlanhaskins: I'm generally hesitant to _add_ a `default` just because I want completeness, but I'll move this since there's already one

Sources/Testing/Events/Event.swift
harlanhaskins: The only thing is I don't actually know how to silence the warning in the clients that use this -- can you put `@available` on a switch case? I guess I can just godbolt that.

harlanhaskins: Oh, well, I tried it, and it just doesn't warn in the switch case but it warns when you try to use the value outside of pattern matching. That's a smart default. https://godbolt.org/z/6rnza4jqq

Sources/Testing/Events/Event.swift
```diff
@@ -542,67 +511,65 @@ extension Event.Kind {
...
+  }
+}
+
+extension Event.Kind.Snapshot {
+  /// Initialize an instance of this type by snapshotting the specified event
+  /// kind.
+  ///
+  /// - Parameters:
+  ///   - kind: The original event kind to snapshot.
+  @diagnose(DeprecatedDeclaration, as: ignored)
```
stmontgomery, line 536: Unfortunately `@diagnose` is only in 6.4 and the 6.3 CI is [failing](https://github.com/swiftlang/swift-testing/actions/runs/27371905382/job/80886077399?pr=1750)

stmontgomery: Could we enclose just the attribute in `#if compiler(>=6.4)`?

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
```diff
@@ -624,18 +605,9 @@ extension Event.HumanReadableOutputRecorder {
...
-      }
-      let duration = descriptionOfDuration(from: iterationStartInstant, to: instant)
-
-      return [
-        Message(
-          symbol: .default,
-          stringValue: "Iteration \(index + 1) ended after \(duration)."
-        )
-      ]
+    case .iterationStarted, .iterationEnded:
```
stmontgomery, line 608: Is this now going to be emitting deprecation diagnostics when we build the package ourselves? (Somewhat related to my other comment about `@diagnose`) Is `@diagnose` usable in this position, on a switch `case` ?

stmontgomery (approved):

Sources/Testing/Events/Recorder/Event.HumanReadableOutputRecorder.swift
harlanhaskins: So, switch cases implicitly suppress deprecation warnings, which is nice.

Sources/Testing/Events/Event.swift
harlanhaskins: Guarded with a hasAttribute, but also it seems that hasAttribute prevents @diagnose from taking effect: https://godbolt.org/z/G5hvahGbP

grynspan: Then can we use `#if compiler` instead?

harlanhaskins: huh, that does work, I swear I tried that too, but maybe I was misreading the output?

harlanhaskins: Well, no, in 6.4 beta 1 it doesn't work, but in godbolt it does, which means maybe the behavior has changed in recent Swift commits? I'm trying nightly. It's still broken within a hasAttribute on nightly.

harlanhaskins:
(CI Failures unrelated)

PR #1751, opened 2026-06-12
Add a `write(toFileAtPath:)` SPI function to `Attachment`.

grynspan:
This PR adds the aforementioned SPI function. Tools can use this function to force an attachment to write itself to a specific file path rather than just an unspecified file name inside a specific directory. This then allows tools to have tighter control over the attachment-saving process.

For attachments created from file URLs via the Foundation overlay, this API will use file cloning where available on Darwin, Linux, and FreeBSD.

Tools that use this API are responsible for ensuring that the file paths they use are unique as the function will throw `EEXIST` if you use it to write to a file that already exists on disk.

### Checklist:

### DIFF
--- a/Sources/Testing/Attachments/Attachment.swift
+++ b/Sources/Testing/Attachments/Attachment.swift
@@ -429,9 +429,9 @@ extension Attachable where Self: ~Copyable {
   ///
   /// - Throws: Any error that prevented writing this instance to `filePath`.
   ///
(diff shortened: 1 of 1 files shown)


grynspan:
Linux failures are due to https://github.com/swiftlang/swift/pull/89354 and will be fixed by https://github.com/swiftlang/swift/pull/89917

jerryjrchen (approved):

PR #1754, opened 2026-06-15
Adopt `objc_copyImageHeaders()` for safer image discovery on Darwin (take 2)

grynspan:

`objc_copyImageHeaders()` is forward-declared as CI doesn't have new-enough SDKs yet.

We must maintain code paths for earlier Apple OSes as well as Apple configurations that disable Objective-C interop. For earlier Apple OSes, we continue to take our existing approach; for no-Objective-C configurations, which are not customer-facing, dyld has lower-level API we can call.

Resolves rdar://169186671.

### Checklist:


Package.swift
```diff
@@ -379,6 +379,9 @@ extension Array where Element == PackageDescription.SwiftSetting {
       // Enabled to allow tests to be added to ~Escapable suites.
       .enableExperimentalFeature("Lifetimes"),
 
+      // Enabled to allow forward-declaring platform functions.
```
harlanhaskins: I don't mind adding this, but are we using this in this PR?

harlanhaskins (approved):

Package.swift
stmontgomery: Looks like `@_extern` was used in an earlier commit but is no longer

stmontgomery (approved):

Package.swift
grynspan: Yup it's a stray.

grynspan:
Linux failures are unrelated.

PR #1755, opened 2026-06-16
Make @_spi ABI visible to corelibs XCTest

jerryjrchen:
### Motivation:

Allow the ABI type to be shared with corelibs XCTest to deduplicate some interoperability logic

### Modifications:

Update CMake build script so that @_spi visible symbols are included in the CMake module. This can then be found by swift-corelibs-xctest when building with CMake.

### Checklist:

### DIFF
--- a/Sources/Testing/CMakeLists.txt
+++ b/Sources/Testing/CMakeLists.txt
@@ -166,3 +166,8 @@ _swift_testing_install_target(Testing)
 # Install the Swift cross-import overlay directory.
 _swift_testing_install_swiftcrossimport(Testing
   Testing.swiftcrossimport)
+
+# Make @_spi visible to other projects building against Testing in the toolchain (e.g. swift-corelibs-xctest)
+set_target_properties(Testing PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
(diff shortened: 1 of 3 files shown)


grynspan:
It is important we don't ship these symbols outside the toolchain. We may need a post-build CMake step (???) that deletes the corresponding module metadata.

stmontgomery:
> It is important we don't ship these symbols outside the toolchain. We may need a post-build CMake step (???) that deletes the corresponding module metadata.

I don't think this will cause module SPI interface content to be exposed to _end users_ of the toolchain. I think all it will do is allow other projects which build and install content into the same toolchain to access the full module, and that's OK to allow.

jerryjrchen:
> I think all it will do is allow other projects which build and install content into the same toolchain to access the full module, and that's OK to allow.

That's my understanding too. I should be able to verify by double checking the build artifacts right?

grynspan:
We should verify the built toolchain does not contain non-public content, yes.

jerryjrchen:
Built the toolchain with the changes here: https://ci.swift.org/job/swift-PR-toolchain-Linux/1770/

No `@_spi` content appears in the Testing swiftinterface files and there isn't a private.swiftinterface file present either. I'm feeling fairly confident this build change doesn't regress in that aspect.

stmontgomery (approved):

grynspan:
This probably means we could ship private Swift module information as a separate download for folks who want to use tools SPI with a production toolchain.

PR #1756, opened 2026-06-16
Quote `WASI` in `CMAKE_SYSTEM_NAME STREQUAL` checks

MaxDesiatov:
Quote the `WASI` operand in three `CMAKE_SYSTEM_NAME STREQUAL WASI` comparisons so that CMake compares against the literal string instead of dereferencing `WASI` as a variable when it is defined (recent CMake's `Platform/WASI-Initialize.cmake` sets `WASI` to 1), which otherwise inverts the WASI build guards.

This means that the build behaves correctly with both CMake 3 and CMake 4, which differ in their handling of these conditions:

- CMake 4.3.3: `Modules/Platform/WASI-Initialize.cmake` has `set(WASI 1)`.
- CMake 3.30.2 (used on CI): that file does not exist, and nothing under `Modules/Platform/` ever sets a `WASI` variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,7 +25,7 @@ project(SwiftTesting
   LANGUAGES CXX Swift)
 
 if(NOT APPLE)
-  if(NOT CMAKE_SYSTEM_NAME STREQUAL WASI)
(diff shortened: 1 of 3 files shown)


harlanhaskins (approved):
[old man yells at cmake]

PR #1759, opened 2026-06-17
Introduce `ABI.Context` and produce human-readable output from the JSON event stream.

grynspan:
This PR adds the ability to generate console output and human-readable text output from the JSON event stream in addition to native Swift Testing event values.

To assist in tracking information between event stream records, it introduces a new `ABI.Context` type that can be passed to the various `init(decoding:)` initializers. This type performs some necessary bookkeeping that cannot be performed on a record-by-record basis (e.g. mapping test IDs back to full test descriptions). It's a vague type by definition so that we can give it additional responsibilities in the future.

### Checklist:


Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -256,5 +256,27 @@ extension Event {
...
   }
+
+  /// Initialize an instance of this type from the given value.
+  ///
+  /// - Parameters:
+  ///   - event: The encoded event to initialize this instance from.
+  ///   - context: A context value that tracks decoded tests and events.
+  ///
+  /// ``testID`` and ``testCaseID`` are always `nil` because we need information
+  /// from the associated `ABI.EncodedTest` to properly decode those values.
```
stmontgomery: I understand how this applies to `testCaseID`, but I see `testID` being assigned under at least some circumstances below. Does this comment still apply to `testID`?

stmontgomery (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
grynspan: Just copied and pasted from the other overload. I'll adjust it.

grynspan:
Linux failures are due to a compiler bug.

PR #1764, opened 2026-06-18
TaskLocal test trait

mbrandonw:
Adds a `.taskLocal` test trait for overriding task locals in tests

Resolves rdar://140371736

### Motivation:

Proposal here: [ST-0026: TaskLocal test trait](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0026-task-local-test-trait.md)

### Checklist:

### DIFF
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -1,6 +1,20 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025–2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
 extension Trait {
-  /// Constructs a trait that overrides a task local value for the duration of a test
+  /// Constructs a trait that binds a task local value for the duration of a test
   /// or suite.
+  ///
+  /// - Parameters:
+  ///   - taskLocal: The task local to bind the value to.
+  ///   - value: The value to set.
   ///
   /// ```swift
   /// @Suite(.taskLocal($myValue, 42))
@@ -9,11 +23,7 @@ extension Trait {
   /// }
   /// ```
   ///
-  /// - Note: The task local must be defined outside the test target where the trait is used.
-  ///
-  /// - Parameters:
-  ///   - taskLocal: The task local to override.
-  ///   - value: The value to set.
+  /// - Note: You must define the task local outside the test target where the trait is used.
   public static func taskLocal<Value>(
     _ taskLocal: TaskLocal<Value>,
     _ value: Value
@@ -23,14 +33,15 @@ extension Trait {
   }
 }
 
-/// A type that that overrides a task local value for the scope of a test.
+/// A type that that binds a task local value for the duration of a test or suite.
 ///
 /// To add this trait to a test, use ``Trait/taskLocal(_:_:)``.
 public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestScoping, TestTrait {
-  public var isRecursive: Bool { true }
+  /// This trait's task local.
+  fileprivate var taskLocal: TaskLocal<Value>
 
-  fileprivate let taskLocal: TaskLocal<Value>
-  fileprivate let value: Value
+  /// This trait's value.
+  fileprivate var value: Value
 
   public func provideScope(
     for test: Test,
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -24,7 +24,7 @@ extension Trait {
   /// ```
   ///
   /// - Note: You must define the task local outside the test target where the trait is used.
-  public static func taskLocal<Value>(
+  public static func taskLocal<Value: Sendable>(
     _ taskLocal: TaskLocal<Value>,
     _ value: Value
   ) -> Self
@@ -36,7 +36,7 @@ extension Trait {
 /// A type that that binds a task local value for the duration of a test or suite.
 ///
 /// To add this trait to a test, use ``Trait/taskLocal(_:_:)``.
-public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestScoping, TestTrait {
+public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestTrait, TestScoping {
   /// This trait's task local.
   fileprivate var taskLocal: TaskLocal<Value>
 
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -14,14 +14,12 @@ extension Trait {
   ///
   /// - Parameters:
   ///   - taskLocal: The task local to bind the value to.
-  ///   - value: The value to set.
+  ///   - value: The value to bind to `taskLocal`.
   ///
-  /// ```swift
-  /// @Suite(.taskLocal($myValue, 42))
-  /// struct MyTests {
-  ///   // ...
-  /// }
-  /// ```
+  ///     This value will only be evaluated if the test this
+  ///     trait is applied to runs, and it will be unbound from
+  ///     the task local and released once the trait is finished
+  ///     providing scope for the test.
   ///
   /// - Note: You must define the task local outside the test target where the trait is used.
   public static func taskLocal<Value: Sendable>(
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -1,0 +1,42 @@
+extension Trait {
+  /// Constructs a trait that overrides a task local value for the duration of a test
+  /// or suite.
+  ///
+  /// ```swift
+  /// @Suite(.taskLocal($myValue, 42))
+  /// struct MyTests {
+  ///   // ...
+  /// }
+  /// ```
+  ///
+  /// - Note: The task local must be defined outside the test target where the trait is used.
+  ///
+  /// - Parameters:
+  ///   - taskLocal: The task local to override.
+  ///   - value: The value to set.
+  public static func taskLocal<Value>(
+    _ taskLocal: TaskLocal<Value>,
+    _ value: Value
+  ) -> Self
+  where Self == TaskLocalTrait<Value> {
+    TaskLocalTrait(taskLocal: taskLocal, value: value)
+  }
+}
+
+/// A type that that overrides a task local value for the scope of a test.
+///
+/// To add this trait to a test, use ``Trait/taskLocal(_:_:)``.
+public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestScoping, TestTrait {
+  public let isRecursive = true
+
+  fileprivate let taskLocal: TaskLocal<Value>
+  fileprivate let value: Value
+
+  public func provideScope(
+    for test: Test,
+    testCase: Test.Case?,
+    performing function: @concurrent () async throws -> Void
+  ) async throws {
+    try await taskLocal.withValue(value, operation: function)
+  }
+}
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -46,7 +46,7 @@ public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestTrait, TestScopin
   public func provideScope(
     for test: Test,
     testCase: Test.Case?,
-    performing function: @concurrent () async throws -> Void
+    performing function: @Sendable () async throws -> Void
   ) async throws {
     try await taskLocal.withValue(value()) {
       try await function()
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -27,7 +27,7 @@ extension Trait {
 ///
 /// To add this trait to a test, use ``Trait/taskLocal(_:_:)``.
 public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestScoping, TestTrait {
-  public let isRecursive = true
+  public var isRecursive: Bool { true }
 
   fileprivate let taskLocal: TaskLocal<Value>
   fileprivate let value: Value
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -1,7 +1,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2025–2026 Apple Inc. and the Swift project authors
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
@@ -51,3 +51,7 @@ public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestScoping, TestTrai
     try await taskLocal.withValue(value, operation: function)
   }
 }
+
+#if DEBUG
+  @TaskLocal var dummyLocal = false
+#endif
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -26,7 +26,7 @@ extension Trait {
   /// - Note: You must define the task local outside the test target where the trait is used.
   public static func taskLocal<Value: Sendable>(
     _ taskLocal: TaskLocal<Value>,
-    _ value: Value
+    _ value: @autoclosure @escaping @Sendable () throws -> Value
   ) -> Self
   where Self == TaskLocalTrait<Value> {
     TaskLocalTrait(taskLocal: taskLocal, value: value)
@@ -41,14 +41,14 @@ public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestTrait, TestScopin
   fileprivate var taskLocal: TaskLocal<Value>
 
   /// This trait's value.
-  fileprivate var value: Value
+  fileprivate var value: @Sendable () throws -> Value
 
   public func provideScope(
     for test: Test,
     testCase: Test.Case?,
     performing function: @concurrent () async throws -> Void
   ) async throws {
-    try await taskLocal.withValue(value, operation: function)
+    try await taskLocal.withValue(value(), operation: function)
   }
 }
 
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -43,15 +43,15 @@ public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestTrait, TestScopin
   /// This trait's value.
   fileprivate var value: @Sendable () throws -> Value
 
+  public var isRecursive: Bool { true }
+
   public func provideScope(
     for test: Test,
     testCase: Test.Case?,
     performing function: @concurrent () async throws -> Void
   ) async throws {
-    try await taskLocal.withValue(value(), operation: function)
+    try await taskLocal.withValue(value()) {
+      try await function()
+    }
   }
 }
-
-#if DEBUG
-  @TaskLocal var dummyLocal = false
-#endif
--- a/Sources/Testing/Traits/TaskLocalTrait.swift
+++ b/Sources/Testing/Traits/TaskLocalTrait.swift
@@ -24,7 +24,7 @@ extension Trait {
   /// - Note: You must define the task local outside the test target where the trait is used.
   public static func taskLocal<Value: Sendable>(
     _ taskLocal: TaskLocal<Value>,
-    _ value: @autoclosure @escaping @Sendable () throws -> Value
+    withValue value: @autoclosure @escaping @Sendable () throws -> Value
   ) -> Self
   where Self == TaskLocalTrait<Value> {
     TaskLocalTrait(taskLocal: taskLocal, value: value)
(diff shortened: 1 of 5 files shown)


Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,42 @@
...
+  /// - Note: The task local must be defined outside the test target where the trait is used.
+  ///
+  /// - Parameters:
+  ///   - taskLocal: The task local to override.
+  ///   - value: The value to set.
+  public static func taskLocal<Value>(
+    _ taskLocal: TaskLocal<Value>,
+    _ value: Value
+  ) -> Self
+  where Self == TaskLocalTrait<Value> {
```
grynspan, line 29: It doesn't make a practical difference, but in existing traits we put the `where` clause on the extension rather than the function. This reduces repetition if we end up with multiple overloads.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,42 @@
...
+  /// or suite.
+  ///
+  /// ```swift
+  /// @Suite(.taskLocal($myValue, 42))
+  /// struct MyTests {
+  ///   // ...
+  /// }
+  /// ```
+  ///
+  /// - Note: The task local must be defined outside the test target where the trait is used.
```
grynspan: Documentation comments should avoid the passive voice per Apple's and Swift's documentation style guides.

Per our own convention, put discussion including callouts _after_ the brief, parameters, returns, and throws sections.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,42 @@
+extension Trait {
+  /// Constructs a trait that overrides a task local value for the duration of a test
+  /// or suite.
+  ///
+  /// ```swift
```
grynspan: An example like this is probably not necessary, but if you want to include it then it needs to go after a paragraph or two describing the trait in more detail. It won't render well in DocC immediately after the brief.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,42 @@
...
+  where Self == TaskLocalTrait<Value> {
+    TaskLocalTrait(taskLocal: taskLocal, value: value)
+  }
+}
+
+/// A type that that overrides a task local value for the scope of a test.
+///
+/// To add this trait to a test, use ``Trait/taskLocal(_:_:)``.
+public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestScoping, TestTrait {
+  public var isRecursive: Bool { true }
```
grynspan: Because task locals are already implicitly applied recursively to subtasks, there's no need for this property's value to be `true`. Having it `true` here just causes us to call `withValue` repeatedly, which does not have a meaningful effect on test code.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,42 @@
+extension Trait {
```
grynspan, line 11: This file needs the standard copyright header used across all Swift source files in the Swift OSS project. Feel free to copy and paste it from another file.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,42 @@
...
+  }
+}
+
+/// A type that that overrides a task local value for the scope of a test.
+///
+/// To add this trait to a test, use ``Trait/taskLocal(_:_:)``.
+public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestScoping, TestTrait {
+  public var isRecursive: Bool { true }
+
+  fileprivate let taskLocal: TaskLocal<Value>
```
grynspan: Please provide documentation comments for these two stored properties. As well, because this is a `struct`, please declare them `var` rather than `let`.

grynspan:
Additional action items:

- [x] Sources/Testing/CMakeLists.txt needs to be updated to reference the new file.
- [x] Some test coverage is needed (we can have a dummy task local in the Testing target for this purpose, guarded with `#if DEBUG`).

grynspan (changes requested):
Off to a great start.

grynspan:
GitHub's multiple assignee feature is broken, I guess.

Sources/Testing/Traits/TaskLocalTrait.swift
stephencelis: There's a generic involved here, so unfortunately that's not an option without generic extensions.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,57 @@
...
+  /// @Suite(.taskLocal($myValue, 42))
+  /// struct MyTests {
+  ///   // ...
+  /// }
+  /// ```
+  ///
+  /// - Note: You must define the task local outside the test target where the trait is used.
+  public static func taskLocal<Value: Sendable>(
+    _ taskLocal: TaskLocal<Value>,
+    _ value: Value
```
stmontgomery: I recommend making the `value` parameter an `@autoclosure` so that its evaluation can be deferred until the point where `provideScope()` is called, _if_ it gets called at all:

```suggestion
    _ value: @autoclosure @escaping @Sendable () throws -> Value
```

A couple reasons:

- If the test is skipped (e.g. it has a `.disabled` trait) the `value` expression will never be evaluated. That can be useful if the evaluation has observable side effects you only want to occur if the test actually runs, especially if that work is expensive.
- The value obtained lazily during `provideScope()` can be released and discarded once the test finishes, which again can be important if the value occupies significant memory or otherwise needs to be deallocated once it's no longer needed.

Making this a lazily-evaluated stored property on `TaskLocalTrait` would also align it more closely with `ConditionTrait`, which stores its condition as a closure which gets evaluated lazily during planning.

If you take this change, I recommend documenting those semantics in the DocC for the `.taskLocal()` function, and discussing the reasoning for using `@autoclosure` in the accompanying proposal.

grynspan:
Keep in mind async autoclosures are broken.

Sources/Testing/Traits/TaskLocalTrait.swift
mbrandonw: > If you take this change, I recommend documenting those semantics in the DocC for the .taskLocal() function, and discussing the reasoning for using @autoclosure in the accompanying proposal.

I've made the change for `@autoclosure`, but on this point I'm not sure what to document. I couldn't find any examples in the Swift ecosystem of autoclosures that are documented. Usually the docs treat the argument as a regular value and do not mention the autoclosure at all, including `enabled(if:)`/`disabled(if:)`.

Sources/Testing/Traits/TaskLocalTrait.swift
grynspan: Oh right, good point.

Sources/Testing/Traits/TaskLocalTrait.swift
stmontgomery: The aspect I would encourage you to document is the lifetime guarantees that lazy evaluation via `@autoclosure` will provide. You could say something like,

```swift
/// - Parameters:
///   ...
///   - value: The value to bind to `taskLocal`.
///
///     This value will only be evaluated if the test this
//      trait is applied to runs, and it will be unbound from
///     the task local and released once the trait is finished
///     providing scope for the test.
```

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,57 @@
...
+  /// This trait's task local.
+  fileprivate var taskLocal: TaskLocal<Value>
+
+  /// This trait's value.
+  fileprivate var value: @Sendable () throws -> Value
+
+  public func provideScope(
+    for test: Test,
+    testCase: Test.Case?,
+    performing function: @concurrent () async throws -> Void
```
stmontgomery: We don't have `@concurrent` on other `provideScope()` implementations in the library. The other implementations do have `@Sendable` though; can we get away with only `@Sendable` ?

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,57 @@
...
+  ) async throws {
+    try await taskLocal.withValue(value(), operation: function)
+  }
+}
+
+#if DEBUG
+  @TaskLocal var dummyLocal = false
```
stmontgomery: Could you include a DocC comment about why this is here? (It's a convenient place to put it for use in an accompanying test, because it needs to be in a separate module)

Also, style nit: we generally outdent code within top-level `#if`s to column 0

Sources/Testing/Traits/TaskLocalTrait.swift
stmontgomery: This is an interesting question to carefully consider, I think! I feel like it'll be common for users to have situations where it's important that the value be reevaluated and re-bound to the task local for every invocation of each test within a suite that a `.taskLocal(...)` trait is applied to. For example, if the value contains some state which accumulates or is modified during each test, it's very important that be reset to a fresh state for each one.

For that reason **I think the safest _default_ for `isRecursive` is `true`**, as currently written. Per the semantics of Test Scoping Traits, that will mean that whenever you add one of these traits to a suite, `value` will be reevaluated (by calling its `@autoclosure`) and (re-)bound to the task local for the invocation of each test within that suite

That said, there could totally be situations when a user _doesn't_ want this work to be redone for each invocation of each test in a suite, they only want the work to happen once per suite. If we want to support that, we could make `isRecursive` a parameter of the `.taskLocal(...)` function, with a default value of `true`, and let it be controlled that way. I'm not certain we need to allow that in the initial version of this API, though; I'd lean towards leaving it as a Future Direction. And either way, I think it's worth discussing this consideration in the proposal.

Sources/Testing/Traits/TaskLocalTrait.swift
mbrandonw: Instead of commenting on this dummy local, what if I move the task local into the test target and don't use the `@TaskLocal` macro? That way I can test it without running into the compiler limitation.

mbrandonw: I just pushed b80766b to show how this could look with defining the task local (without the macro) directly in the test suite. I think it's a bit better since it can not be private.

Sources/Testing/Traits/TaskLocalTrait.swift
mbrandonw: We agree that `isRecursive: true` is the better default and have pushed those changes here b80766b. I also added a test to show the danger of not making it recursive.

Sources/Testing/Traits/TaskLocalTrait.swift
mbrandonw: I think this is `@concurrent` only because it was taken from a library we maintain with this tool, and that library is `NonisolatedNonSendingByDefault` and Swift Testing is not (I believe). Will remove and bring back `@Sendable` as that's the `TestScoping` requirement.

Sources/Testing/Traits/TaskLocalTrait.swift
stmontgomery: Works for me!

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,55 @@
...
+  ) -> Self
+  where Self == TaskLocalTrait<Value> {
+    TaskLocalTrait(taskLocal: taskLocal, value: value)
+  }
+}
+
+/// A type that that binds a task local value for the duration of a test or suite.
+///
+/// To add this trait to a test, use ``Trait/taskLocal(_:_:)``.
+public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestTrait, TestScoping {
```
stmontgomery: In the proposal there's a method `evaluate()`, don't forget to add that to the implementation here:

```swift
  /// Evaluate this trait's bound value.
  public func evaluate() async throws -> Value
```

stmontgomery: And preferably, add a unit test of that method

stmontgomery: @stephencelis mind adding a unit test for `evaluate()` too, just so that it at least has one caller?

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,60 @@
...
+  ///   - taskLocal: The task local to bind the value to.
+  ///   - value: The value to bind to `taskLocal`.
+  ///
+  ///     This value will only be evaluated if the test this
+  ///     trait is applied to runs, and it will be unbound from
+  ///     the task local and released once the trait is finished
+  ///     providing scope for the test.
+  ///
+  /// - Note: You must define the task local outside the test target where the trait is used.
+  public static func taskLocal<Value: Sendable>(
```
grynspan: House style is to always use `where` clauses for generic constraints.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,60 @@
...
+    _ taskLocal: TaskLocal<Value>,
+    withValue value: @autoclosure @escaping @Sendable () throws -> Value
+  ) -> Self
+  where Self == TaskLocalTrait<Value> {
```
grynspan:
```suggestion
  ) -> Self where Self == TaskLocalTrait<Value> {
```

Nitpick, more closely matches existing style.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,60 @@
...
+  /// - Parameters:
+  ///   - taskLocal: The task local to bind the value to.
+  ///   - value: The value to bind to `taskLocal`.
+  ///
+  ///     This value will only be evaluated if the test this
+  ///     trait is applied to runs, and it will be unbound from
+  ///     the task local and released once the trait is finished
+  ///     providing scope for the test.
+  ///
+  /// - Note: You must define the task local outside the test target where the trait is used.
```
grynspan: Wrap comments at 80 columns please.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,60 @@
...
+  ) -> Self
+  where Self == TaskLocalTrait<Value> {
+    TaskLocalTrait(taskLocal: taskLocal, value: value)
+  }
+}
+
+/// A type that that binds a task local value for the duration of a test or suite.
+///
+/// To add this trait to a test, use ``Trait/taskLocal(_:_:)``.
+public struct TaskLocalTrait<Value: Sendable>: SuiteTrait, TestTrait, TestScoping {
```
grynspan: House style is to always use `where` clauses for generic constraints.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,60 @@
...
+  /// This trait's task local.
+  public var taskLocal: TaskLocal<Value>
+
+  /// This trait's value.
+  fileprivate var value: @Sendable () throws -> Value
+
+  /// Evaluate this trait's bound value.
```
grynspan, line 39: Please add a `- Returns:` callout.

grynspan: It would also be worth noting that the result is not cached.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,60 @@
...
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+extension Trait {
```
grynspan: Not critical, but for other traits we've put the `Trait` factory functions after the type. Minor style nitpick, feel free to ignore.

grynspan (changes requested):

Sources/Testing/Traits/TaskLocalTrait.swift
stmontgomery: Added `- Returns:`, `- Throws:`, and a note about not storing the result

stmontgomery (approved):

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+
+// MARK: - TestScoping
+
+/// @Metadata {
+///   @Available(Swift, introduced: 6.5)
+/// }
+extension TaskLocalTrait: TestScoping {
+  public func provideScope(
```
grynspan, line 73: Needs `@Metadata`.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,107 @@
...
+  ) async throws {
+    try await taskLocal.withValue(evaluate()) {
+      try await function()
+    }
+  }
+}
+
+// MARK: -
+
+extension Trait {
```
grynspan, line 86: Needs `@Metadata`.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,107 @@
...
+  /// - Note: You must define the task local outside the test target where the
+  ///   trait is used.
+  ///
+  /// @Metadata {
+  ///   @Available(Swift, introduced: 6.5)
+  /// }
+  public static func taskLocal<Value>(
+    _ taskLocal: TaskLocal<Value>,
+    withValue value: @autoclosure @escaping @Sendable () throws -> Value
+  ) -> Self where Value: Sendable, Self == TaskLocalTrait<Value> {
```
grynspan: Don't need to restate `where Value: Sendable` here, it's implied already.

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,107 @@
...
+  /// - Parameters:
+  ///   - taskLocal: The task local to bind the value to.
+  ///   - value: The value to bind to `taskLocal`.
+  ///
+  ///     This value will only be evaluated if the test this
+  ///     trait is applied to runs, and it will be unbound from
+  ///     the task local and released once the trait is finished
+  ///     providing scope for the test.
+  ///
+  /// - Note: You must define the task local outside the test target where the
```
grynspan:
```suggestion
  /// - Note: You must define `taskLocal` outside the test target where the
```

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,107 @@
...
+  /// test or suite.
+  ///
+  /// - Parameters:
+  ///   - taskLocal: The task local to bind the value to.
+  ///   - value: The value to bind to `taskLocal`.
+  ///
+  ///     This value will only be evaluated if the test this
+  ///     trait is applied to runs, and it will be unbound from
+  ///     the task local and released once the trait is finished
+  ///     providing scope for the test.
```
grynspan: Passive voice and future tense. How about:

```suggestion
  ///     The testing library evaluates this value when the test this trait is applied to runs. The value is bound to `taskLocal`, the test runs, and then the value is unbound.
```

"[P]roviding scope for the test" is maybe too jargony. "binding" in this context is a term of art but I'm not sure if devs are familiar with it, so we might want to spell it some other way.

Sources/Testing/Traits/TaskLocalTrait.swift
grynspan: Also passive voice, try "outside the test target where you use this trait"?

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,107 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// A type that that binds a task local value for the duration of a test or
+/// suite.
+///
+/// When an instance of this trait is applied to a suite, it is recursively
+/// inherited by all child suites and tests.
```
grynspan: Passive voice, and I've been trying to be consistent with terminology (always "test suite" and "test function" since `Test` could be either).

```suggestion
/// When you apply an instance of this trait to a test suite, the testing library recursively applies it to all test suites and test functions within it.
```

Sources/Testing/Traits/TaskLocalTrait.swift
```diff
@@ -0,0 +1,107 @@
...
+///
+/// To add this trait to a test, use ``Trait/taskLocal(_:withValue:)``.
+///
+/// @Metadata {
+///   @Available(Swift, introduced: 6.5)
+/// }
+public struct TaskLocalTrait<Value>: SuiteTrait, TestTrait where Value: Sendable {
+  /// This trait's task local.
```
grynspan: This trait's task-local value? Not sure if "a task local" is grammatical.

Sources/Testing/Traits/TaskLocalTrait.swift
mbrandonw: Fine with either way but we previously received the opposite feedback. That it was preferable to restate the conformance even though it is implicit from `TaskLocal<Value>`.

grynspan: Feel free to disregard. (It has no effect on the compiled product.)

Sources/Testing/Traits/TaskLocalTrait.swift
stmontgomery: Yeah, this is interesting. The abstract of the [`TaskLocal` API documentation](https://developer.apple.com/documentation/swift/tasklocal) is as follows (**emphasis** mine):

> Wrapper type that defines a **task-local value key**.

I think the word "key" there may be what we need; the `TaskLocal` itself isn't a _value_, it's the _key_ used to access the task-local value. Based on that, I'm going with this edit:
```suggestion
  /// This trait's task-local value key.
```

stmontgomery (approved):

PR #1769, opened 2026-06-22
Adopt `objc_copyImageHeaders()` for safer image discovery on Darwin (take 3)

grynspan:

`objc_copyImageHeaders()` is forward-declared as CI doesn't have new-enough SDKs yet.

We must maintain code paths for earlier Apple OSes as well as Apple configurations that disable Objective-C interop. For earlier Apple OSes, we use a circuitous path through `objc_copyImageNames()` and `dlopen()`. For no-Objective-C configurations, which are not customer-facing, dyld has lower-level API we can call.

### Checklist:


grynspan:
@finagolfin @artemcm FYI

finagolfin:
We should probably have a CMake run on the macOS github runners here, to flush out these issues on every pull.

grynspan:
AFAIK I've kicked off everything, but if there's something I've missed please do elaborate.

finagolfin:
What I mean is that we should add a build-script macOS workflow for this repo on GitHub, so you won't have to do these cross-repo runs manually after that.

stmontgomery (approved):

PR #1775, opened 2026-06-26
Add inert generic parameter to TestIssueRecorder

harlanhaskins:
This adds an inert generic parameter to TestIssueRecorder that's always specialized with `Void` to avoid conflicts when loading the package version of Testing (when there is also an SDK-level copy present).

### Motivation:

This is following the same pattern as `Serializer`.

### Modifications:

Adds an inert generic parameter to TestIssueRecorder and instantiates it with `Void`.

### Checklist:

Fixes rdar://179023660


stmontgomery (approved):

harlanhaskins:
CI failures are all the same crash we've been hitting

PR #1779, opened 2026-07-08
Clarify README parameterized-testing example

### LINKED ISSUE #1778: Improve the example test(s) in README.md
### Description

The example [here](https://developer.apple.com/xcode/swift-testing/) that is supposed to demonstrate its “Clear Expressive API” is confusing as heck.

```swift
@Test("Continents mentioned in videos", arguments: [
    "A Beach",
    "By the Lake",
    "Camping in the Woods"
])
func mentionedContinents(videoName: String) async throws {
    let videoLibrary = try await VideoLibrary()
    let video = try #require(await videoLibrary.video(named: videoName))
    #expect(video.mentionedContinents.count <= 3)
}
``` 

What role does `arguments` play?  After a long while, I guess they are getting passed as video names. If so, active names like “Fonzy Jumps the Shark” would help clarify that.  Lastly, are we really testing that there are fewer than 4 continents mentioned in each video?  That's an incredibly implausible test, but if you want to use it, the name of the test should be `“3 or fewer continents mentioned in videos"`.  Otherwise it sounds like its testing that continents are mentioned in videos, which it does not do! A count of zero would pass the test.

### Reproduction

```swift

```

### Expected behavior

Example that demonstrates the clarity of the API

### Environment

Swift Package Manager - Swift 6.3.3

### Additional information

_No response_

leticiamazzoportela:
Clarify the parameterized-testing example in the README.

### Motivation:

Fixes #1778. The example under "Scalable coverage and execution" is meant to showcase how clear and expressive Swift Testing's parameterized tests are, but two things worked against that:
- The test's display name ("Continents mentioned in videos") didn't match what the assertion actually checks: a video mentioning zero continents would still pass, which contradicts what the name implies.
- It wasn't obvious how the `arguments` array maps to the `videoName` parameter on the function below it.

_Note: the same snippet also appears on [developer.apple.com/xcode/swift-testing/](https://developer.apple.com/xcode/swift-testing/#:~:text=Clear%2C%20expressive%20API), but that page isn't part of this repo, so this PR only updates the README._

### Modifications:

- Renamed the test to "Each video mentions 3 or fewer continents", matching the assertion it makes.
- Added a short comment tying the `arguments` array to the `videoName` parameter.

### Checklist:

- [x] If public symbols are renamed or modified, DocC references should be updated. *(n/a — no public symbols changed)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,11 +75,11 @@ can write less code. And all tests integrate seamlessly with Swift Concurrency
 and run in parallel by default.
 
 ```swift
-@Test("Continents mentioned in videos", arguments: [
+@Test("Each video mentions 3 or fewer continents", arguments: [
     "A Beach",
     "By the Lake",
     "Camping in the Woods"
-])
+]) // Runs once per video name above, passing it as `videoName` each time
 func mentionedContinents(videoName: String) async throws {
     let videoLibrary = try await VideoLibrary()
     let video = try #require(await videoLibrary.video(named: videoName))


grynspan:
Thanks for the PR. I'm closing it as it is not a substantive change; likely the real fix for Dave's issue is to rework the readme file's overall content to be clearer. Let's discuss further in #1778 before opening any new PRs. Thanks again!

PR #1781, opened 2026-07-09
Bootstrap a testing harness target.

grynspan:
This PR adds a (stub, unimplemented, incomplete, _etc._) harness executable target that uses Swift Argument Parser. This target will eventually act as a host for tests, but for now just throws an error if invoked.

### Checklist:


Sources/Harness/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+# See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+if(NOT (CMAKE_SYSTEM_NAME IN_LIST SWT_NO_HARNESS_LIST))
+  include(ModuleABIName)
+
+  find_package(ArgumentParser CONFIG REQUIRED)
+
+  add_executable(swift-testing-harness
+    Harness.swift)
+  target_link_libraries(swift-test PRIVATE
```
stmontgomery:
```suggestion
  target_link_libraries(swift-testing-harness PRIVATE
```

Sources/Harness/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+  find_package(ArgumentParser CONFIG REQUIRED)
+
+  add_executable(swift-testing-harness
+    Harness.swift)
+  target_link_libraries(swift-testing-harness PRIVATE
+    ArgumentParser
+    Testing)
+
+  install(TARGETS swift-testing-harness
+    RUNTIME DESTINATION "libexec/swift/pm")
```
stmontgomery: Placing this in the `pm` subdirectory to me would imply that it's solely/primarily intended for use by SwiftPM, and I think this component should be a bit less closely coupled than that. Could we place it in the parent directory, `libexec/swift` ?

grynspan: I am unsure. For the moment I'm aping `swiftpm-testing-helper`.

stmontgomery: That one really is oriented as a helper for SwiftPM, specifically, whereas this is intended to be broader from the outset.

finagolfin:
Oh nice, how would this work for cross-compilation: this new harness runs on the host and communicates to the current test runners on a separate target platform over the network, or this harness runs on the target platform itself? Maybe you don't intend it for cross-compilation altogether, just seeing how we can use it.

grynspan:
@finagolfin Short answer: it should probably be possible to implement cross-device testing for Android with this tool, though we'll need your help to do it.

The exact feature set is currently unspecified; this PR just creates the new target. My initial focus will be on making it work on-device, but it will be leveraging our JSON event stream and there's no reason it couldn't draw events from off-device sources given the right application of elbow grease.

Sources/Harness/CMakeLists.txt
grynspan: Moved to `libexec/swift/testing`.

stmontgomery:
@finagolfin I plan to post a proposal on the forums fairly soon describing the anticipated longer-term trajectory here, this is just a starting point as @grynspan mentioned. It's definitely something I hope can eventually support running tests on remotely-connected devices though, and it's great to hear your interest in that.

stmontgomery (approved):

PR #1782, opened 2026-07-09
Add `FileHandle.read(until:)` to avoid blocking on slow exit test events.

grynspan:
If an exit test runs for a long time, we necessarily block until it terminates before the calling task resumes. We also wait until the end of the exit test to consume any events from its event stream (by calling `FileHandle.readToEnd()` and thus waiting for the write end of the pipe to close).

This PR breaks up reading from the exit test's event stream so that we read one line at a time, then yield before processing the event. The result is that we process events sooner after they occur and we allow other tasks to use the thread we're camped on at least for a little bit of time.

I'll also be able to use `read(until:)` in the harness.

### Checklist:


stmontgomery (approved):

grynspan:
Linux failures are known issues.

PR #1785, opened 2026-07-10
Avoid using members of extensions are unavailable when image attachments are unsupported

tshortli:
On certain platforms, code must be compiled out code when it uses declarations marked `@available(*, unavailable)` because it is unsupported on those platforms.

### Motivation:

The intended behavior of the Swift compiler is that non-type declarations marked `@available(*, unavailable)` are always diagnosed as unavailable, even when used in contexts that are also unavailable. However, there was a regression which caused the compiler to fail to diagnose these declarations when they are members of an extension that is marked unavailable, rather than being marked unavailable directly.

Swift Testing has source code that was inadvertently taking advantage of this regression which needs to change in order to unblock the compiler fix (https://github.com/swiftlang/swift/pull/90588).

### Modifications:

Compile out some code that would access members of universally unavailable extensions.

### Checklist:


grynspan (approved):

grynspan:
Feel free to squash merge at your leisure. Ignore the iOS and jammy CI failures.

stmontgomery (approved):
Thanks!

PR #1787, opened 2026-07-13
Add overload of `#expect(processExitsWith:)` to help the type checker.

### LINKED ISSUE #1786: exit tests seem to force `async` programs
### Description

It appears like only `async` exit tests are possible. But in `async` functions certain blocking functions are unavailable.

But I would like to test some of them. Currently, I'm working around that with a

```
await #expect(...) {
    func wrapper() {
        doNoAsyncStuffHere()
    }
    wrapper()
}
```

but it would be nicer if Swift Testing allowed me to use non-`async` blocking code directly in exit tests.

### Reproduction

```swift
  func eventLoopScheduleAfterShutdown() async {
    let result = await #expect(processExitsWith: .failure, observing: [\.standardErrorContent]) {
        let elg = MultiThreadedEventLoopGroup(numberOfThreads: 1)                                           
         let loop = elg.next()                                                                               
         try! elg.syncShutdownGracefully() // disallowed in async code                                                                   
[...]
    }
}
```

### Expected behavior

there's a way to make this work without workaround

### Environment

Swift 6.4 beta

### Additional information

_No response_

grynspan:
This PR adds overloads of `#expect(processExitsWith:)` and `#require(processExitsWith:)` that take _synchronous_ body closures. These overloads behave identically to the existing ones, but keep the type checker from promoting the closures to `async`. This then allows exit test bodies that are not explicitly `async` to call functions marked `@available(*, noasync)` without diagnosing.

> [!NOTE]
> Whether or not `noasync` functions are safe to call from a particular synchronous context is not known at our layer, and may be undecidable in the general case anyway. However, the language currently permits you to call such a function on any task so long as the immediate caller is not `async`, and we're just preserving that behaviour, not adding new behaviour.

Resolves #1786.

### Checklist:

### DIFF
--- a/Sources/Testing/Expectations/Expectation+Macro.swift
+++ b/Sources/Testing/Expectations/Expectation+Macro.swift
@@ -888,6 +888,52 @@ public macro expect(
   performing expression: @escaping @Sendable () async throws -> Void
 ) -> ExitTest.Result? = #externalMacro(module: "TestingMacros", type: "ExitTestExpectMacro")
 
+/// Check that an expression causes the process to terminate in a given fashion.
+///
+/// - Parameters:
+///   - expectedExitCondition: The expected exit condition.
+///   - observedValues: An array of key paths representing results from within
+///     the exit test that should be observed and returned by this macro. The
+///     ``ExitTest/Result/exitStatus`` property is always returned.
+///   - comment: A comment describing the expectation.
+///   - sourceLocation: The source location to which recorded expectations and
+///     issues should be attributed.
+///   - expression: The expression to be evaluated.
+///
+/// - Returns: If the exit test passes, an instance of ``ExitTest/Result``
+///   describing the state of the exit test when it exited. If the exit test
+///   fails, the result is `nil`.
+///
+/// Use this overload of `#expect()` when an expression will cause the current
+/// process to terminate and the nature of that termination will determine if
+/// the test passes or fails. For example, to test that calling `fatalError()`
+/// causes a process to terminate:
+///
+/// ```swift
+/// await #expect(processExitsWith: .failure) {
+///   fatalError()
+/// }
+/// ```
+///
+/// @Metadata {
+///   @Available(Swift, introduced: 6.2)
+///   @Available(Xcode, introduced: 26.0)
+/// }
+@freestanding(expression)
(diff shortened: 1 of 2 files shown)


Sources/Testing/Expectations/Expectation+Macro.swift
```diff
@@ -888,6 +888,52 @@ public macro expect(
...
+/// ```swift
+/// await #expect(processExitsWith: .failure) {
+///   fatalError()
+/// }
+/// ```
+///
+/// @Metadata {
+///   @Available(Swift, introduced: 6.2)
+///   @Available(Xcode, introduced: 26.0)
+/// }
```
stmontgomery, line 921: `@Available` directives should be adjusted

Sources/Testing/Expectations/Expectation+Macro.swift
```diff
@@ -888,6 +888,52 @@ public macro expect(
...
+@freestanding(expression)
+@_documentation(visibility: private)
+@discardableResult
+#if SWT_NO_EXIT_TESTS
+@_unavailableInEmbedded
+@available(*, unavailable, message: "Exit tests are not available on this platform.")
+#endif
+public macro expect(
```
stmontgomery, line 929: Could include a comment (either DocC or regular) explaining the purpose of these overloads?

Sources/Testing/Expectations/Expectation+Macro.swift
grynspan: Not really. The macros aren't new, we're just helping the type checker a bit. It was valid to pass a synchronous closure previously, so having newer availability would suggest it _wasn't_.

Sources/Testing/Expectations/Expectation+Macro.swift
grynspan: No, because the DocC content here is only visible in Xcode autocomplete and needs to match the "real" overload exactly to avoid developer confusion.

stmontgomery: Ok. I did say "or regular", leaving the door open to a `//`-style comment though.

stmontgomery (approved):

PR #1791, opened 2026-07-14
`TypeInfo` needs to handle generic clauses.

grynspan:
`TypeInfo` can be used with types that have generic clauses. Currently, we can misparse these things. Consider:

```swift
let a = [0, 1, 2]
let b = [1, 2, 3]
```

`a` and `b` are of type `Swift.Array<Swift.Int>`, however the `TypeInfo` value produced for `a` and `b` will report a `fullyQualifiedNameComponents` value of `["Swift", "Array<Swift", "Int>"]` instead of `["Swift", "Array<Swift.Int>"]`.

This PR resolves that issue. We aren't trying to reimplement swift-syntax here, but the new code handles a common case more accurately.

### Checklist:


harlanhaskins (approved):
Nice, thanks as well for the extra comments

PR #1792, opened 2026-07-14
Remove underscore from `iteration` in JSON record for encoded events

harlanhaskins:
I missed this in the implementation of ST-0024, `iteration` on the encoded event needs to be non-underscored.

Fixes rdar://182236523

### Motivation:

Implements the ST-0024 spec.

### Modifications:

Just removes the underscore and adds a regression test.

### Checklist:

### DIFF
--- a/Documentation/ABI/JSON.md
+++ b/Documentation/ABI/JSON.md
@@ -205,6 +205,8 @@ sufficient information to display the event in a human-readable format.
   ["attachment": <attachment>,] ; the attachment (if kind is "valueAttached")
   "messages": <array:message>,
   ["testID": <test-id>,]
+  ["iteration": <number>,] ; the iteration number (if "kind" is
+      ; "testStarted", "testCaseStarted, "testEnded", or "testCaseEnded")
 }
 
 <event-kind> ::= "runStarted" | "testStarted" | "testCaseStarted" |
@@ -249,3 +251,4 @@ sufficient information to display the event in a human-readable format.
 | [ST-0016](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0016-test-cancellation.md#integration-with-supporting-tools) | Added test cancellation. | 6.3 | `"6.3"` |
 | [ST-0019](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0019-include-tags-bugs-and-timeline-in-event-stream.md#json-schema-changes) | Added `tags`, `bugs`, and `timeLimit`. | 6.4 | `"6.4"` |
 | [ST-0020](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0020-sourcelocation-filepath.md#detailed-design) | Added `filePath`. | 6.3 | `"6.3"` |
+| [ST-0024](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0024-per-test-case-repetitions.md)| Added `iteration`. | 6.4 | "6.4" |
--- a/Documentation/ABI/JSON.md
+++ b/Documentation/ABI/JSON.md
@@ -205,8 +205,8 @@ sufficient information to display the event in a human-readable format.
   ["attachment": <attachment>,] ; the attachment (if kind is "valueAttached")
   "messages": <array:message>,
   ["testID": <test-id>,]
-  ["iteration": <number>,] ; the iteration number (if "kind" is
-      ; "testStarted", "testCaseStarted, "testEnded", or "testCaseEnded")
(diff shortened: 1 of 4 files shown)


stmontgomery (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -146,9 +146,11 @@ extension ABI {
         } else {
           kind = .testCaseEnded
         }
+        iteration = eventContext.iteration
       case .testCaseCancelled:
         kind = .testCaseCancelled
       case .testEnded:
+        iteration = eventContext.iteration
```
stmontgomery: Let's only include the new field when using v6.4 or later. Even though it's an additive field, including it by default in older event stream versions may permit clients who still use those versions to begin using it without first opting-in to the broader 6.4 schema and semantics which could lead to unexpected behaviors.

In other words, even for fields which are purely additive, I think should err on the side of caution and always require clients to explicitly opt-in to the newer ABI version to receive events containing those fields.

stmontgomery (changes requested):

Documentation/ABI/JSON.md
```diff
@@ -249,3 +251,4 @@ sufficient information to display the event in a human-readable format.
 | [ST-0016](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0016-test-cancellation.md#integration-with-supporting-tools) | Added test cancellation. | 6.3 | `"6.3"` |
 | [ST-0019](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0019-include-tags-bugs-and-timeline-in-event-stream.md#json-schema-changes) | Added `tags`, `bugs`, and `timeLimit`. | 6.4 | `"6.4"` |
 | [ST-0020](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0020-sourcelocation-filepath.md#detailed-design) | Added `filePath`. | 6.3 | `"6.3"` |
+| [ST-0024](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0024-per-test-case-repetitions.md)| Added `iteration`. | 6.4 | "6.4" |
```
grynspan: Thank you!

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -163,6 +161,17 @@ extension ABI {
       self.messages = messages.map(EncodedMessage.init)
       testID = event.testID.map(EncodedTest.ID.init)
 
+      // Fields introduced in 6.4
+
+      if V.versionNumber >= ABI.v6_4.versionNumber {
+        switch event.kind {
+        case .testStarted, .testCaseStarted, .testEnded, .testCaseEnded:
+          iteration = eventContext.iteration
```
grynspan: Do we need this value for testSkipped, testCancelled, or testCaseCancelled?

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -171,14 +180,11 @@ extension ABI {
...
         case let .valueAttached(attachment):
           _sourceLocation = EncodedSourceLocation<V>(encoding: attachment.sourceLocation)
-        case .testCaseStarted, .testCaseEnded, .testStarted, .testEnded:
-          _iteration = eventContext.iteration
         case let .testCaseCancelled(skipInfo),
           let .testSkipped(skipInfo),
           let .testCancelled(skipInfo):
           _comments = Array(skipInfo.comment).map(\.rawValue)
           _sourceLocation = skipInfo.sourceLocation.map { EncodedSourceLocation(encoding: $0) }
-          _iteration = eventContext.iteration
```
grynspan, line 181: Merp.

grynspan (changes requested):

Documentation/ABI/JSON.md
stmontgomery: nitpicky formatting
```suggestion
| [ST-0024](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0024-per-test-case-repetitions.md) | Added `iteration`. | 6.4 | `"6.4"` |
```

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
harlanhaskins: Good catch. Right now, the clients I know of don't read it off these events, but they could and should be able to.

stmontgomery (approved):

grynspan (approved):

PR #1793, opened 2026-07-14
Simplify and fix encoded event ordering issues with test repetitions

harlanhaskins:
This fixes an issue introduced with #1528 where we now accidentally report `testStarted`/`testEnded` for non-parameterized repeated tests in an inconsistent way.

### Motivation:

(this is copy/pasted from a comment I left in-source, and normally I don't like leaving verbose comments in source but this one's confusing enough.)

For all test cases of a given test, we emit the following set of events internally.

```
testStarted
testCaseStarted (for each case)
...
testCaseEnded (for each case)
...
testEnded
```

For parameterized tests, this is what clients expect; the test itself has a distinct start/end from all the cases.

For non-parameterized tests, however, clients don't need a redundant `testCaseStarted`/`testCaseEnded` for a single case, so we elide it.

However, we don't know which `iteration` we're on until we've started running test cases, and subsequent iterations will post additional `testCaseStarted`/`testCaseEnded` events.

To provide a coherent façade to our clients:
- For non-parameterized tests, elide the outer `testStarted`/`testEnded` events, and replace `testCaseStarted`/`testCaseEnded` with `testStarted`/`testEnded`.
- For parameterized tests, emit all events.

### Modifications:

This extracts the mapping from `Event.Kind` to `EncodedEvent.Kind` into a failable initializer on `EncodedEvent.Kind`, which splits the kind-mapping from the field-population logic, and then simplifies the logic for determining when to substitute/elide `test`/`testCase` events.

### Checklist:

Resolves rdar://182236497

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
@@ -35,6 +35,76 @@ extension ABI {
       case testSkipped
       case testCancelled
       case runEnded
+
+      /// Encodes an ``Event/Kind`` into an ``EncodedEvent/Kind``.
+      ///
+      /// Not all ``Event/Kind`` values map to an encoded kind value.
+      init?(encoding kind: Event.Kind, in eventContext: borrowing Event.Context) {
+        /// For all test cases of a given test, we emit the following set of
+        /// events internally.
+        ///
+        /// ```
+        /// testStarted
+        /// testCaseStarted (for each case)
+        /// ...
+        /// testCaseEnded (for each case)
+        /// ...
+        /// testEnded
+        /// ```
+        ///
+        /// For parameterized tests, this is what clients expect; the test
+        /// itself has a distinct start/end from all the cases.
+        ///
+        /// For non-parameterized tests, however, clients don't need a redundant
+        /// `testCaseStarted`/`testCaseEnded` for a single case, so we elide it.
+        ///
+        /// However, we don't know which `iteration` we're on until we've
+        /// started running test cases, and subsequent iterations will post
+        /// additional `testCaseStarted`/`testCaseEnded` events.
+        ///
+        /// To provide a coherent façade to our clients:
+        /// - For non-parameterized tests, elide the outer
+        ///   `testStarted`/`testEnded` events, and replace `testCaseStarted`/
+        ///   `testCaseEnded` with `testStarted`/`testEnded`.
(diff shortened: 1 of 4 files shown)


Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -110,52 +177,18 @@ extension ABI {
     public var _sourceLocation: EncodedSourceLocation<V>?
 
     init?(encoding event: borrowing Event, in eventContext: borrowing Event.Context, messages: borrowing [Event.HumanReadableOutputRecorder.Message]) {
+      guard let decodedKind = Kind(encoding: event.kind, in: eventContext) else {
+        return nil
+      }
+      kind = decodedKind
```
grynspan: Do you mean `encodedKind`?

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
grynspan, line 224: We should suppress setting this property for synthesized `testStarted`/`testEnded` events.

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -110,52 +177,18 @@ extension ABI {
...
-        if eventContext.test?.isParameterized == false {
-          if let iteration = eventContext.iteration, iteration > 1 {
-            kind = .testEnded
-          } else {
-            return nil
-          }
-        } else {
-          kind = .testCaseEnded
-        }
-      case .testCaseCancelled:
```
grynspan, line 147: Does this event kind need munging too?

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
harlanhaskins: We'll only synthesize `testStarted`/`testEnded` events if `eventContext.test.isParameterized` is false anyway, but I also added a test to make sure this is the case.

grynspan: Such tests do have a test case! It's just uninteresting.

harlanhaskins: Correct, but the test case is only added to the encoded event if `isParameterized` is true, which means we wouldn't have faked the `testStarted` event

stmontgomery (approved):

PR #1798, opened 2026-07-16
Add C settings to the package manifest.

grynspan:
This PR adds C settings to the package manifest. This is necessary because when Swift processes a C header, it does so _as C_ rather than as C++, and our C++ settings don't get applied.

I've also factored out setting the Swift, C, and C++ settings for all targets to a separate `.map {}` call so that we don't have to remember which target types have which factory function arguments and don't accidentally omit settings from new targets in the future.

CMake is not impacted.

### Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -139,8 +139,6 @@ let package = Package(
         buildingForEmbedded ? [] : ["TestingMacros"]
       }(),
       exclude: ["CMakeLists.txt", "Testing.swiftcrossimport"],
-      cxxSettings: .packageSettings(),
-      swiftSettings: .packageSettings() + .enableLibraryEvolution() + .moduleABIName("Testing"),
       linkerSettings: [
         .linkedLibrary("execinfo", .when(platforms: [.custom("freebsd"), .openbsd])),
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -333,12 +313,46 @@ package.targets.append(contentsOf: [
...
+  var swiftSettings = target.swiftSettings ?? []
+  swiftSettings += .packageSettings(isTestTarget: isTestTarget)
+  if !isTestTarget {
+    if target.name == "_Testing_Foundation" {
+      // The Foundation module only has Library Evolution enabled on Apple
+      // platforms, and since this target's module publicly imports Foundation,
+      // it can only enable Library Evolution itself on those platforms.
+      swiftSettings += .enableLibraryEvolution(.whenApple())
+    } else {
+      swiftSettings += .enableLibraryEvolution()
```
stmontgomery: The targets `swift-testing-harness` and `SymbolShowcase` don't currently have or need Library Evolution enabled, but they aren't test targets so I think under this logic they would get that enabled needlessly

stmontgomery (approved):

PR #1799, opened 2026-07-16
Combine a parameterized test case's argument IDs into a single hashed ID

suzannaratcliff:
### Motivation:
`Test.Case.ID.argumentIDs` previously stored one `Argument.ID` per argument, in order. This meant the ID grew with the number of parameters.  I want to simplify this to a single combined ID that identifies the whole combination of arguments.

rdar://182725918

### Checklist:

### DIFF
--- a/Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
+++ b/Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
@@ -99,6 +99,25 @@ extension Test.Case.Argument.ID {
   }
 }
 
+extension Test.Case.Argument.ID {
+  /// Initialize an ID instance by combining several argument IDs into one.
+  ///
+  /// - Parameters:
+  ///   - argumentIDs: The argument IDs to combine, in order.
+  ///
+  /// The bytes of each argument ID are length-prefixed before being hashed
+  /// together, so that argument boundaries are preserved.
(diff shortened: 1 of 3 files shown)


Sources/Testing/Parameterization/CustomTestArgumentEncodable.swift
```diff
@@ -99,6 +99,25 @@ extension Test.Case.Argument.ID {
...
+  ///
+  /// - Parameters:
+  ///   - argumentIDs: The argument IDs to combine, in order.
+  ///
+  /// The bytes of each argument ID are length-prefixed before being hashed
+  /// together, so that argument boundaries are preserved.
+  init(combining argumentIDs: some Sequence<Test.Case.Argument.ID>) {
+    var bytes = [UInt8]()
+    for argumentID in argumentIDs {
+      var count = UInt64(argumentID.bytes.count).littleEndian
```
grynspan: Do we need this ID to be stable across architectures? It may be sufficient to just encode the `Int`.

suzannaratcliff: Probably not, this is a good point.  Thanks! Updated.

grynspan (approved):

stmontgomery (approved):

PR #1801, opened 2026-07-17
Adopt executable path API (sorta) from the stdlib.

grynspan:
https://github.com/swiftlang/swift/pull/90476 added an exported function to the stdlib on non-Apple platforms to get the current executable's path. We should adopt this implementation instead of continuing to use our own.

On Apple platforms, due to back-deployment requirements, that function is not implemented and we instead continue to call `_NSGetExecutablePath()`.

I have cherry-picked the change to 6.4.x [here](https://github.com/swiftlang/swift/pull/90659), so the change should work there too once the toolchain builds a new nightly.

### Checklist:


Sources/Testing/Support/Additions/CommandLineAdditions.swift
```diff
@@ -26,12 +26,27 @@ enum CommandLine {
...
+  private static var executablePathCString: ContiguousArray<FPEncoding.CodeUnit>? {
+    @_silgen_name("_swift_stdlib_executablePathCString") get
+  }
+#endif
+
   /// The path to the current process' executable.
   static var executablePath: String {
     get throws {
-#if SWT_TARGET_OS_APPLE
+#if hasFeature(Embedded) || os(WASI)
```
harlanhaskins, line 45: Should this be its own configuration?

grynspan: It's really a quirk of the stdlib rather than ours. 🤷‍♂️

Sources/Testing/Support/Additions/CommandLineAdditions.swift
```diff
@@ -49,91 +64,21 @@ extension CommandLine {
...
-              throw Win32Error(rawValue: DWORD(ERROR_ILLEGAL_CHARACTER))
-            }
-          case DWORD(ERROR_INSUFFICIENT_BUFFER):
-            bufferCount += Int(MAX_PATH)
-          case let errorCode:
-            throw Win32Error(rawValue: errorCode)
-          }
+      }
+      return try executablePathCString.withUnsafeBufferPointer { executablePathCString in
+        guard let result = String.decodeCString(executablePathCString.baseAddress!, as: FPEncoding.self)?.result else {
```
suzannaratcliff, line 76: Will `executablePathCString.baseAddress!` ever be nil?

Sources/Testing/Support/Additions/CommandLineAdditions.swift
```diff
@@ -49,91 +64,21 @@ extension CommandLine {
...
-          throw CError(rawValue: swt_errno())
-        }
-        return try withUnsafeTemporaryAllocation(of: CChar.self, capacity: bufferCount) { buffer in
-          guard 0 == sysctl(mib.baseAddress!, .init(mib.count), buffer.baseAddress!, &bufferCount, nil, 0) else {
-            throw CError(rawValue: swt_errno())
-          }
-          return String(cString: buffer.baseAddress!)
-        }
-      }
-#elseif os(OpenBSD)
```
suzannaratcliff, line 88: I see this was moved to here: https://github.com/swiftlang/swift/pull/90476/changes#diff-890347f617154979102f47d3bae0b3f9275eb20ff060b44870bf0d35a0cca86bR725

Sources/Testing/Support/Additions/CommandLineAdditions.swift
grynspan: No, because it will always contain at least the trailing NUL character even if it's an empty string.

stmontgomery (approved):

Sources/Testing/Support/Additions/CommandLineAdditions.swift
grynspan: Yup!

grynspan:
Windows' main branch hasn't built since May, so I need to wait for that to get bumped. FreeBSD hasn't built in a while either AFAIK.

PR #1807, opened 2026-07-21
Make the fields of a JSON-encoded event optional.

grynspan:
This PR makes the `absolute` and `since1970` fields of `ABI.EncodedInstant` optional when using the experimental version of the JSON schema. If either is missing, we synthesize them by mapping against the current system's clocks. If both are missing, we use the current time on the current system and cross our fingers.

This logic may be incorrect if events are decoded on a different system or different session from where they were encoded. The expectation in those cases is that if a field is missing, it will be _consistently_ missing from all events and so the offsets we compute will still be correct relative to each other even if they are not strictly accurate in the originating device's reference frame. (We do not attempt to handle relativistic time dilation.)

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedInstant.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedInstant.swift
@@ -20,10 +20,10 @@ extension ABI {
     /// The number of seconds since the system-defined suspending epoch.
     ///
     /// For more information, see [`SuspendingClock`](https://developer.apple.com/documentation/swift/suspendingclock).
-    var absolute: Double
+    var absolute: Double?
 
     /// The number of seconds since the UNIX epoch (1970-01-01 00:00:00 UT).
-    package var since1970: Double
+    package var since1970: Double?
   }
 }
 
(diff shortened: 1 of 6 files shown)


Sources/Testing/ABI/Encoded/ABI.EncodedInstant.swift
```diff
@@ -20,10 +20,10 @@ extension ABI {
     /// The number of seconds since the system-defined suspending epoch.
     ///
     /// For more information, see [`SuspendingClock`](https://developer.apple.com/documentation/swift/suspendingclock).
-    var absolute: Double
+    var absolute: Double?
 
     /// The number of seconds since the UNIX epoch (1970-01-01 00:00:00 UT).
-    package var since1970: Double
+    package var since1970: Double?
```
stmontgomery, line 26: Can we update the DocC to say "if any", at minimum, and ideally also explain why/under what conditions it might be valid for the value of these properties to be `nil`?

stmontgomery (approved):

Sources/Testing/ABI/Encoded/ABI.EncodedInstant.swift
grynspan: They should never be `nil` until we formally allow it via the new schema version; I'll add those comments when I move these changes from `Experimental` to (presumably) `v6_5`.

PR #1815, opened 2026-07-29
Add support for statically linking on Linux

keith:
Manually adding the autolinks is required when linking these statically.
It also requires we remove the SWT_NO_INTEROP define

This is work towards supporting `-static-stdlib` with test targets on Linux

Work towards https://github.com/swiftlang/swift-package-manager/issues/5590

### DIFF
--- a/Sources/Overlays/_Testing_Foundation/CMakeLists.txt
+++ b/Sources/Overlays/_Testing_Foundation/CMakeLists.txt
@@ -33,6 +33,10 @@ if(NOT APPLE)
   target_link_libraries(_Testing_Foundation PUBLIC
     Foundation)
 endif()
+if(NOT BUILD_SHARED_LIBS)
+  target_compile_options(_Testing_Foundation PRIVATE
+    "SHELL:-Xfrontend -public-autolink-library -Xfrontend _TestingInternals")
+endif()
 
 # Note: This does not enable Library Evolution, despite emitting a module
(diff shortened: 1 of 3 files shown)


stmontgomery:
Thanks for the PR. Could you provide more details and context about what problem this is resolving, and what audiences and/or workflows are affected?

keith:
yea sorry, more context on https://github.com/swiftlang/swift/pull/91049 and i updated the description here

keith:
I'm hoping I can test e2e in that PR with a toolchain so we can verify with that before landing anything

finagolfin:
> Could you provide more details and context about what problem this is resolving, and what audiences and/or workflows are affected?

It is currently not possible to test Swift code on non-Darwin when built with the Swift compiler's `-static-stdlib` flag or SwiftPM's `--static-swift-stdlib` flag, because static archives of XCTest and Testing are not provided. Keith is presumably fixing that for linux, which can hopefully be extended to the static linux SDK too, whose executables' tests could so far never be run, swiftlang/swift-docker#396.

grynspan:
I'm hesitant to accept this PR because we don't really have a roadmap for testing with the static stdlib. It's not really as simple as just enabling static linkage; we need to consider whether code should be linking to the test targets at all, and we need to consider the XCTest interop story (removing `SWT_NO_INTEROP` is almost certainly incorrect).

PR #1818, opened 2026-08-02
Add support for Nerd Fonts glyphs in console output

### LINKED ISSUE #702: Add support for Nerd Fonts
We should look into supporting [Nerd Fonts](https://www.nerdfonts.com) as symbol substitutes when using `swift test`.

There's no standard way to detect what font the current terminal is using (if any). Some platform-specific mechanisms:

- For Gnome Terminal, we may be able to use [`GSettings`](https://docs.gtk.org/gio/class.Settings.html).
- On Windows, [`GetCurrentConsoleFont()`](https://learn.microsoft.com/en-us/windows/console/getcurrentconsolefont) may or may not work depending on whether the pipe to SwiftPM interferes with it.
- We could read the font out of `"com.apple.terminal"` for Terminal.app?

Or we could just fall back to an environment variable like `"SWT_NERD_FONTS_ENABLED"` like we do with SF Symbols.

maxches99:
Adds opt-in support for rendering [Nerd Fonts](https://www.nerdfonts.com) glyphs in console output via the `SWT_NERD_FONTS_ENABLED` environment variable.

### Motivation:

The console output recorder represents events with Unicode pictographs, or with SF Symbols on macOS when available. Terminals configured with a Nerd Font can render richer, more legible glyphs instead.

Since there is no standard cross-platform way to detect which font (if any) the current terminal is using, this follows the fallback approach suggested in #702: an explicit opt-in via an environment variable, mirroring the existing `SWT_SF_SYMBOLS_ENABLED` pattern.

Resolves #702.

### Modifications:

- Added `Event.Symbol.nerdFontCharacter` mapping each symbol to a Nerd Fonts glyph (codepoints verified against the official Nerd Fonts `glyphnames.json` registry):

  | Symbol | Codepoint | Glyph name |
  |-|-|-|
  | `.default` | U+F070C | `nf-md-rhombus_outline` |
  | `.skip` | U+F0A9 | `nf-fa-arrow_circle_right` |
  | `.pass` | U+F00C | `nf-fa-check` |
  | `.pass` (with known issues) | U+F068 | `nf-fa-minus` |
  | `.fail` | U+F00D | `nf-fa-times` |
  | `.difference` | U+F0992 | `nf-md-plus_minus` |
  | `.warning` | U+F071 | `nf-fa-exclamation_triangle` |
  | `.details` | U+F060D | `nf-md-subdirectory_arrow_right` |
  | `.attachment` | U+F0C6 | `nf-fa-paperclip` |

- Added `useNerdFonts` to `Event.ConsoleOutputRecorder.Options`. Unlike `useSFSymbols`, it is available on all platforms. Since Nerd Fonts glyphs are only ever enabled explicitly, they take precedence over SF Symbols if both options are enabled. As with SF Symbols, an extra trailing space is added after the glyph when ANSI escape codes are in use.
- The entry point (e.g. `swift test`) now reads the `SWT_NERD_FONTS_ENABLED` environment variable to enable the option. There is no automatic detection.
- Documented the new environment variable in `Documentation/EnvironmentVariables.md`.
- Extended the "Writing events" test matrix in `EventRecorderTests` with a `useNerdFonts` dimension, including checks that Nerd Fonts glyphs replace the Unicode (and SF Symbols) characters when enabled.

### Checklist:

### DIFF
--- a/Documentation/EnvironmentVariables.md
+++ b/Documentation/EnvironmentVariables.md
@@ -28,6 +28,7 @@ names prefixed with `SWT_`.
 | `COLORTERM`\* | `String` | Used to determine if the current terminal supports 24-bit color. Common across UNIX-like platforms. |
 | `NO_COLOR`[\*](https://no-color.org) | `Any?` | If set to any value, disables color output regardless of terminal capabilities. |
 | `SWT_ENABLE_EXPERIMENTAL_CONSOLE_OUTPUT` | `Bool` | Used to enable or disable experimental console output. |
+| `SWT_NERD_FONTS_ENABLED` | `Bool` | Used to explicitly enable or disable [Nerd&nbsp;Fonts](https://www.nerdfonts.com) support. |
 | `SWT_SF_SYMBOLS_ENABLED` | `Bool` | Used to explicitly enable or disable SF&nbsp;Symbols support on macOS. |
 | `TERM`[\*](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html) | `String` | Used to determine if the current terminal supports 4- or 8-bit color. Common across UNIX-like platforms. |
 
--- a/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
+++ b/Sources/Testing/ABI/EntryPoints/EntryPoint.swift
@@ -780,6 +780,13 @@ extension Event.ConsoleOutputRecorder.Options {
     }
 #endif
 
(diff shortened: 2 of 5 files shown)


grynspan:
Thanks for opening this, however we are not looking for changes to our console output at this time as we expect to do some significant refactoring for the next Swift release.

If you have other ideas for contributing to Swift Testing, please reach out on the forums and we can discuss. Thanks!

PR #1820, opened 2026-08-06
Stand up the "6.5" JSON schema version.

grynspan:
This new version doesn't include any new content yet, but we're in 6.5 now, so…

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/ABI.swift
+++ b/Sources/Testing/ABI/ABI.swift
@@ -49,7 +49,7 @@ extension ABI {
   }
 
   /// The current supported ABI version (ignoring any experimental versions.)
-  public typealias CurrentVersion = v6_4
+  public typealias CurrentVersion = v6_5
 
   /// Get the type representing a given ABI version.
   ///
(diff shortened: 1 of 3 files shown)


suzannaratcliff (approved):

Tests/TestingTests/TestCaseIterationTests.swift
```diff
@@ -120,7 +120,7 @@ struct TestCaseIterationTests {
 
   private func assertEncodedEventKinds(
     _ test: Test,
-    equals expected: [ABI.EncodedEvent<ABI.v6_4>.Kind],
+    equals expected: [ABI.EncodedEvent<ABI.CurrentVersion>.Kind],
```
suzannaratcliff, line 123: Why are you using `ABI.CurrentVersion` instead of `v6_5`?

grynspan: Because the rest of the test already uses `CurrentVersion`.

Sources/Testing/ABI/ABI.swift
```diff
@@ -281,6 +283,17 @@ extension ABI {
     }
   }
 
+  /// A namespace and type for ABI version 6.4 symbols.
```
stmontgomery, line 286:
```suggestion
  /// A namespace and type for ABI version 6.5 symbols.
```

grynspan: Ah well.

PR #1823, opened 2026-08-07
Allow decoding a JSON schema version number directly from a C string.

grynspan:
This PR adds an `init` to `ABI.VersionNumber` to allow decoding them directly from C strings without needing to first allocate a Swift string. I toyed with having the existing `String`-based initializer call `withCString` and have the logic all against C strings (to avoid having to allocate a `String` for each C string) but the code was significantly more complex _and_ relied on a system API call to `strtol()` because it wasn't worth writing my own integer parser. The time cost of allocating a temporary `String` shouldn't be any worse than it is today without the new initializer.

This PR also adds a cache for `ABI.VersionNumber.init(String)` which makes that initializer about 10x faster for repeatedly-used version numbers. We can't cache C strings (by address) the same way because we don't know if the C string pointers are immortal, but since we create `String` instances for C strings, we'll still be able to use the cache to avoid _parsing_ them repeatedly.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/ABI.VersionNumber.swift
+++ b/Sources/Testing/ABI/ABI.VersionNumber.swift
@@ -50,10 +50,71 @@ extension ABI.VersionNumber {
 // MARK: - CustomStringConvertible
 
 extension ABI.VersionNumber: CustomStringConvertible {
+  /// A cache of previously-parsed version numbers.
+  private static let _versionNumberCache = Mutex<[String: Self?]>()
+
+  /// Parse an instance of this type from the given string.
+  ///
+  /// - Parameters:
+  ///   - string: The string to parse, such as `"0"` or `"6.3.0"`.
+  ///
+  /// - Returns: An instance of this type, or `nil` if one could not be parsed
+  ///   from `string`.
+  ///
+  /// - Bug: We are not able to reuse the logic from swift-syntax's
+  ///   `VersionTupleSyntax` type here because we cannot link to swift-syntax
+  ///   in this target.
+  private static func _parse(_ string: String) -> Self? {
+    // Check if we've previously encountered this version number.
+    let cachedValue = Self._versionNumberCache.withLock { versionNumberCache in
+      versionNumberCache[string]
+    }
+    if case let .some(cachedValue) = cachedValue {
+      return cachedValue
+    }
+
+    var result: Self?
+    do {
+      // Split the string on "." (assuming it is of the form "1", "1.2", or
+      // "1.2.3") and parse the individual components as integers.
+      let components = string.split(separator: ".", omittingEmptySubsequences: false)
+      func componentValue(_ index: Int) -> Component? {
+        components.count > index ? Component(components[index]) : 0
+      }
+      if let majorComponent = componentValue(0),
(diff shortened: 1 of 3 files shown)


Sources/Testing/ABI/ABI.VersionNumber.swift
```diff
@@ -65,20 +126,11 @@ extension ABI.VersionNumber: CustomStringConvertible {
...
-    // "1.2.3") and parse the individual components as integers.
-    let components = string.split(separator: ".", omittingEmptySubsequences: false)
-    func componentValue(_ index: Int) -> Component? {
-      components.count > index ? Component(components[index]) : 0
-    }
-
-    guard let majorComponent = componentValue(0),
-          let minorComponent = componentValue(1),
-          let patchComponent = componentValue(2) else {
+  public init?(validatingCString string: UnsafePointer<CChar>) {
```
harlanhaskins, line 133: Given that we're calling `validatingCString`, is it worth constructing a C string of invalid unicode sequences in a test and ensuring it returns `nil`?

harlanhaskins (approved):

Sources/Testing/ABI/ABI.VersionNumber.swift
grynspan: I don't think so, as that's really just testing `validatingCString` which is already tested by the stdlib. But if you feel strongly about it, DM me and we can discuss what test would be appropriate to add.

Sources/_TestingInternals/include/Stubs.h
```diff
@@ -60,6 +60,14 @@ static int swt_errno(void) {
   return errno;
 }
 
+/// Set the current C error.
+///
+/// This function is provided because `errno` is a complex macro on some
+/// platforms and cannot be imported directly into Swift.
+static void swt_set_errno(int errorCode) {
```
jerryjrchen: Is this related? I don't see `swt_set_errno` being called elsewhere in this change

Sources/Testing/ABI/ABI.VersionNumber.swift
```diff
@@ -50,11 +54,72 @@ extension ABI.VersionNumber {
...
+
+  /// Parse an instance of this type from the given string.
+  ///
+  /// - Parameters:
+  ///   - string: The string to parse, such as `"0"` or `"6.3.0"`.
+  ///
+  /// - Returns: An instance of this type, or `nil` if one could not be parsed
+  ///   from `string`.
+  ///
+  /// - Bug: We are not able to reuse the logic from swift-syntax's
```
jerryjrchen, line 68: Just curious for a bit more background: why we can't link `swift-syntax` for the `Testing` library?

jerryjrchen (approved):

Sources/Testing/ABI/ABI.VersionNumber.swift
grynspan: It doesn't exist at runtime.

Sources/_TestingInternals/include/Stubs.h
grynspan: Oh, stray from an earlier revision. Will remove. Thanks.

PR #1825, opened 2026-08-09
Remove fallback implementation of `Mutex` using `pthread_mutex_t`.

grynspan:
`os_unfair_lock` is available on all Apple platforms we target (sorry System 7), and `Synchronization` is available everywhere else Swift goes, so we don't really need a `pthread_mutex_t` fallback path.

### Checklist:


stmontgomery (approved):

grynspan:
Checked the diff against Apple's internal fork. (And that's all I can say about that, sorry.)

PR #1826, opened 2026-08-09
[DNM] Remove thunk over `objc_copyImageHeaders()`.

grynspan:
Our CI is now on an Xcode 27 beta, so it can directly invoke `objc_copyImageHeaders()`. No need for the `dlsym()`-calling thunk anymore.

### Checklist:


grynspan:
Primary Swift CI still uses an older Xcode, so need to double-check we don't break the toolchain build first. (May need to hold off on this change for a few years, or use `#if compiler(>=6.4)`.)

PR #1828, opened 2026-08-10
Include the `$s` prefix in mangled names we emit into the JSON event stream.

grynspan:
Make sure we include the `$s` prefix so we can distinguish Swift mangling from other languages' mangling (e.g. C++) at runtime.

### Checklist:

### DIFF
--- a/Tests/TestingTests/TypeInfoTests.swift
+++ b/Tests/TestingTests/TypeInfoTests.swift
@@ -104,10 +104,16 @@ struct TypeInfoTests {
     #expect(typeInfo.fullyQualifiedNameComponents == expectedComponents)
   }
 
-  @Test func mangledTypeName() {
-    #expect(_mangledTypeName(String.self) == TypeInfo(describing: String.self).mangledName)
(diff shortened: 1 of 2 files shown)


Tests/TestingTests/TypeInfoTests.swift
```diff
@@ -104,10 +104,16 @@ struct TypeInfoTests {
...
-    #expect(_mangledTypeName(String.self) == TypeInfo(describing: String.self).mangledName)
-    #expect(_mangledTypeName(String.NestedType.self) == TypeInfo(describing: String.NestedType.self).mangledName)
-    #expect(_mangledTypeName(SomeEnum.self) == TypeInfo(describing: SomeEnum.self).mangledName)
+  @Test(
+    arguments: [
+      String.self,
+      String.NestedType.self,
+      SomeEnum.self,
+    ]
+  ) func mangledTypeName(type: Any.Type) throws {
```
stmontgomery, line 113: Looks like there are some legitimate CI build failures for this test with 6.4

grynspan: Fixed.

stmontgomery (approved):

PR #1829, opened 2026-08-11
Don't call `write()` in new tag-filtering code without file I/O enabled.

grynspan:
Embedded Swift targets may not have access to file I/O, so we need to conditionalize this one line.

### Checklist:


grynspan:
@gmedori FYI.

harlanhaskins (approved):

briancroom (approved):

gmedori:
TIL! Thanks for catching this @grynspan!

Issue #1614, opened 2026-03-06
CI jobs do not use the macros from this repo


finagolfin:
### Description

#1523 was recently merged, then [broke the macOS trunk snapshot toolchain CI](https://ci.swift.org/job/oss-swift-package-macos/4813/console), because it was using macros that aren't  there yet when building the toolchain, which had to be fixed in #1609.

This probably wasn't detected during PR testing because all the current PR CI jobs are using the macros from a prebuilt toolchain instead.

### Reproduction

Do what the linked CI job does, ie build fresh macOS toolchain using macros as in #1523

### Expected behavior

CI to catch such regressions

### Environment

macOS

### Additional information

It may be possible to set up a PR CI job with the prebuilt nightly snapshot toolchain and `build-script`, by passing in the prebuilt toolchain using the `--skip-build-swift --native-swift-tools-path` flags and building and installing this repo and its macros using that script alone, as the toolchain CI does in fuller form after a bunch of pulls are merged every day.

This would warn about such build errors _before_ such pulls are merged.

grynspan:
This was already resolved in #1609 (specifically for `#_sourceLocation` because that's the one that may crop up without being written explicitly.)

finagolfin:
I think you misunderstand, this is about what I raised in that pull, ie adding CI so such mistakes don't happen again.

stmontgomery:
@finagolfin The fix @grynspan made in #1609 actually _will_ guard against this happening in the future because it includes some clever changes to the macro expansion which will detect when it's accidentally misused from a target where it shouldn't be. From the PR description:

> Update the expansion of `#_sourceLocation` to detect it's being used in one of our own library targets and emit an error.

So although we didn't implement the precise change you suggested, we now have CI protection against this. (And we gain this protection without needing to switch to invoking build-script, which would take a lot longer to run!)

finagolfin:
Correct me if I'm wrong, but looking at that pull, you simply add a warning so that `#_sourceLocation` is not invoked in the current debug builds in the CI jobs here.

Whereas what I'm suggesting is to set up a `build-script` CI job that will catch _all_ such disallowed macro usage going forward, including other macros. Are you saying that is not a concern because `#_sourceLocation` is the only macro that might be misused that way, both now and in the foreseeable future?

> we gain this protection without needing to switch to invoking build-script, which would take a lot longer to run!

[...]

PR #1830, opened 2026-08-11
Defer allocating `__Expression` instances until we actually need them.

grynspan:
This PR makes the `expression` argument of the internal `__check()` function family an autoclosure that we only evaluate on expectation failure. This saves us the cost of allocating these values. For tests that heavily rely on `#expect()` (in the hundreds of thousands of invocations!), this has a measurable impact on test time.

The remaining time spent in a successful call to `#expect()` outside the test code itself is dominated by unspecialized generics overhead, which we can't realistically eliminate at our layer. The compiler team will need to look into that separately.

### Metrics

Tested on an Ubuntu 24.04 VM with the latest Swift nightly toolchain (Swift 1f6442f1d1fcbb7 + Swift Testing 2171d0caab16414) and with a toolchain containing this PR built via CI command (Swift 202b1eab5ef91b2 + Swift Testing a4c97b3e3d4d796).

Built and ran the tests in the attached Swift package using `swift test`. The target tests were built without optimizations. 📎 [testing-perf-demo.tgz](https://github.com/user-attachments/files/31043475/testing-perf-demo.tgz)

| | Main Branch | With Patch | Difference |
|-|-:|-:|-|
| Control Group (`fast()`) | 0.097s | 0.097s | None[^really] |
| Run 1 | 0.779s | 0.348s | |
| Run 2 | 0.767s | 0.345s | |
| Run 3 | 0.769s | 0.357s | |
| Average of runs | 0.772s | 0.350s | 2.2x as fast |

[^really]: Yes, I really got the same result.

The remaining time appears to be dominated by Swift generics code in the runtime. At our layer, we cannot realistically optimize these calls away.

Resolves rdar://184618689.

### Checklist:

### DIFF
--- a/Sources/TestingMacros/Support/ConditionArgumentParsing.swift
+++ b/Sources/TestingMacros/Support/ConditionArgumentParsing.swift
@@ -27,6 +27,9 @@ struct Condition {
   /// the testing library's `__Expression` type.
   var expression: ExprSyntax
 
+  /// How many negation operators were applied to the overall expression.
+  var negationCount = 0
+
   init(_ expandedFunctionName: String, arguments: [Argument], expression: ExprSyntax) {
     self.expandedFunctionName = .identifier(expandedFunctionName)
     self.arguments = arguments
@@ -455,6 +458,7 @@ private func _parseCondition(from expr: MemberAccessExprSyntax, for macro: some 
 private func _parseCondition(negating expr: ExprSyntax, isParenthetical: Bool, for macro: some FreestandingMacroExpansionSyntax, in context: some MacroExpansionContext) -> Condition {
   var result = _parseCondition(from: expr, for: macro, in: context)
   result.expression = createExpressionExprForNegation(of: result.expression, isParenthetical: isParenthetical)
+  result.negationCount += 1
   return result
 }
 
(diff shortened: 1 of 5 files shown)


grynspan:
Screenshot 2026-08-13 at 3 21 42 PM

Auspicious!

Sources/TestingMacros/Support/ConditionArgumentParsing.swift
```diff
@@ -455,6 +458,7 @@ private func _parseCondition(from expr: MemberAccessExprSyntax, for macro: some
 private func _parseCondition(negating expr: ExprSyntax, isParenthetical: Bool, for macro: some FreestandingMacroExpansionSyntax, in context: some MacroExpansionContext) -> Condition {
   var result = _parseCondition(from: expr, for: macro, in: context)
   result.expression = createExpressionExprForNegation(of: result.expression, isParenthetical: isParenthetical)
+  result.negationCount += 1
```
harlanhaskins, line 461: Nice little optimization, no need to pay for this traversal twice

grynspan: It's unfortunate the current implementation requires this traversal at all, but such is life. #840, someday!

stmontgomery (approved):

finagolfin:
We are seeing [SwiftPM build failures with Testing on the trunk macOS toolchain CI](https://ci.swift.org/job/oss-swift-package-macos/5142/console), possibly related to this pull, possibly broke even earlier. Let me know what you think of an expanded CMake GitHub workflow with SwiftPM that I suggested in #1614 a couple months ago, which would've likely caught this earlier.

grynspan:
The failures in CI aren't due to this PR, but rather because SwiftPM is misconfigured to use the newly-built macro target with an older library target, and the two aren't compatible. We'd run into the same issue any time we made substantive changes to our macro expansions.

We're already investigating this internally.

PR #1831, opened 2026-08-12
Use Debian 12 in CI instead of Fedora 41.

grynspan:
We aren't building Swift 6.4 toolchains for Fedora 41 at this time, which is causing one of the Wasm jobs to fail. We don't have a separate host OS matrix for Wasm jobs, so just ~~downgrade to Fedora 39~~ switch to Debian 12.

### Checklist:


grynspan:
>   Error response from daemon: manifest for swiftlang/swift:nightly-6.4.x-fedora39 not found: manifest unknown: manifest unknown

welp

grynspan:
Closing as this appears to be a GHA or Docker misconfig, not a problem with our repo.

PR #1832, opened 2026-08-12
Expose as tools SPI an `ABI.Version.eventHandler()` variant that produces JSON.

grynspan:
This PR adds a convenience function that produces an event stream handler that yields JSON directly. It's the moral equivalent of:

```swift
let eventHandler = ABI.___.eventHandler { record in
  guard let json = try? JSONEncoder().encode(record) else {
    // failed to encode (unexpected)
  }
  // write JSON somewhere
}
```

And has a toggle for JSON Lines output specifically.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/ABI.swift
+++ b/Sources/Testing/ABI/ABI.swift
@@ -21,29 +21,47 @@ extension ABI {
   public protocol Version: Sendable {
     /// The numeric representation of this ABI version.
     static var versionNumber: VersionNumber { get }
-  }
 
-  /// A protocol that extends the public ``ABI/Version`` protocol with
-  /// internal-only requirements.
-  protocol _Version: Version {
 #if !SWT_NO_ABI_JSON_SCHEMA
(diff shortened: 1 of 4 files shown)


Sources/Testing/ABI/ABI.swift
```diff
@@ -21,29 +21,47 @@ extension ABI {
   public protocol Version: Sendable {
     /// The numeric representation of this ABI version.
     static var versionNumber: VersionNumber { get }
-  }
 
-  /// A protocol that extends the public ``ABI/Version`` protocol with
-  /// internal-only requirements.
-  protocol _Version: Version {
```
stmontgomery, line 28: Happy to see `_Version` no longer needed

stmontgomery (approved):

Sources/Testing/ABI/ABI.swift
grynspan: I might bring it back in the future if we decide it's needed, but yah.

PR #1837, opened 2026-08-16
Support non-finite floats in exit test captures

### LINKED ISSUE #1836: Exit test value capture fails for non-finite floating-point values
## Description

Starting in Swift 6.3, values explicitly captured by exit test closures are encoded as JSON in the parent process and decoded in the child process.

`Float` and `Double` satisfy the `Codable` and `Sendable` requirements for captured values, but the default `JSONEncoder` rejects `.infinity`, `-.infinity`, and `.nan`. As a result, the exit test reports an encoding error before launching its body.

This became observable when exit-test value capturing was introduced in Swift 6.3.

## Reproduction

```swift
import Testing

@Test
func captureNonFiniteFloatingPointValue() async {
  let value = Double.infinity

  await #expect(processExitsWith: .success) { [value = value as Double] in
    precondition(value == .infinity)
  }
}
```

Run `swift test`.

The test reports an error containing:

```text
A system failure occurred: EncodingError.invalidValue:: inf(Double). Debug description: Unable to encode Double.inf directly in JSON
```

## Expected behavior

Non-finite floating-point values should round-trip through an exit test capture list so the subprocess body can run. Positive and negative infinity should retain their sign, and NaN should remain NaN.

## Environment

```text
Swift Testing Library Version: 1902
Apple Swift version 6.3.3 effective-5.10 (swiftlang-6.3.3.1.3 clang-2100.1.1.101)
Xcode 26.6 (17F113)
macOS 26.5.2 (25F84)
```

The same failure also occurs in CI on macOS 15.5 with Xcode 26.6 / Swift 6.3.3.

## Additional information

Exit-test value capture was introduced by [ST-0012: Exit Test Value Capturing](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0012-exit-test-value-capturing.md).

A workaround is to capture each floating-point value's `bitPattern` as an integer and reconstruct the value in the subprocess closure, but this exposes an implementation detail to every affected test.

```diff
-                await #expect(processExitsWith: .failure) { [min, ideal, max] in
+                await #expect(processExitsWith: .failure) {
+                    [
+                        minBitPattern = Double(min).bitPattern as UInt64,
+                        idealBitPattern = Double(ideal).bitPattern as UInt64,
+                        maxBitPattern = Double(max).bitPattern as UInt64,
+                    ] in
+                    let min = CGFloat(Double(bitPattern: minBitPattern))
+                    let ideal = CGFloat(Double(bitPattern: idealBitPattern))
+                    let max = CGFloat(Double(bitPattern: maxBitPattern))
```

Swift Open Source Slack channel discussion https://swift-open-source.slack.com/archives/C06QXFKB4ER/p1786820881281939

Kyle-Ye:
Allow exit test capture lists to round-trip NaN and positive and negative infinity.

### Motivation:

Exit-test value capturing became available in Swift 6.3. Captured values are JSON-encoded in the parent process and decoded in the subprocess. The default JSON coding strategies reject non-finite floating-point values even though `Float` and `Double` satisfy the `Codable` and `Sendable` requirements for captured values.

Fixes #1836.

### Modifications:

Configure matching non-conforming floating-point encoding and decoding strategies specifically for the exit-test captured-value transport, using an internal coding user-info key so other JSON call sites retain their default behavior.

Use `"NaN"`, `"Infinity"`, and `"-Infinity"` as the JSON string representations, following conventions used by other JSON encoders.

Add a parameterized regression test covering positive infinity, negative infinity, and NaN, including validation that their classification and sign survive the round trip.

### AI Disclosure:

Codex was used to help investigate the exit-test capture path, research JSON prior art, and draft and refine the implementation, tests, and PR text. I reviewed the resulting changes and verified the regression test locally.

### Checklist:

- [x] Code and documentation follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] No public symbols are renamed or modified, so no DocC references need updating.

### Future discussion

I suggest this should also be cherry-picked into release/6.4 and release/6.3.

### DIFF
--- a/Sources/Testing/Support/JSON.swift
+++ b/Sources/Testing/Support/JSON.swift
@@ -25,8 +25,8 @@ enum JSON {
   private static let _prettyPrintingEnabled = Environment.flag(named: "SWT_PRETTY_PRINT_JSON") == true
 
   /// String representations of non-finite floating-point values.
-  private static let _positiveInfinityString = "INF"
-  private static let _negativeInfinityString = "-INF"
+  private static let _positiveInfinityString = "Infinity"
+  private static let _negativeInfinityString = "-Infinity"
   private static let _nanString = "NaN"
 
   /// Encode a value as JSON.
@@ -34,8 +34,6 @@ enum JSON {
   /// - Parameters:
   ///   - value: The value to encode.
   ///   - userInfo: Any user info to pass into the encoder during encoding.
-  ///   - allowNonFiniteFloatingPointValues: Whether or not non-finite
-  ///     floating-point values can be encoded.
   ///   - body: A function to call.
   ///
   /// - Returns: Whatever is returned by `body`.
@@ -44,12 +42,14 @@ enum JSON {
   static func withEncoding<R>(
     of value: some Encodable,
     userInfo: [CodingUserInfoKey: any Sendable] = [:],
-    allowNonFiniteFloatingPointValues: Bool = false,
     _ body: (UnsafeRawBufferPointer) throws -> R
   ) throws -> R {
     let encoder = JSONEncoder()
 
-    if allowNonFiniteFloatingPointValues {
+    // Set user info keys that clients want to use during encoding.
+    encoder.userInfo.merge(userInfo, uniquingKeysWith: { _, rhs in rhs })
+
+    if encoder.userInfo[._allowNonFiniteFloatingPointValuesUserInfoKey] as? Bool == true {
       encoder.nonConformingFloatEncodingStrategy = .convertToString(
         positiveInfinity: _positiveInfinityString,
         negativeInfinity: _negativeInfinityString,
@@ -63,9 +63,6 @@ enum JSON {
       encoder.outputFormatting.insert(.prettyPrinted)
       encoder.outputFormatting.insert(.withoutEscapingSlashes)
     }
-
-    // Set user info keys that clients want to use during encoding.
-    encoder.userInfo.merge(userInfo, uniquingKeysWith: { _, rhs in rhs})
 
     let data = try encoder.encode(value)
     return try data.withUnsafeBytes(body)
@@ -102,8 +99,7 @@ enum JSON {
   /// - Parameters:
   ///   - type: The type of value to decode.
   ///   - jsonRepresentation: The JSON encoding of the value to decode.
-  ///   - allowNonFiniteFloatingPointValues: Whether or not non-finite
-  ///     floating-point values can be decoded.
+  ///   - userInfo: Any user info to pass into the decoder during decoding.
   ///
   /// - Returns: An instance of `T` decoded from `jsonRepresentation`.
   ///
@@ -111,7 +107,7 @@ enum JSON {
   static func decode<T>(
     _ type: T.Type,
     from jsonRepresentation: UnsafeRawBufferPointer,
-    allowNonFiniteFloatingPointValues: Bool = false
+    userInfo: [CodingUserInfoKey: any Sendable] = [:]
   ) throws -> T where T: Decodable {
     try withExtendedLifetime(jsonRepresentation) {
       let byteCount = jsonRepresentation.count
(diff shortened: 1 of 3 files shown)


Sources/Testing/Support/JSON.swift
```diff
@@ -24,19 +24,39 @@ enum JSON {
   /// testing library to improve the readability of JSON output.
   private static let _prettyPrintingEnabled = Environment.flag(named: "SWT_PRETTY_PRINT_JSON") == true
 
+  /// String representations of non-finite floating-point values.
+  private static let _positiveInfinityString = "INF"
```
grynspan: Is there prior art in other languages or using other JSON encoders for encoding infinities and NaN? What strings do they use?

Sources/Testing/Support/JSON.swift
```diff
@@ -24,19 +24,39 @@ enum JSON {
...
   ///   - body: A function to call.
   ///
   /// - Returns: Whatever is returned by `body`.
   ///
   /// - Throws: Whatever is thrown by `body` or by the encoding process.
-  static func withEncoding<R>(of value: some Encodable, userInfo: [CodingUserInfoKey: any Sendable] = [:], _ body: (UnsafeRawBufferPointer) throws -> R) throws -> R {
+  static func withEncoding<R>(
+    of value: some Encodable,
+    userInfo: [CodingUserInfoKey: any Sendable] = [:],
+    allowNonFiniteFloatingPointValues: Bool = false,
```
grynspan: Could we make this part of `userInfo` instead? `userInfo` is intended for this sort of purpose.

Tests/TestingTests/ExitTestTests.swift
```diff
@@ -569,6 +569,20 @@ private import _TestingInternals
     }
   }
 
+  @Test("Capture list (non-finite floating-point values)")
+  func captureListWithNonFiniteFloatingPointValues() async {
+    await #expect(processExitsWith: .success) {
+      [
```
grynspan, line 580: Consider making this a parameterized test that takes a single `Double` argument instead?

grynspan (changes requested):
Thanks for the PR! And thank you for not making the change everywhere we encode/decode JSON (as that would be an ABI break for our JSON event stream and would require Swift Evolution approval).

I am hesitant to approve the change without a better understanding of how unreal values are handled in other systems that use JSON. It would help if you could gather some examples of what other languages, libraries, implementations, etc. are currently doing to handle them.

grynspan:
> I suggest this should also be cherry-picked into release/6.4 and release/6.3.

We are unlikely to take further changes to Swift 6.4.0 at this point in the schedule, however once this PR is merged you can open a new PR targetting `release/6.4.x` and we can see if it meets the criteria for 6.4.1 (or whatever dot release we're on at that point).

I am not aware of any future planned updates to Swift 6.3. The bar for changes there is exceptionally high.

Sources/Testing/Support/JSON.swift
Kyle-Ye: Yes—there is consistent prior art for these spellings. [RFC 8259 §6](https://www.rfc-editor.org/rfc/rfc8259.html#section-6) does not permit Infinity or NaN as JSON numbers, so encoding them as strings keeps the payload valid JSON.
[ProtoJSON](https://protobuf.dev/programming-guides/json/#representation-of-each-type), [.NET System.Text.Json](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonnumberhandling), and [Go encoding/json/v2](https://pkg.go.dev/encoding/json/v2#Marshal) all use "NaN", "Infinity", and "-Infinity".
Based on that prior art, I’ve changed "INF" and "-INF" to "Infinity" and "-Infinity" while retaining "NaN".

Sources/Testing/Support/JSON.swift
Kyle-Ye: OK

Tests/TestingTests/ExitTestTests.swift
Kyle-Ye: Got it.

Sources/Testing/Support/JSON.swift
```diff
@@ -98,8 +120,30 @@ enum JSON {
...
     }
   }
 #endif
 }
+
+#if !SWT_NO_CODABLE
+extension CodingUserInfoKey {
+  /// A coding user info key whose value is a `Bool` indicating whether or not
+  /// non-finite floating-point values are allowed.
+  static var _allowNonFiniteFloatingPointValuesUserInfoKey: Self {
```
grynspan: No underscore needed as this isn't `private`, per the repo's style guide.

stmontgomery (approved):
Aside from @grynspan's remaining style comment, this looks good to me now, thank you!

Sources/Testing/Support/JSON.swift
Kyle-Ye: Done.

Kyle-Ye:
@grynspan The style issue is fixed. Any other suggestion?

grynspan (approved):

grynspan:
Thanks!

PR #1838, opened 2026-08-17
Fix incorrectly exposed overload of `#require(processExitsWith:)`.

grynspan:
There's a hidden overload of `#require(processExitsWith:)` that takes a synchronous body closure (so as to help the type checker handle them). It is supposed to be hidden from documentation. Oops.

### Checklist:


jerryjrchen (approved):

grynspan:
Failure on macOS appears to be due to GitHub instability today and is unrelated to the PR.

PR #1839, opened 2026-08-17
Update Issue JSON schema

jerryjrchen:
Add fields to the EncodedIssue structure

Reviewers: I've split this change up by commit which you might have an easier time reviewing. Let me know any feedback on this structure too, thanks!

### Motivation:

Provider richer issue metadata for tools to work with.

This is the implementation that accompanies the pitch "Include
additional issue metadata in event stream", which contains additional
context and motivation. https://github.com/jerryjrchen/swift-evolution/blob/update-issue-json-abi/proposals/testing/nnnn-add-issue-metadata-event-stream.md

### Modifications:

Add custom Codable conformances to many ABI.Encoded... types as a
consequence of the more complex issue encoding scheme.

For example, the isKnown field can be encoded as either a boolean or a
string depending on the presence of _knownIssueComment.

Issue schema additions:

* Expression
* Time limit exceeded
* Error
* isKnown (modified to encode as _knownIssueComment if present)
* Confirmation miscount

Event schema changes:

* Add comments
* Messages is already optional (#1827), but is now formally included in
  this proposal for 6.5. Update the version checks for
  always[Encode,Decode]MessagesField to < ABI.v6_5.

Resolves rdar://173944877

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
@@ -1,0 +1,87 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if !SWT_NO_ABI_JSON_SCHEMA
+extension ABI {
+  /// A type implementing the JSON encoding of the actual and expected
+  /// confirmation counts for a miscounted confirmation for the ABI entry point
+  /// and event stream output.
+  ///
+  /// This type is not part of the public interface of the testing library. It
+  /// assists in converting values to JSON; clients that consume this JSON are
+  /// expected to write their own decoders.
+  struct EncodedConfirmationMiscount<V>: Sendable where V: ABI.Version {
+    /// The actual number of confirmations that occurred.
+    var actual: Int
+
+    /// The number of confirmations that were expected.
+    var expected: ExpectedCount
+
+    /// An enumeration describing the number of confirmations that were
+    /// expected, which may be a single value or a range of values.
+    enum ExpectedCount {
+      /// The expected count was a single value.
+      case single(Int)
+
+      /// The expected count was a range of values.
+      case range(ABI.EncodedRange<V>)
+    }
+  }
+}
+
+// MARK: - Conversion to/from library types
+
+extension ABI.EncodedConfirmationMiscount {
+  /// Encodes a miscount based on the actual and expected count.
+  /// - Parameter value: A tuple containing actual and expected number of confirmations.
+  ///    If the expected count is a single value, provide it as a single value
+  ///    range, e.g. `5...5`.
+  init?(encoding value: (actual: Int, expected: any RangeExpression)) {
+    guard let range = ABI.EncodedRange<V>(expectedRange: value.expected) else { return nil }
+
+    actual = value.actual
+    if let min = range.min, let max = range.max, min == max {
+      expected = .single(min)
+    } else {
+      expected = .range(range)
+    }
+  }
+}
+
+// MARK: - Codable
+
+extension ABI.EncodedConfirmationMiscount: Codable {
+  private enum _CodingKeys: String, CodingKey {
+    case actual
+    case expected
+  }
+
+  func encode(to encoder: any Encoder) throws {
+    var container = encoder.container(keyedBy: _CodingKeys.self)
+    try container.encode(actual, forKey: .actual)
+    switch expected {
+    case .single(let count):
+      try container.encode(count, forKey: .expected)
+    case .range(let range):
+      try container.encode(range, forKey: .expected)
+    }
+  }
+
+  init(from decoder: any Decoder) throws {
+    let container = try decoder.container(keyedBy: _CodingKeys.self)
+    actual = try container.decode(Int.self, forKey: .actual)
+    if let count = try? container.decode(Int.self, forKey: .expected) {
+      expected = .single(count)
+    } else {
+      expected = .range(try container.decode(ABI.EncodedRange<V>.self, forKey: .expected))
+    }
+  }
+}
+#endif
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedError.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedError.swift
@@ -18,9 +18,6 @@ extension ABI {
   /// This type is not part of the public interface of the testing library. It
   /// assists in converting values to JSON; clients that consume this JSON are
   /// expected to write their own decoders.
-  ///
-  /// - Warning: Errors are not yet part of the JSON schema.
-  @_spi(Experimental)
   public struct EncodedError<V>: Sendable where V: ABI.Version {
     /// The error's description.
     ///
@@ -37,7 +34,10 @@ extension ABI {
     var domain: String?
 
     /// The code of the error.
-    var code: Int
+    var code: Int?
+
+    /// The type info for the error.
+    var typeInfo: EncodedTypeInfo<V>?
 
     // TODO: userInfo (partial) encoding
   }
@@ -56,7 +56,7 @@ extension ABI.EncodedError: Error {
   }
 
   public var _code: Int {
-    code
+    code ?? -1
   }
 
   public var _userInfo: AnyObject? {
@@ -67,7 +67,14 @@ extension ABI.EncodedError: Error {
 
 // MARK: - Codable
 
-extension ABI.EncodedError: Codable {}
+extension ABI.EncodedError: Codable {
+  private enum CodingKeys: String, CodingKey {
+    case code
+    case domain
+    case description
+    case typeInfo = "type"
+  }
+}
 
 // MARK: - CustomTestStringConvertible
 
@@ -76,9 +83,9 @@ extension ABI.EncodedError: CustomTestStringConvertible {
     if let description {
       return description
     } else if let domain {
-      return "\(domain) error \(code)"
+      return "\(domain) error \(_code)"
     }
-    return "error \(code)"
+    return "error \(_code)"
   }
 }
 
@@ -95,6 +102,7 @@ extension ABI.EncodedError {
       self.domain = domain
     }
     code = error._code
+    typeInfo = ABI.EncodedTypeInfo<V>(encoding: TypeInfo(describingTypeOf: error))
   }
 }
 
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
@@ -160,9 +160,7 @@ extension ABI {
     /// issue matcher, and either can be `nil`. In such cases, the secondary
     /// comment(s) are represented via a distinct property depending on the kind
     /// of that event.
-    ///
-    /// - Warning: Comments at this level are not yet part of the JSON schema.
-    var _comments: [String]?
+    var comments: [String]?
 
     /// A source location associated with this event, if any.
     ///
@@ -206,18 +204,30 @@ extension ABI {
         iteration = eventContext.iteration
       }
 
+      // Fields introduced in 6.5
+      if V.versionNumber >= ABI.v6_5.versionNumber {
+        switch event.kind {
+        case let .issueRecorded(recordedIssue):
+          comments = recordedIssue.comments.map(\.rawValue)
+        case let .testCaseCancelled(skipInfo),
+          let .testSkipped(skipInfo),
+          let .testCancelled(skipInfo):
+          comments = Array(skipInfo.comment).map(\.rawValue)
+        default:
+          break
+        }
+      }
+
       // Experimental fields
       if V.includesExperimentalFields {
         switch event.kind {
         case let .issueRecorded(recordedIssue):
-          _comments = recordedIssue.comments.map(\.rawValue)
           _sourceLocation = recordedIssue.sourceLocation.map { EncodedSourceLocation(encoding: $0) }
         case let .valueAttached(attachment):
           _sourceLocation = EncodedSourceLocation<V>(encoding: attachment.sourceLocation)
         case let .testCaseCancelled(skipInfo),
           let .testSkipped(skipInfo),
           let .testCancelled(skipInfo):
-          _comments = Array(skipInfo.comment).map(\.rawValue)
           _sourceLocation = skipInfo.sourceLocation.map { EncodedSourceLocation(encoding: $0) }
         default:
           break
@@ -244,7 +254,7 @@ extension ABI.EncodedEvent: Codable {
     case testID
     case iteration
     case testCase = "_testCase"
-    case comments = "_comments"
+    case comments
     case sourceLocation = "_sourceLocation"
   }
 
@@ -260,7 +270,7 @@ extension ABI.EncodedEvent: Codable {
     try container.encodeIfPresent(testID, forKey: .testID)
     try container.encodeIfPresent(iteration, forKey: .iteration)
     try container.encodeIfPresent(_testCase, forKey: .testCase)
-    try container.encodeIfPresent(_comments, forKey: .comments)
+    try container.encodeIfPresent(comments, forKey: .comments)
     try container.encodeIfPresent(_sourceLocation, forKey: .sourceLocation)
   }
 
@@ -278,7 +288,7 @@ extension ABI.EncodedEvent: Codable {
     testID = try container.decodeIfPresent(ABI.EncodedTest<V>.ID.self, forKey: .testID)
     iteration = try container.decodeIfPresent(Int.self, forKey: .iteration)
     _testCase = try container.decodeIfPresent(ABI.EncodedTestCase<V>.self, forKey: .testCase)
-    _comments = try container.decodeIfPresent([String].self, forKey: .comments)
+    comments = try container.decodeIfPresent([String].self, forKey: .comments)
     _sourceLocation = try container.decodeIfPresent(ABI.EncodedSourceLocation<V>.self, forKey: .sourceLocation)
   }
 }
@@ -387,8 +397,7 @@ extension ABI.Version {
     // If the environment variable above is set to `true`, then even newer
     // schema versions should encode the "messages" field.
 
-    // TODO: fix speculative version number check
-    _alwaysIncludeMessagesField == true || versionNumber < ABI.ExperimentalVersion.versionNumber
+    _alwaysIncludeMessagesField == true || versionNumber < ABI.v6_5.versionNumber
   }
 
   /// Whether or not to require the presence of the `"messages"` field in
@@ -397,8 +406,7 @@ extension ABI.Version {
     // Whether or not the field is required during decoding is solely dependent
     // on the schema version, not on the environment variable.
 
-    // TODO: fix speculative version number check
-    versionNumber < ABI.ExperimentalVersion.versionNumber
+    versionNumber < ABI.v6_5.versionNumber
   }
 }
 #endif
--- a/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
+++ b/Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
@@ -174,11 +174,7 @@ extension ABI {
     /// the known issue matcher. In such cases, the secondary source location(s)
     /// are represented via a distinct property depending on the kind of that
     /// event.
-    ///
-    /// - Warning: Source locations at this level of the JSON schema are not yet
-    ///   part of said JSON schema.
-    @_spi(Experimental)
-    public var _sourceLocation: EncodedSourceLocation<V>?
+    public var sourceLocation: EncodedSourceLocation<V>?
 
     init?(encoding event: borrowing Event, in eventContext: borrowing Event.Context, messages: borrowing [Event.HumanReadableOutputRecorder.Message] = []) {
       guard let encodedKind = Kind(encoding: event.kind, in: eventContext) else {
@@ -208,10 +204,14 @@ extension ABI {
         switch event.kind {
         case let .issueRecorded(recordedIssue):
           comments = recordedIssue.comments.map(\.rawValue)
+          sourceLocation = recordedIssue.sourceLocation.map { EncodedSourceLocation(encoding: $0) }
+        case let .valueAttached(attachment):
+          sourceLocation = EncodedSourceLocation<V>(encoding: attachment.sourceLocation)
         case let .testCaseCancelled(skipInfo),
           let .testSkipped(skipInfo),
           let .testCancelled(skipInfo):
           comments = Array(skipInfo.comment).map(\.rawValue)
+          sourceLocation = skipInfo.sourceLocation.map { EncodedSourceLocation(encoding: $0) }
         default:
           break
         }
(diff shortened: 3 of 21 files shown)


jerryjrchen:
- [ ] Still needs JSON.md updates

Sources/Testing/ABI/Encoded/ABI.EncodedTypeInfo.swift
```diff
@@ -0,0 +1,44 @@
...
+      fullyQualifiedName = typeInfo.fullyQualifiedName
+      unqualifiedName = typeInfo.unqualifiedName
+      mangledName = typeInfo.mangledName
+    }
+  }
+}
+
+// MARK: - Codable
+
+extension ABI.EncodedTypeInfo: Codable {}
```
grynspan, line 43: We ought to be able to provide `init(decoding:)` here, no? That will be valuable when decoding other bits that need a `TypeInfo` structure.

Sources/Testing/ABI/Encoded/ABI.EncodedError.swift
```diff
@@ -67,7 +70,32 @@ extension ABI.EncodedError: Error {
 
 // MARK: - Codable
 
-extension ABI.EncodedError: Codable {}
+extension ABI.EncodedError: Codable {
```
grynspan, line 70: This explicit implementation shouldn't be necessary. `_code` can just fall back to `1` if `code` is `nil`.

Note **1**, not -1, because that's what the Swift stdlib does when an error doesn't explicitly provide an error code.

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -260,7 +271,7 @@ extension ABI.EncodedEvent: Codable {
     try container.encodeIfPresent(testID, forKey: .testID)
     try container.encodeIfPresent(iteration, forKey: .iteration)
     try container.encodeIfPresent(_testCase, forKey: .testCase)
-    try container.encodeIfPresent(_comments, forKey: .comments)
+    try container.encodeIfPresent(comments, forKey: .comments)
     try container.encodeIfPresent(_sourceLocation, forKey: .sourceLocation)
```
grynspan: I thought your pitch included `sourceLocation` for all events? We will need this.

Sources/Testing/ABI/Encoded/ABI.EncodedExpression.swift
```diff
@@ -40,7 +44,30 @@ extension ABI {
 
 // MARK: - Codable
 
-extension ABI.EncodedExpression: Codable {}
+extension ABI.EncodedExpression: Codable {
```
grynspan, line 46: This explicit implementation is not necessary.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -78,35 +81,51 @@ extension ABI {
 
       // >= v6.3
       if V.versionNumber >= ABI.v6_3.versionNumber {
-        severity = switch issue.severity {
-        case .warning: .warning
-        case .error: .error
-        }
+        severity =
```
grynspan: Typo?

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -78,35 +81,51 @@ extension ABI {
...
-        } else {
-          switch issue.kind {
-          case .apiMisused:
-            EncodedError(encoding: APIMisuseError(description: ""))
-          case .system:
-            EncodedError(encoding: SystemError(description: ""))
-          default:
-            nil
+
+        error =
```
grynspan: This indentation isn't idiomatic in Swift.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -115,7 +134,74 @@ extension ABI {
 
 // MARK: - Codable
 
-extension ABI.EncodedIssue: Codable {}
+extension ABI.EncodedIssue: Codable {
```
grynspan, line 142: I don't _think_ this explicit implementation is necessary.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -169,15 +256,18 @@ extension Issue {
       // TODO: improve fidelity of issue kind reporting (especially those without associated values)
       issueKind = .unconditional
     }
-    let severity: Issue.Severity = switch issue.severity {
-    case .warning:
+    let severity: Issue.Severity =
```
grynspan: No.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -169,15 +256,18 @@ extension Issue {
...
-    case .error, nil:
-      // Prior to 6.3, all Issues are errors
+      case .error, nil:
+        // Prior to 6.3, all Issues are errors
         .error
-    }
+      }
     let sourceContext = SourceContext(
-      backtrace: issue._backtrace.map { Backtrace(addresses: $0.symbolicatedAddresses.map(\.address)) },
+      backtrace: issue._backtrace.map {
```
grynspan: House style is to only use anonymous closure arguments on one-liners. You've added newlines, so you need to add `backtrace in` to replace `$0`.

I'd just leave it as a one-liner.

Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
```diff
@@ -0,0 +1,90 @@
...
+
+// MARK: - Conversion to/from library types
+
+extension ABI.EncodedMiscount {
+  /// Encodes a miscount based on the actual and expected count.
+  /// - Parameters:
+  ///   - actual: The actual number of confirmations.
+  ///   - expected: The expected number of confirmations. If the expected count
+  ///    is a single value, provide it as a single value range, e.g. `5...5`.
+  init(actual: Int, expected: any RangeExpression & Sendable) {
```
grynspan:
```suggestion
  init(encoding value: (actual: Int, expected: some RangeExpression<Int>)) {
```

Not sure if it's better, but I think it maybe matches the style we've established a bit better? We don't have any other examples of using a single structure to encode multiple Swift values though.

Note the use of `some` rather than `any`, and no need for `Sendable`.

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
```diff
@@ -0,0 +1,55 @@
...
+    var max: Int?
+
+    // Creates the encoded range from an expected confirmation count.
+    //
+    // For completeness, this supports a superset of ranges allowed by the
+    // Testing library for confirmations. For example,
+    // `confirmation(expectedCount:...10)` without a explicit lower bound is
+    // considered ambiguous, but can be represented as an encoded range with a
+    // nil `min`.
+    init(expectedRange: any RangeExpression & Sendable) {
```
grynspan:
```suggestion
    init(encoding range: some RangeExpression<Int>) {
```

Doesn't need to be `Sendable` at this layer since it's not preserved.

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
```diff
@@ -0,0 +1,55 @@
...
+      if let range = expectedRange as? ClosedRange<Int> {
```
grynspan, line 38: These conditional casts can be expensive. I'm not sure if there's a reasonable alternative though.

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
```diff
@@ -0,0 +1,55 @@
...
+      } else if let range = expectedRange as? Range<Int> {
+        min = range.lowerBound
+        max = range.upperBound - 1
+      } else if let range = expectedRange as? PartialRangeFrom<Int> {
+        min = range.lowerBound
+      } else if let range = expectedRange as? PartialRangeUpTo<Int> {
+        max = range.upperBound - 1
+      } else if let range = expectedRange as? PartialRangeThrough<Int> {
+        max = range.upperBound
+      }
```
grynspan, line 54: Need an `} else {` I think, which might just serve to call `fatalError()`.

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+}
+
+// MARK: - Codable
```
grynspan, line 59: Codable conformances need to be guarded since Embedded Swift doesn't support `Codable`.

Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
```diff
@@ -0,0 +1,90 @@
...
+
+// MARK: - Codable
+
+extension ABI.EncodedMiscount: Codable {
+  private enum _CodingKeys: String, CodingKey {
+    case actual
+    case expected
+  }
+
+  // Encodes an actual count integer value, plus an expected count integer value
```
grynspan: Nitpick: not sure what this comment adds?

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
```diff
@@ -0,0 +1,55 @@
...
+      if let range = expectedRange as? ClosedRange<Int> {
+        min = range.lowerBound
+        max = range.upperBound
+      } else if let range = expectedRange as? Range<Int> {
+        min = range.lowerBound
+        max = range.upperBound - 1
+      } else if let range = expectedRange as? PartialRangeFrom<Int> {
+        min = range.lowerBound
+      } else if let range = expectedRange as? PartialRangeUpTo<Int> {
+        max = range.upperBound - 1
```
grynspan, line 49: This will crash for the partial range `..< .min`. Which is a silly range, but it can exist.

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
```diff
@@ -0,0 +1,55 @@
...
+        max = range.upperBound
+      }
+    }
+  }
+}
+
+// MARK: - Codable
+
+extension ABI.EncodedRange: Codable {}
+
```
grynspan, line 62: Ought to be possible to provide `ClosedRange<Int>.init(decoding:)`. We can assume unspecified bounds are `.min` and `.max`, respectively.

Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
```diff
@@ -0,0 +1,90 @@
...
+    let container = try decoder.container(keyedBy: _CodingKeys.self)
+    actual = try container.decode(Int.self, forKey: .actual)
+    if let count = try? container.decode(Int.self, forKey: .expected) {
+      expected = .single(count)
+    } else {
+      expected = .range(try container.decode(ABI.EncodedRange<V>.self, forKey: .expected))
+    }
+  }
+}
+#endif
```
grynspan, line 87: I guess there's no nominal type for us to provide `init(decoding:)` on. Ah well.

Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
```diff
@@ -0,0 +1,90 @@
...
+  /// Encodes a miscount based on the actual and expected count.
+  /// - Parameters:
+  ///   - actual: The actual number of confirmations.
+  ///   - expected: The expected number of confirmations. If the expected count
+  ///    is a single value, provide it as a single value range, e.g. `5...5`.
+  init(actual: Int, expected: any RangeExpression & Sendable) {
+    self.actual = actual
+
+    let range = ABI.EncodedRange<V>(expectedRange: expected)
+    if let min = range.min, let max = range.max, min == max {
```
grynspan, line 50: Would it make more sense for this logic to live in `ABI.EncodedRange`'s `Codable` conformance? If it did, we wouldn't need the enum here.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
```diff
@@ -156,8 +242,9 @@ extension Issue {
       default:
         issueKind = .errorCaught(error)
       }
-    } else if let expression = issue._expression.flatMap(__Expression.init(decoding:)),
-              let sourceLocation = issue.sourceLocation.flatMap(SourceLocation.init) {
+    } else if let expression = issue.expression.flatMap(__Expression.init(decoding:)),
+      let sourceLocation = issue.sourceLocation.flatMap(SourceLocation.init)
+    {
```
grynspan, line 259: You should add another `else if` clause that checks `confirmationMiscount` and, if present, sets `issueKind` appropriately.

Sources/Testing/ABI/Encoded/ABI.EncodedMiscount.swift
```diff
@@ -0,0 +1,90 @@
...
+#if !SWT_NO_ABI_JSON_SCHEMA
+extension ABI {
+  /// A type implementing the JSON encoding of the actual and expected
+  /// confirmation counts for a miscounted confirmation for the ABI entry point
+  /// and event stream output.
+  ///
+  /// This type is not part of the public interface of the testing library. It
+  /// assists in converting values to JSON; clients that consume this JSON are
+  /// expected to write their own decoders.
+  struct EncodedMiscount<V>: Sendable where V: ABI.Version {
```
grynspan: If you're calling it `"confirmationMiscount"`, should this type be called `EncodedConfirmationMiscount`?

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
```diff
@@ -0,0 +1,55 @@
...
+#if !SWT_NO_ABI_JSON_SCHEMA
+extension ABI {
+  /// A type implementing the JSON encoding of a range for the ABI entry point
+  /// and event stream output.
+  ///
+  /// This type is not part of the public interface of the testing library. It
+  /// assists in converting values to JSON; clients that consume this JSON are
+  /// expected to write their own decoders.
+  struct EncodedRange<V>: Sendable where V: ABI.Version {
```
grynspan, line 19: Since this type is only (currently) used in `EncodedMiscount`, should it be a nested type? In the future if we use ranges elsewhere, we can move it of course.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
grynspan: And I think one checking `exceededTimeLimit`, yes?

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
jerryjrchen: No `sourceLocation` changes in the pitch. The `EncodedIssue` type already has an optional `sourceLocation` field though

Sources/Testing/ABI/Encoded/ABI.EncodedError.swift
jerryjrchen: I'm keeping the `CodingKeys` bit to encode `var typeInfo` -> field named type but can simplify otherwise.

Sources/Testing/ABI/Encoded/ABI.EncodedExpression.swift
jerryjrchen: Simplified to just the CodingKeys changes!

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
grynspan: Right but I intentionally added `_sourceLocation` to all events because it's needed for things other than issues.

Sources/Testing/ABI/Encoded/ABI.EncodedError.swift
grynspan: But why? Just rename `var typeInfo`.

Sources/Testing/ABI/Encoded/ABI.EncodedExpression.swift
grynspan: You don't even need that though. Just rename the properties.

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
stmontgomery: I'm fine with this as-is, it seems generally useful and there have been other Encoded* types which are only used in one place but have been represented as a non-nested type (I'm thinking of `ABI.EncodedError` as one example)

Sources/Testing/ABI/Encoded/ABI.EncodedError.swift
```diff
@@ -18,8 +18,6 @@ extension ABI {
   /// This type is not part of the public interface of the testing library. It
   /// assists in converting values to JSON; clients that consume this JSON are
   /// expected to write their own decoders.
-  ///
-  /// - Warning: Errors are not yet part of the JSON schema.
   @_spi(Experimental)
```
stmontgomery: Should we remove `@_spi(Experimental)` now? (Leaving only the inherited `@_spi(ForToolsIntegrationOnly)`)

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
```diff
@@ -206,18 +204,31 @@ extension ABI {
         iteration = eventContext.iteration
       }
 
+      // Fields introduced in 6.5
+
```
stmontgomery:
```suggestion
```

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
stmontgomery: The proposal this PR is implementing is currently scoped to issues (its title is _"Include additional issue metadata in event stream"_). I agree we ought to expose this field, but it seems like we should defer that to one of our anticipated follow-on proposals.

Sources/Testing/ABI/Encoded/ABI.EncodedExpression.swift
```diff
@@ -16,8 +16,6 @@ extension ABI {
   /// This type is not part of the public interface of the testing library. It
   /// assists in converting values to JSON; clients that consume this JSON are
   /// expected to write their own decoders.
-  ///
-  /// - Warning: Expressions are not yet part of the JSON schema.
   @_spi(Experimental)
```
stmontgomery:
```suggestion
```

stmontgomery (commented):
Looking good, thanks! Especially appreciate the very thorough tests

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
grynspan: `comments` begs to differ.

grynspan: As does `[messages]`.

stmontgomery: Ah, fair enough. Are you suggesting, then, that the proposal remove `sourceLocation` from `<issue>` and add it to `<event>` (i.e. hoist it up)?

grynspan: Indeed!

Sources/Testing/ABI/Encoded/ABI.EncodedExpression.swift
jerryjrchen: Hmm, but wouldn't it be better to preserve the more descriptive `typeInfo` and `runtimeValue` property names?

Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
jerryjrchen: You mean like an `<SomeRangeType>.init(decoding:)`, right? Could do something shaped like:

```swift
static func decode() -> any RangeExpression
```

Although I don't know how useful that would be.

Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
jerryjrchen: This is getting passed the `expected` range from `Issue.Kind`:

```swift
    indirect case confirmationMiscounted(actual: Int, expected: any RangeExpression & Sendable)
```

Since we're only given an existential type of `any RangeExpression` to pass to the initialiser, I'm not sure I can actually change it to `some RangeExpression<Int>` here.

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
jerryjrchen: I'll revert to the original formatting. I will point out though that this is the default formatting offered by `swift-format`

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
jerryjrchen: I'm using it here for the following and I'm not sure there's a way to automatically get the last two:
- Explicitly excludes `knownIssueComment` as an encoded field (could keep as an underscored field)
- Encoding `isKnown` as either a Bool or a String depending on if there's a comment
- Omitting `isKnown` if the value is false + no comment

Sources/Testing/ABI/Encoded/ABI.EncodedIssue.swift
jerryjrchen: Good catch, thanks. I'm updating the implementation and adding some tests to catch these cases.

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
jerryjrchen: Hmm, since the range is inclusive I can't really represent that upper bound. I'll make it a failable init to account for that.

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
jerryjrchen: Isn't that already covered by the `#if !SWT_NO_ABI_JSON_SCHEMA` compilation guard on the whole file?

Sources/Testing/ABI/Encoded/ABI.EncodedEvent.swift
jerryjrchen: Proposal has been updated. PR update will follow soon

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
jerryjrchen: I'm leaning towards a separate file since it's a not-insignificant amount of code

Sources/Testing/ABI/Encoded/ABI.EncodedRange.swift
jerryjrchen: Updated to `return nil`

Sources/Testing/ABI/Encoded/ABI.EncodedConfirmationMiscount.swift
jerryjrchen: If I'm understanding correctly, is this something like a custom codable conformance where the encoding/decoding depends on the bounds?

```
EncodedRange(min:1, max:10) <->  { "min": 1, "max": 10 }

EncodedRange(min:10, max:10) <-> 10
```

I would find it unexpected if a range ever encoded to that single int, even if it's a single item range. Might be worth the tradeoff though

grynspan:
Please adjust your changes to adopt `JSON.Encodable`. DM me for help if needed.

PR #1844, opened 2026-08-19
Remove some old workarounds.

grynspan:
Remove workarounds for `Error` not having underscored properties in Embedded Swift (fixed in https://github.com/swiftlang/swift/pull/88954) and for the lack of constrained existential casts (our minimum deployment targets support it now).

### Checklist:

### DIFF
--- a/Sources/Testing/Issues/Issue.swift
+++ b/Sources/Testing/Issues/Issue.swift
@@ -262,18 +262,6 @@ extension Issue: CustomStringConvertible, CustomDebugStringConvertible {
   }
 }
 
-#if !hasFeature(Embedded)
-/// An empty protocol defining a type that conforms to `RangeExpression<Int>`.
-///
-/// In the future, when our minimum deployment target supports casting a value
-/// to a constrained existential type ([SE-0353](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0353-constrained-existential-types.md#effect-on-abi-stability)),
-/// we can remove this protocol and cast to `RangeExpression<Int>` instead.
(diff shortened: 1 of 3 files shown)


Sources/Testing/Issues/Issue.swift
```diff
@@ -262,18 +262,6 @@ extension Issue: CustomStringConvertible, CustomDebugStringConvertible {
   }
 }
 
-#if !hasFeature(Embedded)
-/// An empty protocol defining a type that conforms to `RangeExpression<Int>`.
-///
-/// In the future, when our minimum deployment target supports casting a value
-/// to a constrained existential type ([SE-0353](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0353-constrained-existential-types.md#effect-on-abi-stability)),
```
stmontgomery, line 269: Ok just checking: I didn't think our deployment target was high enough to remove this workaround yet, however when I looked, I realized it was the _similar_ protocol named `ExpectedCount` in a test file `ConfirmationTests.swift` which we cannot (yet) remove because it was working around a crash using this runtime cast feature in conjunction with collection APIs (rdar://163980446). Does that match your understanding?

(To be clear, I've concluded this is safe as-is, I just wanted to confirm because I initially thought it wasn't)

stmontgomery (approved):

Sources/Testing/Issues/Issue.swift
grynspan: Yes, however if you think the crash is still occurring and would affect us around this change, I can leave it alone.

PR #1848, opened 2026-08-21
Add new "metadata recorded" event.

grynspan:
This PR breaks out the metadata details we collect in the human-readable output recorder and instead emits them as separate `.metadataRecorded` events.

This allows us to collect this data even when not emitting a human-readable output stream. If/when we add it to the JSON schema, consuming tools can gather that information and present it in their own interfaces.

I've taken (some) care to ensure that, in existing JSON schemas, the `messages` field for the `.runStarted` event kind still contains this metadata. In the experimental schema, which does not generate the `messages` field by default, they are instead broken out into a new record kind, `"_metadata"`.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/ABI.Record+Streaming.swift
+++ b/Sources/Testing/ABI/ABI.Record+Streaming.swift
@@ -18,19 +18,14 @@ extension ABI.Version {
       humanReadableOutputRecorder = Event.HumanReadableOutputRecorder()
     }
     return { [humanReadableOutputRecorder] event, context in
-      if case .testDiscovered = event.kind, let test = context.test {
-        let testRecord = ABI.Record<Self>(encoding: test)
-        recordHandler(testRecord)
-      } else {
-        var messages: [Event.HumanReadableOutputRecorder.Message] = []
-        if let humanReadableOutputRecorder {
-          var configuration = Configuration()
-          configuration.verbosity = 0
-          messages = humanReadableOutputRecorder.record(event, in: context, configuration: configuration)
-        }
-        if let eventRecord = ABI.Record<Self>(encoding: event, in: context, messages: messages) {
-          recordHandler(eventRecord)
-        }
+      var messages: [Event.HumanReadableOutputRecorder.Message] = []
+      if let humanReadableOutputRecorder {
+        var configuration = Configuration()
+        configuration.verbosity = 0
+        messages = humanReadableOutputRecorder.record(event, in: context, configuration: configuration)
+      }
+      if let record = ABI.Record<Self>(encoding: event, in: context, messages: messages) {
+        recordHandler(record)
       }
     }
   }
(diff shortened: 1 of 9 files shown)


Sources/Testing/Events/Event.Metadata.swift
```diff
@@ -0,0 +1,97 @@
...
+  public var description: String {
+    "\(name): \(value)"
+  }
+}
+
+// MARK: - "Standard" metadata
+
+extension Event.Metadata {
+  /// The names of various metadata that the testing library collects.
+  private static var _swiftStandardLibraryVersionName: String { "Swift Standard Library Version" }
```
harlanhaskins, line 59: I am the teeniest bit concerned about these names being display names vs identifiers, do we want clients relying on display names? If they, for example, wanted to compose a string like, `(swiftlang: \(compilerVersion), stdlib: \(stdlibVersion), running on \(platform) (version \(osVersion)))"`, they'd have to reach into the metadata and look by display names. Is this something we intend for clients to be able to do?

grynspan: It's meant to be human-readable only, which is mentioned in a few places. I can add it to more places?

harlanhaskins: I think it's understandable from the current comments that the _values_ are human-readable, but not clear that the names are also human-readable. A doc comment on the encoded event's `name` property would be fine, I think.

harlanhaskins (approved):
This looks good now, thanks for the doc comment clarification

Sources/Testing/ABI/Encoded/ABI.EncodedMetadata.swift
```diff
@@ -0,0 +1,56 @@
...
+#if !SWT_NO_ABI_JSON_SCHEMA
+@_spi(Experimental)
+extension ABI {
+  /// A type implementing the JSON encoding of ``Event/Metadata`` for the ABI
+  /// entry point and event stream output.
+  ///
+  /// This type is not part of the public interface of the testing library. It
+  /// assists in converting values to JSON; clients that consume this JSON are
+  /// expected to write their own decoders.
+  public struct EncodedMetadata<V>: Sendable where V: ABI.Version {
```
stmontgomery, line 22:
```swift
  /// - Warning: Metadata is not yet part of the JSON schema.
```

Sources/Testing/ABI/ABI.Record+Streaming.swift
```diff
@@ -18,19 +18,14 @@ extension ABI.Version {
       humanReadableOutputRecorder = Event.HumanReadableOutputRecorder()
     }
     return { [humanReadableOutputRecorder] event, context in
-      if case .testDiscovered = event.kind, let test = context.test {
```
stmontgomery, line 21: Sorry if I'm overlooking it, but is there some other change in this PR which facilitates removing the `if case .testDiscovered = event.kind` condition here?

stmontgomery (approved):

Sources/Testing/ABI/ABI.Record+Streaming.swift
grynspan: It's handled in `ABI.Record.init(encoding:in)` now.

grynspan:
Linux failure is unrelated. Filed rdar://185716260.

PR #1849, opened 2026-08-24
Add a dedicated argument parser system.

grynspan:
This PR adds `CommandLineArgumentList`, a type for parsing and interpreting command-line arguments that provides a (very constrained!) subset of the sort of functionality you'd find in Swift Argument Parser.

Swift Testing cannot directly link to Swift Argument Parser because it would cause cyclical dependencies in Swift Argument Parser's own test suites, and because Swift Argument Parser is not formally included in the Swift toolchain[^everybodyUsesItAnyway].

We _do_ need to parse command-line arguments, however, so we need a mechanism for doing so. Right now, we have logic in the exported entry point function that does some basic parsing and validation, but it's very naïve (even more so than this new implementation) and cannot be used with other targets in Swift Testing that might need to parse arguments such as the experimental testing harness in
 #1784.

[^everybodyUsesItAnyway]: A number of components in the toolchain directly link to their own copies of Swift Argument Parser. When they build using CMake, they end up building Swift Argument Parser multiple times. Improving that is beyond the scope of this PR or Swift Testing's mandate.

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/EntryPoints/CommandLineArgumentList.swift
+++ b/Sources/Testing/ABI/EntryPoints/CommandLineArgumentList.swift
@@ -1,0 +1,320 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2023–2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+/// A type that can perform basic parsing of command-line arguments.
+///
+/// The testing library uses this type to parse command-line arguments in its
+/// entry point function and in its associated executable targets.
+///
+/// - Important: This type is provided because the testing library cannot link
+///   directly to Swift Argument Parser. Other tools and libraries should
+///   generally prefer to use Swift Argument Parser and should not use or copy
+///   this type or its logic.
+@_spi(ForToolsIntegrationOnly)
+public struct CommandLineArgumentList: Sendable {
+  /// An enumeration describing how to handle a command-line argument.
+  public enum Descriptor: Sendable, Equatable, Hashable {
+    /// An argument that has no label.
+    case anonymous
+
+    /// A subcommand argument (one that does not take a value and does not start
+    /// with a dash character).
+    ///
+    /// - Parameters:
+    ///   - commandName: The name of the subcommand.
+    ///
+    /// The testing library does not enforce that subcommand arguments must come
+    /// before other arguments.
+    case subcommand(_ commandName: String)
+
+    /// An argument that takes a value.
+    ///
+    /// - Parameters:
+    ///   - label: The option's label, including leading dashes.
+    ///
+    /// If an option is specified more than once, the testing library builds an
+    /// array of values for that option in the order the values are parsed.
+    case option(_ label: String)
+
+    /// An argument that does not take a value.
+    ///
+    /// - Parameters:
+    ///   - label: The flag's label, including leading dashes.
+    ///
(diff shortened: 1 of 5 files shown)


grynspan:
Linux failure is unrelated.

Sources/Testing/ABI/EntryPoints/CommandLineArgumentList.swift
```diff
@@ -0,0 +1,320 @@
...
+  /// - Returns: A tuple containing the argument descriptor matching `arg` and,
+  ///   if `arg` is anonymous or is an option, its value.
+  ///
+  /// - Throws: Any error that occurs while parsing `arg`.
+  private static func _parseArgument(
+    _ arg: String,
+    describedBy descriptors: [Descriptor],
+    describingUnrecognizedArgumentWith describeUnrecognizedArgument: ((String) throws -> Descriptor)?,
+    gettingValueWith getValue: () -> String?
+  ) throws -> (descriptor: Descriptor, value: String?) {
```
plemarquand, line 98: Can we use a dedicated `_ParsedArgument` struct to remove the ambiguity around the value's optionality? You'd be able to remove the force unwraps on `value` in `init`.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -1020,6 +995,12 @@ private enum _EntryPointError: Error {
   /// - Parameters:
   ///   - versionNumber: The experimental ABI version number.
   case experimentalABIVersion(_ versionNumber: VersionNumber)
+
+  /// A failure occurred while parsing the command line arguments list.
+  ///
+  /// - Parameters:
+  ///   - error: The underlying error that occurred.
+  case commandLineParsingFailed(_ error: CommandLineArgumentList.ParseError)
```
plemarquand, line 1003: Nothing seems to leverage this (yet?)

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
```diff
@@ -427,7 +388,24 @@ func parseCommandLineArguments(from args: [String]) throws -> __CommandLineArgum
...
+      .option("--xunit-output"),
+      .option("--attachments-path"), .option("--experimental-attachments-path"),
+      .flag("--parallel"), .flag("--no-parallel"),
+      .option("--experimental-maximum-parallelization-width"),
+      .option("--symbolicate-backtraces"),
+      .option("--verbosity"), .flag("--verbose"), .flag("-v"), .flag("--very-verbose"), .flag("--vv"), .flag("--quiet"), .flag("-q"),
+      .option("--filter"), .option("--skip"),
+      .option("--repetitions"), .option("--repeat-until"),
+    ],
+    describingUnrecognizedArgumentWith: { _ in .anonymous }
```
plemarquand, line 407: Do we need a closure here? The way this is used it could be modelled as an enum of `.anonymous` or `.throwing`.

Sources/Testing/ABI/EntryPoints/CommandLineArgumentList.swift
grynspan: I think that's overkill right now. Since this isn't API, we can change it in the future if we want.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: A caller may choose to handle other arguments differently. The current code doesn't, but I am leaving it open for now.

Sources/Testing/ABI/EntryPoints/EntryPoint.swift
grynspan: Yet.

stmontgomery (approved):

PR #1857, opened 2026-08-27
Cache a shared instance of `JSONEncoder` and `JSONDecoder`.

grynspan:
This PR uses a single instance of `JSONEncoder` and of `JSONDecoder` for most of our encoding/decoding operations. These classes are thread-safe and can be used concurrently.

This reduces the time to e.g. decode 1,000,000 JSON records from about 5.2s to about 4.5s on my system in a debug build, so an approximately 13% speedup.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/JSON.swift
+++ b/Sources/Testing/Support/JSON.swift
@@ -29,6 +29,22 @@ enum JSON {
   private static let _negativeInfinityString = "-Infinity"
   private static let _nanString = "NaN"
 
+  /// A JSON encoder to use with the default encoding configuration.
+  ///
+  /// We reuse this object to avoid allocating encoders repeatedly.
+  private static let _defaultEncoder: JSONEncoder = {
+    let encoder = JSONEncoder()
+
+    // Keys must be sorted to ensure deterministic matching of encoded data.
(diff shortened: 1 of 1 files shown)


jerryjrchen:
> These classes are thread-safe and can be used concurrently.

Is this based on both classes conforming to `Sendable`?

> This reduces the time to e.g. decode 1,000,000 JSON records from about 5.2s to about 4.5s on my system in a debug build, so an approximately 13% speedup.

I'm also curious if this approximate speedup applied to encoding as well.

jerryjrchen (approved):

stmontgomery (approved):

grynspan:
> > These classes are thread-safe and can be used concurrently.
> 
> Is this based on both classes conforming to `Sendable`?

Yes, and we know we're only using the base classes so we don't need to be concerned with subclasses sneaking in and ruining our day.

> > This reduces the time to e.g. decode 1,000,000 JSON records from about 5.2s to about 4.5s on my system in a debug build, so an approximately 13% speedup.
> 
> I'm also curious if this approximate speedup applied to encoding as well.

By about the same amount, yes.

PR #1859, opened 2026-08-27
Promote `#sourceLocation` to API.

grynspan:
As promised! For [ST-0027](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0027-sourcelocation-macro.md).

### Checklist:

### DIFF
--- a/Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
+++ b/Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
@@ -25,10 +25,15 @@
 /// default argument to a function.
 ///
 /// ```swift
-/// func cookBurger(sourceLocation: SourceLocation = #_sourceLocation) {
+/// func cookBurger(sourceLocation: SourceLocation = #Testing::sourceLocation) {
 ///   // ...
 /// }
 /// ```
+///
+/// @DeprecationSummary {
+///   Use [`#Testing::sourceLocation`](``Testing/sourceLocation()``) instead.
+/// }
+@available(swift, deprecated: 100000.0, message: "Use '#Testing::sourceLocation' instead.")
 @freestanding(expression) public macro _sourceLocation() -> SourceLocation = #externalMacro(module: "TestingMacros", type: "SourceLocationMacro")
 
 /// Get the current source location.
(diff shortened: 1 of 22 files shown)


Sources/Testing/Testing.docc/Expectations.md
```diff
@@ -95,6 +95,6 @@ the test when the code doesn't satisfy a requirement, use
 ### Representing source locations
 
 - ``SourceLocation``
-<!-- - ``sourceLocation()`` -->
+- ``sourceLocation()``
```
suzannaratcliff, line 98: The style of this looks a little different, is this expected?

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
```diff
@@ -25,10 +25,15 @@
...
-/// func cookBurger(sourceLocation: SourceLocation = #_sourceLocation) {
+/// func cookBurger(sourceLocation: SourceLocation = #Testing::sourceLocation) {
 ///   // ...
 /// }
 /// ```
+///
+/// @DeprecationSummary {
+///   Use [`#Testing::sourceLocation`](``Testing/sourceLocation()``) instead.
+/// }
+@available(swift, deprecated: 100000.0, message: "Use '#Testing::sourceLocation' instead.")
```
suzannaratcliff, line 36: Why is this 100000.0?  I do see it used other places in the code as well

suzannaratcliff (approved):

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
```diff
@@ -25,10 +25,15 @@
...
 @freestanding(expression) public macro _sourceLocation() -> SourceLocation = #externalMacro(module: "TestingMacros", type: "SourceLocationMacro")
```
stmontgomery, line 37:
```suggestion
@freestanding(expression) public macro _sourceLocation() -> SourceLocation = #Testing::sourceLocation()
```

stmontgomery:
Can we please keep at least one test which intentionally calls the (now to-be-deprecated) `#_sourceLocation` macro, to verify it continues working? Or if we have such a test, can you point it out?

Sources/Testing/Testing.docc/Expectations.md
grynspan: Yes, this is the style used by DocC for symbol lists (and for macros in particular).

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
grynspan: It's equal to `API_TO_BE_DEPRECATED` and has special handling in Xcode to treat it as deprecated without being noisy about it.

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
grynspan: This breaks the build. :)

grynspan:
> Can we please keep at least one test which intentionally calls the (now to-be-deprecated) `#_sourceLocation` macro, to verify it continues working? Or if we have such a test, can you point it out?

We have one already:

```swift
  @Test("#_sourceLocation and #Testing::sourceLocation are equivalent")
  func newAndOldMacrosAreEquivalent() {
    let lhs = #_sourceLocation
    var rhs = #Testing::sourceLocation
    rhs.line -= 1
    #expect(lhs == rhs)
  }
```

stmontgomery (approved):

PR #1860, opened 2026-08-30
Support semver prerelease IDs in `ABI.VersionNumber`

grynspan:
This PR adds support for representing development versions of the JSON schema independently of release versions by specifying the `"-dev"` suffix, e.g. `--event-stream-version 6.5-dev`.

This is useful if we want to enable functionality only on a development branch, for instance. A separate PR will adopt this new logic when parsing `VERSION.txt` so we always know what our own branch is, which will make `ABI.version(forVersionNumber:)` play nicer whenever `main` moves to a new version (shout-out to @bkhouri for noticing that `"--event-stream-version 6.5"` didn't work after we branched `release/6.4` and retargetted `main` until we explicitly added `ABI.v6_5`.).

This PR adds the necessary infrastructure to support other suffixes in the future that can be used to enable or disable functionality dynamically. (No, I don't have specific ideas in mind, it's just a useful tool for the toolbox.)

### Checklist:

### DIFF
--- a/Sources/Testing/ABI/ABI.VersionNumber.swift
+++ b/Sources/Testing/ABI/ABI.VersionNumber.swift
@@ -42,12 +42,44 @@ extension ABI {
 
     /// The patch, revision, or bug fix version.
     var patchComponent: Component = 0
+
+    /// A type describing flags for version numbers.
+    ///
+    /// When converting a version number to or from a string, all flags are
+    /// represented as "pre-release identifiers" per the semantic versioning
+    /// specification.
+    ///
+    /// - Important: The semantic versioning specification requires that the
+    ///   order of prerelease IDs be preserved so that it can factor into
+    ///   version number comparisons. This type does _not_ follow that rule so
+    ///   we can save space: if we preserved ordering, the stride of
+    ///   ``ABI/VersionNumber`` would need to be at least as large as two
+    ///   pointers to hold the numeric components, the flags, and their
+    ///   ordering.
+    struct Flags: OptionSet {
+      var rawValue: Component.Magnitude
+
+      /// The owning version number represents a development build of the
+      /// testing library.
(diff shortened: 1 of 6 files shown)


Sources/Testing/ABI/ABI.VersionNumber.swift
```diff
@@ -215,4 +303,73 @@ extension ABI.VersionNumber: Codable {
...
+  /// The suffix to apply to the owning instance of ``ABI/VersionNumber``
+  /// when converting it to a string.
+  fileprivate var prereleaseIDSuffix: String {
+    if self.isEmpty {
+      return ""
+    }
+    return "-" + prereleaseIDs.joined(separator: ".")
+  }
+
+  init?<S>(prereleaseID: S) where S: StringProtocol, S.SubSequence == Substring {
```
stmontgomery, line 438: Documentation?

Sources/Testing/Events/Event.Metadata.swift
stmontgomery, line 1: I'm a little unclear how the changes in this file connect to the rest of the PR

stmontgomery (approved):

Sources/Testing/ABI/ABI.VersionNumber.swift
grynspan: Eh, seemed like a trivial initializer to me.

Sources/Testing/Events/Event.Metadata.swift
grynspan: I can move these changes out to a separate PR, but the intent is that we'll be reading `testingLibraryVersion` from `VERSION.txt` at compile time now, so at the same time I'd want to split out the commit hash to a separate row so the version string is (theoretically, not an ABI-stable thing) recoverable.

PR #1862, opened 2026-08-31
Add support for `SIGINFO`.

grynspan:
This PR adds support for `SIGINFO` when running the testing library directly from the command line via its entry point function.

On Linux, we use `SIGUSR1` which is a pseudo-convention in place of `SIGINFO`. On Windows, we listen for Ctrl+Break instead.

Note that this feature won't be user-accessible from `swift test` until/unless it's updated to forward `SIGINFO` etc. to the test process.

### Checklist:


stmontgomery (approved):
Looks great!

Since it has some new `#if`s and platform conditionals, mind doing a Swift CI build before landing?

Sources/Testing/Support/SIGINFOHandler.swift
```diff
@@ -0,0 +1,99 @@
...
+private let _all = Mutex<[Weak<SIGINFOHandler<Void>>]>()
+
+/// The dispatch source that listens for `SIGINFO` (or the platform-specific
+/// equivalent).
+///
+/// This declaration is annotated `nonisolated(unsafe)` because dispatch sources
+/// do not conform to `Sendable` on non-Darwin targets.
+private nonisolated(unsafe) let _siginfoSource: Any = {
+#if SWT_TARGET_OS_APPLE || os(FreeBSD) || os(OpenBSD)
+  let source = DispatchSource.makeSignalSource(signal: SIGINFO, queue: .main)
```
briancroom, line 42: What's the thinking behind using the main queue for this?

In a "my tests are stuck" situation, it seems like that's the execution context with the greatest risk of being involved in the deadlock/bug, and having the handlers never getting a chance to run.

briancroom (approved):

Sources/Testing/Support/SIGINFOHandler.swift
grynspan: I've actually already changed it over in the swift-argument-parser PR for this reason. `.main` was just out of convenience.

grynspan:
Moving to draft as I think I'll want to have this only in the harness layer and want to see what shakes out of Swift Argument Parser.

PR #1866, opened 2026-09-03
Remaining Embedded Swift support work

grynspan:
Most of the remaining work for Embedded Swift lieth here:

- Get `@Test` compiling, including basic support for parameterized tests.
  - `#if` out almost all code around parameter value capture and stringification (will revisit in the future though, I have some ideas about how I might at least partially resolve this constraint).
  - Mark `@Suite` unavailable.
- Provide the C declaration for `_swift_testing_writeJSON()`.
- Update TestContent.md and StyleGuide.md to cover some Embedded Swift topics.
- Fix some functional calls that the Embedded compiler doesn't like (`.map` etc.)
- Workaround for a lack of `sterror()` (although I'd like the core PAL to provide a function for this).
- Mark `.serialized` as unavailable.


Sources/_TestingInternals/include/EmbeddedPlatform.h
```diff
@@ -205,6 +205,49 @@ SWT_EXTERN SWT_NODISCARD bool _swift_testing_getConsoleCapabilities(swift_testin
...
+/// flockfile(f); {
+///   fwrite(json, 1, count, f);
+///   if (terminator) {
+///     fputc(*terminator, f);
+///   }
+/// } funlockfile(f);
+/// ```
+///
+/// If your platform does not support writing JSON or consuming it later, you
+/// can implement this function as a no-op.
```
stmontgomery, line 241: If a platform doesn't support writing JSON, would we want some other PAL annex function to indicate that? I'm wondering about the overhead of serializing the JSON content if it's not going to ultimately be written anywhere.

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
```diff
@@ -130,16 +130,22 @@ extension ABI {
...
 
       // TODO: define an encodable form of Test.Case.ID
       id = String(describing: testCase.id)
+#if !hasFeature(Embedded)
       displayName = arguments.lazy
         .map { $0.value }
         .map(String.init(describingForTest:))
         .joined(separator: ", ")
+#else
+      displayName = ""
```
stmontgomery, line 147: Could we print some unknown sentinel instead of empty?
```suggestion
      displayName = "<unknown arguments>"
```

Sources/Testing/Test+Macro.swift
```diff
@@ -144,12 +157,17 @@ public macro Test(
   _ traits: any TestTrait...
 ) = #externalMacro(module: "TestingMacros", type: "TestDeclarationMacro")
 
+// TODO: @Suite support in some form
```
stmontgomery: I'm unsure why this TODO comment belongs here, in addition to the same one above

Sources/_TestingInternals/include/EmbeddedPlatform.h
grynspan: I considered it, but decided instead to go with some sort of configuration function that could say "enable JSON output"; if they don't then enable it, we don't emit the function call after optimization and dead code stripping.

grynspan: (And adding that function requires some dedicated thought which is why I didn't include it here.)

Sources/Testing/Test+Macro.swift
grynspan: I think it's just a bad rebase? Will adjust.

stmontgomery (approved):

stmontgomery:
LGTM, modulo a few comments above

Sources/Testing/ABI/Encoded/ABI.EncodedTest.swift
grynspan: Nothing actually uses this string as it's on the experimental path and our "real" implementation of test case JSON is going to be wildly different. So I'm not worried.

PR #1870, opened 2026-09-04
Add custom event symbols with custom colours.

grynspan:
This change adds a `case custom` to `Event.Symbol`. This type is used internally to render test diamonds, checkmarks, etc. before console output. I intend to use custom symbols to augment the existing output over in yonder experimental testing harness.

I've also done some refactoring to reduce our direct dependence on ANSI escape codes. I've moved the `Tag.Color` type up to just `Color` and repurposed it for console output (in addition to the experimental tag color feature). I'm now using the `Color` type to represent colours and transforming them to ANSI escape codes on demand. (There's no measurable performance cost there since it's just a bit of math and is swamped by the cost of I/O to `stderr`.) Built-in symbols continue to use 4-bit ANSI escape codes, but custom symbols will render at up to 24 bits depending on the terminal app's capabilities.

### Checklist:


Sources/Testing/Events/Recorder/Event.Symbol.swift
```diff
@@ -76,15 +97,18 @@ extension Event.Symbol {
...
   }
 
 #if os(macOS) || (os(iOS) && targetEnvironment(macCatalyst))
   /// The SF&nbsp;Symbols character corresponding to this instance.
   ///
-  /// This property is not part of the public interface of the testing library.
-  /// Developers should use ``sfSymbolName`` instead.
-  var sfSymbolCharacter: Character {
+  /// - Important: This character only renders as an [SF&nbsp;Symbol](https://developer.apple.com/sf-symbols/)
+  ///   if the user has the SF&nbsp;Symbols app installed. Developers should
```
plemarquand, line 109: &nbsp in SF Symbols to keep the name together in the docs is truly a nice touch.

Sources/Testing/Events/Recorder/Event.ConsoleOutputRecorder.swift
```diff
@@ -75,14 +81,14 @@ extension Event {
       /// The value of this property is ignored unless the value of
       /// ``useANSIEscapeCodes`` is `true` and the value of
       /// ``ansiColorBitDepth`` is greater than `1`.
```
plemarquand, line 83: Entrypoint.swift seems to remove this greater than 1 check

grynspan: It's just part of `useColorANSIEscapeCodes` now, since `> 1` is equivalent to `>= 4` (there is no 3-bit color mode).

Sources/Testing/Events/Recorder/Event.Symbol.swift
grynspan: It's an old DevPubs trick!

grynspan: Back in the day, "Mac OS X" was supposed to be spelled `Mac&nbsp;OS&nbsp;X`. Not that anybody did it outside apple.com.

briancroom (approved):

stmontgomery (approved):

Issue #155, opened 2023-12-11
Support for expected failures


dabrahams:
### Description

Users shouldn't have to concoct their own versions of [this facility](https://github.com/loftware/CheckXCAssertionFailure).  You're going to need it for your own tests anyway; it should be exposed.

stmontgomery:
Thanks @dabrahams. swift-testing currently has a set of APIs named `withKnownIssue` which are meant to serve this purpose. See documentation at:

>https://swiftpackageindex.com/apple/swift-testing/0.1.0/documentation/testing/expectthrows#recording-known-issues-in-tests

Can you check whether `withKnownIssue` meets your needs? If not, I'd love to know what functionality it's missing.

dabrahams:
I think while those _might_ _do_ what I am asking for (I am not sure), their names certainly imply that these issues are supposed to get fixed at some point.  The facility I'm talking about is expressly for testing that some testing facility works, and should reflect the intent that the proper operation of that facility under the circumstances is to report a test failure.  The intention expressed by a piece of test code is important, and I would say `withKnownIssue` expresses the wrong thing to a reader.

stmontgomery:
Any chance you could provide a small example of the usage scenario you have in mind? Just to help us understand your use case a bit better. Are you attempting to test a library which provides testing utilities of its own?

dabrahams:
@stmontgomery Did you look at the README on the page I linked to?  I think it answers all those questions and indeed describes some behavior you might want for `withKnownIssue` as well. Is there some information missing from the README that I can add?

You can see more robust examples of where this function is useful in these [tests of the StandardLibraryProtocolChecks package](https://github.com/loftware/StandardLibraryProtocolChecks/blob/main/Tests/CollectionTests.swift).

grynspan:
Do you have a specific interface in mind here, @dabrahams?

dabrahams:
Not really; I haven't internalized your interface style.

stmontgomery:
Thank you. I think we understand the use case better here, now. We think it _could_ be useful to improve this area, but consider this fairly niche and not a priority for the project while it's still in an early stage of development. We will hold this Issue for further consideration, and in the mean time, if you would like to propose a specific SPI or API enhancement along these lines, we can consider it further. I'll also re-title this issue to try to clarify how it's different than a general-purpose "expected failure" API, which for most users is `withKnownIssue`.

[...]

PR #1871, opened 2026-09-06
Add `withExpectedIssue` for validating the content of recorded Issues (#155)

Aaravanand00:
Introduce `withExpectedIssue(_:isIntermittent:sourceLocation:_:matching:)` and its async counterpart as `@_spi(Experimental)` API.

### Motivation :- 

When writing tests for custom testing utilities (e.g. a helper that calls `Issue.record()`), there was no first-class way to assert that the helper correctly records an `Issue` and to inspect the recorded content.

`withKnownIssue()` is semantically wrong for this purpose because it implies a bug to be fixed in the future, whereas the intent here is that recording an issue is the **correct and intended behavior** of the code under test. This is the use case tracked in #155.

### Modifications:

The new API follows the same `@TaskLocal` scope pattern as `KnownIssueScope`:

- **`ExpectedIssueScope`** (internal) is pushed onto the task-local stack by `withExpectedIssue`. Any `Issue` that passes the caller-supplied matcher is suppressed (not forwarded to the event handler or `failureBreakpoint`) and the first captured issue is returned to the caller for content inspection.

- **`Issue.record()`** gains an 8-line guard that checks `ExpectedIssueScope.current` before posting the event, mirroring the existing `KnownIssueScope` check. No other existing logic is modified.

- If no matching issue is recorded and `isIntermittent` is `false`, a `.knownIssueNotRecorded` issue is recorded so the test fails visibly consistent with `withKnownIssue` behavior.

**Files changed:**
- `Sources/Testing/Issues/ExpectedIssue.swift` - new file with `ExpectedIssueScope` + sync/async `withExpectedIssue`
- `Sources/Testing/Issues/Issue+Recording.swift` - +8 lines in `record()`
- `Tests/TestingTests/ExpectedIssueTests.swift` - new file with 12 tests covering all behaviors

### Checklist:

- [x] If public symbols are renamed or modified, DocC references should be updated....

### DIFF
--- a/Sources/Testing/Issues/ExpectedIssue.swift
+++ b/Sources/Testing/Issues/ExpectedIssue.swift
@@ -1,0 +1,236 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+#if canImport(Synchronization)
+private import Synchronization
+#endif
+
+/// A type that represents an active ``withExpectedIssue(_:isIntermittent:sourceLocation:_:matching:)``
+/// call.
+///
+/// When a test function calls ``withExpectedIssue(_:isIntermittent:sourceLocation:_:matching:)``,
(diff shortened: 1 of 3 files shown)


grynspan:
Thank you for your PR. This appears to just be copied whole-cloth from `withKnownIssue()`.

I would recommend you engage with us on the Swift forums first. New API in particular must go through the Swift Evolution process.

PR #1874, opened 2026-09-10
Replace our uses of `AnySequence` with a custom type-erasing box.

grynspan:
This PR replaces our use of `AnySequence` in `Test` with a custom type `AnySendableSequence` that conforms to `Sequence` and `Sendable` and type-erases an underlying sequence type.

`AnySendableSequence` is a structure (whereas `AnySequence` is backed by a class) which saves us from having to heap-allocate the `AnySequence`. The underlying existential in `AnySendableSequence` might still heap-allocate if its concrete type is more than 24 bytes long, but it doesn't need to for smaller types and _in particular_ doesn't impose an extra heap allocation for unparameterized tests (`CollectionOfOne<Void>` has a `size` of `0` and `stride` of `1`). `Array` and `Dictionary` both also fit in an existential's inline storage (they store their contents out-of-line) so we eliminate the extra heap allocation for them too.

This also helps us in Embedded Swift where there are significantly more limits on what you can do with an existential box. The implementation of `AnySendableSequence` already has `#if` branches for Embedded Swift. That implementation has different heap allocation characteristics out of necessity, but _c'est la vie_.

### Checklist:

### DIFF
--- a/Sources/Testing/CMakeLists.txt
+++ b/Sources/Testing/CMakeLists.txt
@@ -93,6 +93,7 @@ add_library(Testing
   Support/Additions/MutexAdditions.swift
   Support/Additions/NumericAdditions.swift
   Support/Additions/ResultAdditions.swift
+  Support/Additions/SequenceAdditions.swift
   Support/Additions/TaskAdditions.swift
   Support/Additions/WinSDKAdditions.swift
   Support/Color/Color.swift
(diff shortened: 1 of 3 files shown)


harlanhaskins (approved):
Understandable, if unfortunate. Thanks for pulling on the threads to minimize allocations unless necessary.

PR #1880, opened 2026-09-10
Implement `Array.binarySearch(_:)` in Swift.

grynspan:
We currently use the C standard library's `bsearch()` function to implement a binary search. That's _fine_, but `bsearch()` may not be implemented on embedded platforms.

I've ~~stolen~~ borrowed a Swift implementation from swift-collections. This ought to have at least the same performance as `bsearch()`, but probably better because it can be inlined and optimized.

The implementation can be adapted to any `RandomAccessCollection` or to any sequence with contiguous storage if needed in the future.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/Additions/ArrayAdditions.swift
+++ b/Sources/Testing/Support/Additions/ArrayAdditions.swift
@@ -10,14 +10,6 @@
 
 private import _TestingInternals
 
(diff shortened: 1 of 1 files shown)


harlanhaskins (approved):
🎉

PR #1881, opened 2026-09-10
Abstract calls to `FileHandle.stderr.write()`.

grynspan:
This PR abstracts our console output, which we currently do by writing to `stderr`, so that they can be implemented with a different destination. Embedded Swift in particular does not guarantee the existence of `stderr`.

We accomplish this noble goal by introducing what I'm terming an _annex_[^annexWat] to the Platform Abstraction Layer introduced [here](https://forums.swift.org/t/introducing-the-embedded-swift-platform-abstraction-layer/89248). The Platform Abstraction Layer testing annex ("Annex T" for short? Anybody?) will include those functions that are needed in Embedded Swift but which we cannot implement ourselves because they depend on target-specific functionality.

[^annexWat]: The C language uses the term "annex" to refer to supplemental and generally optional components of the language specification such as the infamous "Annex K" describing nominally secure interfaces that nobody implements.

The first member of the annex is `_swift_testing_writeToConsole()` which, as the name suggests, writes to the "console", whatever that might be. There's also a wrapper around this function on the Swift side, `writeToConsole(_:useStandardOutputIfAvailable:)`, that takes a Swift string; in non-Embedded Swift, when `SWT_NO_FILE_IO` is not set, this wrapper just writes to `stderr` as we did previously. Call sites then just need to pass the string to this single function.

### Checklist:

### DIFF
--- a/Sources/Testing/EmbeddedPlatform.swift
+++ b/Sources/Testing/EmbeddedPlatform.swift
@@ -1,0 +1,42 @@
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+//
+
+internal import _TestingInternals
+
+/// This file contains abstractions over functionality that, under Embedded
+/// Swift, is provided by Swift Testing's Platform Abstraction Layer annex.
+
+/// Writes a Swift string as UTF-8 to the current system's console.
+///
+/// - Parameters:
+///   - string: The string to write.
+///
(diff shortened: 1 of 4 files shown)


Sources/Testing/EmbeddedPlatform.swift
```diff
@@ -0,0 +1,42 @@
...
+/// This file contains abstractions over functionality that, under Embedded
+/// Swift, is provided by Swift Testing's Platform Abstraction Layer annex.
+
+/// Writes a Swift string as UTF-8 to the current system's console.
+///
+/// - Parameters:
+///   - string: The string to write.
+///
+/// The testing library uses this function to write a _human-readable_
+/// transcript of a test run. This function is a convenience over
```
harlanhaskins: tiny nitpick, not important: on non-embedded, this isn't true. Should we call that out or no?

harlanhaskins (approved):

PR #1883, opened 2026-09-10
Don't use keypaths as arguments to functional monady things like `map(_:)`.

grynspan:
Embedded Swift does not like using keypaths as function references, so it errors on patterns like `array.map(\.whatever)`. Change these uses to `{ $0.whatever }` to satisfy the Embedded Swift compiler.

Don't bother touching the macros target since it always runs host-side.

### Checklist:

### DIFF
--- a/Sources/Overlays/_Testing_AppKit/Attachments/NSImage+AttachableAsImage.swift
+++ b/Sources/Overlays/_Testing_AppKit/Attachments/NSImage+AttachableAsImage.swift
@@ -79,7 +79,7 @@ extension NSImage: AttachableAsImage, AttachableAsCGImage {
 
     // Check whether the image contains any representations that we don't think
     // are safe. If it does, then make a "safe" copy.
-    let allImageRepsAreSafe = representations.allSatisfy(\.isEffectivelySendable)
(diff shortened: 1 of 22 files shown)


harlanhaskins (approved):

Sources/Overlays/_Testing_AppKit/Attachments/NSImage+AttachableAsImage.swift
```diff
@@ -79,7 +79,7 @@ extension NSImage: AttachableAsImage, AttachableAsCGImage {
 
     // Check whether the image contains any representations that we don't think
     // are safe. If it does, then make a "safe" copy.
-    let allImageRepsAreSafe = representations.allSatisfy(\.isEffectivelySendable)
+    let allImageRepsAreSafe = representations.allSatisfy { $0.isEffectivelySendable }
```
stmontgomery, line 82: We don't _need_ to change these ones which are in cross-import overlays for high-level ~~Apple~~ frameworks that aren't relevant in Embedded mode, do we?

grynspan: Probably not, but then, who's to say somebody doesn't want to use the overlay in Embedded Swift someday? :)

PR #1888, opened 2026-09-11
Start enabling writing JSON from Embedded Swift.

grynspan:
This PR starts plumbing through the logic to write the JSON event stream under Embedded Swift. Since we don't have arbitrary file handles here, we provide a new PAL annex function `_swift_testing_writeJSON()` that is responsible for routing the JSON to the right place.

At this time, we haven't yet decoupled the JSON event stream Swift structures (`ABI.Record` etc.) from `Codable`, so this code is nonfunctional. Decoupling will come in a future PR, as will support for actually encoding JSON in the first place.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/FileHandle.swift
+++ b/Sources/Testing/Support/FileHandle.swift
@@ -586,6 +586,30 @@ extension FileHandle {
     }
   }
 
+  /// Write a single byte to this file handle.
+  ///
+  /// - Parameters:
+  ///   - byte: The byte to write.
+  ///   - flushAfterward: Whether or not to flush the file (with `fflush()`)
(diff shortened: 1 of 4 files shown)


Sources/Testing/Support/FileHandle.swift
```diff
@@ -607,10 +630,9 @@ extension FileHandle {
         }
       }
 
-      try string.withCString { string in
-        if EOF == fputs(string, file) {
-          throw CError(rawValue: swt_errno())
-        }
+      var string = string
```
grynspan, line 633: Drive-by perf improvement to avoid copying `string` into a null-terminated C string.

harlanhaskins (approved):

PR #1232, opened 2025-07-21
Serialization traits for data dependencies

grynspan:
Experimental `.serialized(for:)` trait. Example use cases:

```swift
import Foundation

@Test(.serialized(for: \ProcessInfo.environment))
func `HAS_FREEZER environment variable`() {
  _ = setenv("HAS_FREEZER", "1", 1)
  #expect(FoodTruck.hasFreezer)
  _ = setenv("HAS_FREEZER", "0", 1)
  #expect(!FoodTruck.hasFreezer)
}

@Test(.serialized(for: *))
func `Uses lots of different state, seriously, just scads of it`() { ... }
```

Resolves rdar://135288463.

### Checklist:

- [x] Code and documentation should follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references should be updated.

PR #1897, opened 2026-09-12
[Testing] Introduce `EnvironmentTrait` for scoped environment variable overrides

Dev10-sys:
Add `EnvironmentTrait` and `Trait.environment(_:)` factory methods to
allow test functions and suites to declare environment variable overrides
that are automatically applied before the test runs and restored
afterwards.

### Motivation

Tests that rely on environment variables today require manual setup and
teardown, which is error-prone and verbose. This change provides a
first-class, declarative way to scope environment mutations to a test or
suite, matching the ergonomics of `TaskLocalTrait`.

### Implementation

- **`EnvironmentTrait`** conforms to `SuiteTrait`, `TestTrait`, and
  `TestScoping`. Its `provideScope(for:testCase:performing:)` method
  saves the current values of the requested keys, sets the overrides,
  runs the test body, and restores original values in a `defer` block.
- **Serialization:** because POSIX `setenv`/`unsetenv` are MT-unsafe,
  all environment-scoped tests are serialized through a shared
  `Serializer<Void>`. A `@TaskLocal` flag prevents re-entrant tasks
  from deadlocking on the serializer.
- **`ReducibleTrait`:** when multiple `EnvironmentTrait` instances are
  inherited (e.g. suite trait + function trait), they are merged so the
  innermost key wins. This is gated on `!hasFeature(Embedded)` to
  match the pattern used by `ParallelizationTrait`.
- **Three factory methods** on `Trait` cover the common call-site shapes:
  - `.environment("KEY", "value")` — single key/value pair
  - `.environment(["KEY": "value"])` — dictionary of `String: String`
  - `.environment(["KEY": nil])` — dictionary of `String: String?` (for unsetting)
- Documentation and `CMakeLists.txt` are updated accordingly.

### Checklist

- [x] Code and documentation follow the style of the [Style Guide](https://github.com/apple/swift-testing/blob/main/Documentation/StyleGuide.md).
- [x] If public symbols are renamed or modified, DocC references are updated.

### DIFF
--- a/Documentation/EnvironmentVariables.md
+++ b/Documentation/EnvironmentVariables.md
@@ -69,4 +69,11 @@ names prefixed with `SWT_`.
 | `SWT_EXPERIMENTAL_SERIALIZED_TRAIT_APPLIES_GLOBALLY` | `Bool` | Whether or not `.serialized` applies globally or just to its branch of the test graph. |
 | `SWT_EXPERIMENTAL_MAXIMUM_PARALLELIZATION_WIDTH` | `Int` | The default parallelization width when parallelized testing is enabled. |
 
+## Scoped environment variables in tests
+
+To apply custom environment variables to tests or suites, use the `Trait/environment(_:)`
+trait. Tests using this trait run serially, preventing data races in the process'
+environment block, and previous environment variable values are automatically restored
+after the test finishes.
+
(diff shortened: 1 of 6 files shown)


grynspan:
Thank you for the PR. Environment variables are global mutable state and care must be taken when working with them. Our planned approach for them and similar global dependencies is in #1232.

PR #1898, opened 2026-09-12
Remove old deprecated SPI and guard deprecated API from Embedded Swift.

grynspan:
This PR deletes old deprecated tools SPI we removed two years ago, and guards some deprecated API in `SourceLocation` and (unfortunately) `__CommandLineArguments_v0` with `#if !hasFeature(Embedded)` so it isn't present there.

### Checklist:


Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
```diff
@@ -35,6 +36,7 @@
 /// }
 @available(swift, deprecated: 100000.0, message: "Use '#Testing::sourceLocation' instead.")
 @freestanding(expression) public macro _sourceLocation() -> SourceLocation = #externalMacro(module: "TestingMacros", type: "SourceLocationMacro")
```
stmontgomery, line 38: Would this produce a better diagnostic as `@_unavailableInEmbedded`?

I'm bringing it up because of the deprecated stuff being excluded from Embedded in this PR, this one feels the most likely to be hit by the casual user, particularly those trying to port their existing code to Embedded mode.

Sources/Testing/Expectations/Expectation+Macro.swift
```diff
@@ -573,6 +573,7 @@ public macro require<E, R>(
   performing expression: () async throws -> R
 ) -> E = #externalMacro(module: "TestingMacros", type: "RequireThrowsMacro") where E: Error & Equatable
 
+#if !hasFeature(Embedded)
```
stmontgomery: Can you elaborate on why these need to be excluded in Embedded? Is it because to do anything useful with an arbitrary `any Error` you'd need to conditionally downcast it?

grynspan: Well it's deprecated, primarily. But also yes, the `any Error` is very unergonomic in Embedded Swift.

Sources/Testing/SourceAttribution/SourceLocation+Macro.swift
grynspan: My expectation is that somebody getting started with Embedded Swift is going to need to adapt a whole bunch of code _anyway_, so it's just part and parcel really.

stmontgomery (approved):

PR #1899, opened 2026-09-13
Adjust `CustomTestStringConvertible` and `CustomIssueRepresentable` for Embedded Swift.

grynspan:
Adjusts the implementations of these two protocols to work in Embedded Swift. The former needs to explicitly access `description` instead of calling `String(describing:)` because that's implemented generically in Embedded Swift (and you can't open an existential there nor cast between generic types). The latter needs to hard-code the list of conforming types, but only for Embedded Swift.

`CustomTestReflectable` is already unavailable in Embedded Swift. `CustomTestArgumentEncodable` will be hidden from Embedded Swift in its entirety in a future PR as it's not possible to implement it there.

### Checklist:

### DIFF
--- a/Sources/Testing/Support/CustomIssueRepresentable.swift
+++ b/Sources/Testing/Support/CustomIssueRepresentable.swift
@@ -32,6 +32,38 @@ protocol CustomIssueRepresentable: Error {
   func customize(_ issue: consuming Issue) -> Issue
 }
 
+/// Customize the given issue if its type conforms to ``CustomIssueRepresentable``.
+///
+/// - Parameters:
+///   - issue: The issue to customize. The function consumes this value.
+///
+/// - Returns: A customized copy of `issue`, or `nil` if its type does not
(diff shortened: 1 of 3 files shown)


Sources/Testing/Support/CustomIssueRepresentable.swift
```diff
@@ -32,6 +32,38 @@ protocol CustomIssueRepresentable: Error {
   func customize(_ issue: consuming Issue) -> Issue
 }
 
+/// Customize the given issue if its type conforms to ``CustomIssueRepresentable``.
```
stmontgomery: "…if it represents a caught error of a type which conforms to…" or something similar?

It's not the _issue_ which conforms

stmontgomery (approved):

Sources/Testing/Support/CustomIssueRepresentable.swift
grynspan: Sure.

PR #1901, opened 2026-09-13
Honor `SWT_NO_*` envvars that we don't explicitly list.

grynspan:
The table of `SWT_NO_*` compiler conditions in Package.swift's `appendFeatureFlags()` function is not comprehensive. The current logic for forwarding environment variables only checks for those in the table. Adjust the logic so we consider compiler conditions not in the table.

### Checklist:


Package.swift
jerryjrchen, line 539: It could help to expand the doc comment with the new behaviour:
- Applies feature flag for any `SWT_NO_*` env var overrides, including for any defines not already "known"
- Otherwise, appends the platform-specific feature flag value

jerryjrchen (approved):

Package.swift
grynspan: We're pretty loosey-goosey on documentation for this file. Would be worth a whole quality pass of its own. Will keep it in mind for a separate PR. :)

PR #1905, opened 2026-09-15
Update the `@Tag` macro to work in Embedded Swift.

grynspan:
This PR modifies the code `@Tag` expands to when building for Embedded Swift so that it successfully compiles. Note that, because we cannot reliably get fully qualified type names at macro expansion time, we cannot currently support tags in nested types inside `Tag`. So this works:

```swift
extension Tag {
  @Tag static var foo: Self
}
```

But this does not:

```swift
extension Tag {
  enum Widgets {
    @Tag static var foo: Tag
  }
}
```

### Checklist:


briancroom (approved):
Can we include a recovery suggestion in this error, directing to avoid using the nested type?

Sources/TestingMacros/Support/Additions/MacroExpansionContextAdditions.swift
```diff
@@ -12,6 +12,16 @@ import SwiftSyntax
 import SwiftSyntaxBuilder
 import SwiftSyntaxMacros
 import SwiftDiagnostics
+import SwiftIfConfig
+
+extension MacroExpansionContext {
+  /// Whether or not the target is building for Embedded Swift.
+  var isTargetEmbedded: Bool {
+    (try? buildConfiguration?.hasFeature(name: "Embedded")) ?? false
```
stmontgomery, line 20: If it throws an error that could be something legitimate e.g. from the underlying macro system; it may be better to propagate that all the way up (and make this a `throws` getter). But returning `false` is reasonable if `buildConfiguration` is `nil`.

stmontgomery (approved):

Sources/TestingMacros/Support/Additions/MacroExpansionContextAdditions.swift
grynspan: It doesn't actually ever throw any errors. See `StaticBuildConfiguration`. I suppose in the future the swift-syntax code owners could start throwing an error, but that would likely break our code in other places too, so we'll have to cross that bridge when we come to it.

grynspan:
> Can we include a recovery suggestion in this error, directing to avoid using the nested type?

I'd want to do that with a Fix-It, but we can't because we can't apply Fix-Its to nodes in the lexical context (they're not part of the actual source-derived graph as far as swift-syntax is concerned.

For now I'm going to leave it out, but we can revisit. We need to do a documentation and diagnostic editorial pass for Embedded Swift anyway.

PR #1906, opened 2026-09-15
Disable error backtrace capturing in Embedded Swift.

grynspan:
Embedded Swift does not invoke the `_swift_willThrow` hook, so the code we have that implements said hook and performs backtrace capture is dead code. Turn it off in Embedded Swift.

`Backtrace` in general would not be supportable without a new PAL annex function, which I have opted not to require/implement at this time. As such, if you attempt to get a backtrace in Embedded Swift, you always get one with an empty backing store. (We may wish to change this behavior and add a PAL annex function in the future if developers ask for it.)

### Checklist:


briancroom (approved):

briancroom:
Are we documenting the Embedded Swift limitations/restrictions anywhere? This seems like something that would be good to make note of.

stmontgomery (approved):

grynspan:
> Are we documenting the Embedded Swift limitations/restrictions anywhere? This seems like something that would be good to make note of.

It's still a moving target; I want to write up documentation once we're able to actually build and run tests as we'll have, presumably, a stable or metastable definition of "what's possible" by then.

PR #1907, opened 2026-09-15
Remove `defaultSynchronousIsolationContext` from Embedded Swift.

grynspan:
`defaultSynchronousIsolationContext` is an existential `any Actor` property on `Configuration` and cannot be used as-is in Embedded Swift. Since we've broadly moved to `nonisolated(nonsending)`, this is mostly fine.

I would have liked to continue explicitly supporting `MainActor`, but it isn't declared in Embedded Swift. If it gets added in the future, we'll want to amend this change to allow specifying `MainActor.shared`.

This PR also includes a drive-by diagnostic fix for `_embeddedTargetInfoCString`.

### Checklist:


Sources/Testing/Test+Macro.swift
```diff
@@ -561,14 +561,6 @@ extension Test {
   value
 }
 
-/// The current default isolation context.
-///
-/// - Warning: This property is used to implement the `@Test` macro. Do not call
-///   it directly.
-public var __defaultSynchronousIsolationContext: (any Actor)? {
```
stmontgomery, line 568: Is this no longer unused?

stmontgomery (approved):

Sources/Testing/Test+Macro.swift
grynspan: It's no longer _used_, if that's what you mean.

PR #1914, opened 2026-09-17
Implement a custom replacement for `JSONEncoder`.

grynspan:
This PR implements a custom JSON encoder without external library dependencies that we can use in Embedded Swift. `JSONEncoder` is not expected to be brought over to Embedded Swift, and although the community has discussed a replacement, a concrete solution isn't coming in the near future.

### Checklist:


Sources/Testing/ABI/Encoded/ABI.EncodedError.swift
```diff
@@ -65,12 +65,28 @@ extension ABI.EncodedError: Error {
...
+    try encoder.encodeJSONEncodableValue(self)
+  }
+}
 #endif
 
+extension ABI.EncodedError: JSON.Encodable {
+  func jsonValue(in context: borrowing JSON.EncodingContext) -> JSON.Value {
+    var result = [String: JSON.Value]()
+
+    result["description"] = description?.jsonValue(in: context)
```
briancroom, line 82: Did you consider writing a macro to synthesize these simple ones?

:ducks:

briancroom (approved):
This is great to see, Jonathan! I hope we don't have to carry this code forever, but in the meantime, thanks for putting together.

Sources/Testing/ABI/Encoded/ABI.EncodedError.swift
grynspan: I did actually, but we can't use macros in our library implementation (we'll fail to build in the toolchain.)

grynspan:
Merging #yolo

PR #1916, opened 2026-09-18
Fix repetition handling for non-failure issues

### LINKED ISSUE #1915: Repetition policy treats warnings and known issues as failures
### Description

The per-test-case repetition policy currently treats every recorded issue as a reason to continue or stop repeating, including warning-severity issues and issues recorded by `withKnownIssue`.

`Configuration.RepetitionPolicy.ContinuationCondition` is documented in terms of issues that affect test success, and `Issue.isFailure` defines that boundary as a non-known issue with severity `error` or greater. The repetition issue recorder in `Sources/Testing/Running/Runner+Repetition.swift` currently records any `.issueRecorded` event without checking `Issue.isFailure`.

### Reproduction

Using Swift Testing `main` at `30babbc2ea2ce1ce1e05b5f6bbf9bc5320ee1a46` (Swift 6.4 toolchain), add the following to the repository's `Tests/TestingTests` target alongside `TestCaseIterationTests.swift` and run it with a maximum iteration count of 3:

```swift
@testable @_spi(Experimental) @_spi(ForToolsIntegrationOnly) import Testing
import Synchronization

let runs = Mutex(0)
configuration.repetitionPolicy = .repeating(
  .whileIssueRecorded,
  maximumIterationCount: 3
)

await Test {
  runs.withLock { $0 += 1 }
  withKnownIssue {
    Issue.record("expected defect")
  }
}.run(configuration: configuration)

print(runs.withLock { $0 })
```

The same behavior occurs with `Issue.record("warning", severity: .warning)`.

### Actual behavior

The known-issue and warning cases run 3 times with `.whileIssueRecorded`. With `.untilIssueRecorded`, they stop after 1 run.

### Expected behavior

Known issues and warning issues should not control repetition because they are not failures:

- `.whileIssueRecorded`: stop after the first iteration when only a known or warning issue was recorded.
- `.untilIssueRecorded`: continue to the maximum iteration count when no failure issue was recorded.
- A non-known issue with severity `error` or greater should retain the current repetition behavior.

### Additional information

The existing repetition tests cover ordinary failure and no-issue cases, but not the distinction between failure, warning, and known issues. A focused regression matrix for those three issue kinds would make the intended contract explicit.

This appears to be a regression in the per-test-case repetition path introduced by the implementation tracked by #1392 and landed in #1528/#1718; it is separate from the already-implemented change from plan-level to per-test-case repetition.

Hokila:
Fixes #1915

### Motivation:

Per-test-case repetition currently records every `issueRecorded` event as a repetition signal. Warning-severity issues and issues recorded by `withKnownIssue` therefore make `whileIssueRecorded` repeat unexpectedly and make `untilIssueRecorded` stop before a failure occurs.

`Issue.isFailure` is the existing source of truth for whether an issue is an unknown failure (severity `error` or greater). Repetition should use that same classification.

### Modifications:

- Record a test case for repetition only when the recorded issue has `isFailure == true`.
- Add focused coverage for known issues and warning issues under both repetition continuation conditions.

### Non-goals:

- No public API, ABI, event schema, or command-line changes.
- No changes to the per-test-case repetition model introduced by #1528/#1718.

### Validation:

- `swift test --filter TestCaseIterationTests` — 9 tests in 1 suite passed (including 5 parameterized cases).
- `swift test --filter 'TestCaseIterationTests|RunnerTests|EventHandlingInteropTests|SwiftPMTests'` — 81 tests passed.
- `swift test` — 613 tests in 75 suites passed; macro suite also passed with 43 tests in 5 suites.
- `git diff --check` — passed.

### API/release impact:

This is a behavioral correctness fix for the existing repetition policy. No API or release-note changes are required.


Hokila:
@swift-ci test

Hokila:
Added commit `7ec74c5` with a focused regression test, `Failures cause repetition while issue recorded`. It records a real error-level issue and asserts that `.whileIssueRecorded` runs the test for all 3 configured iterations. Validation: `swift test --filter TestCaseIterationTests` (13 tests) and full `swift test` (613 tests in 75 suites; macro suite 43 tests in 5 suites), plus `git diff --check`.

Tests/TestingTests/TestCaseIterationTests.swift
```diff
@@ -116,6 +116,67 @@ struct TestCaseIterationTests {
...
+      case .warning:
+        Issue.record("Warning", severity: .warning)
+      }
+    }.run(configuration: configuration)
+
+    return iterations.load(ordering: .sequentiallyConsistent)
+  }
+
+  @Test
+  func `Failures cause repetition while issue recorded`() async {
```
harlanhaskins: Small suggestion: could this be a parameterized test?

```swift
@Test(arguments: [
    (.failure, .whileIssueRecorded, 3),
    // Other cases...
] as [(IssueKind, Configuration.RepetitionPolicy.ContinuationCondition, Int))
func `Ensure only failures cause repetition`(
    issueKind: IssueKind,
    condition: Configuration.RepetitionPolicy.ContinuationCondition,
    expectedIterations: Int
) {
  // ...
}
```

Then you could inline the helper as well.

harlanhaskins:
Thanks for this PR and for bringing this to our attention. Prior to the per-test-case repetition work in #1528, this used to check `!issue.isKnown`, so there's definitely a behavioral regression here. I think changing to `issue.isFailure` is a reasonable change, but it would be an intentional behavioral change.

stmontgomery (approved):
I think we should fix this, even while recognizing that technically it is a behavior change for warning-severity issues.

I think @harlanhaskins's suggestion about adopting a parameterized test is worthwhile, so I'll hold off on merging until that's done or we discuss further and decide not to for some reason

stmontgomery:
This is probably also worth back-porting to the 6.4.x series releases.

@Hokila are you willing/able to create cherry-pick PRs for those releases? There are instructions on https://www.swift.org/documentation/monthly-non-darwin-release/#pull-requests-for-release-branch, and the current list of relevant release branches can be found [here](https://forums.swift.org/t/swift-6-4-release-process/85421/22).

harlanhaskins (approved):

Tests/TestingTests/TestCaseIterationTests.swift
Hokila: Fixed in [42a3898](https://github.com/swiftlang/swift-testing/commit/42a38981a5369a1a46431d8f2fd64fcce23af101). I converted the five repetition checks into a single parameterized test covering all issue kinds and continuation conditions.

Hokila:
@stmontgomery 
Yes, I’m happy to handle the 6.4.x backport(s) after this PR merges into main. I’ll prepare the cherry-pick PR(s) and run the required release-branch validation.

stmontgomery (approved):

Issue #1917, opened 2026-09-18
Define handling for XML 1.0 disallowed control characters in JUnit output


Hokila:
### Description

`Event.JUnitXMLRecorder` can emit XML 1.0 documents that are not parseable when a test records an issue containing a control character such as U+001B (ESC).

The recorder's XML escaping helper handles markup characters and some Unicode values, but XML 1.0 disallows several control-character ranges entirely. Encoding those values as numeric character references would not make them valid XML, so the output policy needs to be decided explicitly.

### Reproduction

On Swift Testing `main` at `30babbc2ea2ce1ce1e05b5f6bbf9bc5320ee1a46`, add a test to the existing `EventRecorderTests` target which records an issue containing U+001B, run it with `--xunit-output`, and parse the generated file with `XMLParser`:

```swift
@Test
func recordsControlCharacter() {
  Issue.record(Comment(rawValue: "A\u{1B}B"))
}
```

The generated failure element contains the literal control character:

```xml
<failure message="Issue recorded (error): A B" />
```

`XMLParser` rejects the document with `NSXMLParserErrorDomain Code=111`.

### Expected behavior

JUnit XML output should always be a valid XML 1.0 document, with a documented policy for characters that XML 1.0 cannot represent. The policy could remove or replace disallowed controls, or otherwise preserve the diagnostic through a valid representation.

### Additional information

This is separate from escaping `&`, `<`, and `"` in `testcase` attributes; those are unambiguous XML escaping fixes and can be handled independently. It is also separate from #1110, which tracks the inability to finalize XML after a process crash.

The issue needs maintainer guidance because XML 1.0 disallowed characters cannot be repaired by simply adding numeric character references, and changing diagnostic text can affect downstream CI consumers.

grynspan:
Probably the easiest change is to just switch to XML 1.1.

PR #1918, opened 2026-09-18
Escape JUnit XML test names and class names

Hokila:
## Motivation

The JUnit XML recorder writes test IDs directly into the `classname` and `name` attributes. A test name containing XML-significant characters such as `&`, `<`, or `"` therefore produces malformed XML.

## Changes

- Escape the JUnit XML `classname` and `name` attributes with the recorder's existing XML escaping helper.
- Add a regression test covering ampersand, less-than, and quotation mark characters and validating the resulting XML with Foundation's parser.

## Non-goals

- This change does not define handling for XML 1.0-disallowed control characters. That policy question is tracked separately in #1917.
- No public API changes.

## Validation

- `swift test --filter EventRecorderTests`
- `swift test` (609 tests in 75 suites; macro suite 43 tests in 5 suites)
- `git diff --check`

Related: #1917


Hokila:
@swift-ci test

grynspan (approved):

stmontgomery (approved):

PR #1919, opened 2026-09-18
Update the migration doc to discuss the new way to attach encodable values.

grynspan:
We recently updated the API for attaching `Codable` and `NSSecureCoding` types to tests. This PR updates our "Migrating from XCTest" document to advise using the new API rather than the old one, which is now to-be-deprecated.

See also [ST-0028](https://github.com/swiftlang/swift-evolution/blob/main/proposals/testing/0028-revise-attachment-encodable-interfaces.md).

### Checklist:


Sources/Testing/Testing.docc/MigratingFromXCTest.md
```diff
@@ -106,7 +106,7 @@ cross-library issues:
 The default interoperability mode depends on your toolchain and the
 `swift-tools-version` declared in your package.
 
-| Toolchain version | `swift-tools-version` | Default interoperability mode |
```
grynspan, line 109: There's an invisible control character here that Xcode stripped. 🤷

briancroom (approved):

PR #1922, opened 2026-09-20
Add a "Showcase" target for Embedded Swift.

grynspan:
This PR adds an executable target for Embedded Swift that runs a few basic test functions.

We add a to-be-renamed-or-refactored entry point function for Embedded Swift, `swift_testing_embeddedMain(_:_:)`. This function acts much like the existing entry point functions, but is intended specifically for use by an Embedded Swift target that is not using Swift Package Manager. It is exported as a C function so that it can be directly passed `argv` and `argc` from a C context if needed; if not possible, a Swift overload is provided that takes the caller's source location and synthesizes a very minimal command line argument list.

This change enables `--filter` and `--skip` when `_StringProcessing` isn't available, with limited string-matching abilities. (Better than nothing!)

This change also applies some minor fixups to previous Embedded Swift-related commits now that I'm actually building them against an Embedded Swift target and the compiler's actually complaining.

You can't quite use this branch yet because the Swift Concurrency static library is broken (see https://github.com/swiftlang/swift/pull/92355). In the mean time, you can look at the `jgrynspan/embedded-showcase-with-hackery` branch which can be used with non-`async` tests, though it's not in shippable condition.

### Checklist:


Sources/Testing/ABI/EntryPoints/EmbeddedEntryPoint.swift
```diff
@@ -0,0 +1,63 @@
...
+
+private import _TestingInternals
+
+#if hasFeature(Embedded)
+/// Keep the process running until all asynchronous work has completed.
+///
+/// This function is exported by the Concurrency module. We use it in Embedded
+/// Swift because `_runAsyncMain()` is not available there.
+@_silgen_name("swift_task_asyncMainDrainQueue")
+private func _asyncMainDrainQueue() -> Never
```
stmontgomery, line 19: Do you consider this a temporary workaround?

Package.swift
```diff
@@ -107,6 +107,22 @@ let package = Package(
...
+    result += [
+      .executable(
+        name: "EmbeddedShowcase",
+        targets: ["EmbeddedShowcase"]
+      ),
+      .executable(
+        name: "SymbolShowcase",
+        targets: ["SymbolShowcase"]
+      ),
+
```
stmontgomery: nit: stray newline

Package.swift
```diff
@@ -137,10 +153,8 @@ let package = Package(
       dependencies: [
         "_TestDiscovery",
         "_TestingInternals",
-      ] + {
-        // TODO: get macro target building for host when the target is embedded
-        buildingForEmbedded ? [] : ["TestingMacros"]
-      }(),
+        "TestingMacros",
+      ],
```
stmontgomery, line 156: Just checking: this workaround hasn't been necessary for a while, is that right? It's just that you changed how you were performing the build which resolved this, and it's not a code change elsewhere in this PR which removed the need for this?

stmontgomery (approved):

Sources/Testing/ABI/EntryPoints/EmbeddedEntryPoint.swift
grynspan: It's likely semi-permanent but I do want to remove it, obviously. We'll need to discuss with the core team how best to approach the problem.

Package.swift
grynspan: Fixed it a PR or two back, yeah.

